PR #14, opened 2023-09-03
Bump and patch

carlopi:


Mytherin:
Thanks!

Mytherin:
As an aside this is why I generally recommend not to place the entire error message in the `statement error`:)

PR #15, opened 2023-09-22
Move documentation to the DuckDB documentation and add reference

szarnyasg:
Depends on https://github.com/duckdb/duckdb-web/pull/1214 being merged, which moves the Iceberg instructions to the DuckDB docs.


samansmink:
looks good!

PR #28, opened 2023-11-09
pin run-vcpkg

samansmink:
run-vcpkg was updated and 11.2 is incompatible with our manylinux image


Fokko (approved):
LGTM, thanks for fixing this!

PR #30, opened 2023-11-27
Parse schema from table metadata

samansmink:
This PR changes the way we scan Iceberg tables. Before we would simply scan them as a list of parquet files, using the schema we found in the first parquet file scanned.

Now the iceberg metadata is actually scanned to parse the schema and pass that to the parquet_scan in duckdb using the `schema` parameter: https://github.com/duckdb/duckdb/pull/9123

This effectively adds support for schema evolution and resolves issues where the parquet schema is different from the iceberg schema.

To test it, the spark-based test data generation script was updated to include schema evolution. Note that it is not yet run from CI, but that will follow soon in conjunction with the REST catalog support and the testing infrastructure added in https://github.com/duckdb/duckdb_iceberg/pull/27

### DIFF
--- a/src/common/iceberg.cpp
+++ b/src/common/iceberg.cpp
@@ -90,9 +90,9 @@ vector<IcebergManifestEntry> IcebergTable::ReadManifestEntries(const string &pat
 	return ret;
 }
 
-unique_ptr<SnapshotParseInfo> IcebergSnapshot::GetParseInfo(yyjson_doc *metadata_json) {
+unique_ptr<SnapshotParseInfo> IcebergSnapshot::GetParseInfo(yyjson_doc &metadata_json) {
 	SnapshotParseInfo info {};
-	auto root = yyjson_doc_get_root(metadata_json);
+	auto root = yyjson_doc_get_root(&metadata_json);
 	info.iceberg_version = IcebergUtils::TryGetNumFromObject(root, "format-version");
 	info.snapshots = yyjson_obj_get(root, "snapshots");
 
@@ -121,7 +121,7 @@ unique_ptr<SnapshotParseInfo> IcebergSnapshot::GetParseInfo(yyjson_doc *metadata
 unique_ptr<SnapshotParseInfo> IcebergSnapshot::GetParseInfo(const string &path, FileSystem &fs) {
 	auto metadata_json = ReadMetaData(path, fs);
 	auto doc = yyjson_read(metadata_json.c_str(), metadata_json.size(), 0);
-	auto parse_info = GetParseInfo(doc);
+	auto parse_info = GetParseInfo(*doc);
 
 	// Transfer string and yyjson doc ownership
 	parse_info->doc = doc;
--- a/src/include/iceberg_metadata.hpp
+++ b/src/include/iceberg_metadata.hpp
@@ -66,8 +66,7 @@ public:
 	                                     vector<yyjson_val *> &schemas);
 	static string ReadMetaData(const string &path, FileSystem &fs);
 	static yyjson_val *GetSnapshots(const string &path, FileSystem &fs);
-	//! Note: caller keeps ownership of yyjson doc
-	static unique_ptr<SnapshotParseInfo> GetParseInfo(yyjson_doc *metadata_json);
+	static unique_ptr<SnapshotParseInfo> GetParseInfo(yyjson_doc &metadata_json);
 
 protected:
 	//! Internal JSON parsing functions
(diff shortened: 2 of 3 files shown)


src/common/schema.cpp
```diff
@@ -0,0 +1,185 @@
...
+		return LogicalType::TIMESTAMP_TZ;
+	}
+	if (type_str == "string") {
+		return LogicalType::VARCHAR;
+	}
+	if (type_str == "uuid") {
+		return LogicalType::UUID;
+	}
+	if (StringUtil::StartsWith(type_str, "fixed")) {
+		// Is this the right conversion ?
```
Fokko: Similar to the decimal, you want to store the length of the fixed field. `fixed[22]` tells that the blobs have fixed lengths of 22.

src/common/schema.cpp
```diff
@@ -0,0 +1,185 @@
...
+	if (StringUtil::StartsWith(type_str, "decimal")) {
+		D_ASSERT(type_str[7] == '(');
+		D_ASSERT(type_str.back() == ')');
+		auto start = type_str.find('(');
+		auto end = type_str.rfind(')');
+		auto raw_digits = type_str.substr(start+1, end - start);
+		auto digits = StringUtil::Split(raw_digits, ',');
+		D_ASSERT(digits.size() == 2);
+
+		auto width = std::stoi(digits[0]);
```
Fokko, line 132: Checked if it strips whitespace and it looks like it does.

src/common/iceberg.cpp
```diff
@@ -164,6 +165,8 @@ IcebergSnapshot IcebergSnapshot::ParseSnapShot(yyjson_val *snapshot, idx_t icebe
 	ret.timestamp_ms = Timestamp::FromEpochMs(IcebergUtils::TryGetNumFromObject(snapshot, "timestamp-ms"));
 	ret.manifest_list = IcebergUtils::TryGetStrFromObject(snapshot, "manifest-list");
 	ret.iceberg_format_version = iceberg_format_version;
+	ret.schema_id = schema_id;
+	ret.schema = ParseSchema(schemas, ret.schema_id);
```
Fokko, line 199: Some context here. The schema can be in the `schema` or `schemas` depending if it is a V1 or V2 table:

image

It is recommended now to use `schemas` because that will give information about the historical schemas. In PyIceberg, when we see a schema, we'll add it to the schemas and set the `current-schema-id` to the schema. This way we don't really care which version the table is, and we can treat them the same.

Fokko (approved):
Looking good @samansmink

src/common/iceberg.cpp
samansmink: Ah good catch! I updated it to use your suggested method to support both v1 and v2.

src/common/schema.cpp
samansmink: So this is good as is right?

src/common/schema.cpp
samansmink: Added a fixme for now

src/common/iceberg.cpp
```diff
@@ -62,7 +63,8 @@ vector<IcebergManifest> IcebergTable::ReadManifestListFile(string path, FileSyst
 	return ret;
 }
 
-vector<IcebergManifestEntry> IcebergTable::ReadManifestEntries(string path, FileSystem &fs, idx_t iceberg_format_version) {
+vector<IcebergManifestEntry> IcebergTable::ReadManifestEntries(string path, FileSystem &fs,
```
Tishj: `const string &path`
I think clang-tidy will complain about this as well?

src/include/iceberg_metadata.hpp
```diff
@@ -23,20 +55,28 @@ class IcebergSnapshot {
...
 	static IcebergSnapshot GetSnapshotById(string &path, FileSystem &fs, idx_t snapshot_id);
 	static IcebergSnapshot GetSnapshotByTimestamp(string &path, FileSystem &fs, timestamp_t timestamp);
 
-	static IcebergSnapshot ParseSnapShot(yyjson_val *snapshot, idx_t iceberg_format_version);
+	static IcebergSnapshot ParseSnapShot(yyjson_val *snapshot, idx_t iceberg_format_version, idx_t schema_id,
+	                                     vector<yyjson_val *> &schemas);
 	static string ReadMetaData(string &path, FileSystem &fs);
+	static yyjson_val *GetSnapshots(string &path, FileSystem &fs);
+	//! Note: caller keeps ownership of yyjson doc
+	static unique_ptr<SnapshotParseInfo> GetParseInfo(yyjson_doc *metadata_json);
```
Tishj: Maybe this should take a reference instead of a raw pointer, that might better convey that this function doesn't take ownership over the yyjson doc

samansmink: good one, that seems nicer

src/common/iceberg.cpp
samansmink: clang-tidy is not currently run on extensions, but indeed that is cleaner, fixed it!

PR #42, opened 2024-03-06
Support to read gzip compressed metadata files

devendrasr:
The current version of iceberg extension does not support reading iceberg tables that were written with gzip compressed metadata files. Adding one more flag in the commands to consider the metadata files as gz while loading metadata.

```
SELECT * FROM iceberg_scan("s3://my-bucket/icebergwh/someschema/t01", metadata_compression_codec="gzip") limit 10;
``` 
As of now  *metadata_compression_codec* can be none or gzip; default is none.

### DIFF
--- a/src/common/iceberg.cpp
+++ b/src/common/iceberg.cpp
@@ -165,7 +165,6 @@ IcebergSnapshot IcebergSnapshot::GetSnapshotByTimestamp(const string &path, File
 
 string IcebergSnapshot::ReadMetaData(const string &path, FileSystem &fs, string metadata_compression_codec) {
 	string metadata_file_path;
-	printf("got metadata_compression_codec=%s\n", metadata_compression_codec.c_str());
 	if (StringUtil::EndsWith(path, ".json")) {
 		metadata_file_path = path;
 		// check if metadata is gz compressed file?
--- a/src/common/iceberg.cpp
+++ b/src/common/iceberg.cpp
@@ -13,7 +13,7 @@
 namespace duckdb {
 
 IcebergTable IcebergTable::Load(const string &iceberg_path, IcebergSnapshot &snapshot, FileSystem &fs,
-                                bool allow_moved_paths) {
+                                bool allow_moved_paths, string metadata_compression_codec) {
 	IcebergTable ret;
 	ret.path = iceberg_path;
 	ret.snapshot = snapshot;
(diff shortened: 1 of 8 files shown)


src/common/utils.cpp
```diff
@@ -12,6 +14,41 @@ string IcebergUtils::FileToString(const string &path, FileSystem &fs) {
 	return ret_val;
 }
 
+// Function to decompress a gz file content string
+string IcebergUtils::GzFileToString(const string &path, FileSystem &fs) {
```
samansmink, line 19: Have you considered using DuckDB's [GZipFilesystem](https://github.com/duckdb/duckdb/blob/main/src/include/duckdb/common/gzip_file_system.hpp) here instead? I think that would be the preferred way to go here.

samansmink (commented):
Hey @devendrasr thanks a lot for the PR! In general it looks good.

I added one comment regarding the actual unzipping and I have a more general comment: I would like to see some tests here! Maybe you can add some small test data with a test that confirms this feature works?

src/common/utils.cpp
devendrasr: Thank you Sam, I will try that out.

devendrasr: @samansmink as per your feedback, i am using  [GzipFileSystem](https://github.com/duckdb/duckdb/blob/main/src/include/duckdb/common/gzip_file_system.hpp) to decompress the gz file and added test data and test cases on top of above feature. Please let me know in case you need more info.

src/common/iceberg.cpp
```diff
@@ -130,61 +130,71 @@ unique_ptr<SnapshotParseInfo> IcebergSnapshot::GetParseInfo(const string &path,
...
+	return ParseSnapShot(snapshot, info->iceberg_version, info->schema_id, info->schemas, metadata_compression_codec);
 }
 
-string IcebergSnapshot::ReadMetaData(const string &path, FileSystem &fs) {
+string IcebergSnapshot::ReadMetaData(const string &path, FileSystem &fs, string metadata_compression_codec) {
 	string metadata_file_path;
-
 	if (StringUtil::EndsWith(path, ".json")) {
 		metadata_file_path = path;
+		// check if metadata is gz compressed file?
```
samansmink: I feel this code can be simplified a bit, it is currently somewhat unclear.

By moving the metadata url generation to a separate function we can check `metadata_compression_codec` only once in this function, something like:

```
        if (StringUtil::EndsWith(path, ".json")) {
		metadata_file_path = path;
	} else {
		auto table_version = GetTableVersion(path, fs);
		metadata_file_path = GenerateMetaDataUrl(fs, path, table_version, metadata_compression_codec);
	}

	if (metadata_compression_codec == "gzip") {
		return IcebergUtils::GzFileToString(metadata_file_path, fs);
	}
	return IcebergUtils::FileToString(metadata_file_path, fs);
```

samansmink (commented):
Hey @devendrasr Thanks for the changes, looks good! I have one more comment then I think this is good to go

src/common/iceberg.cpp
devendrasr: Hi Sam, 
I am done with the said changes. Please let me know in case anything else is missing.

Dev

devendrasr:
@samansmink  Please let me know if you are good to go. Once this is done, we can think about https://github.com/duckdb/duckdb_iceberg/pull/43

samansmink:
looks good now, thanks again @devendrasr!

Issue #39, opened 2024-02-13
0.10.0 Regression - Cannot parse metadata 


harel-e:
Starting with DuckDB 0.10.0, I can no longer use the extension to scan iceberg files

select * from iceberg_scan('s3://zones/iceberg/test/zones_43_1b7d1e67-8f53-42ab-ba55-a66f438f344f/metadata/00006-0d6ca258-8039-446f-843a-69e21344f272.metadata.json');

**Error: IO Error: Invalid field found while parsing field: required**

The above works fine with DuckDB 0.9.2

I attached the metadata file, which was created by Iceberg.
[00006-0d6ca258-8039-446f-843a-69e21344f272.metadata.json](https://github.com/duckdb/duckdb_iceberg/files/14271324/00006-0d6ca258-8039-446f-843a-69e21344f272.metadata.json)

harel-e:
It seems that if the schema contains a required field ("required" : true) the extension fails.
I tried it with https://duckdb.org/data/iceberg_data.zip
I changed one of the fields in v2.metadata.json from **"required" : false** to **"required" : true**

0.9.2 - No problems

0.10.0
```
SELECT * FROM iceberg_scan('data/iceberg/lineitem_iceberg', allow_moved_paths = true);
Error: IO Error: Invalid field found while parsing field: required
```

Sol-Hee:
I'm experiencing the same problem. 
I upgraded the code that used to work at 0.9.2 to 0.10.0 and got IOException. 

`IOException: IO Error: Invalid field found while parsing field: type`

samansmink:
Thanks for reporting @harel-e and @Sol-Hee. 

This is a known issue that is a result from https://github.com/duckdb/duckdb_iceberg/pull/30. In that PR we switched from parsing the parquet schemas to actually parsing the iceberg schema. However, since duckdb's `schema` param for the parquet reader does not yet support nested types, this support was dropped. The way forward here is to:
- add support for nested types in duckdb parquet reader's `schema` option
- add support for iceberg nested types in this extension

harel-e:
@samansmink Thank you for this wonderful extension.
I wonder if there an estimate when this issue will be fixed and are there open issues on the two items above?
I can assist with testing if needed.
Thanks

harel-e:
Ok, thanks for the update. Too bad I cannot upgrade. I'll stick to DuckDB 0.9.2 until the issue is fixed.

veca1982:
[...]

PR #50, opened 2024-04-10
Change yyjson_get_tag() to yyjson_get_type() when doing type comparisons.

rustyconover:
fix: change yyjson comparisions to type rather than tag.
    
Many comparisons were performed using `yyjson_get_tag()` rather than `yyjson_get_type()`.  The tag can have additional information set using bits beyond just the type, causing these type comparisons to fail and parsing the metadata to fail.
    
fix: fix the extension to build with current duckdb main branch.

Fix a few `std::move()` calls and a call to `fs.OpenFile()`.


harel-e:
Issue #39 is fixed with this PR

samansmink:
Looks good to me, thanks a lot!

PR #59, opened 2024-06-28
add generated_data to repo to facilitate CI

peterboncz:
- Commit the generated data into the repo, because this facilitates running the tests in CI
  (since generating the data requires to equip the test runners with pyspark, which at least motherduck prefers not to).

- The data is minimized to 5MB, by setting the scale factor 10x smaller, and from expected_results/ only keeping last/. 
   This motivates rewriting some testcases that used expected_results/q08 into expected_results/last (which is the same)

- Removed one testcase, which tests if we get an error when parquet is not loaded (in MotherDuck, when we run iceberg in so-called "remote_only" mode, this will actually not raise an error).
   It was kind of a trivial test, so I hope this is OK..


Makefile
```diff
@@ -9,11 +9,11 @@ include extension-ci-tools/makefiles/duckdb_extension.Makefile
 
 # Custom makefile targets
 data: data_clean
-	python3 scripts/test_data_generator/generate_iceberg.py 0.01 data/iceberg/generated_spec1_0_01 1
-	python3 scripts/test_data_generator/generate_iceberg.py 0.01 data/iceberg/generated_spec2_0_01 2
+	python3 scripts/test_data_generator/generate_iceberg.py 0.001 data/iceberg/generated_spec1_0_01 1
```
samansmink: nit: the name of the directory is now no longer correct which could be a little confusing

samansmink:
windows CI is failing due to some upstream issue, however the tests also seem to fail due to only the json metadata files being added and not the avro and parquet files

peterboncz:
> windows CI is failing due to some upstream issue, however the tests also seem to fail due to only the json metadata files being added and not the avro and parquet files

apologies for that messup. I have made the name change and made sure all files are pushed now.

samansmink:
Thanks for the changes! looks good!

PR #70, opened 2024-10-01
Use last instance of /data/ and /metadata/ path

### LINKED ISSUE #69: Cannot query when there is a folder named 'data' in path when allow_moved_paths = true
I wrote 2 identical iceberg tables but at different path:
- /Users/henneberger/data/duckdb
- /Users/henneberger/not-data/duckdb

However this query will work:
`select * from iceberg_scan('/Users/henneberger/data/duckdb/default_database/mytable', allow_moved_paths = true);`

But this query will fail:
`select * from iceberg_scan('/Users/henneberger/not-data/duckdb/default_database/mytable', allow_moved_paths = false);`

It appears its when there is a folder name 'data', it will do something weird.

The jdbc driver will give me:
```
IO Error: No files found that match the pattern \"/Users/henneberger/data/duckdb/default_database/mytable/data/duckdb/default_database/mytable/data/00000-0-e2ea7c00-ed3d-4d30-a230-847e4029cadf-00001.parquet
```

henneberger:
Resolves #69

### DIFF
--- a/src/common/utils.cpp
+++ b/src/common/utils.cpp
@@ -22,12 +22,12 @@ string IcebergUtils::GzFileToString(const string &path, FileSystem &fs) {
 }
 
 string IcebergUtils::GetFullPath(const string &iceberg_path, const string &relative_file_path, FileSystem &fs) {
-	std::size_t found = relative_file_path.find("/metadata/");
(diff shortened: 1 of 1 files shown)


samansmink:
LGTM, thanks!

PR #74, opened 2024-10-09
exploit iceberg row-count metadata

peterboncz:
Examine the row-counts in the manifest and count how many rows are in the (existing or added) data resp. deletion files.

Use these two counts to pass the new named_parameter "explicit_cardinality" into the generated parquet_scans.

Note that because iceberg is passing in an explicit schema into the parquet_scan, it does not open any file during bind. That means the parquet_scans it generates did not have any cardinality information during query optimization.

That can cause rather bad query plans. This PR (pb/explicit-iceberg-cardinality) fixes that. Note that this PR needs the DuckDB PR pb/explicit-parquet-cardinality that adds the "explicit_cardinality" named_parameter to parquet_scan.


peterboncz:
This CI is expected to fail (unknown named_parameter 'explicit_cardinality') until https://github.com/duckdb/duckdb/pull/14292 would have been merged

Issue #29, opened 2023-11-21
Can't use the extension if my data catalog did not create a version-hint.text file


jacopotagliabue:
My s3 bucket with iceberg (picture below) cannot be queried with

```iceberg_scan('s3://bucket/iceberg', ALLOW_MOVED_PATHS=true)```

nor 

```iceberg_scan('s3://bucket/iceberg/*', ALLOW_MOVED_PATHS=true)```

In particular the system is trying to find a very specific file (so the * pattern gets ignored):

```duckdb.duckdb.Error: Invalid Error: HTTP Error: Unable to connect to URL
https://bucket.s3.amazonaws.com/iceberg/metadata/version-hint.text```

Unfortunately that file does not exist in my iceberg/ folder, nor in any of the iceberg/sub/metadata folders. Compared to the data zip in duckdb docs about iceberg, it is clear "my iceberg tables" are missing that file, which is important for the current implementation. 

That said, version-hint seems something we do not really need, as that info can default to a version or being an additional parameter perhaps (instead of failing if the file is not found)?

[Original discussion with @Alex-Monahan in dbt Slack is here](https://getdbt.slack.com/archives/C039D1J1LA2/p1700599053331209): note that I originally got pointed to [this](https://github.com/duckdb/duckdb_iceberg/issues/25) as a possible cause, so perhaps reading a table that is formally Iceberg is not really independent from the data catalog it belongs to?

![s3_structure](https://github.com/duckdb/duckdb_iceberg/assets/3503170/8e11c0b1-bc46-4232-87ec-104a6e17af6b)

jacopotagliabue:
Sorry to be a bit clearer: even if we fix the `version-hint` problem, the fact that the system is looking at `https://bucket.s3.amazonaws.com/iceberg/metadata/` as a base path seems to be not aligned with the state of my data lake (see the picture above for the current layout, written by Spark Nessie). 

Happy to help debug this if there's something we can quickly try out.

harel-e:
I ran into similar issue using AWS with Glue as the catalog for Iceberg.

The metadata files stored in S3 are of the following pattern:

00000-0b4430d2-fbee-4b0d-90c9-725f013d6f82.metadata.json
00001-6e3b4909-7e6b-486f-bf81-b1331eba3ac8.metadata.json

I suspect Glue holds the pointer to the current metadata.

rustyconover:
It does.

You can see the current pointer in table properties if you call Glue’s
DescribeTable

[...]

PR #79, opened 2024-11-12
Issues/29: Add version guessing functionality

teaguesterling:
This adds a new (default) version option that can be passed to iceberg table functions.

When `version="?"`, some simple logic will go into effect that does the following:
 - Checks for a `version-hint.text` file, using it if it exists.
 - If not, it will Glob the metadata directory using the `version_format` patterns, and select the lexicographically "largest" metadata filename (e.g., `v00002-xxxxxxxxxxxxx.metadata.json` will be picked over `v00001-xxxxxxxxxxxxx.metadata.json` ).

There's additional tests needed here and I'm running into a seemingly unrelated parquet bug when testing. There aslo some room for improvement on the logic of how versions are found and selected. However, at the very least, this works as we've all hoped from #29 :

```
D SELECT file_path FROM iceberg_metadata('lineitem_iceberg_no_hint');
┌────────────────────────────────────────────────────────────────────────────────────┐
│                                     file_path                                      │
│                                      varchar                                       │
├────────────────────────────────────────────────────────────────────────────────────┤
│ lineitem_iceberg/data/00041-414-f3c73457-bbd6-4b92-9c15-17b241171b16-00001.parquet │
│ lineitem_iceberg/data/00000-411-0792dcfe-4e25-4ca3-8ada-175286069a47-00001.parquet │
└────────────────────────────────────────────────────────────────────────────────────┘
D SELECT file_path FROM iceberg_metadata('lineitem_iceberg_no_hint', version='1');
┌────────────────────────────────────────────────────────────────────────────────────┐
│                                     file_path                                      │
│                                      varchar                                       │
├────────────────────────────────────────────────────────────────────────────────────┤
│ lineitem_iceberg/data/00000-411-0792dcfe-4e25-4ca3-8ada-175286069a47-00001.parquet │
└────────────────────────────────────────────────────────────────────────────────────┘
D SELECT file_path FROM iceberg_metadata('lineitem_iceberg_no_hint', version='2');
┌────────────────────────────────────────────────────────────────────────────────────┐
│                                     file_path                                      │
│                                      varchar                                       │
├────────────────────────────────────────────────────────────────────────────────────┤
│ lineitem_iceberg/data/00041-414-f3c73457-bbd6-4b92-9c15-17b241171b16-00001.parquet │
│ lineitem_iceberg/data/00000-411-0792dcfe-4e25-4ca3-8ada-175286069a47-00001.parquet │
└────────────────────────────────────────────────────────────────────────────────────┘
D .system ls lineitem_iceberg_no_hint/metadata
10eaca8a-1e1c-421e-ad6d-b232e5ee23d3-m0.avro
10eaca8a-1e1c-421e-ad6d-b232e5ee23d3-m1.avro
cf3d0be5-cf70-453d-ad8f-48fdc412e608-m0.avro
snap-3776207205136740581-1-cf3d0be5-cf70-453d-ad8f-48fdc412e608.avro
snap-7635660646343998149-1-10eaca8a-1e1c-421e-ad6d-b232e5ee23d3.avro
v1.metadata.json
v2.metadata.json

### CODE REVIEW

**Comment by teaguesterling** [comment]
> Thanks for the PR @teaguesterling, I think this makes sense! There do seem to be some tests failing related to this PR
> 

Thanks, @samansmink. I'm trying to debug the test failures. I added some new ones to cover the new functionality introduced here but was seeing failures in untouched parts of the code related to the `explicit_cardinality` parameter in parquet scanning. Removing those bits caused tests to not throw exceptions but instead return incorrect results, perhaps related to deletion handing. I've been working off the main branch. Is there a different branch to consider using for testing?

**Comment by teaguesterling** [comment]
@samansmink It seems there was some weirdness with the test data and cases in my branch. I've cleaned those up and made some fixes to get this to pass.

File: src/common/iceberg.cpp [hunk 1]
```diff
@@ -188,19 +188,32 @@ string GenerateMetaDataUrl(FileSystem &fs, const string &meta_path, string &tabl
...
-	}
-
-	auto meta_path = fs.JoinPath(path, "metadata");
-	string version_hint;
-		if(StringUtil::EndsWith(table_version, ".text")||StringUtil::EndsWith(table_version, ".txt")) {
-		version_hint = GetTableVersion(meta_path, fs, table_version);
-	} else {
+	} else if (!fs.DirectoryExists(meta_path)) {
+		// Make sure we have a metadata directory to look in
+		throw IOException("Cannot open \""+path+"\": Metadata directory does not exist");
```
Tishj: Bit of a nitpick, feel free to ignore this if there is no other reason to touch the branch

Formatting can be used within an exception message:
This could instead be:`"Cannot open \"%s\": Metadata directory does not exist", path`

src/common/iceberg.cpp
```diff
@@ -188,19 +188,32 @@ string GenerateMetaDataUrl(FileSystem &fs, const string &meta_path, string &tabl
...
+	} else if(StringUtil::EndsWith(table_version, ".text")||StringUtil::EndsWith(table_version, ".txt")) {
+		// We were given a hint filename
+		version_hint = GetTableVersionFromHint(meta_path, fs, table_version);
+		return GenerateMetaDataUrl(fs, meta_path, version_hint, metadata_compression_codec, version_format);
+	} else if(!StringUtil::StartsWith(table_version, "?")) {
```
Tishj: Could you elaborate in a comment what it implies when a table version does not start with `?`

teaguesterling:
Happy to do that! Thank you for the review

src/common/iceberg.cpp
```diff
@@ -188,19 +188,32 @@ string GenerateMetaDataUrl(FileSystem &fs, const string &meta_path, string &tabl
...
+		// We were given an explicit version number
 		version_hint = table_version;
+		return GenerateMetaDataUrl(fs, meta_path, version_hint, metadata_compression_codec, version_format);
+	} else if(fs.FileExists(fs.JoinPath(meta_path, DEFAULT_VERSION_HINT_FILE))) {
+		// We're guessing, but a version-hint.text exists so we'll use that
+		version_hint = GetTableVersionFromHint(meta_path, fs, DEFAULT_VERSION_HINT_FILE);
+		return GenerateMetaDataUrl(fs, meta_path, version_hint, metadata_compression_codec, version_format);
+	} else {
+		// We need to guess from file paths
+		return GuessTableVersion(meta_path, fs, table_version, metadata_compression_codec, version_format);
```
Tishj, line 217: I've talked with Sam about this, we feel that this could be unsafe.
Potentially you could read an uncommitted version, so we agreed that enabling this behavior should probably be put behind an `unsafe_ ...` setting that is opt-in.

I plan on making this change to your branch, just to let you know 👍

src/common/iceberg.cpp
```diff
@@ -251,6 +264,39 @@ string IcebergSnapshot::GetTableVersion(const string &meta_path, FileSystem &fs,
...
+	throw IOException(
+	        "Could not guess Iceberg table version using '%s' compression and format(s): '%s'",
+	        metadata_compression_codec, version_format);
+}
+
+string IcebergSnapshot::PickTableVersion(vector<string> &found_metadata, string &version_pattern, string &glob) {
+	// TODO: Different "table_version" strings could customize this
+	// For now: just sort the versions and take the largest
+	std::sort(found_metadata.begin(), found_metadata.end());
+	return found_metadata.back();
```
Tishj: <https://en.cppreference.com/w/cpp/container/vector/back>
> Calling back on an empty container causes [undefined behavior](https://en.cppreference.com/w/cpp/language/ub).

We should probably check if `found_metadata` is empty and return `string()` explicitly instead
Or add a `D_ASSERT(!found_metadata.empty())` because this *is* guaranteed by the caller it seems

src/common/iceberg.cpp
Tishj: If this relates to `DEFAULT_TABLE_VERSION` can we instead turn this into `table_version != DEFAULT_TABLE_VERSION` ?

src/common/iceberg.cpp
samansmink: Yea so after reading [this](https://iceberg.apache.org/spec/#metastore-tables) part of the spec, and reading some discussion on Static Tables in the Iceberg repository, we concluded that unless you use the (deprecated!) [file system tables](https://iceberg.apache.org/spec/#file-system-tables) listing the version files the way its currently done in this PR would break ACID guarantees. It's probably sensible to lock this behind a flag forcing the users to be aware of this.

src/common/iceberg.cpp
teaguesterling: Thanks! I had moved that up to the caller and probably shouldn't have.

src/common/iceberg.cpp
teaguesterling: I fully agree with having this behind some sort of unsafe flag. After seeing lots of continued discussion around issue #29 I wanted to see if we could provide some finality to the functionality to close the issue. I'm not sure if there is any other way to address this without using a catalog or a hint file, but I can review the spec again (thank you for linking).

src/common/iceberg.cpp
teaguesterling: Another artifact of cleaning up my implementation to fix the tests. Originally, I had considered having two different "?" implementations (most recently created vs. lexicographically highest). I will update to the simpler proposal.

src/common/iceberg.cpp
teaguesterling: A good nitpick. I felt dirty as I wrote that but didn't go back to clean it up.

src/common/iceberg.cpp
samansmink: As far as I understand it @teaguesterling there is simply no "safe" way to do this in iceberg without a catalog. The way to read a static table is by supplying the full path to the metadata file of a manually picked version

teaguesterling: That's my understanding as well. The number of people that wanted this (and the duckdb "spirit" of just making it work) motivated me to implement this.

djouallah:
a user here, please we need this, we are already doing that hack by listing the metadata and picking the latest metadata, unfortunately, it does not works with movable path.

src/common/iceberg.cpp
teaguesterling: Thank you again for implementing the config variable. I've moved things around a bit for (what I perceive as) increased readability.

teaguesterling:
This PR addresses the review comments from @Tishj and @samansmink and updates the test cases to cover the new config variable.

However, it seems that the OSX build is failing before getting building or testing the extension.

samansmink (approved):
Looks great, thanks a lot @teaguesterling

teaguesterling:
> Looks great, thanks a lot @teaguesterling

Thanks for all your support on this! I had been looking for a way to familiarize myself with the DuckDB extension system.

PR #87, opened 2024-12-06
Fix parsing of list and map schemas from metadata JSON file

### LINKED ISSUE #47: IOException when querying table with a list<int> column.
Hello, thanks for your great work on duckdb-iceberg!

I'm encountering an issue where attempting to query an iceberg table with `iceberg_scan` fails if the table has a column which is a list of ints (`list<int>`). 

```python
table_with_lists # a pyiceberg.table.Table object
table_without_lists # another pyiceberg.table.Table object

duckdb.sql("INSTALL ICEBERG; LOAD ICEBERG;")

# querying the table that doesn't feature any lists works fine
duckdb.sql(f"SELECT * FROM iceberg_scan('{table_without_lists.metadata_location}') LIMIT 10")

# querying the table that *does* have lists throws an error
duckdb.sql(f"SELECT * FROM iceberg_scan('{table_with_lists.metadata_location}') LIMIT 10")
# IOException: IO Error: Invalid field found while parsing field: type
```

This seems to be related to an error [parsing the table schema](https://github.com/duckdb/duckdb_iceberg/blob/5a6b6e738f2009a09b92d948595a160955b826af/src/common/schema.cpp#L73-L81). I was wondering if there maybe was no logic for complex types in the duckdb-iceberg extension, but it appears that [there is](https://github.com/duckdb/duckdb_iceberg/blob/5a6b6e738f2009a09b92d948595a160955b826af/src/common/schema.cpp#L76-L77).

jhump:
I don't think this change broke any tests. At the least, running `make` succeeds with this change. Maybe there is a particular target I should be running to run the tests?

If there are in fact no tests covering this, any recommendation on where I should add them? I've only worked in a few C++ repos (I am much more at home in Go and Java repos), so if there are any patterns or advice on writing unit tests, that would be great. But even just pointing me in the direction of similar-ish tests, so I know what new tests should roughly look like, would be great.

Fixes #47.

### DIFF
--- a/src/common/schema.cpp
+++ b/src/common/schema.cpp
@@ -6,7 +6,7 @@ namespace duckdb {
 // https://iceberg.apache.org/spec/#schemas
 
 // forward declaration
-static LogicalType ParseType(yyjson_val *type);
+static LogicalType ParseTypeValue(yyjson_val *type);
 
 static LogicalType ParseStruct(yyjson_val *struct_type) {
 	D_ASSERT(yyjson_get_type(struct_type) == YYJSON_TYPE_OBJ);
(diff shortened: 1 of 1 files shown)


samansmink:
LGTM, thanks!

PR #91, opened 2024-12-28
Check json parse nullptr and fix segfault

dentiny:
Attempt to fix issue https://github.com/duckdb/duckdb-iceberg/issues/88

From gdb stacktrace (attached below), I found the issue is we don't check return value, which is nullptr in certain error cases.

Problem reproduction:
```
D SELECT COUNT(a) FROM iceberg_scan('/home/ubuntu/table.json') AS (a INT);
Segmentation fault (core dumped)
```

After the fix, duckdb output becomes
```
D SELECT COUNT(a) FROM iceberg_scan('/home/ubuntu/table.json') AS (a INT);
Invalid Input Error: Fails to parse iceberg metadata from /home/ubuntu/table.json
```

gdb stacktrace:
```
(gdb) bt
#0  duckdb::IcebergSnapshot::GetParseInfo (metadata_json=...) at /home/ubuntu/duckdb-iceberg/src/common/iceberg.cpp:95
#1  0x0000555557043e92 in duckdb::IcebergSnapshot::GetParseInfo (path="/home/ubuntu/table.json", fs=..., metadata_compression_codec=...) at /home/ubuntu/duckdb-iceberg/src/common/iceberg.cpp:124
#2  0x00005555570440ff in duckdb::IcebergSnapshot::GetLatestSnapshot (path="/home/ubuntu/table.json", fs=..., metadata_compression_codec="none", skip_schema_inference=skip_schema_inference@entry=false)
    at /home/ubuntu/duckdb-iceberg/src/common/iceberg.cpp:135
#3  0x0000555557030c27 in duckdb::IcebergScanBindReplace (context=..., input=...) at /home/ubuntu/duckdb-iceberg/src/iceberg_functions/iceberg_scan.cpp:263
<omit others>
``` 

According to https://github.com/duckdb/duckdb/blob/19864453f7d0ed095256d848b46e7b8630989bac/third_party/yyjson/include/yyjson.hpp#L937, nullptr is returned when an error is met.

Test command:
```
[~/duckdb_iceberg] (hjiang/check-json-parse-nullptr) 
ubuntu@hjiang-devbox-pg$ ./build/release/test/unittest /home/ubuntu/duckdb_iceberg/test/sql/iceberg_scan_generated_data_0_001.test
```


teaguesterling:
I was able to build and test this as expected.

samansmink:
LGTM, thanks 🫶

PR #95, opened 2025-01-15
Add Iceberg catalog support

ediril:
This PR adds support for attaching an Iceberg catalog and be able to read iceberg tables from a datalake. It specifically supports the following sql commands:

See README-ADDENDUM.md for added features and roadmap.


samansmink:
The main parts of this has been merged a part of https://github.com/duckdb/duckdb-iceberg/pull/98. Thanks again @ediril for the work!

ediril:
@samansmink Thank you! How do we proceed with the unmerged parts, which include Create/Drop Schema, Create/Drop Table and Create Table As?

buremba:
Thanks both! I'm also very interested in write support @ediril originally implemented. Is there any plan to create another PR for the remaining changes?

ediril:
@buremba We had an internal discussion about the "write" changes. It is not done in a way that jives with the duckdb philosophy so it will take a while for a re-write of it to go into the official extension. In the meantime, if you don't care about the specific implementation, you could use it via our fork: https://github.com/fivetran/duckdb-iceberg/releases. It is "unofficial" so you'll need to follow instructions here: https://github.com/fivetran/duckdb-iceberg/blob/iceberg-catalog/README-ADDENDUM.md#installation

NatElkins:
@ediril Out of curiosity, can you explain why the write parts of the PR are incompatible with the DuckDB way of doing things?

ediril:
@NatElkins There is no C++ implementation of iceberg spec, only python and Java unfortunately. I didn't have enough time for this project to create the C++ implementation but wanted to figure out the bits in the duckdb code to fill in to make it work. So I resorted to what you see

dentiny:
> There is no C++ implementation of iceberg spec

Is this the iceberg C++ implementation https://github.com/apache/iceberg-cpp?

ediril:
@dentiny that repo did exist but it barely had any code in it back when I was working on this PR. I see there has been a lot more activity in there recently, which is great.

PR #97, opened 2025-01-28
feat: add tests showing the result of iceberg conversion to DuckDB

EpsilonPrime:


EpsilonPrime:
@samansmink Any thoughts on this PR?  Is it worth codifying the behavior of the iceberg conversion to DuckDB?

samansmink:
yea looks good! I've started CI

corleyma:
Semi-related question from a lurker on this project: does this test suite imply the extension does not support reading of Iceberg struct and array types?  It seems like they should map rather straight-forwardly to Duckb's struct and list types, respectively.

EpsilonPrime:
They should but they weren't available in the table.  I could construct another dataset with struct and lists in them.

EpsilonPrime:
rebased

PR #101, opened 2025-02-14
Update the `iceberg_scan` internals to make use of the MultiFileReader API

Tishj:
These changes should enable future optimizations that would not be possible (or very difficult/limited) with the old bind_replace based method.

The `duckdb` submodule has been updated to point to `v1.2-histrionicus`

One [change](https://github.com/duckdb/duckdb-iceberg/commit/2fba66608ef043f4692a871cac318a70bb47820a) is made that fixes an issue on `main` (git hash 6571dca3c05e7cb4b047fc81175580fff392c86a) related to `reader_data.column_indexes`

That is also relevant for the `duckdb_delta` extension.

Thanks to @Tmonster for working on the testing improvements, wouldn't have found a lot of the bugs that crept in with this rework without that work 🚀


test/sql/local/iceberg_scans/iceberg_scan.test
```diff
@@ -42,22 +41,22 @@ SELECT count(*) FROM ICEBERG_SCAN('data/persistent/iceberg/lineitem_iceberg', ve
 #   1 = 2023-02-15 15:07:54.504
 #	2 = 2023-02-15 15:08:14.73
 query I
-SELECT count(*) FROM ICEBERG_SCAN('data/persistent/iceberg/lineitem_iceberg', '2023-02-15 15:07:54.504'::TIMESTAMP, ALLOW_MOVED_PATHS=TRUE);
+SELECT count(*) FROM ICEBERG_SCAN('data/persistent/iceberg/lineitem_iceberg', snapshot_from_timestamp='2023-02-15 15:07:54.504'::TIMESTAMP, ALLOW_MOVED_PATHS=TRUE);
```
samansmink, line 44: This could potentially break things for people, we should think about whether thats ok

src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -0,0 +1,555 @@
...
+		names.push_back(schema_entry.name);
+		return_types.push_back(schema_entry.type);
+	}
+}
+
+unique_ptr<MultiFileList> IcebergMultiFileList::ComplexFilterPushdown(ClientContext &context,
+                                                                      const MultiFileReaderOptions &options,
+                                                                      MultiFilePushdownInfo &info,
+                                                                      vector<unique_ptr<Expression>> &filters) {
+	FilterCombiner combiner(context);
```
samansmink: Can we just return nullptr here instead if filter pushdown is not yet implemented? Or just not override this method at all? the base class method already returns nullptr

src/common/iceberg.cpp
```diff
@@ -10,26 +10,42 @@
...
+	unique_ptr<ManifestReader> manifest_reader;
+	unique_ptr<ManifestEntryReader> manifest_entry_reader;
+	if (snapshot.iceberg_format_version == 1) {
+		manifest_entry_reader = make_uniq<ManifestEntryReaderV1>(iceberg_path, snapshot.manifest_list, fs, options);
+		manifest_reader = make_uniq<ManifestReaderV1>(iceberg_path, snapshot.manifest_list, fs, options);
+	} else if (snapshot.iceberg_format_version == 2) {
+		manifest_entry_reader = make_uniq<ManifestEntryReaderV2>(iceberg_path, snapshot.manifest_list, fs, options);
+		manifest_reader = make_uniq<ManifestReaderV2>(iceberg_path, snapshot.manifest_list, fs, options);
+	} else {
+		throw InvalidInputException("TODO");
```
samansmink: nit: error msg

src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -0,0 +1,555 @@
...
+}
+
+unique_ptr<NodeStatistics> IcebergMultiFileList::GetCardinality(ClientContext &context) {
+	auto total_file_count = IcebergMultiFileList::GetTotalFileCount();
+
+	if (total_file_count == 0) {
+		return make_uniq<NodeStatistics>(0, 0);
+	}
+
+	// FIXME: visit metadata to get a cardinality count
```
samansmink, line 80: Maybe we can check if we can get this one in asap, good cardinality estimates do make a big difference

duckdb
samansmink, line 1: This should probably be v1.2.0

src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -0,0 +1,555 @@
...
+			vec.Flatten(count);
+		}
+
+		auto names = FlatVector::GetData<string_t>(result.data[0]);
+		auto row_ids = FlatVector::GetData<int64_t>(result.data[1]);
+		for (idx_t i = 0; i < count; i++) {
+			auto &name = names[i];
+			auto &row_id = row_ids[i];
+
+			auto &deletes = delete_data[name.GetString()];
```
samansmink: The files are guaranteed to be ordered, we don't have to look up the filename for every row here.

Instead we can just store an iterator into the delete data and check if the file is still the same as the last file to update the iterator.

src/iceberg_functions/iceberg_multi_file_reader.cpp
Tishj: I agree, the `added_rows_count` and `existing_rows_count` contain this information per manifest, which we can get by just scanning the manifest list.
Optional in v1 but required by >=v2

Tishj: Would like to do that after <https://github.com/duckdb/duckdb-iceberg/pull/110> lands and we integrate those changes in to this PR.

I don't really want to figure out how to extend the avro-cpp reading logic to read this information, especially if we're going to replace it anyways

samansmink (approved):
great work @Tishj and @Tmonster, byebye bind_replace 👋

PR #105, opened 2025-02-17
Replace internal table secret on conflict

danklynn:
I was trying the latest Iceberg REST Catalog support and ran into an issue where it was attempting to recreate the secret every time you run a SELECT statement and throwing due to the conflict setting.

This PR is less than ideal--the right solution would be to cache table info and vended credentials to avoid extra API calls--but it allows repeated queries against the same table.


Tmonster (approved):
Thanks! I've run into this issue as well, pretty sure this is the correct fix

PR #106, opened 2025-02-19
Cleaner Data Generation

Tmonster:
Generate data in a Generate Data in a cleaner way. Make generating new data simple and straightforward.


Makefile
```diff
@@ -12,11 +12,11 @@ include extension-ci-tools/makefiles/duckdb_extension.Makefile
...
+	mkdir -p data_generated/data
+	python3 scripts/data_generators/generate_data.py
 
 data_large: data data_clean
-	python3 scripts/test_data_generator/generate_iceberg.py 1 data/iceberg/generated_spec2_1 2
+	python3 scripts/data_generators/generate_data.py
 
 data_clean:
-	rm -rf data/iceberg/generated_*
+	rm -rf data_generated/
```
samansmink: nit: I think it's a bit cleaner to have the generated data live in `data/generated`

edit: nvm i see this is already done: then i think this is just a leftover

scripts/data_generators/tmp_data/tmp.parquet
samansmink, line 1: Can we remove this?

scripts/start-rest-catalog.sh
```diff
@@ -1,38 +1,22 @@
...
-UNPARTITIONED_TABLE_PATH=$(curl -s http://127.0.0.1:8181/v1/namespaces/default/tables/table_unpartitioned | jq -r '."metadata-location"')
+if test ! -f "./scripts/docker-compose.yml"
+then
+  # in CI
+  echo "Please run from duckdb root."
+  exit 1
+fi
 
-SQL=$(cat <<-END
+mkdir -p data/generated/iceberg/spark-rest
```
samansmink, line 9: nit: I think this can be cleaned up

test/sql/local/iceberg_scan_generated_data_1.test_slow
```diff
@@ -9,33 +9,35 @@ require iceberg
 require-env DUCKDB_ICEBERG_HAVE_TEST_DATA
```
samansmink: nit: can we name this `DUCKDB_ICEBERG_HAVE_GENERATED_DATA`

.github/workflows/Rest.yml
```diff
@@ -47,12 +51,16 @@ jobs:
...
 
       - name: Start Rest Catalog
-        working-directory: scripts/
         run: |
-          ./start-rest-catalog.sh
+          ./scripts/start-rest-catalog.sh
+
+      - name: Generate data
+        run: |
+          python3 scripts/data_generators/generate_data.py
```
samansmink: nit: can this use the makefile target instead?

.github/workflows/Rest.yml
```diff
@@ -47,12 +51,16 @@ jobs:
           make release
 
       - name: Start Rest Catalog
-        working-directory: scripts/
         run: |
-          ./start-rest-catalog.sh
+          ./scripts/start-rest-catalog.sh
```
samansmink: Can we maybe add a makefile target for starting the rest catalog too?

I think it's nice if a dev only needs to look at the makefile to understand what to run

samansmink: While doing this, I think its nice to make clear that the requirements to run `make data` are installed in the start-rest-catalog.sh script. Perhaps we can also install the dependencies in a makefile target separately

PR #107, opened 2025-02-21
Run TPCH on local s3 IRC

Tmonster:
Mostly to test that we produce the correct results when reading iceberg tables. 
Future work will/would be
- Benchmarking
- Running actual tpch power tests where deletes and inserts take place. This will better test our insert and delete functionality.


samansmink (approved):

samansmink:
lookin great, thanks!

PR #109, opened 2025-02-24
Tests to verify reading from tables with deletes

Tmonster:
Adding more tests and generated tables to verify that we can read from iceberg tables that have deletes.


Tmonster:
tests are now in 
https://github.com/duckdb/duckdb-iceberg/pull/101

PR #110, opened 2025-02-27
Replace `avro-cpp` with `duckdb_avro`

Tishj:
The ReadManifestList and ReadManifestEntries methods now start manual scans using `read_avro`

The bind data (return names+types) are checked for correctness using the iceberg format version.

We then read through the produced chunks to parse the relevant information from them.

This currently reads the entire avro file, with no projection pushdown.


extension_config.cmake
```diff
@@ -6,4 +6,10 @@ duckdb_extension_load(iceberg
     LOAD_TESTS
 )
 
-duckdb_extension_load(tpch)
\ No newline at end of file
+duckdb_extension_load(avro
+    LOAD_TESTS
+    GIT_URL https://github.com/tishj/duckdb_avro
```
Tishj: Will change later, needed this branch (PR is submitted upstream) to detect the vcpkg-installed avro header file path

src/common/iceberg.cpp
```diff
@@ -3,90 +3,301 @@
...
 	IcebergTable ret;
 	ret.path = iceberg_path;
 	ret.snapshot = snapshot;
 
+	auto &fs = FileSystem::GetFileSystem(context);
 	auto manifest_list_full_path = allow_moved_paths
 	                                   ? IcebergUtils::GetFullPath(iceberg_path, snapshot.manifest_list, fs)
 	                                   : snapshot.manifest_list;
-	auto manifests = ReadManifestListFile(manifest_list_full_path, fs, snapshot.iceberg_format_version);
+	auto manifests = ReadManifestListFile(context, manifest_list_full_path, snapshot.iceberg_format_version);
```
Tishj: Note: we need the ClientContext to set up the `read_avro` scans

src/common/iceberg.cpp
```diff
@@ -3,90 +3,301 @@
...
+	if (!name_to_vec.count("record_count")) {
+		return false;
+	}
+	if (iceberg_format_version >= 2 && !name_to_vec.count("content")) {
+		return false;
+	}
+	return true;
+}
+
+vector<IcebergManifestEntry> IcebergTable::ReadManifestEntries(ClientContext &context, const string &path,
```
Tishj: FIXME: we can probably create a single scan for all manifest entries, which can be executed lazily

vcpkg_ports/avro.patch
```diff
@@ -0,0 +1,377 @@
+diff --git a/lang/c/CMakeLists.txt b/lang/c/CMakeLists.txt
```
Tishj, line 1: Note: this is copied from `duckdb_avro`, without this we get errors about `avro_reader_reader` because the patch doesn't get applied to `avro-c` even though `duckdb_avro` is being built as a dependency

I imagine the automatic merging of `vcpkg.json` doesn't take `ports` into account

vcpkg_ports/vcpkg.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "avro-c",
```
Tishj, line 2: Copied from `duckdb_avro` see the comment on `avro.patch` for context

src/common/iceberg.cpp
```diff
@@ -3,90 +3,301 @@
...
+	if (iceberg_format_version >= 2 && !name_to_vec.count("sequence_number")) {
+		return false;
+	}
+	if (iceberg_format_version >= 2 && !name_to_vec.count("content")) {
+		return false;
+	}
+	return true;
+}
+
+vector<IcebergManifest> IcebergTable::ReadManifestListFile(ClientContext &context, const string &path, idx_t iceberg_format_version) {
```
Tishj: This function along with ReadManifestEntries can probably be deduplicated, using a functor to inject the verification check and the chunk-parsing logic

Fokko:
FWIW, @wgtmac is making the [necessary changes in Avro](https://github.com/apache/avro/pulls/wgtmac) to make it work for Iceberg (fetching the field-IDs and such). This is part of the new [iceberg-cpp](https://github.com/apache/iceberg-cpp) library.

src/common/iceberg.cpp
```diff
@@ -3,90 +3,301 @@
...
+	std::function<void(DataChunk &input, case_insensitive_map_t<ColumnIndex> &name_to_vec, vector<IcebergManifest> &result)> produce_manifests;
 	if (iceberg_format_version == 1) {
-		schema = avro::compileJsonSchemaFromString(MANIFEST_SCHEMA_V1);
-		avro::DataFileReader<c::manifest_file_v1> dfr(std::move(stream), schema);
-		c::manifest_file_v1 manifest_list;
-		while (dfr.read(manifest_list)) {
-			ret.emplace_back(IcebergManifest(manifest_list));
-		}
+		produce_manifests = [](DataChunk &input, case_insensitive_map_t<ColumnIndex> &name_to_vec, vector<IcebergManifest> &result) {
+			auto manifest_paths = FlatVector::GetData<string_t>(input.data[0]);
```
Tishj: This is wrong, not using the `name_to_vec`

wgtmac:
@Fokko Thanks for pinging me! IIRC, DuckDB leverages the Apache Avro C library so my effort may not be applicable here.

samansmink:
@wgtmac @Fokko correct! that uses the avro c library. 

> (fetching the field-IDs and such

Are these things we're going to run into if we leverage the C avro library in our avro extension, or is that already supported there?

src/iceberg_manifest.cpp
```diff
@@ -0,0 +1,178 @@
...
+		entry.content = IcebergManifestEntryContentType::DATA;
+		entry.file_path = file_path[i].GetString();
+		entry.file_format = file_format[i].GetString();
+		entry.record_count = record_count[i];
+
+		result.push_back(entry);
+	}
+}
+
+bool IcebergManifestEntryV1::VerifySchema(const case_insensitive_map_t<ColumnIndex> &name_to_vec) {
```
Tishj, line 133: This checks the existence of the column in the metadata, but it doesn't (and can't) check what the type of it is, maybe that's a problem..

src/common/iceberg.cpp
Tishj: It says outdated because I refactored this method to deduplicate the `read_avro` scan boilerplate

wgtmac:
@samansmink Sorry I'm not familiar with the C library.

Tishj:
> @samansmink Sorry I'm not familiar with the C library.

That's fine, can you shed some light on why custom attributes are required for iceberg?

wgtmac:
See: https://iceberg.apache.org/spec/#avro

- `"field-id"` is required for every column. Same for `"key-id"`, `"value-id"` and `"element-id"`. This is used by both data file and manifest file.
- `"logicalType": "map"`is required for map type which is actually an Avro `array<record<key,value>>` type, especially when the key type is not a string.

Tishj:
Aha so that's the `100` for the `file_path` in the `data_file` for example?
<https://iceberg.apache.org/spec/#manifests>

Or I guess more broadly this section, describing the Avro format requirements: <https://iceberg.apache.org/spec/#appendix-a-format-specific-requirements>, the "Field IDs" section?

Tishj:
> When writing data to Avro files, these IDs must be stored in the Avro schema to support ID-based column pruning.

It seems to me that this is not relevant for us, we don't support Avro as data files currently, only parquet
For the metadata it looks to me like these attributes aren't relevant

wgtmac:
The schema of a manifest file (which is an Avro file) is a struct called `manifest_entry` with the following fields:

| v1         | v2         | Field id, name                | Type                                                      | Description |
| ---------- | ---------- |-------------------------------|-----------------------------------------------------------|-------------|
| _required_ | _required_ | **`0  status`**               | `int` with meaning: `0: EXISTING` `1: ADDED` `2: DELETED` | Used to track additions and deletions. Deletes are informational only and not used in scans. |
| _required_ | _optional_ | **`1  snapshot_id`**          | `long`                                                    | Snapshot id where the file was added, or deleted if status is 2. Inherited when null. |
|            | _optional_ | **`3  sequence_number`**      | `long`                                                    | Data sequence number of the file. Inherited when null and status is 1 (added). |
|            | _optional_ | **`4  file_sequence_number`** | `long`                                                    | File sequence number indicating when the file was added. Inherited when null and status is 1 (added). |
| _required_ | _required_ | **`2  data_file`**            | `data_file` `struct` (see below)                          | File path, partition tuple, metrics, ... |

See https://iceberg.apache.org/spec/#manifests.

Tishj:
So you're saying the `Field id` portion of the spec is what's written in the custom attributes?
For example the `0` for the `status` field is written there?

That would be relevant for projection pushdown into the avro scan, but apart from that I don't see any hurdles in implementing pushdown into the iceberg scan by not supporting custom attributes

wgtmac:
> So you're saying the Field id portion of the spec is what's written in the custom attributes?
For example the 0 for the status field is written there?

Yes, the only way to read and write `field-id` is via custom attributes. (Please correct me if I was wrong, @Fokko)

> That would be relevant for projection pushdown into the avro scan

Doesn't it affect column projection for manifest file? Though we can ignore field id by just using the field name, it depends on whether the field name is reliable.

Tishj:
There is no projection pushdown in read_avro (yet(?)). Mainly because it won't result in any significant performance gains since the format is row-based, the data would still need to be read.
It could help reduce some of the mapping and lookups we have to do now, but that's something we can look into in the future.

I am relying on the names of the columns, given that these are mentioned alongside the field-ids in the spec I would be very confused if these weren't binding.

If they aren't then we can look into using the field-ids, similar to how we handle them in parquet

samansmink:
The avro extension is now available from core, see https://github.com/duckdb/duckdb/pull/16708.

I *think* we can already make this PR happen cleanly without breaking peoples workflows by explicitly installing and loading it. 

Alternatively, we can wait for v1.2.2 where avro will be autoloadable

Tishj:
> I think we can already make this PR happen cleanly without breaking peoples workflows by explicitly installing and loading it.

I'll leave that decision to you, but it sounds reasonable

Tishj:
I have a branch where I integrated the `multi_file_reader` with this branch - I was expecting these merge conflicts already

I'll merge main into that, pray for no further conflicts and merge that branch into this 👍

extension_config.cmake
Tishj: Changed to `duckdb/duckdb_avro`

PR #117, opened 2025-03-11
add logging to iceberg catalog api requests

samansmink:
add some logging to the IRC requests


Tmonster (commented):
Do we not want to use `enable_http_logging=true` anymore?
https://duckdb.org/docs/stable/configuration/overview.html

samansmink:
> Do we not want to use `enable_http_logging=true` anymore? https://duckdb.org/docs/stable/configuration/overview.html

I think that can be deprecated now that we have generic logging infrastructure

Tmonster (approved):
LGTM :+1:

PR #120, opened 2025-03-12
Do not use "null" for session key

Tmonster:
If there was no session key, it would be NULL, which is converted to the string "Null", which is not a valid session token. 
So if there is no session token, we now pass "".

This also fixes some issues where two threads were doing the same work.


Tmonster:
Hmm, there may be an issue here.

Tmonster:
nvm I think my local environment was borked

PR #124, opened 2025-03-14
feat: Add support to other Glue catalog ID styles

fabito:
See:  https://docs.aws.amazon.com/glue/latest/dg/connect-glu-iceberg-rest.html#prefix-catalog-path-parameters

### Manually tested with:

```sql
-- Default catalog
ATTACH ':' AS my_catalog (TYPE ICEBERG, ENDPOINT_TYPE 'GLUE');

-- Default catalog in a specific account
ATTACH '111122223333' AS my_catalog2 (TYPE ICEBERG, ENDPOINT_TYPE 'GLUE');
```


Tmonster:
Hi @fabito,

This looks great and is very much on the right track. Reading the glue documentation a bit more, it seems like Glue Catalogs can also be backed by redshift data. Currently DuckDB can't read redshift data, so I worry this will cause some undefined behavior if someone attempts to attach a redshift backed catalog.

That being said, if there are Glue catalogs with these catalog path parameters, then we should be able to connect with them. I'm going to try and create a redshift backed glue catalog to see what our behavior is, and I may recommend some patches here. Do you maybe have access to such a catalog?

fabito:
Hi @Tmonster ,

Good point, I didn't know about the redshift backed catalogs. We should definitely refactor the implementation to "reject" them until they are supported by the extension.

> Do you maybe have access to such a catalog?

No, sorry. We don't use Redshift

goober:
FYI. I have tested this PR and it works fine for our Athena created iceberg tables with both of the alternative attaching method that is mention above. We are using regular s3 buckets under the hood.

src/iceberg_extension.cpp
```diff
@@ -61,18 +62,23 @@ static void SetCatalogSecretParameters(CreateSecretFunction &function) {
...
-	if (bucket_sep_correct) {
-		return true;
-	}
-	throw IOException("Invalid Glue Catalog Format: '" + warehouse + "'. Expected '<account_id>:s3tablescatalog/<bucket>");
+    // See: https://docs.aws.amazon.com/glue/latest/dg/connect-glu-iceberg-rest.html#prefix-catalog-path-parameters
+
+	const std::regex patterns[] = {
+        std::regex("^:$"),                   // Default catalog ":" in current account
+        std::regex("^\\d{12}$"),             // Default catalog in a specific account
+        std::regex("^\\d{12}:[^:/]+$"),      // Specific catalog in a specific account
```
Tishj: This one doesn't seem to be listed in the docs ?
If we have `accountID:catalog`, don't we also need `catalog` (`^[^:/]+$`) to access a specific catalog in the current account ?

Ah actually never mind, found this <https://docs.aws.amazon.com/glue/latest/dg/security_iam_id-based-policy-examples.html#security_iam_id-based-policy-examples-deny-by-name-prefix>

```json
            "Resource": [
                "arn:aws:glue:us-west-2:123456789012:catalog",
                "arn:aws:glue:us-west-2:123456789012:database/dev-*",
                "arn:aws:glue:us-west-2:123456789012:database/prod-*",
                "arn:aws:glue:us-west-2:123456789012:table/dev-*/*",
                "arn:aws:glue:us-west-2:123456789012:table/*/dev-*",
                "arn:aws:glue:us-west-2:123456789012:table/prod-*/*",
                "arn:aws:glue:us-west-2:123456789012:table/*/prod-*",
                "arn:aws:glue:us-west-2:123456789012:userDefinedFunction/dev-*/*",
                "arn:aws:glue:us-west-2:123456789012:userDefinedFunction/*/dev-*",
                "arn:aws:glue:us-west-2:123456789012:userDefinedFunction/prod-*/*",
                "arn:aws:glue:us-west-2:123456789012:userDefinedFunction/*/prod-*",
                "arn:aws:glue:us-west-2:123456789012:connection/dev-*",
                "arn:aws:glue:us-west-2:123456789012:connection/prod-*"
            ]
```

Which does use `accountID:catalog` 👍 
Do you know if `catalog` (which would logically access a specific catalog in the current account) is a valid way of accessing, and should thus be supported ?

Tishj:
A note for everyone that is eager for this to be merged, there is a workaround available in 1.2.2 (released yesterday, might need `FORCE INSTALL ICEBERG;` to update the installed version), replacing the `ENDPOINT_TYPE glue` with what it expands to:

endpoint_type glue:
```sql
statement ok
attach '840140254803:s3tablescatalog/pyiceberg-blog-bucket' as my_datalake (
    TYPE ICEBERG,
    ENDPOINT_TYPE 'GLUE'
);
```

becomes:
```sql
attach '840140254803:s3tablescatalog/pyiceberg-blog-bucket' as my_datalake (
	TYPE ICEBERG,
	AUTHORIZATION_TYPE 'sigv4',
	ENDPOINT 'glue.us-east-1.amazonaws.com/iceberg'
);
```

joaocrebelo:
fyi - I used this workaround to register a regular Glue Catalog and it worked. But ran into the following two issues:

- Reading from some tables gave error `Invalid field found while parsing field: type` , which I think relates to this issue https://github.com/duckdb/duckdb-iceberg/issues/47. I'm not sure the workaround suggested there applies.
- Running "show tables;" on a database with just a few tables takes around 1min.

Tishj:
> Invalid field found while parsing field: type
This is fixed by #207 as mentioned there

PR #130, opened 2025-03-25
Feat: query the config endpoint when attaching the catalog. And add APIUtils for requests

Tmonster:
Currently we infer the prefix for a catalog as the warehouse value. This was a temporary stop-gap to get things working, but according to the Iceberg rest specification, we should be hitting the config endpoint first "to get catalog configuration properties from the server to configure the catalog and its HTTP client"

[see the spec here](https://raw.githubusercontent.com/apache/iceberg/main/open-api/rest-catalog-open-api.yaml).

Now, when we attach, we create the catalog and call the config endpoint. 

I have also created an APIUtils class that should contain the GET/POST request code since it was exploding the catalog_api.cpp file.

This also fixes bugs with AWS glue which were the result of not using the prefix provided by hitting the config endpoint.


Tishj:
I have a commit touching similar paths and making overlapping changes 😅 
https://github.com/Tishj/duckdb_iceberg/commit/406ec2e972db929207e8f065ffb414c103d1b9ef

Tmonster:
Hmm, alright. Looking at it, the conflicts don't seem too bad. I feel like maybe merging this and then resolving conflicts won't be too bad?

kevinjqliu:
If youre looking for ways to test integrations with the rest catalog the [`iceberg-rest-fixture`](https://hub.docker.com/r/apache/iceberg-rest-fixture) docker image is the "official" image built from the https://github.com/apache/iceberg repo

We use this for various sub-projects such as [pyiceberg](https://github.com/apache/iceberg-python/blob/278f7643cd62f9e14496177632cb48d9b52e553d/dev/docker-compose-integration.yml#L43)

Tmonster:
@kevinjqliu Thanks! We will definitely add it to our test suite 👍

Issue #108, opened 2025-02-22
Querying empty table fails with `No snapshots found`


buremba:
I have an empty table where the `metadata.json` is as follows:

```
{
  "format-version" : 2,
  "table-uuid" : "06018e7d-c2e2-4446-afce-37b26cd6fa8b",
  "location" : "s3://universql-test",
  "last-sequence-number" : 0,
  "last-updated-ms" : 1740192346580,
  "last-column-id" : 15,
  "current-schema-id" : 0,
  "schemas" : [ {
    "type" : "struct",
    "schema-id" : 0,
    "fields" : [ {
      "id" : 1,
      "name" : "PTS",
      "required" : true,
      "type" : "timestamp"
    }]
  } ],
  "default-spec-id" : 0,
  "partition-specs" : [ {
    "spec-id" : 0,
    "fields" : [ ]
  } ],
  "last-partition-id" : 999,
  "default-sort-order-id" : 0,
  "sort-orders" : [ {
    "order-id" : 0,
    "fields" : [ ]
  } ],
  "properties" : {
    "format-version" : "2"
  },
  "current-snapshot-id" : -1,
  "snapshots" : [ ],
  "snapshot-log" : [ ]
}
```

When I try to query the file as follows:

```
CREATE OR REPLACE VIEW universql_demo.table_name AS SELECT * FROM ICEBERG_SCAN('s3://.../metadata.json')
```

The query fails with `IO Error: No snapshots found`. Ideally it would be great to return empty table as as per Iceberg spec, snapshots can be added later on.

buremba:
Thank you @Tmonster !

PR #131, opened 2025-03-26
BugFix: Allow read from empty tables

Tmonster:
fixes https://github.com/duckdb/duckdb-iceberg/issues/108

If there are no snapshots, then we are reading from an empty table. Early out from the `InitializeFiles` call so that we can populate the schema and just return.

### DIFF
--- a/src/iceberg_functions/iceberg_multi_file_reader.cpp
+++ b/src/iceberg_functions/iceberg_multi_file_reader.cpp
@@ -191,16 +191,18 @@ void IcebergMultiFileList::InitializeFiles() {
 		throw InvalidInputException("Reading from Iceberg version %d is not supported yet", snapshot.iceberg_format_version);
 	}
 
+	if (snapshot.snapshot_id == DConstants::INVALID_INDEX) {
+		// we are in an empty table
+		current_data_manifest = data_manifests.begin();
+		current_delete_manifest = delete_manifests.begin();
+		return;
+	}
+
 	// Read the manifest list, we need all the manifests to determine if we've seen all deletes
 	auto manifest_list_full_path = options.allow_moved_paths
 	                                   ? IcebergUtils::GetFullPath(iceberg_path, snapshot.manifest_list, fs)
 	                                   : snapshot.manifest_list;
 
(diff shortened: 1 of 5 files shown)


src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -195,6 +195,12 @@ void IcebergMultiFileList::InitializeFiles() {
 	auto manifest_list_full_path = options.allow_moved_paths
 	                                   ? IcebergUtils::GetFullPath(iceberg_path, snapshot.manifest_list, fs)
 	                                   : snapshot.manifest_list;
+
+	if (manifest_list_full_path.empty()) {
```
Tishj: I think we can check `snapshot.manifest_list.empty()` earlier (line 149), we don't need a manifest reader, manifest producer, entry producer etc.. when we're reading an empty table

Tishj: I also think this is leaving `current_data_manifest` and `current_delete_manifest` uninitialized, but they are being compared to `data_manifests.end()` and `delete_manifests.end()` respectively, which is undefined behavior (works on mac, won't work on linux)

Tishj: I think we want to set `snapshot_id` to `DConstants::INVALID_INDEX` and check for that to see if no snapshot could be found, before we do comparisons against uninitialized values (or make sure that these initialization steps do still run, but that's less future-proof I feel)

```c++
string IcebergMultiFileList::GetFile(idx_t file_id) {
	if (!initialized) {
		InitializeFiles();
	}

	if (snapshot.snapshot_id == DConstants::INVALID_INDEX) {
		//! No snapshots found, empty scan
		return string();
	}
```

Tishj (changes requested):
Thanks, I think this makes sense, mentioned one point about comparing uninitialized values that should be addressed, other than that LGTM 👍

Tishj (approved):
LGTM

Tmonster:
Thanks!

PR #132, opened 2025-03-27
Add support for the optional `storage-credentials` of the LoadTableResult and explicit scope/oauth2_server_uri

Tishj:
This PR adds scoped secrets created from the `storage-credentials`.
It also adds the ability to provide a `OAUTH2_SCOPE` to the attach options, this is used for the oauth2 endpoint.
It defaults to the `PRINCIPAL_ROLE:ALL` scope which is the default for Polaris.

Along with the `SCOPE` parameter, it also makes the `OAUTH2_SERVER_URI` option available, now that the Iceberg REST Catalog API is moving away from the `/v1/oauth/tokens` endpoint (https://github.com/apache/iceberg/issues/10537)

These changes are tested locally against Lakekeeper, I'll add a CI run for this after we have synced up and made sure this doesn't break the codepaths for the existing supported catalogs.


src/catalog_api.cpp
```diff
@@ -75,30 +72,130 @@ static IRCAPIColumnDefinition ParseColumnDefinition(yyjson_val *column_def) {
...
-		result.session_token = IcebergUtils::TryGetStrFromObject(warehouse_credentials, "s3.session-token", false);
-		if (catalog_credentials) {
-       		auto kv_secret = dynamic_cast<const KeyValueSecret &>(*catalog_credentials->secret);
-			auto region = kv_secret.TryGetValue("region").ToString();
-			result.region = region;
+
+	// Mapping from config key to a duckdb secret option
+
+	case_insensitive_map_t<Value> config_options;
+	if (catalog_credentials) {
```
Tishj: Should we restore the old behavior where we only check these if the `config` was set in the result?

src/catalog_api.cpp
```diff
@@ -75,30 +72,130 @@ static IRCAPIColumnDefinition ParseColumnDefinition(yyjson_val *column_def) {
...
+				throw InternalException("required property 'prefix' is missing from the StorageCredential schema");
+			}
+
+			CreateSecretInfo create_secret_info(OnCreateConflict::REPLACE_ON_CONFLICT, SecretPersistType::TEMPORARY);
+			auto prefix_string = yyjson_get_str(sc_prefix);
+			if (!prefix_string) {
+				throw InternalException("property 'prefix' of StorageCredential is NULL");
+			}
+			create_secret_info.scope.push_back(string(prefix_string));
+			create_secret_info.type = "s3";
```
Tishj, line 163: This might need a name, because that's probably used for the conflict resolution.
if we have multiple "unnamed" secrets, they replace eachother

src/storage/irc_table_entry.cpp
```diff
@@ -54,67 +53,51 @@ TableFunction ICTableEntry::GetScanFunction(ClientContext &context, unique_ptr<F
...
+		}
+
+		//! Limit the scope to the metadata location if no explicit scope was set
+		if (info.scope.empty()) {
+			std::string lc_storage_location;
+			lc_storage_location.resize(table_data->storage_location.size());
+			std::transform(table_data->storage_location.begin(), table_data->storage_location.end(), lc_storage_location.begin(), ::tolower);
+			size_t metadata_pos = lc_storage_location.find("metadata");
+			if (metadata_pos != std::string::npos) {
+				info.scope = {lc_storage_location.substr(0, metadata_pos)};
```
Tishj: This looks like it will have a bigger scope than the "scoped" secrets we create from the StorageCredential items, which might cause it to override the scoped ones.

```c++
(duckdb::vector<std::string>) {
  std::__1::vector<std::__1::string, std::__1::allocator<std::__1::string> > = size=1 {
    [0] = "s3://examples/initial-warehouse/0195ce0f-4c9e-7612-b69a-566f703d535c/0195ce0f-4fac-7cb1-ad20-45c047a2d2fc/"
  }
}
```

src/catalog_api.cpp
Tishj: We probably should only create a secret out of the config if there were no `storage-credentials` provided.

src/iceberg_extension.cpp
```diff
@@ -100,11 +108,25 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
...
 	}
 	auto warehouse = info.path;
+
+	if (scope.empty()) {
+		//! Default to the Polaris scope: 'PRINCIPAL_ROLE:ALL'
+		scope = "PRINCIPAL_ROLE:ALL";
+	}
+
+	if (oauth2_server_uri.empty()) {
+		//! If no oauth2_server_uri is provided, default to the (deprecated) REST API endpoint for it
```
Tishj, line 120: Maybe log

src/include/catalog_api.hpp
```diff
@@ -41,6 +42,7 @@ struct IRCAPITableCredentials {
 	string secret;
```
Tishj: remove these

src/catalog_api.cpp
```diff
@@ -75,30 +72,130 @@ static IRCAPIColumnDefinition ParseColumnDefinition(yyjson_val *column_def) {
...
+	parameters.push_back(StringUtil::Format("%s=%s", "client_secret", secret));
+	parameters.push_back(StringUtil::Format("%s=%s", "scope", scope));
+
+	string post_data = StringUtil::Format("%s", StringUtil::Join(parameters, "&"));
+	string api_result = APIUtils::PostRequest(context, uri, post_data);
+	//! FIXME: the oauth/tokens endpoint returns, on success;
+	// { 'access_token', 'token_type', 'expires_in', <issued_token_type>, 'refresh_token', 'scope'}
 	std::unique_ptr<yyjson_doc, YyjsonDocDeleter> doc(ICUtils::api_result_to_doc(api_result));
 	auto *root = yyjson_doc_get_root(doc.get());
 	return IcebergUtils::TryGetStrFromObject(root, "access_token");
```
Tishj: Verify that the `token_type` is `bearer`

src/storage/irc_catalog.cpp
```diff
@@ -135,7 +135,6 @@ DatabaseSize IRCatalog::GetDatabaseSize(ClientContext &context) {
 
 IRCEndpointBuilder IRCatalog::GetBaseUrl() const {
 	auto base_url = IRCEndpointBuilder();
-	base_url.SetPrefix(prefix);
 	base_url.SetWarehouse(warehouse);
```
Tishj: check if this can be removed

Tishj: check if this can be removed

src/storage/irc_table_entry.cpp
```diff
@@ -54,67 +53,51 @@ TableFunction ICTableEntry::GetScanFunction(ClientContext &context, unique_ptr<F
...
+			} else {
+				throw std::runtime_error("Substring not found");
+			}
+		}
 
 		if (StringUtil::StartsWith(ic_catalog.host, "glue")) {
```
Tishj, line 76: add `!info.options.count("endpoint")` to this condition

src/storage/irc_table_entry.cpp
```diff
@@ -54,67 +53,51 @@ TableFunction ICTableEntry::GetScanFunction(ClientContext &context, unique_ptr<F
...
-
-		std::string lc_storage_location;
-		lc_storage_location.resize(table_data->storage_location.size());
-		std::transform(table_data->storage_location.begin(), table_data->storage_location.end(), lc_storage_location.begin(), ::tolower);
-		size_t metadata_pos = lc_storage_location.find("metadata");
-		if (metadata_pos != std::string::npos) {
-			info.scope = {lc_storage_location.substr(0, metadata_pos)};
-		} else {
-			throw std::runtime_error("Substring not found");
+		} else if (StringUtil::StartsWith(ic_catalog.host, "s3tables")) {
```
Tishj, line 83: Only do this for the "initial" secret created from the config

Tishj:
> This could cause some confusion with duckdb's preexisting `scope` parameter used in secrets https://duckdb.org/docs/stable/configuration/secrets_manager.html#creating-multiple-secrets-for-the-same-service-type

hmm fair enough, that is the name that iceberg uses though..
maybe AUTH_SCOPE or RBAC_SCOPE could work?

src/storage/irc_table_entry.cpp
Tishj: Solved by only creating this if there are no `storage-credentials` provided (as those will have their own specific scope already)

samansmink:
`AUTH_SCOPE` sounds good to my ears

Tishj:
I figured since it's specific to oauth2 (possibly other authorization methods we would support later don't have a scope) I thought it best to name it `OAUTH2_SCOPE`

Tishj:
@Tmonster https://github.com/duckdb/duckdb-iceberg/actions/runs/14174779634/job/39714472407?pr=132 CI looks broken, but unrelated to this PR, do you agree?

src/catalog_api.cpp
```diff
@@ -71,35 +70,150 @@ static IRCAPIColumnDefinition ParseColumnDefinition(yyjson_val *column_def) {
...
+
+	auto *storage_credentials = yyjson_obj_get(root, "storage-credentials");
+	auto storage_credentials_size = yyjson_arr_size(storage_credentials);
+	if (storage_credentials && storage_credentials_size > 0) {
+		yyjson_val *storage_credential;
+		size_t index, max;
+		yyjson_arr_foreach(storage_credentials, index, max, storage_credential) {
+			auto *sc_prefix = yyjson_obj_get(storage_credential, "prefix");
+			if (!sc_prefix) {
+				throw InternalException("required property 'prefix' is missing from the StorageCredential schema");
```
Tmonster: Should this be an InternalException? I feel like since this is a catalog error, we shouldn't invalidate the whole database

src/storage/irc_table_entry.cpp
```diff
@@ -86,42 +72,37 @@ TableFunction ICTableEntry::GetScanFunction(ClientContext &context, unique_ptr<F
...
-		std::string lc_storage_location;
-		lc_storage_location.resize(table_data->storage_location.size());
-		std::transform(table_data->storage_location.begin(), table_data->storage_location.end(),
-		               lc_storage_location.begin(), ::tolower);
-		size_t metadata_pos = lc_storage_location.find("metadata");
-		if (metadata_pos != std::string::npos) {
-			info.scope = {lc_storage_location.substr(0, metadata_pos)};
-		} else {
-			throw std::runtime_error("Substring not found");
+		if (StringUtil::StartsWith(ic_catalog.host, "glue")) {
```
Tmonster, line 76: nit: You can also check the `catalog_type` if you want. But I think eventually we will be getting rid of that.

src/storage/irc_table_entry.cpp
```diff
@@ -86,42 +72,37 @@ TableFunction ICTableEntry::GetScanFunction(ClientContext &context, unique_ptr<F
...
+			                                      ? ""
+			                                      : kv_secret.TryGetValue("session_token").ToString()},
+			                {"region", region},
+			                {"endpoint", endpoint}};
 		}
 		auto my_secret = secret_manager.CreateSecret(context, info);
 	}
 
+	for (auto &info : table_credentials.storage_credentials) {
+		auto my_secret = secret_manager.CreateSecret(context, info);
```
Tmonster: nit: I think you can remove the `auto my_secret =` if you want (and there are no compilation errors etc.). The variable isn't used.

Tmonster (commented):
Thanks! Just have some small comments, then we can see if CI does pass or not

src/catalog_api.cpp
Tishj: I'd go for IOException, but I don't know what else fits, because this means the REST catalog is broken, that's a required property

Tishj: went with InvalidInput

PR #134, opened 2025-03-31
Replace IOExceptions caused by invalid external input with InvalidInputExceptions

Flogex:
The iceberg extension currently throws `IOExceptions` when it fails to read iceberg data.
This makes it hard to distinguish between errors that are caused by malformed or non-existent input files from iceberg, and system-level I/O failures in DuckDB.
With this PR, an `InvalidInputException` is thrown when invalid external input is read, similar to e.g. the [JSON reader](https://github.com/duckdb/duckdb/blob/6d19e746d302372560397a6f20151cafd817a231/extension/json/json_reader.cpp#L516).


src/common/iceberg.cpp
```diff
@@ -64,7 +64,7 @@ unique_ptr<SnapshotParseInfo> IcebergSnapshot::GetParseInfo(yyjson_doc &metadata
 	} else {
 		auto schema = yyjson_obj_get(root, "schema");
 		if (!schema) {
-			throw IOException("Neither a valid schema or schemas field was found");
+			throw InvalidInputException("Neither a valid schema or schemas field was found");
```
Tmonster: Is there a reason this is not `InvalidConfigurationException`? I agree `IOException` isn't correct, but usually with an IOException a user can infer it was their fault, in this case, I feel like something is wrong with the rest catalog server, not necessarily user input

Tmonster (commented):
Thanks for the PR! I agree many of the IOError messages weren't correct error messages. I feel some of the `InvalidInputException` messages are not correct. 

One comment though, many of the errors in `iceberg.cpp` are thrown because the catalog has invalid data, not because the user entered invalid input.  Could those be changed to `InvalidConfigurationException`?

src/common/iceberg.cpp
Flogex: As discussed, I updated the exception types to `InvalidConfigurationException` in https://github.com/duckdb/duckdb-iceberg/pull/134/commits/ddccfbf240ec44480ee757904ba7150a45513180 everywhere except for irc_catalog.cpp.

src/common/api_utils.cpp
```diff
@@ -45,7 +47,8 @@ string APIUtils::GetRequest(ClientContext &context, const IRCEndpointBuilder &en
 		                 curl_easy_strerror(res));
 		if (res != CURLcode::CURLE_OK) {
 			string error = curl_easy_strerror(res);
-			throw IOException("Curl Request to '%s' failed with error: '%s'", url, error);
+			throw HTTPException(
+			    Exception::ConstructMessage("Curl Request to '%s' failed with error: '%s'", url, error));
```
Tmonster: Any reason you chose `Exception::ConstructMessage`
over `StringUtil::Format`?

Looking at duckdb/duckdb it looks like we prefer `StringUtil::Format`

Flogex: No particular reason. I have replaced the calls with `StringUtil::Format` in https://github.com/duckdb/duckdb-iceberg/pull/134/commits/98d69527134c38dfded99584dbf7ce338f19d7eb.

Tmonster (approved):
lgtm, thanks!

Tmonster:
@Flogex Looks like CI is failing? 
I think it's just a matter of modifying the test cases

Flogex:
> Looks like CI is failing? I think it's just a matter of modifying the expected results for the test cases

Sorry, missed this. Fixed in https://github.com/duckdb/duckdb-iceberg/pull/134/commits/1c8de5c2f24f4d427766266063a630b75c2c46d8.

The only thing I'm not sure about are the [changes in test_glue.test](https://github.com/duckdb/duckdb-iceberg/pull/134/commits/1c8de5c2f24f4d427766266063a630b75c2c46d8#diff-33c43d8717a5858df6b0d211ba253ecab3afa0de0a81d14983758c2849f2d57c) because I'm not sure what exception is expected here in the first place.

Tmonster:
I think `InvalidConfigurationException` is fine. We may change it to `InvalidInputException` later. The failure is that the glue endpoint doesn't match normal glue endpoint, so it's either invalid configuration or invalid input.

Flogex:
_test_mutliple_keys.test_ is still failing with an HTTP error. Is this expected or an issue with the test setup in CI?

```
Query failed, but error message did not match expected error message: Invalid Configuration Error (/home/runner/work/duckdb-iceberg/duckdb-iceberg/test/sql/local/irc/test_mutliple_keys.test:57)!
attach '840140254803:s3tablescatalog/pyiceberg-blog-bucket' as glue_lake (
    TYPE ICEBERG,
    ENDPOINT_TYPE 'GLUE'
);
Actual result:
HTTP Error: Failed to query https://glue.us-east-1.amazonaws.com/iceberg/v1/config?warehouse=840140254803%3As3tablescatalog%2Fpyiceberg-blog-bucket, http error 403 thrown. Message: {"message":"The security token included in the request is invalid."}
```

Failing in these lines:
```
statement ok
CREATE or replace SECRET amazing_secret (
    TYPE S3,
    KEY_ID 'admin',
    SECRET 'password',
    ENDPOINT '127.0.0.1:9000',
    URL_STYLE 'path',
    USE_SSL 0,
    REGION 'us-east-1'
);

# new secret has a region, but amazing secret is still used. The call to /config? has invalid credentials
statement error
attach '840140254803:s3tablescatalog/pyiceberg-blog-bucket' as glue_lake (
    TYPE ICEBERG,
    ENDPOINT_TYPE 'GLUE'
);
----
Invalid Configuration Error
```

Tmonster:
@Flogex This error is expected. You can run this test locally since it's in `test/sql/local`. The changed error message is this one

https://github.com/duckdb/duckdb-iceberg/pull/134/files#diff-c251a93dfb43dc99d59ea9d01d34d034e8d9391fbc9aa8b1ebea8b799bd4b276R122

Tmonster:
Thanks!

PR #136, opened 2025-04-01
Fix CREATE SECRET for Polaris catalog

ediril:
Fixes: https://github.com/duckdb/duckdb-iceberg/issues/137

This PR fixes logic in `IcebergCatalogAttach()` to work correctly for Polaris usecase.

### DIFF
--- a/src/iceberg_extension.cpp
+++ b/src/iceberg_extension.cpp
@@ -31,7 +31,7 @@ static unique_ptr<BaseSecret> CreateCatalogSecretFunction(ClientContext &context
 		auto lower_name = StringUtil::Lower(named_param.first);
 
 		if (lower_name == "key_id" || lower_name == "secret" || lower_name == "endpoint" ||
-		    lower_name == "aws_region" || lower_name == "oauth2_scope" || lower_name == "oauth2_server_uri") {
+		    lower_name == "region" || lower_name == "oauth2_scope" || lower_name == "oauth2_server_uri") {
 			result->secret_map[lower_name] = named_param.second.ToString();
 		} else {
 			throw InternalException("Unknown named parameter passed to CreateIRCSecretFunction: " + lower_name);
@@ -169,6 +169,11 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
 		catalog->catalog_type = catalog_type;
 		catalog->GetConfig(context);
 		return std::move(catalog);
+	}
+
+	// Check no endpoint type has been passed.
+	if (!endpoint_type.empty()) {
+		throw IOException("Unrecognized endpoint type: %s. Expected either S3_TABLES or GLUE", endpoint_type);
 	}
 
 	catalog_type = ICEBERG_CATALOG_TYPE::OTHER;
--- a/src/iceberg_extension.cpp
+++ b/src/iceberg_extension.cpp
@@ -31,7 +31,7 @@ static unique_ptr<BaseSecret> CreateCatalogSecretFunction(ClientContext &context
 		auto lower_name = StringUtil::Lower(named_param.first);
 
 		if (lower_name == "key_id" || lower_name == "secret" || lower_name == "endpoint" ||
-		    lower_name == "region" || lower_name == "endpoint_type") {
+		    lower_name == "region") {
 			result->secret_map[lower_name] = named_param.second.ToString();
 		} else {
 			throw InternalException("Unknown named parameter passed to CreateIRCSecretFunction: " + lower_name);
@@ -53,7 +53,6 @@ static void SetCatalogSecretParameters(CreateSecretFunction &function) {
 	function.named_parameters["key_id"] = LogicalType::VARCHAR;
 	function.named_parameters["secret"] = LogicalType::VARCHAR;
 	function.named_parameters["endpoint"] = LogicalType::VARCHAR;
-	function.named_parameters["endpoint_type"] = LogicalType::VARCHAR;
 	function.named_parameters["region"] = LogicalType::VARCHAR;
 	function.named_parameters["token"] = LogicalType::VARCHAR;
 }
@@ -82,6 +81,8 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
 
 	string account_id;
 	string service;
+	string endpoint_type;
+	string endpoint;
 
 	// check if we have a secret provided
 	string secret_name;
@@ -91,6 +92,11 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
 			// already handled
 		} else if (lower_name == "secret") {
 			secret_name = StringUtil::Lower(entry.second.ToString());
+		} else if (lower_name == "endpoint_type") {
+			endpoint_type = StringUtil::Lower(entry.second.ToString());
+		} else if (lower_name == "endpoint") {
+			endpoint = StringUtil::Lower(entry.second.ToString());
+			StringUtil::RTrim(endpoint, "/");
 		} else {
 			throw BinderException("Unrecognized option for iceberg catalog attach: %s", entry.first);
 		}
@@ -106,8 +112,6 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
 	}
 	const auto kv_secret = dynamic_cast<const KeyValueSecret &>(*secret_entry->secret);
 	auto region = kv_secret.TryGetValue("region");
-	auto endpoint_type_val = kv_secret.TryGetValue("endpoint_type");
-	string endpoint_type = endpoint_type_val.IsNull() ? "" : endpoint_type_val.ToString();
 
 	if (endpoint_type == "glue" || endpoint_type == "s3_tables") {
 		if (endpoint_type == "s3_tables") {
@@ -147,17 +151,14 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
 	} 
 
 	catalog_type = ICEBERG_CATALOG_TYPE::POLARIS;
-	Value region_val = kv_secret.TryGetValue("region");
-	credentials.region = region_val.IsNull() ? "" : region_val.ToString();
-	Value endpoint_val = kv_secret.TryGetValue("endpoint");
-	if (endpoint_val.IsNull()) {
-		throw IOException("Assumed catalog secret " + secret_entry->secret->GetName() + " for catalog " + name +
-							" does not have an endpoint");
-	}
-	string endpoint = endpoint_val.ToString();
-
-	if (endpoint_type.empty() && endpoint.empty()) {
-		throw IOException("No endpoint type or endpoint provided");
+	credentials.region = region.IsNull() ? "" : region.ToString();
+	if (endpoint.empty()) {
+		Value endpoint_val = kv_secret.TryGetValue("endpoint");
+		if (endpoint_val.IsNull()) {
+			throw IOException("Assumed catalog secret " + secret_entry->secret->GetName() + " for catalog " + name +
+								" does not have an endpoint");
+		}
+		endpoint = endpoint_val.ToString();
 	}
 
 	// Default IRC path
--- a/src/iceberg_extension.cpp
+++ b/src/iceberg_extension.cpp
@@ -148,7 +148,7 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
 		catalog->catalog_type = catalog_type;
 		catalog->GetConfig(context);
 		return std::move(catalog);
-	} 
+	}
 
 	catalog_type = ICEBERG_CATALOG_TYPE::POLARIS;
 	credentials.region = region.IsNull() ? "" : region.ToString();
--- a/src/iceberg_extension.cpp
+++ b/src/iceberg_extension.cpp
@@ -150,7 +150,7 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
 		return std::move(catalog);
 	}
 
-	catalog_type = ICEBERG_CATALOG_TYPE::POLARIS;
+	catalog_type = ICEBERG_CATALOG_TYPE::OTHER;
 	credentials.region = region.IsNull() ? "" : region.ToString();
 	if (endpoint.empty()) {
 		Value endpoint_val = kv_secret.TryGetValue("endpoint");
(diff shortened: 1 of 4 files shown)


src/include/storage/irc_catalog.hpp
```diff
@@ -28,7 +28,7 @@ class ICRClearCacheFunction : public TableFunction {
 	static void ClearCacheOnSetting(ClientContext &context, SetScope scope, Value &parameter);
 };
 
-enum class ICEBERG_CATALOG_TYPE { AWS_S3TABLES, AWS_GLUE, OTHER, INVALID };
+enum class ICEBERG_CATALOG_TYPE { AWS_S3TABLES, AWS_GLUE, POLARIS, INVALID };
```
Tmonster: Any reason this needs to specifically be Polaris? `AWS_S3Tables` and `AWS_GLUE` exist because there is some special handling when building the endpoint, and also to trigger the logic to make a SigV4 request instead of a regular CURL request. 

I think for catalogs that support Oauth, `OTHER` should be fine. (For instance Lakekeeper also uses Oauth)

src/iceberg_extension.cpp
```diff
@@ -98,12 +98,21 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
...
-			throw BinderException("Unrecognized option for PC attach: %s", entry.first);
+			throw BinderException("Unrecognized option for iceberg catalog attach: %s", entry.first);
 		}
 	}
 	auto warehouse = info.path;
 	auto catalog_type = ICEBERG_CATALOG_TYPE::INVALID;
 
+	// Lookup a secret we can use to access the rest catalog.
+	// if no secret is referenced, this throw
+	auto secret_entry = IRCatalog::GetSecret(context, secret_name);
```
Tmonster, line 117: We are working on refactoring some of our iceberg secret functionality, we don't want to throw if no secret is available, but rather if no iceberg secret is available. This will also require some changes to the `GetSecret` function, I'll come back to this

src/iceberg_extension.cpp
```diff
@@ -147,25 +150,18 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
 		return std::move(catalog);
 	}
 
-	// Check no endpoint type has been passed.
-	if (!endpoint_type.empty()) {
-		throw IOException("Unrecognized endpoint point: %s. Expected either S3_TABLES or GLUE", endpoint_type);
```
Tmonster, line 173: nit: I would like to keep this in for UX. `ENDPOINT_TYPE` is for glue and s3tables, but may later be expanded.

src/iceberg_extension.cpp
```diff
@@ -147,25 +150,18 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
...
-	}
-	if (endpoint_type.empty() && endpoint.empty()) {
-		throw IOException("No endpoint type or endpoint provided");
+	catalog_type = ICEBERG_CATALOG_TYPE::POLARIS;
+	credentials.region = region.IsNull() ? "" : region.ToString();
+	if (endpoint.empty()) {
+		Value endpoint_val = kv_secret.TryGetValue("endpoint");
```
Tmonster, line 169: Re: refactoring our secret functionality. 

It looks like here you get the catalog `endpoint` from the secret if it is not defined in the `Attach` statement. Is there a specific reason you need to get it from the secret?

src/storage/irc_catalog.cpp
```diff
@@ -137,13 +137,13 @@ DatabaseSize IRCatalog::GetDatabaseSize(ClientContext &context) {
 
 IRCEndpointBuilder IRCatalog::GetBaseUrl() const {
 	auto base_url = IRCEndpointBuilder();
-	base_url.SetPrefix(prefix);
```
Tmonster: Does Polaris not need the prefix when querying the catalog? I know for hitting the config endpoint, the prefix is not needed, but the `GetConfig` function handles this. 

I can imagine a scenario where the prefix is an empty string "", which may break the endpoint builder, but then you can handle that in `IRCEndpointBuilder::GetUrl()`

Tmonster (commented):
I know for other extensions, like Postgres, the endpoint is part of the secret. For Iceberg, we may change where endpoints are defined because the Iceberg REST spec is deprecating OAUTH. My understanding is that this means the oauth authentication endpoint may be different from the catalog endpoint depending on how Polaris is set up/what authentication services are used. (Although I have not encountered this scenario anywhere yet).

This is to say some of this code may be changing very soon.

Tmonster:
@ediril can you also modify the [Polaris test](https://github.com/duckdb/duckdb-iceberg/blob/main/test/sql/local/irc/test_polaris.test) so it reflects a bit more what your workflow is? It's not tested in CI right now, but we are working towards adding it. This way we can avoid breaking Polaris connections in the future, and we can also stay up to date with current Polaris usage

src/storage/irc_catalog.cpp
Tishj: this is already gone on main now

src/iceberg_extension.cpp
Tishj: This is an interesting observation though

If glue/s3tables, we look up the s3 secret
But then in the non-glue/s3tables path, we *also* get the s3 secret

So this is essentially just a deduplication of the current logic.
But we shouldn't be looking up an s3 secret in the non-glue/s3tables path, as Tom said, we should be looking for an `iceberg` secret instead.
I have already started working on a branch for this.

The s3 secret in the non-glue/s3tables path is currently used to look up a key+secret to provide the `/v1/oauth/tokens` endpoint - in which case the key+secret is not actually an s3 secret.
Which is why we're changing this.

src/iceberg_extension.cpp
Tishj: There is no reason to change this in this PR

src/iceberg_extension.cpp
Tishj: To continue on my comment from above, the fact that you're looking for the catalog endpoint through the  `endpoint` of an s3 secret shows again that this shouldn't be an s3 secret.

See [here](https://duckdb.org/docs/stable/extensions/httpfs/s3api#overview-of-s3-secret-parameters) what the `endpoint` means for an s3 secret.

Tishj: The current plan for the secret rework does not include the ability to provide the `endpoint` in the iceberg secret.. but thinking about it, perhaps this is a good idea.

A single endpoint can provide access to multiple catalogs (or "warehouses"), since every catalog has a different prefix.
So putting this into the secret, together with the authorization credentials, could be useful for reusability purposes.

src/include/storage/irc_catalog.hpp
Tishj: This change is not necessary for this PR

src/storage/irc_catalog.cpp
```diff
@@ -137,13 +137,13 @@ DatabaseSize IRCatalog::GetDatabaseSize(ClientContext &context) {
...
 	base_url.SetWarehouse(warehouse);
 	base_url.SetVersion(version);
 	base_url.SetHost(host);
 	switch (catalog_type) {
 	case ICEBERG_CATALOG_TYPE::AWS_GLUE:
 	case ICEBERG_CATALOG_TYPE::AWS_S3TABLES: {
+		base_url.SetPrefix(prefix);
```
Tishj: Please remove this.
I think this change was made in an attempt to preserve old behavior, but the `SetPrefix` call here has been removed entirely on main already.

src/iceberg_extension.cpp
ediril: Note that, when I made this PR, I just wanted to fix the POLARIS code path, so I did not question the non-polaris code path. Just took it as is as much as possible.

src/iceberg_extension.cpp
ediril: This check is unnecessary for the non-polaris code path as ENDPOINT TYPE is not used there. Therefore the check is removed as it won't be provided

src/iceberg_extension.cpp
ediril: You are correct, the ENDPOINT doesn't need to live in the secret since it does not need to be a secret, and can be provided as part of the ATTACH command. We can make that as a separate change though if you don't mind

src/storage/irc_catalog.cpp
ediril: No it does not, the warehouse is the same thing so it gets included twice and leads to an error

src/include/storage/irc_catalog.hpp
ediril: Reverted

src/storage/irc_catalog.cpp
ediril: Removed

src/iceberg_extension.cpp
Tishj: Which is why we're throwing if `endpoint_type` is set here, because it's not recognized - otherwise it wouldn't get here 👍

src/storage/irc_catalog.cpp
Tishj:
```c++
IRCEndpointBuilder IRCatalog::GetBaseUrl() const {
	auto base_url = IRCEndpointBuilder();
	base_url.SetVersion(version);
	base_url.SetHost(host);
	switch (catalog_type) {
	case ICEBERG_CATALOG_TYPE::AWS_GLUE:
	case ICEBERG_CATALOG_TYPE::AWS_S3TABLES: {
		base_url.AddPathComponent("iceberg");
		base_url.AddPathComponent(version);
		break;
	}
	default:
		break;
	}
	return base_url;
}
```

You should fetch main and merge with it 👍

src/iceberg_extension.cpp
ediril: It was already in the secret, then it was removed as part of the GLUE/S3_TABLES changes but I don't think they thought through all the implications at that time

ediril:
> @ediril can you also modify the [Polaris test](https://github.com/duckdb/duckdb-iceberg/blob/main/test/sql/local/irc/test_polaris.test) so it reflects a bit more what your workflow is? It's not tested in CI right now, but we are working towards adding it. This way we can avoid breaking Polaris connections in the future, and we can also stay up to date with current Polaris usage

@Tmonster I'm afraid I don't have time to do that at the moment for this PR

src/iceberg_extension.cpp
ediril: Ok the check is back

ediril:
@Tmonster I don't quite follow how the test works.. Is there any documentation? how can i run it on my laptop?

ediril:
@Tmonster I modified polaris test to reflect our workflow

src/iceberg_extension.cpp
```diff
@@ -30,20 +30,14 @@ static unique_ptr<BaseSecret> CreateCatalogSecretFunction(ClientContext &context
...
-		    lower_name == "aws_region" || lower_name == "oauth2_scope" || lower_name == "oauth2_server_uri") {
+		if (lower_name == "client_id" || lower_name == "client_secret" || lower_name == "endpoint" ||
+		    lower_name == "region" || lower_name == "oauth2_scope" || lower_name == "oauth2_server_uri") {
 			result->secret_map[lower_name] = named_param.second.ToString();
 		} else {
 			throw InternalException("Unknown named parameter passed to CreateIRCSecretFunction: " + lower_name);
 		}
 	}
 
-	// Get token from catalog
```
ediril, line 41: This is moved to `IcebergCatalogAttach()`

src/iceberg_extension.cpp
```diff
@@ -116,15 +110,17 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
 		oauth2_scope = "PRINCIPAL_ROLE:ALL";
 	}
 
-	if (oauth2_server_uri.empty()) {
```
ediril, line 119: This is moved to `IcebergCatalogAttach()`

test/sql/local/irc/test_polaris.test
```diff
@@ -16,15 +16,17 @@ mode skip
 
 statement ok
 create secret polaris_secret (
-  type s3,
-  KEY_ID '',
-  SECRET ''
+  type ICEBERG,
```
ediril, line 19: updated the test to reflect our workflow with Polaris catalog

src/iceberg_extension.cpp
Tmonster: I am still not convinced having a catalog endpoint stored in an `ICEBERG`/`S3` secret is the best idea. This logic also makes that code path the only way to add a catalog endpoint. 
@ediril can you provide a bit more of an explanation as to what the implications are to not having the catalog endpoint in the secret? 

Also, https://github.com/duckdb/duckdb-iceberg/pull/142 and https://github.com/duckdb/duckdb-iceberg/pull/150 add some more functionality to test Polaris

Tmonster (commented):
Hi @ediril ,

Thanks for the PR. We want to fix the Polaris use cases, but I think https://github.com/duckdb/duckdb-iceberg/pull/142 provides a more sustainable solution that is more catalog agnostic. 

Looking at this PR from a general standpoint, I'm unsure if the changes provide a benefit. I agree that `key_id` should be changed to `client_id`, and `secret_id` to `client_secret`. What I still don't understand is why the catalog url can only be declared in a secret and no longer in the Attach statement. Maybe I'm missing something though, is there a reason the catalog URL **needs** to be in the Secret?

Also, we have recently added Polaris to the CI, so if you merge with main you should see Polaris automatically tested as well

CC @Tishj

ediril:
> I am still not convinced having a catalog endpoint stored in an ICEBERG/S3 secret is the best idea. 

You are correct, it doesn't have to be in the secret. It can be specified in the ATTACH command. #142 makes sense to me as well.

The priority of this PR is to fix the broken aspects of iceberg extension with regards to Polaris catalog, minimizing any other changes, so it preserves the endpoint in the secret.

Tmonster:
> How do you recommend we proceed?

I think we can close this PR in favor of https://github.com/duckdb/duckdb-iceberg/pull/142 if you agree that it also solves your problem. You can take a look at tests there to make sure things are fixed, another option is to download the latest iceberg version from CORE_NIGHTLY once the PR is merged to verify fixes

Sorry about any issues caused by lack of/bad communication on my part

ediril:
@Tmonster I'll check out #142 tomorrow and test it to see if it fixes everything in #136. I'll report back here and either close this PR or let you know what needs to be moved over there.

ediril:
PR #142 supercedes this PR, it already has the same fixes and more.

PR #139, opened 2025-04-01
Fix Format Check CI

Tishj:
The submodule was not being checked out, so it had no `extension-ci-tools` to use in the workflow.


Tmonster:
Thanks! Can you also remove the `make generate-file` rule?

Tishj:
@Tmonster that's a different issue, looks like ninja needs to be installed
just had a look at other `extension-ci-tools`-using extensions, none of them have a step dedicated to installing ninja-build, but it looks like that is what we need to do, from the error.

PR #142, opened 2025-04-02
Iceberg attach rework

Tishj:
This PR changes the way secret and attach options function for iceberg.

Previously the `secret` in the attach options expected an S3 secret, this makes sense for s3tables/glue, but not so much for more traditional Iceberg REST Catalogs, that instead rely on OAuth2 to authorize access to the catalog.

In which case the `key_id` and `secret` from the s3 secret were used in request to the authorization server, in the client_credentials flow, which is wrong and confusing.

#### Options handled by the root:

- `endpoint`
The main endpoint of the Iceberg REST Catalog.
- `endpoint_type`
Streamlined way of attaching recognized catalog types.
- `authorization_type`
The method used to authorize access to the catalog.

`AUTHORIZATION_TYPE` defaults to `oauth2`.

#### Options handled by the endpoint type:
Based on the endpoint type, any additional options may be passed, see "Endpoint Types"

#### Options handled by the authorization type:
Based on the authorization type, any additional options may be passed, see "Authorization Types"

### ICEBERG Secret Options

The `ICEBERG` secret uses the same options as the `OAuth2` `AUTHORIZATION_TYPE`, with the addition of `endpoint`, to infer the `OAUTH2_SERVER_URI` if it's not given.

#### Glue (`glue`)
Attach Options:
- `secret`
The explicit name of a recognized storage secret to use.

NOTE:
The found secret (either through discovery or by explicitly naming it) needs to have `region` set.

`endpoint` will be set to `glue.{region}.amazonaws.com/iceberg`.
`authorization_type` will be set to `sigv4`.

#### S3Tables (`s3_tables`)

The region is discovered from the provided warehouse (`path` of the attached catalog)

`endpoint` will be set to `s3tables.{region}.amazonaws.com/iceberg`.
`authorization_type` will be set to `sigv4`.

#### OAuth2
Attach Options:
- `secret`
The explicit name of an `ICEBERG` secret to use.
- `oauth2_scope`
The `scope` for the authorization request.
- `oauth2_server_uri`
The endpoint of the authorization server to send the request to.
- `client_id`
The `client_id` for the authorization request.
- `client_secret`
The `client_secret` for the authorization request.
- `oauth2_grant_type`
The `grant_type` for the authorization request.

NOTE:
`secret` can not be combined with any of the other options.

#### SigV4
Attach Options:
- `secret`
The explicit name of a recognized storage secret to use.

### TL;DR

The main changes are:
- Addition of the `AUTHORIZATION_TYPE`.
- Fixing the `ICEBERG` secret type so it can be created by the user.
- Supporting `token` in the `oauth2` authorization type options.


corleyma:
I think this is headed in the right direction!  One question:

if I read this correctly, can storage_secret be e.g. a GCS secret, if someone is running a rest catalog for iceberg data on GCS, and hasn't enabled credential vending?

Tishj:
> I think this is headed in the right direction! One question:
> 
> if I read this correctly, can storage_secret be e.g. a GCS secret, if someone is running a rest catalog for iceberg data on GCS, and hasn't enabled credential vending?

Hmm I'm not sure, I think we'll need further changes to support other storage secret types than S3.
In this scenario, does the catalog expect authorization though the oauth endpoint or is it more like glue/s3tables?

corleyma:
@Tishj If by oauth endpoint you mean the catalog-hosted oauth endpoint, that's going away in the REST spec.  The catalog does still expect authentication and will need its own secret.  And while in practice for REST catalog servers, the way to auth has been via oauth client credentials flow (because that's what the major iceberg client reference implementations in Java implemented), that is also changing [now.](https://github.com/apache/iceberg/pull/12197) with the introduction of the AuthManager API.  Similar changes are planned for pyiceberg to support more flexible auth.

I think instead of thinking of it as an "Iceberg Secret", you could think of a few secret types:
 - sigv4 (this is the kind of secret that s3tables/glue/etc use, what you've called "s3" secret so far -- standard auth mechanism for AWS services)
 - oauth client credentials flow (currently, what you're calling Iceberg secret type)
 - oauth device code flow (human user flow, would require user to auth in browser)
 - token (just give me a token however you got it, I will submit this as Bearer token to catalog)
 
 If you consider a catalog like Lakekeeper that implements OIDC with external identity providers, all that matters is that it gets a signed access (or identity) jwt with the correct claims (like audience, etc).
 
 So I think it might be smart to have an interface that works like:
 - catalog secret (always required)
 - storage secret (only required if catalog does not vend credentials)
 
 And catalog secret can be one of:
 - sigv4
 - Oauth2Credentials (initially, just support client credentials flow as you do now, could be updated in future to make client_secret optional and add more optional fields to support e.g. oauth2 device code flow, etc;  I would also advocate for dropping the legacy {endpoint}/v1/oauth/tokens support here so it really is just generic oauth2 client credentials, since this endpoint has been deprecated in the spec for awhile now and I am not aware of production catalogs that require it.)
 - Token
 etc

By letting users configure a token secret type that you will pass directly as bearer token when making requests to catalog, you provide the opportunity for others to implement their own custom client-side auth flows, including if needed e.g. device code flow (though i do think that's something that would be good to be duckdb native in the future), and just give you a ready to use token.  This will help people who have weird requirements like accessing their catalogs behind authenticating proxies, etc (it's a thing).

src/storage/irc_catalog.cpp
```diff
@@ -156,24 +156,45 @@ void IRCatalog::ClearCache() {
...
+		secret_entry = context.db->GetSecretManager().GetSecretByName(transaction, "__default_s3");
+	} else {
 		secret_entry = context.db->GetSecretManager().GetSecretByName(transaction, secret_name);
 	}
+
 	if (!secret_entry) {
 		auto secret_match = context.db->GetSecretManager().LookupSecret(transaction, "s3://", "s3");
 		if (!secret_match.HasMatch()) {
-			throw IOException("Failed to find a secret and no explicit secret was passed!");
+			throw IOException("Failed to find an S3 secret and no explicit secret was passed!");
```
Tmonster: nit: Maybe `InvalidConfigurationException` is a better exception type?

src/storage/irc_catalog.cpp
```diff
@@ -156,24 +156,45 @@ void IRCatalog::ClearCache() {
...
+		//! Try to find any secret with the 'iceberg' type
+		auto secret_match = context.db->GetSecretManager().LookupSecret(transaction, "", "iceberg");
+		if (!secret_match.HasMatch()) {
+			return nullptr;
+		}
+		secret_entry = std::move(secret_match.secret_entry);
+	} else {
+		secret_entry = context.db->GetSecretManager().GetSecretByName(transaction, secret_name);
+	}
+	return secret_entry;
```
Tmonster, line 203: Should this throw if there is no IcebergSecret?

test/sql/cloud/test_bad_connections.test
```diff
@@ -60,6 +60,5 @@ attach 'quickstart_catalog' as my_datalake (
     ENDPOINT 'http://does_not_exist.com/api/catalog'
 )
 ----
-IO Error
-
+Invalid Input Error: Please provide either a 'client_id' + 'client_secret' pair or the name of an ICEBERG secret as 'secret' / 'catalog_secret'
```
Tmonster: I think a similar error is in Main now https://github.com/duckdb/duckdb-iceberg/pull/161

src/iceberg_extension.cpp
```diff
@@ -80,52 +142,63 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
...
 			endpoint = StringUtil::Lower(entry.second.ToString());
 			StringUtil::RTrim(endpoint, "/");
 		} else if (lower_name == "oauth2_scope") {
-			oauth2_scope = StringUtil::Lower(entry.second.ToString());
+			oauth2_scope = entry.second.ToString();
 		} else if (lower_name == "oauth2_server_uri") {
-			oauth2_server_uri = StringUtil::Lower(entry.second.ToString());
+			oauth2_server_uri = entry.second.ToString();
+		} else if (lower_name == "client_id") {
+			client_id = entry.second.ToString();
```
Tmonster: Not sure how I feel about about `client_id` and `client_secret` being allowed in an Attach statement. Are we sure this is something we want to allow?

src/iceberg_extension.cpp
```diff
@@ -136,20 +209,20 @@ static unique_ptr<Catalog> IcebergCatalogAttach(StorageExtensionInfo *storage_in
...
 		// if there is no secret, an error will be thrown
-		auto secret_entry = IRCatalog::GetSecret(context, secret_name);
+		auto secret_entry = IRCatalog::GetS3Secret(context, storage_secret);
 		auto kv_secret = dynamic_cast<const KeyValueSecret &>(*secret_entry->secret);
 		auto region = kv_secret.TryGetValue("region");
 
 		if (region.IsNull()) {
-			throw IOException("Assumed catalog secret " + secret_entry->secret->GetName() + " for catalog " + name +
-			                  " does not have a region");
+			throw IOException("Assumed catalog secret '%s' for catalog '%s' does not have a region",
```
Tmonster: nit: InvalidConfiguration?

test/sql/local/iceberg_on_tpch.test
```diff
@@ -15,22 +15,26 @@ require httpfs
...
-    SECRET 'password',
     ENDPOINT '127.0.0.1:9000',
     URL_STYLE 'path',
+    KEY_ID 'admin',
+    SECRET 'password',
     USE_SSL 0
 );
 
+# Note: the 'CLIENT_SECRET' is not verified at all by the authorization server,
+# so it's not actually testing authorization
```
Tmonster, line 28: Do you mean for just this example? Or for all catalogs?

src/include/storage/irc_catalog.hpp
```diff
@@ -77,7 +77,9 @@ class IRCatalog : public Catalog {
 		return "iceberg";
 	}
 
-	static unique_ptr<SecretEntry> GetSecret(ClientContext &context, const string &secret_name);
+	static unique_ptr<SecretEntry> GetS3Secret(ClientContext &context, const string &secret_name);
```
Tmonster: I think this can be `GetAWSSecret`, the httpfs extension now also supports `AWS` secrets, which is also in the tests now. `GetAWSSecret` also makes a bit more sense since the UX of using an S3 secret for an Iceberg catalog doesn't make a lot of sense

test/sql/local/irc/irc_catalog_read_deletes.test
```diff
@@ -26,6 +26,8 @@ CREATE SECRET (
 statement ok
 ATTACH '' AS my_datalake (
 	TYPE ICEBERG,
+    CLIENT_ID 'admin',
```
Tmonster: nit: whitespace

test/sql/local/irc/test_create_secret_attach_combinations.test
```diff
@@ -17,14 +17,16 @@ set enable_logging=true
...
 CREATE SECRET iceberg_secret (
 	TYPE ICEBERG,
 	CLIENT_ID 'admin',
 	CLIENT_SECRET 'password',
 	OAUTH2_SERVER_URI 'http://127.0.0.1:8181/v1/oauth/tokens',
-    Authorization_type 'Oauth2'
+	Authorization_type 'Oauth2'
 );
+----
+Invalid Input Error: Could not get token from http://127.0.0.1:8181/v1/oauth/tokens
```
Tmonster: Is there a reason this shouldn't pass? It should if you have the apache iceberg server running right?

test/sql/local/irc/test_create_secret_attach_combinations.test
```diff
@@ -37,12 +39,17 @@ CREATE SECRET storage_secret(
...
 
-statement ok
+statement error
 ATTACH '' AS my_datalake (
     TYPE ICEBERG,
     CATALOG_SECRET iceberg_secret,
     ENDPOINT 'http://127.0.0.1:8181'
 );
+----
+Invalid Input Error: No ICEBERG secret by the name of 'iceberg_secret' could be found
```
Tmonster: If the `create secret` statement passes, then this one will too I think

test/sql/local/irc/test_incorrect_secret_and_attach.test
```diff
@@ -47,23 +47,24 @@ CREATE SECRET iceberg_secret (
...
-statement ok
+# FIXME: We can't avoid this error, unless we delay the GetToken call until attach
+statement error
 CREATE SECRET iceberg_secret (
 	TYPE ICEBERG,
 	CLIENT_ID 'admin',
 	CLIENT_SECRET 'password'
 );
+----
+Invalid Input Error: No 'oauth2_server_uri' was provided, and no 'endpoint' was provided to fall back on
```
Tmonster: Hmmm, I feel like we can improve this error message to also say what is needed to the error.
Maybe something like this?
`AUTHORIZATION_TYPE is set to 'OAUTH2'. No  'oauth2_server_uri' was provided, and no 'endpoint' was provided to fall back on. Please provide an 'oauth2_server_uri' or set AUTHORIZATION_TYPE to SIGV4`

Tmonster (commented):
Thanks for adding the tests, I think some of them should be passing though.

We may have to talk to AWS about this, but I'm wondering what happens in the case where someone creates an `ICEBERG` secret to communicate with a `S3Tables` or `GLUE` catalog. It will most likely require add `SigV4` to the Authorization Types for Iceberg secrets

Tmonster:
> This is shaping up nicely to me, but wanted to bump again the discussion of supporting storage secrets for object stores other than s3 -- GCS, Azure, etc? Is that planned in this round of work or future work?

@corleyma That won't happen in this PR, sorry. We will get to it though (hopefully soon, definitely before DuckDB v1.3), I can ping you when we have a PR up 👍

src/storage/irc_catalog.cpp
Tishj: Yes, and it does, at a later place:
```c++
		if (!catalog_secret.empty()) {
			throw InvalidInputException("No ICEBERG secret by the name of '%s' could be found", catalog_secret);
		}
```

test/sql/cloud/test_bad_connections.test
Tishj: I don't follow?

Tishj:
```
type ICEBERG,
endpoint XXXX
```
This needs credentials, either a storage secret or a catalog secret

src/iceberg_extension.cpp
Tishj: I don't see why not?
You left another comment how `ENDPOINT` shouldn't be necessary in the creation of an `ICEBERG` secret, which isn't possible, because we need it to infer where to get the token from.

The other possibility is to not create an ICEBERG secret and instead provide the client_id+client_secret in the attach options, then an `ICEBERG` secret will be created during the attach

test/sql/local/iceberg_on_tpch.test
Tishj: This test, the "local" IRC's authorization server is a dummy, you can give it any secret and it will grant a token.

test/sql/local/irc/test_create_secret_attach_combinations.test
Tishj: I'll add the `require-env` then 👍

src/include/storage/irc_catalog.hpp
Tishj: I'm confused, is there a PR that changes all the internal logic that expect S3 secrets to AWS secrets? If not then we can't support this in this PR, and it will need a separate PR

test/sql/local/irc/test_create_secret_attach_combinations.test
```diff
@@ -160,30 +143,38 @@ CREATE SECRET iceberg_secret (
...
+	ENDPOINT 'http://127.0.0.1:8181'
 );
 
-query I
+# FIXME: move this up
+statement error
 select count(*) from my_datalake.default.pyspark_iceberg_table_v2;
 ----
-6592
+Invalid Configuration Error: Provided 'STORAGE_SECRET' is of type 'iceberg', expected 's3'
```
Tmonster: The [storage secret ](https://github.com/duckdb/duckdb-iceberg/pull/142/commits/de44243753feec274dbf49dd11330cef5de8a096#diff-64323b66fdccf6ee922c599b49e023c340e1e81a8f2b8f4d415d2955db8d2b25R32) is type s3. I think there is something else wrong if this is the error. Also the Attach and Select on lines 122 and 132 work with the same `STORAGE_SECRET`

test/sql/local/irc/test_create_secret_attach_combinations.test
```diff
@@ -160,30 +143,38 @@ CREATE SECRET iceberg_secret (
...
-statement ok
+
+statement error
 CREATE SECRET simple_iceberg_secret (
 	TYPE ICEBERG,
 	CLIENT_ID 'admin',
 	CLIENT_SECRET 'password'
 );
+----
+Invalid Input Error: No 'oauth2_server_uri' was provided, and no 'endpoint' was provided to fall back on (or consider changing the 'authorization_type')
```
Tmonster: Thanks, this is more informative, but I feel it doesn't tell the user the authorization type was assumed to be `Oauth2`. I think something like

```
AUTHORIZATION_TYPE is 'oauth2', yet no 'oauth2_server_uri' was provided, and no 'endpoint' was provided to fall back on. Please provide one or change the 'authorization_type'.
```

test/sql/cloud/test_aws_secret_type.test
```diff
@@ -0,0 +1,61 @@
+# name: test/sql/cloud/test_aws_secret_type.test
+# description: test integration with iceberg catalog read
+# group: [cloud]
+
+# require-env ICEBERG_AWS_REMOTE_AVAILABLE
```
Tmonster: We should uncomment these

test/sql/cloud/test_aws_secret_type.test
```diff
@@ -0,0 +1,61 @@
...
+);
+
+# GLUE / S3 Table type should be able to find AWS secrets
+statement error
+attach '840140254803:s3tablescatalog/duckdblabs-iceberg-testing' as glue_catalog (
+    TYPE ICEBERG,
+    ENDPOINT_TYPE 'GLUE'
+);
+----
+Invalid Configuration Error: Failed to find an S3 secret and no explicit secret was passed!
```
Tmonster: If the `GetAWSSecret` function ends up working this should pass, let me know if it doesn't and I can give you credentials

test/sql/local/irc/test_create_secret_attach_combinations.test
```diff
@@ -0,0 +1,180 @@
...
+require httpfs
+
+# Test catalog secret and storage secret
+statement ok
+set enable_logging=true
+
+statement ok
+set logging_level='debug'
+
+require-env ICEBERG_SERVER_AVAILABLE
```
Tmonster: nit: can we move this either to the top of the test file or above the first `statement ok`s?

test/sql/local/irc/test_incorrect_secret_and_attach.test
```diff
@@ -0,0 +1,70 @@
...
+# we don't allow oauth2 without oauth2_server_uri
+statement error
+CREATE SECRET iceberg_secret (
+	TYPE ICEBERG,
+	CLIENT_ID 'admin',
+	CLIENT_SECRET 'password',
+	AUTHORIZATION_TYPE 'oauth2'
+);
+----
+Invalid Input Error: No 'oauth2_server_uri' was provided
```
Tmonster: This error message can be the same as the one in `test_create_secret_attach_combinations.test`
```
AUTHORIZATION_TYPE is 'oauth2', yet no 'oauth2_server_uri' was provided, and no 'endpoint' was provided to fall back on. Please provide one or change the 'authorization_type'.
```

test/sql/local/irc/test_lakekeeper_combinations.test
```diff
@@ -0,0 +1,157 @@
...
+    ENDPOINT 'http://127.0.0.1:8181/catalog'
+);
+
+query I
+SELECT message FROM duckdb_logs where type='iceberg.Catalog.Curl.HTTPRequest' and message like '%POST%' order by timestamp
+----
+POST http://127.0.0.1:8181/v1/oauth/tokens (curl code 'No error')
+
+query I
+SELECT message FROM duckdb_logs where type='iceberg.Catalog.Curl.HTTPRequest' and message like '%POST%' order by timestamp
```
Tmonster: This should be `message like `%GET%`, my bad

test/sql/local/irc/test_lakekeeper_combinations.test
```diff
@@ -0,0 +1,157 @@
...
+
+query III
+select * from my_datalake.my_namespace.my_table;
+----
+1	a-string	2.2
+
+# simple iceberg secret should be ok, catalog attach statement has opportunity to say what
+# authorization type is.
+statement ok
+CREATE SECRET simple_iceberg_secret (
```
Tmonster: You can remove this, I think we've covered it

Tmonster (commented):
Tests are looking good, mostly nit comments, but there is one failure in test/sql/local/irc/test_create_secret_attach_combinations.test that doesn't make sense

Tmonster:
The DuckDB submodule pointer just updates the pointer to latest v1.2 right? I see it snuck in with a merge with upstream/main so I just wanna make sure.

Working on fixing Polaris CI [here](https://github.com/duckdb/duckdb-iceberg/pull/169)

Any reason why a `secret` parameter and a `storage_secret` parameter can't exist in the same attach? I think `secret` can just be an alias for `catalog_secret` right?

src/include/storage/irc_catalog.hpp
Tmonster: No just that `TYPE AWS` is basically an alias to `TYPE S3`

Tishj:
> The DuckDB submodule pointer just updates the pointer to latest v1.2 right? I see it snuck in with a merge with upstream/main so I just wanna make sure.
> 
> Working on fixing Polaris CI [here](https://github.com/duckdb/duckdb-iceberg/pull/169)
> 
> Any reason why a `secret` parameter and a `storage_secret` parameter can't exist in the same attach? I think `secret` can just be an alias for `catalog_secret` right?

`secret` was a shorthand for supplying both, no?
I thought that's what we settled on

I'd prefer to just rip out `secret` entirely and only have `catalog_secret` and `storage_secret`, but that's not what we settled on

Tmonster:
So I think I've figured out the issue with the cloud tests when running them locally. The issue is with immediately initializing the url to `{service}.{region}.amazonaws.com/iceberg/v1`. Doing do causes the AWS SDK to treat `{service}.{region}.amazonaws.com/iceberg/v1` as the authority, meaning `iceberg` and `v1` are not treated as path elements that get separately encoded during the request. The fix should be to go back to calling `AddPathComponent("iceberg"); AddPathComponent(version)` for AWS catalogs. 

We can figure out a better way to fix it in a future PR. If we eventually get rid of the AWS extension all together, this problem won't exist anymore (I think)

src/storage/authorization/sigv4.cpp
```diff
@@ -0,0 +1,35 @@
...
+		auto lower_name = StringUtil::Lower(entry.first);
+		if (lower_name == "secret") {
+			if (!result->secret.empty()) {
+				throw InvalidInputException("Duplicate 'secret' option detected!");
+			}
+			result->secret = StringUtil::Lower(entry.second.ToString());
+		}
+	}
+	//! Delete the 'secret' from the options if it was present - it has been handled
+	input.options.erase("secret");
```
Tishj: input.options is not case-insensitive, we should populate the `remaining_options` in the loop instead

ediril:
I just built and ran this version of duckdb against my data lake with Polaris. It works, except that if `ENDPOINT` is provided in the secret, it asks for it again in the `ATTACH` command. If you allow endpoint in the secret, then you should also use it in the ATTACH command even if an explicit endpoint is not provided.

ediril:
I just built and ran this version of duckdb against my data lake with Polaris. It mostly works, except that: 

1. If `ENDPOINT` is provided in the secret, it asks for it again in the `ATTACH` command. If you allow endpoint in the secret, then you should also use it in the ATTACH command even if an explicit endpoint is not provided. So the following should work but doesnt:

```
D CREATE SECRET (
       TYPE ICEBERG,
       CLIENT_ID '...',
       CLIENT_SECRET '...',
       ENDPOINT 'https://polaris.fivetran.com/api/catalog'); 
┌─────────┐
│ Success │
│ boolean │
├─────────┤
│ true    │
└─────────┘
D ATTACH 'turnpike_stable' AS my_catalog (TYPE ICEBERG);
Invalid Configuration Error:
No 'endpoint_type' or 'endpoint' provided
```

2. The following should work but I get an error:
```
CREATE SECRET my_secret (
       TYPE ICEBERG,
       CLIENT_ID '....',
       CLIENT_SECRET '...', 
       OAUTH2_SERVER_URI 'https://polaris.fivetran.com/api/catalog/v1/oauth/tokens'); 
┌─────────┐
│ Success │
│ boolean │
├─────────┤
│ true    │
└─────────┘
D ATTACH 'turnpike_stable' AS my_catalog (TYPE ICEBERG, ENDPOINT 'https://polaris.fivetran.com/api/catalog', SECRET 'my_secret');
D select * from my_catalog.metrics.data_points;
Invalid Configuration Error:
Found a secret by the name of 'my_secret', but it is not of an accepted type for a 'secret', accepted types are: 's3', 'aws', 'r2' and 'gcs', found 'iceberg'
D 
```

ediril:
I also ran into this issue with 1.2.2 which would be nice to get fixed in this PR: https://github.com/duckdb/duckdb-iceberg/issues/172

Tishj:
> I just built and ran this version of duckdb against my data lake with Polaris. It mostly works, except that:

1. We are aware, that's unfortunately expected. The reason being is that the token request has to be made during the creation of the secret, because the token is part of the secret.
2. Hmm that does look unexpected, I'll have a look

1.2.2 doesn't have an associated iceberg release, I believe we're looking to merge this PR and then push the release for 1.2.2.

Tishj:
@ediril I can't reproduce your issue

```sql
require avro

require parquet

require iceberg

require httpfs

require-env ICEBERG_SERVER_AVAILABLE

statement ok
CREATE SECRET local_catalog_secret (
    TYPE S3,
    KEY_ID 'admin',
    SECRET 'password',
    ENDPOINT '127.0.0.1:9000',
    URL_STYLE 'path',
    USE_SSL 0
);

statement ok
CREATE SECRET my_secret (
	TYPE ICEBERG,
	CLIENT_ID 'admin',
	CLIENT_SECRET 'password',
	OAUTH2_SERVER_URI 'http://127.0.0.1:8181/v1/oauth/tokens'
); 

statement ok
ATTACH 'turnpike_stable' AS my_catalog (
	TYPE ICEBERG,
	ENDPOINT 'http://127.0.0.1:8181',
	SECRET 'my_secret'
);

statement ok
use my_catalog.default

query III
select * from table_more_deletes;
----
2023-03-01	1	a
2023-03-02	2	b
2023-03-03	3	c
2023-03-10	10	j
2023-03-11	11	k
2023-03-12	12	l
```

ediril:
I'm running it against a real data lake, and definitely see it happening.. we can do a google meet if you'd like me to show it to you 🤷

Tishj:
> Please try removing the S3 secret before the ICEBERG secret. It is not necessary when using the ICEBERG secret (Polaris will take care of it), shouldn't be there.. it's probably the reason why it works for you

No, it's not, this isn't Polaris.
I'll test it with Polaris/Lakekeeper, but the S3 secret is never referenced by the catalog

test/sql/local/irc/test_create_secret_attach_combinations.test
```diff
@@ -0,0 +1,177 @@
...
+select count(*) from my_datalake.default.pyspark_iceberg_table_v2;
+----
+6592
+
+statement ok
+Detach my_datalake;
+
+# create bad storage secret to test storage secret
+statement ok
+CREATE SECRET storage_secret_2 (
```
Tmonster: This create secret statement can be removed since it's not used anymore along with the drop 2 statements later

test/sql/local/irc/test_create_secret_attach_combinations.test
```diff
@@ -0,0 +1,177 @@
...
+drop secret storage_secret_2
+
+statement ok
+Detach my_datalake;
+
+# specifically use storage secret
+
+statement ok
+ATTACH '' AS my_datalake (
+	TYPE ICEBERG,
```
Tmonster, line 107: This attach is identical to the one on line 99. 
Maybe you can change this one to test that the oauth workflow defaults to the deprecated `{endpoint}/v1/oauth/token` endpoint?

src/storage/authorization/oauth2.cpp
```diff
@@ -0,0 +1,168 @@
...
+			    "these options are mutually exclusive");
+		}
+		if (!result->client_secret.empty()) {
+			throw InvalidConfigurationException("Please provide either a client_id+client_secret pair, or 'secret', "
+			                                    "these options are mutually exclusive");
+		}
+
+		auto &kv_iceberg_secret = dynamic_cast<const KeyValueSecret &>(*iceberg_secret->secret);
+		token = kv_iceberg_secret.TryGetValue("token");
+		result->uri = kv_iceberg_secret.TryGetValue("token").ToString();
```
Tmonster: this doesn't look right to me

src/storage/authorization/oauth2.cpp
```diff
@@ -0,0 +1,168 @@
...
+}
+
+unique_ptr<OAuth2Authorization> OAuth2Authorization::FromAttachOptions(ClientContext &context,
+                                                                       IcebergAttachOptions &input) {
+	auto result = make_uniq<OAuth2Authorization>();
+
+	unordered_map<string, Value> remaining_options;
+	string secret;
+	for (auto &entry : input.options) {
+		auto lower_name = StringUtil::Lower(entry.first);
```
Tishj, line 89: this is missing `token`

src/storage/authorization/oauth2.cpp
```diff
@@ -0,0 +1,168 @@
...
+			remaining_options.emplace(std::move(entry));
+		}
+	}
+
+	Value token;
+	// Check if any of the options are given that indicate intent to provide options inline, rather than use a secret
+	bool user_intends_to_use_secret = true;
+	if (!result->scope.empty() || !result->uri.empty() || !result->client_id.empty() ||
+	    !result->client_secret.empty()) {
+		user_intends_to_use_secret = false;
```
Tishj: I think this can be much less complicated
We can just put apart all the options for creating the `ICEBERG` secret, if they're empty we try to look up a secret, if they're not then we don't

src/storage/authorization/oauth2.cpp
Tmonster: Also if you add `input.endpoint = kv_iceberg_secret.TryGetValue("endpoint").ToString();` under this, and move [this check](https://github.com/duckdb/duckdb-iceberg/pull/142/files#diff-629ad72ca849b0e01b7afe5a855c65677fb5e27a1423b0c58abdd5f3f20a8793R178) to after parsing the oauth workflow, we can support (1) from [this comment](https://github.com/duckdb/duckdb-iceberg/pull/142#issuecomment-2794915560)

Tmonster: We can also add a log statement here to say we are assuming the endpoint from iceberg secret named 'xyz'

Tmonster (commented):
Almost ready to merge, just some comments on tests and one UX thing

ediril:
@Tmonster please don't merge yet, I'm getting HTTP 400 error when I try to query my data lake even though I'm not getting any errors during CREATE SECRET and ATTACH. Let me look into it

ediril:
This is what I'm getting:
```
D CREATE SECRET my_secret (
     TYPE ICEBERG,
     CLIENT_ID '...',
     CLIENT_SECRET '...',
     OAUTH2_SERVER_URI 'https://polaris.fivetran.com/api/catalog/v1/oauth/tokens'
  );
┌─────────┐
│ Success │
│ boolean │
├─────────┤
│ true    │
└─────────┘
D ATTACH 'land_inappropriate' AS my_catalog (
      TYPE ICEBERG, SECRET my_secret,
      ENDPOINT 'https://polaris.fivetran.com/api/catalog'
  );
D select * from my_catalog.fivetran_metadata_land_inappropriate.account limit 1;
HTTP Error:
HTTP GET error on 'https://george-fivetran-data-lake-test.s3.amazonaws.com/my_data_lake/fivetran_metadata_land_inappropriate/account/metadata/00002-3637a3f6-dc84-4df1-bfd8-978dcff9b477.metadata.json' (HTTP 400)
```

ediril:
Btw, there was a AWS_REGION parameter in the ICEBERG secret, that needs to be put back somewhere.. It could go into the ATTACH command, doesn't need to be in the SECRET

ediril:
It looks like REGION moved to GLUE and S3_TABLES but Polaris needs it as well.

ediril:
Here is a patch that fixes this to give you inspiration

[fix-region-for-polaris.patch](https://github.com/user-attachments/files/19712838/fix-region-for-polaris.patch)

Tishj:
> Here is a patch to give you inspiration that fixes the region issue above:
> 
> [fix-region-for-polaris.patch](https://github.com/user-attachments/files/19712838/fix-region-for-polaris.patch)

There is a `client.region` (see <https://editor-next.swagger.io/?url=https://raw.githubusercontent.com/apache/iceberg/main/open-api/rest-catalog-open-api.yaml>) listed in the recognized properties, that we don't support currently.

Would adding support for that fix your problem?
I really don't see why the "region" would need to be added by us in this manner, the catalog has many opportunities to give it in a sensible fashion, i.e through `/v1/config` or returned as part of the `config` or `storage-credentials` in the `/v1/{prefix}/namespaces/{namespace}/tables/{table}` response

ediril:
> I really don't see why the "region" would need to be added by us in this manner

Without it, it doesn't work for data lakes not in `us-east-1`. Might be a Polaris bug, I'm not sure. How you accept the region is upto you and it is optional.

The following is the key fix. How you get the `region` there is upto you. My patch is just one way, if you don't like it, that's fine. But we do need it, please make it possible to specify region:

```
diff --git a/src/storage/irc_table_entry.cpp b/src/storage/irc_table_entry.cpp
index f716ec3..82950b6 100644
--- a/src/storage/irc_table_entry.cpp
+++ b/src/storage/irc_table_entry.cpp
@@ -97,6 +97,8 @@ TableFunction ICTableEntry::GetScanFunction(ClientContext &context, unique_ptr<F
 			                                      : kv_secret.TryGetValue("session_token").ToString()},
 			                {"region", region},
 			                {"endpoint", endpoint}};
+		} else {
+			info.options["region"] = ic_catalog.region;
 		}
 
 		(void)secret_manager.CreateSecret(context, info);
```

Tishj:
> > I really don't see why the "region" would need to be added by us in this manner
> 
> Without it, it doesn't work for data lakes not in `us-east-1`. Might be a Polaris bug, I'm not sure. The following is the key fix. How you get the `region` there is upto you. My patch is just one way, if you don't like it, that's fine. But we do need it, please make it possible to specify region for the "else" case. The region is handled indirectly for GLUE and S3_TABLES. We need to provide it explicitly for Polaris when it differs from the default us-east-1
> 
> ```
> diff --git a/src/storage/irc_table_entry.cpp b/src/storage/irc_table_entry.cpp
> index f716ec3..82950b6 100644
> --- a/src/storage/irc_table_entry.cpp
> +++ b/src/storage/irc_table_entry.cpp
> @@ -97,6 +97,8 @@ TableFunction ICTableEntry::GetScanFunction(ClientContext &context, unique_ptr<F
>  			                                      : kv_secret.TryGetValue("session_token").ToString()},
>  			                {"region", region},
>  			                {"endpoint", endpoint}};
> +		} else {
> +			info.options["region"] = ic_catalog.region;
>  		}
>  
>  		(void)secret_manager.CreateSecret(context, info);
> ```

Can you please verify that the `/v1/config` result doesn't contain the region somehow?
And the LoadTableResult (the other endpoint I mentioned) can also have it in the `config` property, or in `storage-credentials`

Tishj:
@ediril can the `region` be a default, or does it have to be an override
i.e if the catalog returns credentials with a certain region, does that take precedence or should the user-supplied `region` override that?

ediril:
> I can add an override_region / default_region to the oauth2 authorization_type.
This would not have a default value (empty if not supplied)

Yes I just tested this, it works, thanks!

It doesn't need to have a default value, and you can even call it just `REGION`. You added a "not empty" check for it, that's perfect.

ediril:
> the catalog has many opportunities to give it in a sensible fashion, i.e through /v1/config or returned as part of the config or storage-credentials in the /v1/{prefix}/namespaces/{namespace}/tables/{table} response

No the region is not returned via either of these unfortunately

Tishj:
> > the catalog has many opportunities to give it in a sensible fashion, i.e through /v1/config or returned as part of the config or storage-credentials in the /v1/{prefix}/namespaces/{namespace}/tables/{table} response
> 
> No the region is not returned via either of these unfortunately. I agree `v1/config` would be a good place to return it. Might be a bug or oversight, I'm not sure.

I am confident this is a bug in Polaris, because this just doesn't respect vended credentials, there shouldn't be any intervention on the user side outside of the official channels (config, storage-credentials...)

PowerPlop:
Is authorization_type NONE supported?

Nessie catalog can run with authentication disabled: https://projectnessie.org/nessie-latest/authentication/

PR #144, opened 2025-04-02
Fix scanning from the "local" REST catalog we create in CI

Tishj:
The way this catalog works is that we create an S3 secret that should inform the options passed to the scoped secret we create from the `config` of the `LoadTableResult`

```sql
statement ok
CREATE SECRET (
    TYPE S3,
    KEY_ID 'admin',
    SECRET 'password',
    ENDPOINT '127.0.0.1:9000',
    URL_STYLE 'path',
    USE_SSL 0
);

statement ok
ATTACH '' AS my_datalake (
    TYPE ICEBERG,
);
```

Without this change, `url_style`, `use_ssl`, `endpoint`, etc.. would not be copied over, causing the scan to fail.


Tishj:
#149 fixed the compilation, this PR fixes the existing tests against the "local REST catalog"

PR #148, opened 2025-04-02
Fix compilation for linux arm64

Tishj:


CMakeLists.txt
```diff
@@ -45,6 +45,9 @@ find_package(CURL REQUIRED)
 find_package(AWSSDK REQUIRED COMPONENTS core sso sts)
 include_directories(${CURL_INCLUDE_DIRS})
 
+# Reset the TARGET_NAME, the AWS find_package build could bleed into our build - overriding `TARGET_NAME`
+set(TARGET_NAME iceberg)
```
Tishj, line 49: `TARGET_NAME` was set to `AWS::crypto` at this point, causing an issue further down, where we use `TARGET_NAME` assuming its value is `iceberg`

Tmonster:
Thanks!

PR #149, opened 2025-04-02
Add temporary bandaid to fix cmake 4.0.0 compilation

Tishj:


Tishj:
This passes, other than the issue in https://github.com/duckdb/duckdb-iceberg/pull/144
So I'm merging this so #144 can be rerun and check if it fixes the issue fully

PR #150, opened 2025-04-03
Add Polaris catalog test to CI

Tmonster:
Setup a CI local polaris catalog and then run some tests. This will later be extended to run tpch and more


Tishj (approved):
Thanks for taking the time to figure out all the CI headaches to get this working, this is great!

PR #153, opened 2025-04-04
Fix cloud tests. 

Tmonster:
Discovered some issues when testing against s3tables and glue locally. 

Here is a PR to fix them. We now run 6/22 tpch queries. These queries take about 30 seconds on my machine

### DIFF
--- a/src/catalog_api.cpp
+++ b/src/catalog_api.cpp
@@ -135,10 +135,12 @@ IRCAPITableCredentials IRCAPI::GetTableCredentials(ClientContext &context, IRCat
 
 	case_insensitive_map_t<Value> config_options;
 	auto *config_val = yyjson_obj_get(root, "config");
-	if (config_val && catalog_credentials) {
+	if (catalog_credentials) {
 		auto kv_secret = dynamic_cast<const KeyValueSecret &>(*catalog_credentials->secret);
 		for (auto &option : kv_secret.secret_map) {
-			config_options.emplace(option);
+			if (option.first != "refresh_info" && option.first != "refresh") {
+				config_options.emplace(option);
+			}
 		}
 	}
 	ParseConfigOptions(config_val, config_options);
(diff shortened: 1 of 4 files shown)


src/catalog_api.cpp
```diff
@@ -135,10 +135,12 @@ IRCAPITableCredentials IRCAPI::GetTableCredentials(ClientContext &context, IRCat
 
 	case_insensitive_map_t<Value> config_options;
 	auto *config_val = yyjson_obj_get(root, "config");
-	if (config_val && catalog_credentials) {
+	if (catalog_credentials) {
```
Tmonster, line 143: @Tishj this change I'm not 100% sure about. The issue is that if an S3 secret is created for s3tables, we need to create a temporary secret for the storage (with scope etc.) that also has the region which is inferred by the warehouse. 
So in this case, `config` is empty, but we have catalog credentials, and we want to use those

src/catalog_api.cpp
```diff
@@ -135,10 +135,12 @@ IRCAPITableCredentials IRCAPI::GetTableCredentials(ClientContext &context, IRCat
...
 		auto kv_secret = dynamic_cast<const KeyValueSecret &>(*catalog_credentials->secret);
 		for (auto &option : kv_secret.secret_map) {
-			config_options.emplace(option);
+			if (option.first != "refresh_info" && option.first != "refresh") {
```
Tmonster, line 149: Also it seems like if we add refresh info here, the httpfs extension throws an error. We can talk about this

src/catalog_api.cpp
Tishj: I think this was only following the old behavior, I'm fine with this change. 👍

src/catalog_api.cpp
Tmonster: The httpfs extension assumes the refresh_info object is a `map` and then calls `MapToStruct(named_param.second);`. Since it is already a `struct`, the function errors internally. But I've also added a comment as to why we can ignore this information when we copy the catalog secret

PR #154, opened 2025-04-04
Add Lakekeeper to CI

Tishj:
This PR adds Lakekeeper to CI, the steps to set up Lakekeeper are largely taken from <https://github.com/lakekeeper/lakekeeper/tree/main/examples/access-control-simple>

Using the existing `data_generators` in the scripts directory we populate Lakekeeper with TPCH and run a subset of the TPCH tests on it (`test_lakekeeper_tpch.test`).


.github/workflows/LakekeeperTesting.yml
```diff
@@ -0,0 +1,282 @@
...
+            attempt=$((attempt + 1))
+          done
+          echo "Lakekeeper is healthy"
+
+      - name: Populate Lakekeeper
+        env:
+          JAVA_HOME: /usr/lib/jvm/java-21-openjdk-amd64
+        run: |
+          # install java
+          sudo apt install -y -qq openjdk-21-jre-headless
```
Tmonster: Are the java installs not duplicated from setup lake keeper step?

.github/workflows/PolarisTesting.yml
```diff
@@ -132,4 +131,5 @@ jobs:
         run: |
           export POLARIS_CLIENT_ID=$(cat polaris_client_id.txt)
           export POLARIS_CLIENT_SECRET=$(cat polaris_client_secret.txt)
-          make test_release
\ No newline at end of file
+          ./build/release/test/unittest "$PWD/test/*" --list-test-names-only || true
```
Tmonster, line 134: any specific reason we want to list the test names and then run the tests?

scripts/data_generators/generate_spark_rest/generate_iceberg_spark_rest.py
```diff
@@ -25,7 +25,6 @@
 import os
 from pyspark import SparkContext
 from pathlib import Path
-import duckdb
```
Tmonster, line 28: DuckDB is used in the `GenerateTPCH` step, don't know if this should be taken out. I think a command like 
```
python3 scripts/data_generators/generate_data.py spark-rest 
```
might fail otherwise

scripts/lakekeeper/setup_lakekeeper_catalog.py
```diff
@@ -0,0 +1,137 @@
...
+from pyspark.sql import SparkSession
+import pandas as pd
+
+CATALOG_URL = "http://localhost:8181/catalog"
+MANAGEMENT_URL = "http://localhost:8181/management"
+KEYCLOAK_TOKEN_URL = "http://localhost:30080/realms/iceberg/protocol/openid-connect/token"
+WAREHOUSE = "demo"
+
+CLIENT_ID = "spark"
+CLIENT_SECRET = "2OR3eRvYfSZzzZ16MlPd95jhLnOaLM52"
```
Tmonster, line 14: Are we sure this secret is always the same?

test/sql/local/irc/test_lakekeeper_tpch.test
```diff
@@ -0,0 +1,55 @@
...
+
+statement ok
+create secret my_secret (
+	TYPE s3,
+	KEY_ID 'spark',
+	SECRET '2OR3eRvYfSZzzZ16MlPd95jhLnOaLM52'
+)
+
+statement ok
+attach 'demo' as my_datalake (
```
Tmonster, line 27: We also support the `OAUTH2_SCOPE` and `OAUTH2_SERVER_URI` in the `create secret` statement right? 
Can we also get a test where those values are initialized there?

Tmonster (commented):
Looks good, just a couple of comments

.github/workflows/PolarisTesting.yml
Tishj: To verify that the tests we care about are actually being run

scripts/data_generators/generate_spark_rest/generate_iceberg_spark_rest.py
Tishj: This was duplicate

scripts/lakekeeper/setup_lakekeeper_catalog.py
Tishj: Yes 👍

test/sql/local/irc/test_lakekeeper_tpch.test
Tishj: I don't think we do currently, that's part of #142

PR #159, opened 2025-04-07
Add User-agent header to AWS and regular Curl Requests

Tmonster:


Tishj (approved):
Great stuff, I think we eventually want to deduplicate these scripts, only the configuration is catalog-specific, populating the catalog isn't, so we can avoid duplicating this 3 (or more) times

PR #161, opened 2025-04-07
Fix formatting dependencies in CI

Tishj:
versions should be locked.


Tishj:
@Tmonster did this possibly break the CI workflow? I can't imagine that's the case, but I can't see anything else that happened with the `CodeQuality.yml` file

Unless it's just the "scheduled brownout" that is causing this workflow to break

Tmonster:
@Tishj I think it's the ubuntu brown-out, we can make CodeQuality run on ubuntu-latest

PR #162, opened 2025-04-07
Attempt to fix cloud tests #2

Tmonster:
For some reason glue is behaving weirdly.

Edit: I think it's a race condition, nothing major

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,9 +38,6 @@ format-fix:
 	rm -rf src/amalgamation/*
 	python3 scripts/format.py --all --fix --noconfirm
 
-format:
-	python3 scripts/format.py --all --fix --noconfirm
-
 format-check:
 	python3 scripts/format.py --all --check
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -38,9 +38,6 @@ format-fix:
 	rm -rf src/amalgamation/*
 	python3 scripts/format.py --all --fix --noconfirm
 
-format:
-	python3 scripts/format.py --all --fix --noconfirm
```
Tishj, line 42: I added this because somehow our `main` didn't have correct formatting
I use `make format-main` to avoid reformatting everything normally - but `make format` doesn't work correctly for the case where this *is* necessary

So I would prefer to keep this, the `extension-ci-tools` format is not doing what it's supposed to, I prefer overriding it with this.

Tmonster: does `format-fix` not do the exact same thing?
And you also can use `make format-head` to just fix files that have changed that are tracked right?

Tishj: ah you're entirely right, I never use that, I was expecting that to have a different meaning 😅  and I got thrown off by the `rm -rf` in my initial scan, looking for the functionality

Tishj (approved):
Thanks 👍

PR #163, opened 2025-04-08
Update Iceberg to work with the latest MultiFileReader changes (for v1.3)

Tishj:
These latest changes have broken the old way we handled positional deletes, which consisted of:
- pushing the `file_row_number` option onto the underlying reader.
- saving the "global" index of this `file_row_number`
- in `FinalizeChunk` look up the "local" index of the `file_row_number` column and use it to apply the positional deletes (with a Slice)

The new method is:
- in the `FinalizeBind` set the `deletion_filter` of the `BaseFileReader` (underlying reader base class)
- Inside the parquet scan this now uses this `deletion_filter` when it's performing the scan, in the same path that it handles pushed down filters.

This is both less intrusive and less complicated, and more performant as now these rows don't even have to be scanned from the file, instead of being scanned and then later removed.


extension_config.cmake
```diff
@@ -2,8 +2,8 @@
 
 duckdb_extension_load(avro
         LOAD_TESTS
-        GIT_URL https://github.com/duckdb/duckdb_avro
-        GIT_TAG aa2741544c5b53cc8fda6a85ca56bd0353116de4
+        GIT_URL https://github.com/tishj/duckdb_avro
+        GIT_TAG 9735db41b8fa1d409eb4d0e20e3e16499da503f9
```
Tishj: This should point at `duckdb/duckdb_avro`, for that the PR needs to be merged there first

.github/workflows/MainDistributionPipeline.yml
```diff
@@ -14,21 +14,21 @@ concurrency:
...
-    uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@v1.2.1
+    uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@main
     with:
       extension_name: iceberg
-      duckdb_version: v1.2.2
-      ci_tools_version: v1.2.1
-      exclude_archs: 'wasm_mvp;wasm_eh;wasm_threads;windows_amd64_rtools;windows_amd64_mingw'
+      duckdb_version: main
+      ci_tools_version: main
+      exclude_archs: 'wasm_mvp;wasm_eh;wasm_threads;windows_amd64_rtools;windows_amd64_mingw;linux_amd64_musl;linux_amd64_gcc4'
```
Tishj: `linux_amd64_musl` can likely be enabled again once we add:
```
with:
    extra_toolchains: 'python3'
```

src/common/iceberg.cpp
```diff
@@ -46,9 +46,22 @@ IcebergTable IcebergTable::Load(const string &iceberg_path, IcebergSnapshot &sna
 	return ret;
 }
 
-unique_ptr<SnapshotParseInfo> IcebergSnapshot::GetParseInfo(yyjson_doc &metadata_json) {
-	SnapshotParseInfo info {};
-	auto root = yyjson_doc_get_root(&metadata_json);
+unique_ptr<IcebergMetadata> IcebergMetadata::Parse(const string &path, FileSystem &fs,
```
Tishj: This includes changes merged into `main`, we could merge `main` into `feature` again to remove these changes from this diff, if that's preferable?

src/include/iceberg_multi_file_reader.hpp
```diff
@@ -147,39 +123,38 @@ struct IcebergMultiFileReader : public MultiFileReader {
...
-	void BindOptions(MultiFileReaderOptions &options, MultiFileList &files, vector<LogicalType> &return_types,
+	void BindOptions(MultiFileOptions &options, MultiFileList &files, vector<LogicalType> &return_types,
 	                 vector<string> &names, MultiFileReaderBindData &bind_data) override;
 
-	void CreateColumnMapping(const string &file_name, const vector<MultiFileReaderColumnDefinition> &local_columns,
-	                         const vector<MultiFileReaderColumnDefinition> &global_columns,
-	                         const vector<ColumnIndex> &global_column_ids, MultiFileReaderData &reader_data,
-	                         const MultiFileReaderBindData &bind_data, const string &initial_file,
-	                         optional_ptr<MultiFileReaderGlobalState> global_state) override;
+	// FIXME: none of this is virtual anymore
```
Tishj: should remove this

PR #164, opened 2025-04-08
Make local polaris easier

Tmonster:


Tmonster:
Closing for now to reduce clutter, will reopen when needed

PR #166, opened 2025-04-08
[Dev] Fix crash with malformed `version_name_format` parameter.

Tishj:
Further down the codepath this is used with `StringUtil::Format` which throws if these requirements aren't met.


Tmonster (approved):
Looks good to me

PR #170, opened 2025-04-09
Only Load avro on Create secret (TYPE ICEBERG) or ATTACH '' (TYPE ICEBERG)

Tmonster:
Currently if duckdb main is compiled with iceberg, avro is compiled but not linked. Then when a command is executed in iceberg, all extensions are auto loaded, including iceberg. Iceberg then attempts to load avro, which isn't available, causing an exception. This exception is thrown for any command. To precent this, we only load avro when code in the iceberg extension is actually run, i.e during a create secret statement, or an attach statement.


src/iceberg_extension.cpp
```diff
@@ -23,7 +23,18 @@
 
 namespace duckdb {
 
+bool AVRO_LOADED = false;
+
 static unique_ptr<BaseSecret> CreateCatalogSecretFunction(ClientContext &context, CreateSecretInput &input) {
+	if (!AVRO_LOADED) {
+		auto &db_instance = context.db;
+		ExtensionHelper::AutoLoadExtension(*db_instance, "avro");
```
Tishj: AutoLoadExtension throws if it can't, the line below this will never be reached (I know that's also there on main, but if we're changing it, let's fix this as well)

PR #187, opened 2025-04-16
[Dev] Move all the CURL-specific logic to `curl.cpp`

Tishj:
As the title says, this creates a `RequestInput` class, which contains all the info needed by curl to create and perform the request.

The existing logic for this has been moved to `GET`, `POST` and `DELETE`.
The reason for this change is to make it easier for WASM-specific logic to be implemented, since the curl package doesn't work there.

Note that this still doesn't fully enable WASM builds, for that we also need to wrap AWS, which will be done in a follow-up PR.

And note also that this is likely not going to be permanent, because ideally we move the burden of this into HTTPFS and just delegate to that.


Tishj:
Follow up for AWS is [here](https://github.com/Tishj/duckdb_iceberg/tree/wrap_aws_for_wasm)

carlopi:
I will not have a proper chance to review this, I trust you can make a good call, and I can pick up what's needed wasm side.

CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ set(EXTENSION_SOURCES
     src/manifest_reader.cpp
     src/catalog_api.cpp
     src/catalog_utils.cpp
+	src/curl.cpp
```
Tmonster: nit: whitespace

src/curl.cpp
```diff
@@ -0,0 +1,183 @@
...
+	if (res != CURLcode::CURLE_OK) {
+		string error = curl_easy_strerror(res);
+		throw HTTPException(StringUtil::Format("Curl DELETE Request to '%s' failed with error: '%s'", url, error));
+	}
+	return result;
+}
+
+string RequestInput::PostRequest(ClientContext &context, const string &post_data) {
+	CURLRequestHeaders curl_headers;
+	for (auto &header : headers) {
```
Tmonster: nit: maybe this could be a reusable function? Something like 
```
CURLRequestHeaders curl_headers = InitializeCurlHeaders();
```
I see this in Post, Get & Delete. We may end up adding extra calls like PUT or HEAD too

Tmonster (commented):
just some small comments, otherwise looks good to me

src/curl.cpp
Tishj: moved it to the constructor 👍

Issue #185, opened 2025-04-15
Invalid Input Error: Function with name "read_avro" not found in ExtensionUtil::GetTableFunction


Fokko:
Since yesterday we've noticed an issue with the PyIceberg CI when we do integration testing with PyIceberg:

```python
    @pytest.mark.integration
    def test_duckdb_url_import(warehouse: Path, arrow_table_with_null: pa.Table) -> None:
        os.environ["TZ"] = "Etc/UTC"
        time.tzset()
        tz = pytz.timezone(os.environ["TZ"])
    
        catalog = SqlCatalog("test_sql_catalog", uri="sqlite:///:memory:", warehouse=f"/{warehouse}")
        catalog.create_namespace("default")
    
        identifier = "default.arrow_table_v1_with_null"
        tbl = _create_table(catalog, identifier, {}, [arrow_table_with_null])
        location = tbl.metadata_location
    
        import duckdb
    
        duckdb.sql("INSTALL iceberg; LOAD iceberg;")
>       result = duckdb.sql(
            f"""
        SELECT *
        FROM iceberg_scan('{location}')
        """
        ).fetchall()
E       duckdb.duckdb.InvalidInputException: Invalid Input Error: Function with name "read_avro" not found in ExtensionUtil::GetTableFunction
```

Does this sounds a bell?

kevinjqliu:
Weird, this is definitely on the duckdb side. Its using the `read_avro` function from the avro extension
https://github.com/search?q=repo%3Aduckdb%2Fduckdb-iceberg+ExtensionUtil%3A%3AGetTableFunction+read_avro&type=code

I'd assume its auto loaded. let me open an issue on duckdb side

kevinjqliu:
Looks like we need to add auto load `avro` extension here
https://github.com/duckdb/duckdb-iceberg/blob/94d95da21bb8d839743c39796101fd458de377a0/src/iceberg_extension.cpp#L238-L242

Tishj:
Yea this sounds like what we have in our tests:
```sql
require avro

require parquet

require iceberg
```

It essentially loads the extensions or fails (skips the test) if the extension isn't built/available

kevinjqliu:
I think this should be fixed internally by duckdb. Otherwise, clients have to always do `INSTALL avro` when using the iceberg extension

Tmonster:
[...]

PR #189, opened 2025-04-16
only load avro when during the avro scan call

Tmonster:
fixes https://github.com/duckdb/duckdb-iceberg/issues/185

### DIFF
--- a/src/iceberg_manifest.cpp
+++ b/src/iceberg_manifest.cpp
@@ -1,5 +1,7 @@
 #include "manifest_reader.hpp"
 #include "iceberg_manifest.hpp"
+#include "iceberg_extension.hpp"
+#include "duckdb/main/extension_helper.hpp"
 #include "duckdb/main/database.hpp"
 
 //! Iceberg Manifest scan routines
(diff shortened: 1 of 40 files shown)


Tmonster:
Ah, that makes sense, I was wondering why it was passing locall

Tishj:
This is *almost* patchable, because we have `extension_directory`, but the problem with that is that it still expects a certain directory structure starting at `extension_directory`, which doesn't align with how extensions are placed in `build/<build_type>/repository`

src/iceberg_manifest.cpp
```diff
@@ -221,6 +223,11 @@ void IcebergManifestEntryV2::PopulateNameMapping(idx_t column_id, const LogicalT
 
 AvroScan::AvroScan(const string &scan_name, ClientContext &context, const string &path) : context(context) {
 	auto &instance = DatabaseInstance::GetDatabase(context);
+	if (!__AVRO_LOADED__) {
+		ExtensionHelper::AutoLoadExtension(instance, "avro");
+		__AVRO_LOADED__ = true;
+	}
+
 	auto &avro_scan_entry = ExtensionUtil::GetTableFunction(instance, "read_avro");
```
kevinjqliu, line 231: nit: perhaps we want to wrap extension loading and `GetTableFunction` into a single helper function. Otherwise, I can see this problem reoccurring the next time `read_avro` is used in a different path

kevinjqliu (approved):
LGTM!

src/iceberg_manifest.cpp
Tmonster: Normally we like to load all necessary extensions in `LoadInternal` in `iceberg_extension.cpp` so that avro loading happens during the iceberg extension load. The problem is that in our CI, we test compiling/linking all the extensions together to verify behavior between extensions. This uncovered a problem where Avro has conflicts with some of our other extensions, so we don't link it. Not linking the extension means we can't auto load it in our tests (but `require avro` still works). Since avro cannot be auto-loaded, all tests that auto-load iceberg fail if avro gets autoloaded. The temporary solution for now is to only load avro when the avro scan function is used. In the future we plan to link the avro extension so we can load it when the iceberg extension gets loaded

PR #191, opened 2025-04-16
Wrap AWS for WASM

Tishj:
Follow up to #187


src/common/api_utils.cpp
```diff
@@ -1,131 +1,23 @@
...
-	} else {
-		Aws::StringStream resBody;
-		resBody << res->GetResponseBody().rdbuf();
-		throw HTTPException(StringUtil::Format("Failed to query %s, http error %d thrown. Message: %s",
-		                                       req->GetUri().GetURIString(true), res->GetResponseCode(),
-		                                       resBody.str()));
-	}
+string &APIUtils::GetCURLCertPath() {
+	static string cert_path = "";
+	return cert_path;
```
Tishj: Btw, I just noticed we have a bit of complexity around the curl cert path, by changing it to a singleton with a static lifetime we could also change it to:
```c++
static string cert_path = SelectCURLCertPath();
return cert_path;
```

Because of the `static` lifetime, this guarantees that `SelectCURLCertPath()` is only called once, to initialize the value

Issue #147, opened 2025-04-02
Column projection support


regadas:
Hello!

First, of all thanks for all of the hard work on this. It's shaping up quite nicely!

Not sure if this is tracked in some other issue, might have missed, but I noticed that it's currently missing support for column projection, more precisely support for `schema.name-mapping.default` property.

I noticed this because some of my Iceberg Tables have this property in order to do column projection on data files that lack field-ids.

Let me know if I'm missing something and support is already here 😄 

Thank you!

Tishj:
Support is there for name-mapping in duckdb core, that is the default column mapping way in Delta, but I don't think we have any way to signal that name-mapping should be used in iceberg currently.

kevinjqliu:
we had a similar issue in the [iceberg slack channel](https://apache-iceberg.slack.com/archives/C029EE6HQ5D/p1743659464231449)

> It would be great if you could add information here about how we could generate a table that uses this, so we can test against it, or upload a table that uses this that we can use 👍


The best way to do this is to use pyiceberg's `add_files` API https://py.iceberg.apache.org/api/#add-files

kevinjqliu:
copying over from the slack thread, this can be a good example to reproduce the issue

```
import duckdb
import os
from pyiceberg.catalog import load_catalog
import pyarrow.parquet as pq
import pyarrow as pa

# Ensure necessary directories exist
os.makedirs("warehouse/example", exist_ok=True)

# Create a sample Parquet file
sample_data = pa.table({
    "id": [1, 2, 3],
    "name": ["Alice", "Bob", "Charlie"],
    "age": [25, 30, 35]
})
pq.write_table(sample_data, "warehouse/example/mock_data.parquet")

# Connect to DuckDB and install necessary extensions
conn = duckdb.connect("duck.local.db")
conn.execute("""
    INSTALL iceberg;
    LOAD iceberg;
""")

def add_parquet_file_to_table(uri: str, namespace: str, catalog_name: str, table_name: str, parquet_path: str):
    """Registers a Parquet file into an Iceberg table."""
    
    # Load Iceberg catalog
    catalog = load_catalog(catalog_name, uri=uri, warehouse='warehouse')
    
    # Read and clean Parquet schema
    parquet_table = pq.read_table(parquet_path)
    parquet_schema = parquet_table.schema.remove_metadata()
    
    # Create ns, and table
    catalog.create_namespace_if_not_exists(namespace)
[...]

PR #192, opened 2025-04-17
Column name mapping support

Tishj:
This PR implements #147


data/column_mapping/warehouse/default.db/my_table/metadata/00000-aa288ee6-3a21-4880-a1a4-91333f273075.metadata.json
```diff
@@ -0,0 +1 @@
+{"location":"data/column_mapping/warehouse/default.db/my_table","table-uuid":"199c7f6d-3808-4a95-84db-f3cf06322240","last-updated-ms":1744881705813,"last-column-id":3,"schemas":[{"type":"struct","fields":[{"id":1,"name":"id","type":"long","required":false},{"id":2,"name":"name","type":"string","required":false},{"id":3,"name":"age","type":"long","required":false}],"schema-id":0,"identifier-field-ids":[]}],"current-schema-id":0,"partition-specs":[{"spec-id":0,"fields":[]}],"default-spec-id":0,"last-partition-id":999,"properties":{},"snapshots":[],"snapshot-log":[],"metadata-log":[],"sort-orders":[{"order-id":0,"fields":[]}],"default-sort-order-id":0,"refs":{},"statistics":[],"format-version":2,"last-sequence-number":0}
```
Tmonster, line 1: nit: could you put these files in data/persistent/column_mapping?

Tmonster (commented):
Awesome thanks!
Couple of questions,

Is there no way to get the tests working with our current test infra?

Can we also check renaming of columns to different (previously valid names?) and that  the contents of the tables is consistent? Something like the following tests should be good

check column drop also drops values
```
create table t1 as select range a, random()*1000 b from range(10000);
alter table t1 drop column b;
alter table t1 add column b;
# in test
select count(*) from t1 where b is null
```

check column rename to other (previously valid) column name also preserves correct values.
```
create table t1 as select range a, random()*1000 b from range(10000);
alter table t1 rename column a to c;
# in test
select c from t1;
# verify selecting a is an error
select a from t1;
```

```
create table t1 as select range a, random()*1000 b from range(10000);
alter table t1 drop column b;
alter table t1 rename column a to b;
# in test
# verify a is the range column
# verify selecting a results in error
select a from t1;
```

kevinjqliu:
heres where you can find some tests related to name-mapping
https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/mapping/TestMappingUpdates.java
https://github.com/apache/iceberg/blob/main/core/src/test/java/org/apache/iceberg/TestSchemaAndMappingUpdate.java

PR #193, opened 2025-04-17
MainDistributionPipeline.yml: publish also v1.2-histrionicus

carlopi:
Enable publishing to `core_nightly` also v1.2-histrionicus branch.

PR #198, opened 2025-04-22
Only load avro when avro is needed

Tmonster:
This is basically a merge of v1.2-histrionicus, without the PR https://github.com/duckdb/duckdb-iceberg/pull/193.

I would normally open a straight merge of v1.2, but #193 (also in v1.2) changes main distribution pipeline to also publish to v1.2-histrionicus, which we don't want in main


kevinjqliu (approved):
lgtm

Tishj (approved):
Looks good, just have a look at the PR I referenced, I cleaned up the loading code a little, most notably the global we're using here is not needed

Tishj (approved):

PR #201, opened 2025-04-24
Implement equality deletes

### LINKED ISSUE #5: Support Equality based deletes
Currently the extension supports only position based delete files, it doesn't handle equality based deletes.

Tishj:
This PR should fix #5 

I have had a hard time generating test data for this, contributions would be very welcome in this regard.
That being said I have verified the correctness with a table, it's sadly not something I can add to the repository however.


src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -474,6 +587,112 @@ void IcebergMultiFileReader::FinalizeChunk(ClientContext &context, const MultiFi
 	// Base class finalization first
 	MultiFileReader::FinalizeChunk(context, bind_data, reader, reader_data, input_chunk, output_chunk, executor,
 	                               global_state);
+
+	D_ASSERT(global_state);
```
Tishj: Move to a separate method (ApplyEqualityDeletes)

PR #202, opened 2025-04-25
Add support for the r2 catalog

Tmonster:
Tried to get the R2 catalog to work with our current iceberg implementation. Turns out we were basically all of the way there, except the R2 catalog returns "/" as a prefix, which breaks our scoped secrets functionality. When looking for the secret again when executing the remote read on file 's3://r2-catalog' the scope "/" doesn't match, so we don't retrieve the correct secret. 

This PR just checks if "/" is the prefix, and if so, uses a different value for the prefix. Also happy to just not use a scope at all in these cases.


Tmonster:
> I also wonder why / as a prefix doesn't work. As long as there are no credentials with a more specific scope matching the path it should still get used, no?

I think it has to do with our implementation of scope, where scope has to match the beginning of the files, so it needs to start with `s3://`, `gcs://`. Maybe @samansmink will have a better idea? 

Also, if you want to test this yourself, I found this blog to be super easy to follow, you will need to sign up for cloud flare yourself, but maybe Gabor can let us use the company account, and then we can also add this to CI
https://developers.cloudflare.com/r2/data-catalog/

Tmonster:
Hmm, maybe a better fix then is to look for `gcs://`, `s3://` etc in the provided prefix? If none of those are found, we can assume the file type from the secrets, and then we can construct the prefix ourselves.

Tishj:
> Hmm, maybe a better fix then is to look for `gcs://`, `s3://` etc in the provided prefix? If none of those are found, we can assume the file type from the secrets, and then we can construct the prefix ourselves.

That feels a little needlessly limiting, I think this should be fixed in core.
When receiving a scope for a secret, that scope has to start with the prefix of the filesystem, otherwise we add it.
So `/` would become `s3://` if the secret type is S3

PR #204, opened 2025-04-28
Estimate Cardinalities based on manifest list

Tmonster:
Using the Added count we can get a good idea of what the cardinality of the table is. 

Ideally, it would be better to get the record count from the manifest file instead of the `added_record_counts` from the manifest list, but the two track similar information, it's just one is on the snapshot level, and the other is at the data level.

After a little bit of testing this seems to be sorta accurate, which is the most we can hope for. Potentially in the next release we read from the manifest files.


src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -71,15 +71,14 @@ idx_t IcebergMultiFileList::GetTotalFileCount() {
 }
 
 unique_ptr<NodeStatistics> IcebergMultiFileList::GetCardinality(ClientContext &context) {
-	auto total_file_count = IcebergMultiFileList::GetTotalFileCount();
-
-	if (total_file_count == 0) {
-		return make_uniq<NodeStatistics>(0, 0);
+	idx_t cardinality = 0;
```
Tishj, line 79: I think this needs to call GetFile just like `GetTotalFileCount` was doing, because we need to make sure we've looked through all manifests at this point

src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -312,7 +311,8 @@ void IcebergMultiFileReader::FinalizeBind(MultiFileReaderData &reader_data, cons
 	auto &data_file = multi_file_list.data_files[file_id];
 
 	// The path of the data file where this chunk was read from
-	auto &file_path = data_file.file_path;
+	// Copy the file path otherwise it gets overwritten somewhere
+	const auto file_path = data_file.file_path;
```
Tishj: This sounds wrong to me, that would indicate that the MultiFileList is either not const, or not living long enough (if it gets destroyed before we use the data_file referenced here)

src/iceberg_manifest.cpp
```diff
@@ -58,6 +60,7 @@ idx_t IcebergManifestV2::ProduceEntries(DataChunk &chunk, idx_t offset, idx_t co
 		manifest.manifest_path = manifest_path[index].GetString();
 		manifest.content = IcebergManifestContentType(content[index]);
 		manifest.sequence_number = sequence_number[index];
+		manifest.added_rows_count = added_rows_count[index];
```
Tishj, line 70: Judging from https://iceberg.apache.org/spec/#manifest-lists, this looks like we should also use the `existing_rows_count`, for files that aren't added by the snapshot, but are still part of the manifest (?)

src/iceberg_functions/iceberg_multi_file_reader.cpp
Tishj: Not immediately relevant here, but after we implement partition-based filtering we should look to improve this, *if* `GetCardinality` is called after `ComplexFilterPushdown`, then we know which manifests are relevant for the scan.

src/include/iceberg_types.hpp
```diff
@@ -70,6 +70,8 @@ struct IcebergManifest {
 	int64_t sequence_number;
 	//! either data or deletes
 	IcebergManifestContentType content;
+	//! added rows in the manifest
+	idx_t added_rows_count;
```
Tishj: I noticed `IcebergManifestV1::ProduceEntries` doesn't do anything with the `added_rows_count`, which is fine, since the field is optional, I'm fine with ignoring it in V1 for now. But then this should be initialized with a default value.

Tishj:
We're missing the `many_adds_delete` table from the generate script

src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -312,7 +322,7 @@ void IcebergMultiFileReader::FinalizeBind(MultiFileReaderData &reader_data, cons
 
 	// The path of the data file where this chunk was read from
 	// Copy the file path otherwise it gets overwritten somewhere
```
Tishj: TODO: remove this

src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -71,15 +71,24 @@ idx_t IcebergMultiFileList::GetTotalFileCount() {
...
-	if (total_file_count == 0) {
-		return make_uniq<NodeStatistics>(0, 0);
+	if (snapshot.iceberg_format_version == 1) {
+		//! We collect no cardinality information from manifests for V1 tables.
+		return nullptr;
 	}
 
-	// FIXME: visit metadata to get a cardinality count
+	//! Make sure we have fetched all manifests
+	(void)GetTotalFileCount();
```
Tishj, line 87: We should look into using the `summary` field of the snapshot (https://iceberg.apache.org/spec/?h=optional+sn#optional-snapshot-summary-fields), probably the `total-records` is the current cardinality

kevinjqliu: the summary fields `total-records` is not entirely accurate, see https://github.com/apache/iceberg/issues/12823

Tishj: Hmm, can a different part of the summary be used for the cardinality estimation?

We're mostly concerned with the cost of expanding the manifests to get this information

Though admittedly this could be optimized to only read the manifest_list, without traversing into the manifests's avro files

kevinjqliu: if its just an estimate, you might get away with using the summary. 

> We're mostly concerned with the cost of expanding the manifests to get this information

[`record_count` in the manifest](https://iceberg.apache.org/spec/#manifests:~:text=103-,record_count,-long) is the correct cardinality. If `GetCardinality` is on the read path, I think you'd have to read the manifest files at a later time anyways. So maybe we can cache it and amortize the cost

Tishj: Thanks, we'll stick with `existing_rows_count` + `added_rows_count` - `deleted_rows_count` 👍

Fokko: This can still be off, since with merge-on-read delete files, rows might be deleted twice since Iceberg uses optimistic concurrency. If an off-by-one isn't a problem, I think that's a pretty good approach.

Tishj: It's not a problem, it's an estimate

samansmink (approved):

PR #206, opened 2025-04-29
Refactor `generate_data`

Tishj:
The previous code for `generate_data.py` coupled a set of tests to a connection, which isn't necessary and resulted in a lot of duplication / syncing issues.

This new code separates `tests` and `connections` (while preserving (almost) all the existing logic)

An `IcebergTest` has a `generate` method, and an `__init__` (constructor), where it sets up any prerequisites it might need.
Most tests consists of running a set of `.sql` files, but more varied logic is possible, as seen from the `tpch` test.

### To add a test (generate an iceberg table):
- Create a new folder in `scripts/data_generators/tests`
- Write the `__init__.py`, which is usually as simple as:
```py
from scripts.data_generators.tests.base import IcebergTest
import pathlib

@IcebergTest.register()
class Test(IcebergTest):
    def __init__(self):
        path = pathlib.PurePath(__file__)
        super().__init__(path.parent.name)
```
- Add the `.sql` files to the folder (one statement per file)

### To add a new connection (new iceberg catalog type to test against):
- Add an option to the choices for `targets`, in `generate_data.py`
- Create a new folder in `scripts/data_generators/connections`
- Create an `__init__.py` file in the folder
- Add a new derived class of `IcebergConnection`
- Add the `@IcebergConnection.register(CONNECTION_KEY)` decorator to it, where `CONNECTION_KEY` is the choice added in step 1
- Define the `get_connection` method for the new class


Tishj:
It would be great if we can automate the testing as well, since all of these connections can make use of the exact same tables now, the only custom logic required for each connection is setting it up.

Perhaps we can move the related `.test` files into the `scripts/data_generators/tests/{test}` folders and generate the final `.test` files instead.
Putting them into `test/sql/generated/{target}/{test}` folders, so as to separate them from the rest of the tests, so we can delete them, exclude them from the repo as well

PR #78, opened 2024-10-16
Add predicate pushdown (updated)

mike-luabase:
This PR enhances query performance by filtering data at the metadata level, reducing the amount of data read during scans. It currently only addresses queries containing a single Iceberg table.

### **Key Changes**

- **Extended `IcebergManifestEntry`:**
  - Added `lower_bounds` and `upper_bounds` maps to store column statistics.
  
- **Utility Function:**
  - Implemented `IcebergUtils::GetFullPath` to resolve file paths accurately.

- **Metadata Retrieval:**
  - Added `GetEntries` template in `IcebergTable` to fetch relevant manifest entries, excluding deleted ones.

- **Predicate Evaluation:**
  - Created `EvaluatePredicateAgainstStatistics` to assess if data files satisfy query predicates based on their statistics.
  - For each predicate:
    - Identifies the column involved.
    - Checks if the column has defined lower and upper bounds.
    - Based on the comparison type (e.g., =, >, <), determines if the predicate can be satisfied given the file's bounds.
    - If any predicate fails, the file is excluded from the scan.

- **Scan Expression Modification:**
  - Updated `MakeScanExpression` to filter `data_file_entries` using predicates before scanning.

- **Binding Function Enhancements:**
  - Enhanced `IcebergScanBindReplace` with additional logging and prepared data files based on predicate results.

This implementation optimizes Iceberg table scans by leveraging metadata for early data filtering, significantly improving query efficiency and resource usage.

PR #209, opened 2025-04-30
Filter `data_file`s using bounds (`lower_bound` and `upper_bound`)

Tishj:
This PR is a revival of #78 made to work against the latest scan changes

Relies on <https://github.com/duckdb/duckdb/pull/17315>


samansmink (approved):
Lookgs great @Tishj!

One note is that I think testing is quite primitive in this PR still. I'm fine with just pushing forwards here and adding more tests in a follow up PR though, because it would likely require adding some logging code similar to what we do in delta to properly test this.

Tishj:
@samansmink I added logging and that is being used in `test/sql/local/iceberg_scans/filtering_on_bounds.test` to test how many records are skipped by the filter.

But yes, the tests are very minimal, we would need a lot more coverage to test the various types.
I think I would like to add pytest tests at some point, mostly because pytest lends itself much better to parametrized tests for coverage

PR #212, opened 2025-05-01
Filter `manifest_file`s using partitions

### LINKED ISSUE #210: Does partition pruning work with iceberg?
Hi There!

I have an Iceberg table on S3 that is partitioned by date and by id, using Hive-style partitioning:
```
poc/iceberg/{dataset}/{table}/partition_date=2025-03-31/partition_id=0/some_file.parquet
```

When trying to query this table with DuckDB, the query never finishes. For reference, the size of the Parquet file inside the partition is 10MB.

Since DuckDB doesn't complete the query, what I suspect is that it's not accessing the partition specified in the query. Below is my code:
```python
def main():
    START_PARTITION_DATE = "2025-03-01"
    PARTITION_ID = "0"

    con = duckdb.connect(database=':memory:')
    
    # Install and load required extensions
    con.execute("INSTALL aws; LOAD aws;")
    con.execute("INSTALL httpfs; LOAD httpfs;")
    con.execute("UPDATE EXTENSIONS;")
    con.execute("SET unsafe_enable_version_guessing = true;")
    con.execute("SET memory_limit = '1GB';")
    con.execute("SET threads = 4;")
    con.execute("SET enable_progress_bar = true;")
    con.execute("SET allocator_background_threads = true;")
    
    # Configure AWS credentials and enable version guessing for Iceberg
    con.execute(f"""
        SET s3_region='{REGION}';
        SET s3_access_key_id='{ACCESS_KEY_ID}';
        SET s3_secret_access_key='{SECRET_ACCESS_KEY}';
    """)
    
    # Then query with filters
    sql = f"""
        SELECT COUNT(*)
        FROM iceberg_scan(
            's3://{BUCKET}/poc/iceberg/{DATABASE}/{TABLE}'
        )
        WHERE partition_date = '{START_PARTITION_DATE}'
        AND partition_id = '{PARTITION_ID}'
    """
    
    res = con.execute(sql)
    print(res.fetchdf().to_clipboard())
```

I ran a test recreating the same table with Avro as the data file format, and in this test, DuckDB returned the following error:
```
Traceback (most recent call last):
  File "/home/Wuerike.Cavalheiro/stone/scripts/iceberg_poc/engines/duckdb/iceberg.py", line 99, in <module>
    main()
  File "/home/Wuerike.Cavalheiro/stone/scripts/iceberg_poc/engines/duckdb/iceberg.py", line 77, in main
    res = con.execute(sql)
          ^^^^^^^^^^^^^^^^
duckdb.duckdb.InvalidInputException: Invalid Input Error: No magic bytes found at end of file 's3a://pay-data-layer-warehouse-stg/poc/iceberg/poc_iceberg/poc_iceberg_charges_events_avro/data/partition_date=2025-03-31/partition_id=279/00002-7643-1ad08722-0d3c-434f-9240-7b3fcadcfc07-0-00003.avro'
```

What I suspect is that DuckDB might not yet support reading Iceberg tables with the Avro format—could someone confirm that, please?

But more importantly, the main point we can extract from this error log is the attempt to access `partition_date=2025-03-31/partition_id=279` even though the query specified `partition_date = '2025-03-01' and partition_id = '0'`.

Does partition pruning work with duckdb + iceberg?

Tishj:
This PR fixes #210 and #2

This PR implements filtering based on the partitioning information in the `manifest_file`.
Also clean up a problem with the `is_unpartitioned` check for equality deletes, copying the logic for "unpartitioned" from the iceberg java reference implementation. (I could not find anything about the definition of the "unpartitioned" partition spec in the docs)


src/common/iceberg.cpp
```diff
@@ -8,45 +8,121 @@
...
 	IcebergTable ret;
 	ret.path = iceberg_path;
 	ret.snapshot = snapshot;
 
-	auto &fs = FileSystem::GetFileSystem(context);
-	auto manifest_list_full_path = options.allow_moved_paths
-	                                   ? IcebergUtils::GetFullPath(iceberg_path, snapshot.manifest_list, fs)
-	                                   : snapshot.manifest_list;
 	vector<IcebergManifest> manifests;
+	unique_ptr<ManifestReader> manifest_reader;
```
Tishj, line 18: These changes are just to get this code more in line with the manifest reading logic used by the `iceberg_scan` path

src/iceberg_manifest.cpp
```diff
@@ -306,7 +421,8 @@ idx_t IcebergManifestEntryV2::ProduceEntries(DataChunk &chunk, idx_t offset, idx
 				entry.sequence_number = sequence_numbers[index];
 			} else {
 				//! Value should only be NULL for ADDED manifest entries, to support inheritance
-				D_ASSERT(entry.status == IcebergManifestEntryStatusType::ADDED);
+				//! FIXME: the 'persistent' tables have DELETED entries with no sequence number, while the metadata says
+				//! v2 According to the iceberg spec, those tables are corrupt
```
Tishj: This is really odd, and should be looked into
the `sequence_number` is allowed to be NULL for an entry with the status `ADDED`, so it can inherit from the manifest.
Once it's not ADDED, it should have a sequence number, to indicate that it was created by that sequence

From the spec:
> When writing an existing file to a new manifest or marking an existing file as deleted, the data and file sequence numbers must be non-null and set to the original values that were either inherited or provided at the commit time.

src/iceberg_manifest.cpp
```diff
@@ -20,14 +20,65 @@ idx_t IcebergManifestV1::ProduceEntries(DataChunk &chunk, idx_t offset, idx_t co
...
+				FieldSummary summary;
+				auto list_idx = list_entry.offset + j;
+				if (contains_null) {
+					summary.contains_null = contains_null[list_idx];
+				}
+				if (contains_nan) {
+					summary.contains_nan = contains_nan[list_idx];
+				}
+				if (lower_bound) {
+					summary.lower_bound = lower_bound[list_idx].GetString();
```
samansmink: this needs the same changes as v2 below

src/iceberg_manifest.cpp
samansmink: perhaps this was just written incorrectly by spark at the time this data was generated? we could try to regen the data

Tishj: Made #215 for this 👍

PR #221, opened 2025-05-04
Case sensitive names support

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/4664

Spark can't create tables with case-sensitive names, so I've had to use pyiceberg to create the table (script is included).
This means we can't test the REST catalog support currently.

We might want to expand this test to include tests for struct fields that need deduplication, for now I just wanted to implement basic support.


corleyma:
> Spark can't create tables with case-sensitive names, so I've had to use pyiceberg to create the table (script is included).
This means we can't test the REST catalog support currently.

pyiceberg should be able to write to a REST catalog....

Tishj:
> > Spark can't create tables with case-sensitive names, so I've had to use pyiceberg to create the table (script is included).
> > This means we can't test the REST catalog support currently.
> 
> pyiceberg should be able to write to a REST catalog....

While I'm sure that's true, it's not helpful because our generate script uses Spark, so not being able to use Spark means we can't use the generate script - hence it won't be tested on REST catalogs.

Tishj:
@samansmink I think this is still a problem if we have a struct with case-sensitive field names

But I think that would require *many* changes in core, not sure if we even want to ecknowledge that can of worms

test/sql/local/case_sensitive_names.test
```diff
@@ -0,0 +1,24 @@
...
+require avro
+
+require parquet
+
+require iceberg
+
+require httpfs
+
+query II
+select * from ICEBERG_SCAN('data/persistent/case_sensitive_names/default.db/case_sensitive_names/metadata/00001-a7a3a44c-4aac-4619-bebd-11be37b27351.metadata.json');
```
Tmonster, line 13: can we also get a `describe` statement to test the multiple column names? I.e

```
query I
select column_name from (describe from  ICEBERG_SCAN('data/persistent/case_sensitive_names/default.db/case_sensitive_names/metadata/00001-a7a3a44c-4aac-4619-bebd-11be37b27351.metadata.json'));
----
user_id
uSeR_Id_1
```

Tmonster (commented):
just one nit about selecting column names, otherwise lgtm!

PR #223, opened 2025-05-06
Add partition-based default values and improve partitioning support in general

Tishj:
https://iceberg.apache.org/spec/#column-projection

> Return the value from partition metadata if an [Identity Transform](https://iceberg.apache.org/spec/#partition-transforms) exists for the field and the partition value is present in the partition struct on data_file object in the manifest.
This allows for metadata only migrations of Hive tables.

These `partition` values (the fields of the struct) use primitive types matching the spec of "Single Value Serialization" (<https://iceberg.apache.org/spec/#appendix-d-single-value-serialization>)
Meaning the partition field for a `DATE` column will have an `INTEGER` Value representing the days since epoch (1970-01-01)

This PR adds support for partitioning on:
- `BOOL`
- `INTEGER`
- `BIGINT`
- `FLOAT`
- `DOUBLE`
- `VARCHAR`
- `DATE`
- `TIME`
- `TIMESTAMP`
- `TIMESTAMPTZ`
- `DECIMAL` (smallint, integer, bigint, hugeint)
- `BINARY`
- `FIXED(L)`

`UUID` can't be created by PyIceberg or Spark, would need to be generated using the Java implementation, that's a TODO.
One thing to note there is that the spec says `UUID` is stored as a big-endian 16 byte binary, so this will likely require some transformation of the bytes to little endian.

`TIMESTAMP_NS` and `TIMESTAMPTZ_NS` are not added, because these are part of v3:
> Types variant, geometry, geography, unknown, timestamp_ns, and timestamptz_ns are added in v3.


kevinjqliu:
> See TransformPartitionValue, where the INTEGER that represents the serialized DATE value gets deserialized.

we have similar logic to transform identity partition value here 
https://github.com/apache/iceberg-python/blob/dfae694e151b52608f94a4b20b380afe6d1cf871/pyiceberg/partitioning.py#L426-L461

Tishj:
Thanks that's somewhat helpful, though what I'm looking for here is the reverse, `_from_partition_representation`

Tishj:
Ah never mind, the spec is clear on this after all, I just had trouble finding it
<https://iceberg.apache.org/spec/#appendix-d-single-value-serialization>

Tishj:
I'm not a huge fan of the large amount of files this adds, but neither pyiceberg or spark omits the column that is partitioned on in the parquet file.
So we have to manually strip the column(s), which is more difficult in a non-persistent environment

mike-luabase:
This is amazing, thanks @Tishj !

extension_config.cmake
```diff
@@ -1,5 +1,6 @@
 # This file is included by DuckDB's build system. It specifies which extension to load
 
+duckdb_extension_load(icu)
```
Tishj, line 3: Note: `icu` is added because we're adding a test for `timestamptz`, for timezones we need `icu` in duckdb

src/manifest_reader.cpp
```diff
@@ -25,11 +25,13 @@ void ManifestReader::SetPartitionSpecID(int32_t partition_spec_id_p) {
...
 	const bool first_init = scan == nullptr;
 	scan = std::move(scan_p);
-	if (first_init) {
-		scan->InitializeChunk(chunk);
-	} else {
-		chunk.Reset();
+	if (!first_init) {
+		chunk.Destroy();
 	}
+	//! Reinitialize for every new scan, the schema isn't guaranteed to be the same for every scan
```
Tishj, line 31: Note: this came up because of the evolved partition spec.
The schema of the `partition` column of the `data_file` is affected by this.

This will likely proof problematic when we in the future try to use the MultiFileReader for the manifest reading, as the schemas can't be cleanly unified.

src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -727,6 +761,121 @@ static void ApplyFieldMapping(MultiFileColumnDefinition &col, vector<IcebergFiel
 	}
 }
 
+static Value TransformPartitionValue(const Value &value, const LogicalType &type) {
```
Tishj, line 621: This is kind of unfortunate actually, this uses the exact same spec as `lower_bound` and `upper_bound` serialization (https://iceberg.apache.org/spec/#appendix-d-single-value-serialization).

But the problem is that the `partition` field is a struct with an actual type, rather than just `BLOB`.
We might want to force reading the non-nested fields as BLOB, just so we can re-use the existing logic.

Tishj: Cooked this up to be able to reuse the same logic:
```c++
static Value TransformPartitionValueFromBlob(const string_t &blob, const LogicalType &type) {
	auto result = IcebergValue::DeserializeValue(blob, type);
	if (result.HasError()) {
		throw InvalidConfigurationException(result.GetError());
	}
	return result.GetValue();
}

template <class T>
static Value TransformPartitionValueTemplated(const Value &value, const LogicalType &type) {
	T val = value.GetValue<T>();
	string_t blob((const char *)&val, sizeof(T));
	return TransformPartitionValueFromBlob(blob, type);
}

static Value TransformPartitionValue(const Value &value, const LogicalType &type) {
	D_ASSERT(!value.type().IsNested());
	switch (value.type().InternalType()) {
	case PhysicalType::BOOL:
		return TransformPartitionValueTemplated<bool>(value, type);
	case PhysicalType::INT8:
		return TransformPartitionValueTemplated<int8_t>(value, type);
	case PhysicalType::INT16:
		return TransformPartitionValueTemplated<int16_t>(value, type);
	case PhysicalType::INT32:
		return TransformPartitionValueTemplated<int32_t>(value, type);
	case PhysicalType::INT64:
		return TransformPartitionValueTemplated<int64_t>(value, type);
	case PhysicalType::INT128:
		return TransformPartitionValueTemplated<hugeint_t>(value, type);
	case PhysicalType::FLOAT:
		return TransformPartitionValueTemplated<float>(value, type);
	case PhysicalType::DOUBLE:
		return TransformPartitionValueTemplated<double>(value, type);
	case PhysicalType::VARCHAR: {
		return TransformPartitionValueFromBlob(value.GetValueUnsafe<string_t>(), type);
	}
	default:
		throw NotImplementedException("TransformPartitionValue: Value: '%s' -> '%s'", value.ToString(),
		                              type.ToString());
	}
}
```

PR #226, opened 2025-05-08
Schema evolution

Tishj:
This PR adds tests for schema evolution:
- int -> bigint
- float -> double
- decimal(12,8) -> decimal(18,8)
- struct rename, evolve, add, drop column
- map value evolution (key is not allowed to evolve)
- list element evolution

EDIT:
As pointed out by @samansmink, Iceberg can evolve structs inside of the list `element` and the map `value`, which we're currently missing support for in core (in the MultiFileReader), once this is fixed in core we need to add tests and support for that here.


src/common/schema.cpp
```diff
@@ -6,46 +6,49 @@ namespace duckdb {
...
-		return ParseStruct(type.struct_type);
+unique_ptr<IcebergColumnDefinition>
+IcebergColumnDefinition::ParseType(const string &name, int32_t field_id, bool required, rest_api_objects::Type &type,
+                                   optional_ptr<rest_api_objects::PrimitiveTypeValue> initial_default) {
+	auto res = make_uniq<IcebergColumnDefinition>();
+	res->id = field_id;
+	res->required = required;
+	res->name = name;
+	//! FIXME: parse the 'initial_default' into a Value
+
```
Tishj, line 17: This has changed the most, it now produces (potentially) nested `IcebergColumnDefinition`s
All the other changes kind of stem from this, previously we had `vector<IcebergColumnDefinition>`, which is now `vector<unique_ptr<IcebergColumnDefinition>`

Because of that, `IRCAPITable` can't be copied, and `IcebergSnapshot` as well. I've wrapped `IcebergSnapshot` into a `shared_ptr` because it is assigned to the new MultiFileList in `ComplexFilterPushdown`.
We likely want to cache this somewhere (the transaction?) instead, and just use `optional_ptr` in these places, but that can happen separately.

src/common/schema.cpp
```diff
@@ -6,46 +6,49 @@ namespace duckdb {
...
-	if (type.has_struct_type) {
-		return ParseStruct(type.struct_type);
+unique_ptr<IcebergColumnDefinition>
+IcebergColumnDefinition::ParseType(const string &name, int32_t field_id, bool required, rest_api_objects::Type &type,
+                                   optional_ptr<rest_api_objects::PrimitiveTypeValue> initial_default) {
+	auto res = make_uniq<IcebergColumnDefinition>();
+	res->id = field_id;
+	res->required = required;
+	res->name = name;
+	//! FIXME: parse the 'initial_default' into a Value
```
Tishj, line 16: We didn't implement this before either, but this is definitely something important we're missing

Tishj:
We might want to add map/list inside the struct and outside, with some evolution of the `element` and `key` and `value` types

src/common/schema.cpp
Tishj: Just an extra note:
> Note that default values are serialized using the JSON single-value serialization in [Appendix D](https://iceberg.apache.org/spec/#appendix-d-single-value-serialization).

This uses the same (de)serialization as we use for the `lower_bound` and `upper_bound`, so we can reuse that logic

samansmink (approved):

PR #228, opened 2025-05-09
Fixed some race conditions in caching logic

NiclasHaderer:
Even if `HasCachedValue` returns true, it is possible that `GetCachedValue` still throws an exception because the cached value has expired between the two function calls. 

I also added a mutex to prevent conflicts when the cached value is simultaneously written and read

### DIFF
--- a/src/catalog_api.cpp
+++ b/src/catalog_api.cpp
@@ -45,10 +45,12 @@ static string GetTableMetadataCached(ClientContext &context, IRCatalog &catalog,
 	url.AddPathComponent(schema);
 	url.AddPathComponent("tables");
 	url.AddPathComponent(table);
-	if (catalog.HasCachedValue(url.GetURL())) {
-		return catalog.GetCachedValue(url.GetURL());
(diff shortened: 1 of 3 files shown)


Tishj:
Thanks this looks great!

I have been meaning to rework this a bit, because currently we load the metadata once and never refresh, which is problematic.
If the catalog has been updated since we attached, we would never get that update until we detach and re-attach

NiclasHaderer:
> I have been meaning to rework this a bit, because currently we load the metadata once and never refresh, which is problematic.

That's great to hear! I think you refresh once the cache expired, or did I read that wrong?

Tishj:
> > I have been meaning to rework this a bit, because currently we load the metadata once and never refresh, which is problematic.
> 
> That's great to hear! I think you refresh once the cache expired, or did I read that wrong?

Hmm maybe? But we should just always fetch the latest metadata for the table when the catalog gets hit

Tishj (approved):
Thanks, the changes look good to me!

PR #229, opened 2025-05-10
Improve predicate transform architecture

Tishj:
This PR paves the way for implementing transforms other than "identity".

### Refactor

We now parse the `transform` into an `IcebergTransform` structure (which parses the modulo for `bucket` and the width for `truncate`)
Adding functors for transforms in the filter matching logic, which has been moved to `iceberg_predicate.cpp`

### Bug fixes

Also fix a bug where we now properly check the transform type when filtering on the lower+upper bounds of the `data_file`, before we seemed to always assume this used an `identity` transform.

### DIFF
--- a/src/common/iceberg.cpp
+++ b/src/common/iceberg.cpp
@@ -123,14 +123,14 @@ bool IcebergPartitionSpec::IsUnpartitioned() const {
 	return !IsPartitioned();
 }
 
-const IcebergPartitionSpecField &IcebergPartitionSpec::GetFieldBySourceId(idx_t field_id) const {
+const IcebergPartitionSpecField &IcebergPartitionSpec::GetFieldBySourceId(idx_t source_id) const {
 	for (auto &field : fields) {
-		if (field.source_id == field_id) {
+		if (field.source_id == source_id) {
 			return field;
 		}
 	}
 	throw InvalidConfigurationException("Field with source_id %d doesn't exist in this partition spec (id %d)",
-	                                    field_id, spec_id);
+	                                    source_id, spec_id);
 }
 
 static void ParseFieldMappings(yyjson_val *obj, vector<IcebergFieldMapping> &mappings, idx_t &mapping_index,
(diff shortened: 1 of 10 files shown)


src/common/iceberg.cpp
```diff
@@ -123,6 +123,16 @@ bool IcebergPartitionSpec::IsUnpartitioned() const {
 	return !IsPartitioned();
 }
 
+const IcebergPartitionSpecField &IcebergPartitionSpec::GetFieldBySourceId(idx_t field_id) const {
```
Tishj: whoops, this should be `source_id` instead of `field_id`

src/iceberg_predicate.cpp
```diff
@@ -0,0 +1,88 @@
...
+		auto &conjunction_and_filter = filter.Cast<ConjunctionAndFilter>();
+		return MatchBoundsConjunctionAndFilter<TRANSFORM>(conjunction_and_filter, lower_bound, upper_bound, transform);
+	}
+	default:
+		//! Conservative approach: we don't know what this is, just say it doesn't filter anything
+		return true;
+	}
+}
+
+bool IcebergPredicate::MatchBounds(TableFilter &filter, const Value &lower_bound, const Value &upper_bound,
```
Tishj: Perhaps we want to abstract the `lower_bound` + `upper_bound` into a `IcebergPredicateInput` struct, because we likely want to add `nan_count` and `null_count` at some point.

src/iceberg_predicate.cpp
```diff
@@ -0,0 +1,88 @@
...
+                                   const IcebergTransform &transform) {
+	switch (transform.Type()) {
+	case IcebergTransformType::IDENTITY:
+		return MatchBoundsTemplated<IdentityTransform>(filter, lower_bound, upper_bound, transform);
+	case IcebergTransformType::BUCKET:
+		throw NotImplementedException("Transform '%s' not implemented", transform.RawType());
+	case IcebergTransformType::TRUNCATE:
+		throw NotImplementedException("Transform '%s' not implemented", transform.RawType());
+	case IcebergTransformType::YEAR:
+		throw NotImplementedException("Transform '%s' not implemented", transform.RawType());
```
Tishj: Maybe these should just return `true` while they're not implemented, this would preserve the existing behavior of:
```c++
		if (field.transform != "identity") {
			//! FIXME: add support for different transformations
			continue;
		}
```

PR #230, opened 2025-05-10
Add basic support for the `year` transform

Tishj:
Discovered some more details about the iceberg metadata, only the `FieldSummary` bounds are transformed with the `transform` of the partition.

`lower_bounds` and `upper_bounds` of the `data_file` aren't transformed, which makes perfect sense, but we didn't make this distinction before.

`year` is supported for `TIMESTAMP` `TIMESTAMP_TZ` and `DATE`
Note that Spark doesn't support the `year` transform for `DATE` so this path is untested for now.

### `IS_NULL` and `IS_NOT_NULL` filter
Added support for these types, in practice these go through the `ExpressionFilter` it seems.


test/sql/local/partitioning/year/year_timestamp.test
```diff
@@ -0,0 +1,69 @@
...
+----
+12345
+54321
+67890
+88888
+NULL
+
+# Test filtering by year
+query I
+select user_id from ICEBERG_SCAN('data/generated/iceberg/spark-local/default/year_timestamp')
```
samansmink, line 99: could we perhaps use logging to verify that we successfully skip files by the partition transform?

src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -187,23 +187,33 @@ unique_ptr<NodeStatistics> IcebergMultiFileList::GetCardinality(ClientContext &c
...
+static void DeserializeBounds(const Value &lower_bound, const Value &upper_bound, const string &name,
+                              const LogicalType &type, IcebergPredicateStats &out) {
+	if (lower_bound.IsNull()) {
+		out.lower_bound = Value(type);
+	} else {
+		D_ASSERT(lower_bound.type().id() == LogicalTypeId::BLOB);
+		auto lower_bound_blob = lower_bound.GetValueUnsafe<string_t>();
+		auto deserialized_lower_bound = IcebergValue::DeserializeValue(lower_bound_blob, type);
+		if (deserialized_lower_bound.HasError()) {
+			throw InvalidConfigurationException("Column %s lower bound deserialization failed: %s", name,
```
samansmink, line 199: Note sure about `InvalidConfigurationException`, I would much prefer an `IOException`

samansmink (commented):
LGTM!

src/iceberg_functions/iceberg_multi_file_reader.cpp
Tishj: @Tmonster we had some run-ins before with the use of IOException being a source of contention

Do you agree with using InvalidConfiguration here or is IO suitable here?

Tishj:
@samansmink added logging to show many records were skipped at the manifest level for each query

src/iceberg_functions/iceberg_multi_file_reader.cpp
Tmonster: Yea, mother duck was complaining that throwing an IOException everywhere made it hard to distinguish between system-level I/O failures in duckdb vs malformed Iceberg configurations. 

We may have corrected too hard by putting InvalidConfiguration in a lot of places. Here we are reading a parquet file, so I think IOException is ok.

[Original Issue](https://github.com/duckdb/duckdb-iceberg/pull/134)

Tishj: We're not, we're deserializing a blob coming from either the metadata, the manifest_file or the data_file

Vasampa23:
Will adding below code to DayTransform in iceberg_transform.hpp fix 
switch (constant.type().id()) {
		case LogicalTypeId::TIMESTAMP: {
			auto val = constant.GetValue<timestamp_t>();
			auto days_since_epoch = Timestamp::GetEpochDays(val);
			return Value::INTEGER(days_since_epoch);
		}
		case LogicalTypeId::TIMESTAMP_TZ: {
			auto val = constant.GetValue<timestamp_tz_t>();
			auto days_since_epoch = Timestamp::GetEpochDays(val);
			return Value::INTEGER(days_since_epoch);
		}
		case LogicalTypeId::DATE: {
			auto val = constant.GetValue<date_t>();
			auto days_since_epoch = Date::EpochDays(val);
			return Value::INTEGER(days_since_epoch);
		}
		
along with timestamp.cpp adding below
int64_t Timestamp::GetEpochDays(timestamp_t timestamp) {
	D_ASSERT(Timestamp::IsFinite(timestamp));
	return static_cast<int32_t>(timestamp.value / Interval::MICROS_PER_DAY);
}

PR #234, opened 2025-05-13
Initial defaults for primitives

Tishj:
This PR implements parsing `initial-default` values to get used when the field is missing from the parquet file.

All types are added (including `uuid`, using the unreleased latest version of pyiceberg) to the table with an initial default.
Also added all types as fields to a struct, to test that the initial defaults are also propagated there correctly.

### Misc

Updated the `jar` we ship for our test generation, this was in an attempt to get Spark to set `initial-default` when a column is added with a DEFAULT value - it did not help, but it can't hurt to update to the latest version.

The generate script was updated because defaults for `long` where being truncated to `int` (int32_t), I've decided that the spec is underspecified so I'm just making the necessary edits to the `api.yaml` file.

UPDATE:
`yyjson` assigns `uint` to a value if it's not negative, which makes sense.
So if the format is `int64` we now first check `yyjson_is_sint`, if that fails, we fall back to `yyjson_is_int`


Tishj:
This is annoying..
The spec has a `format` to specify a clearer format for things that are ambiguous in json's type system
Such as `integer`

So it annotates things with `int64` to make it clear that it's a signed integer.
And now tests are failing because we respect that `format`, by using `yyjson_is_sint` when the format is `int64` - and then we receive a `uint` for those integers..
```
Invalid Input Error: TableMetadata property 'last_updated_ms' is not of type 'integer', found 'uint' instead
```
        last-updated-ms:
          type: integer
          format: int64
```

Not sure what to do here, except add a fallback of `yyjson_get_int` for every `integer` that has a `format`....

PR #235, opened 2025-05-14
Bump to latest duckdb

Tmonster:


Tishj:
Does this fix the avro issues?
I think Mark might have just merged something that we need to fix in iceberg

Tishj (approved):

PR #239, opened 2025-05-15
[Dev] Clean up IRC `Catalog` code

Tishj:
- `default_schema` is removed.
- `internal_name` was never set, removed.

The `pc_clear_cache` method was removed, this will be replaced with proper transaction-level caching (not in this PR), no need for user level interference to invalidate caching.

Moved the `Attach` code for the IRC Catalog to `irc_catalog.cpp`, this feels like a much more logical place for it than `iceberg_extension.cpp`


src/storage/irc_catalog.cpp
```diff
@@ -115,34 +56,59 @@ optional_ptr<SchemaCatalogEntry> IRCatalog::LookupSchema(CatalogTransaction tran
...
+                                                       unique_ptr<LogicalOperator> plan) {
+	throw NotImplementedException("IRCatalog BindCreateIndex");
+}
+
 bool IRCatalog::InMemory() {
 	return false;
 }
 
 string IRCatalog::GetDBPath() {
-	return internal_name;
```
Tmonster, line 123: Are we sure returning warehouse is what we want? If I run 

```
ATTACH 'warehouse' as my_icebergdb (
   ....
);
show all tables;
```

Does `warehouse` show up or `my_icebergdb`? 
I think `my_icebergdb` is better in this case, since the warehouse could be a confusing name in some cases (for example R2)

Tmonster (commented):
just the one comment. Looks good otherwise. `pc_clear_cache` can be removed, I don't know why it's in there anymore

src/storage/irc_catalog.cpp
Tishj: It was returning `internal_name`, which was never set
so it returned an empty string, whatever `GetDBPath` was used for

But yea sure I can change it to the name of the catalog 👍

Tishj: It seems to be used here:
```c++
void AttachedDatabase::Close() {
	D_ASSERT(catalog);
	if (is_closed) {
		return;
	}
	is_closed = true;

	if (!IsSystem() && !catalog->InMemory()) {
		db.GetDatabaseManager().EraseDatabasePath(catalog->GetDBPath());
	}
```

So I think we want it to be empty actually

Tishj: ahh
```c++
	//! A set containing all attached database path
	//! This allows to attach many databases efficiently, and to avoid attaching the
	//! same file path twice
	case_insensitive_set_t db_paths;
```

This is necessary to prevent attaching the catalog twice
I'll look into what it should actually be

Tishj: I think `warehouse` is actually correct:

in iceberg:
```c++
attach_options.warehouse = info.path;
```

In core:
```c++
	if (options.db_type.empty()) {
		InsertDatabasePath(context, info.path, attached_db->name);
	}
```
`path` here is `info.path` above
```c++
	bool path_exists;
	{
		lock_guard<mutex> path_lock(db_paths_lock);
		path_exists = db_paths.find(path) != db_paths.end();
	}
```

PR #253, opened 2025-05-22
[REST Catalog] Return catalog entry with the correct table schema for time travel queries

### LINKED ISSUE #252: Internal error reading from snapshot
In duckdb v1.3.0 I encountered an error executing 

```
select * from my_datalake.default.pyspark_iceberg_table_v2 AT (VERSION => 8930676650273007362);
```

The snapshot is the last snapshot_id in the manifest.json

Tishj:
This PR fixes #252 

This PR ended up doing a bunch of clean up to get this working correctly
We do away with the `IRCAPITable` struct, as it wasn't adding any value

The `IRCTableSet` entries are now `IcebergTableInformation`, which now holds the `unique_ptr<ICTableEntry>`, 1 for every schema that the table has.

`LoadEntries` no longer creates a `unique_ptr<ICTableEntry>`, this used to make a dummy, so we had something to return as a "candidate" for error messages, retrieved with `Scan`.
Instead we only create the entries with `FillEntry`, and `Scan` now calls `FillEntry`, to keep the same candidate information.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,7 @@ set(EXTENSION_SOURCES
     src/iceberg_extension.cpp
     src/iceberg_functions.cpp
     src/iceberg_manifest.cpp
+    src/iceberg_snapshot_lookup.cpp
     src/catalog_api.cpp
     src/iceberg_logging.cpp
     src/catalog_utils.cpp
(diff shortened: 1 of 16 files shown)


Tishj:
Because we now use `IcebergTableMetadata` in the `IcebergTableInformation` (which is reachable from the created `ICTableEntry`s we can do some optimizing of the `GetScanFunction` path, through the `TableFunctionInfo` we can pass this down to the `iceberg_scan`.

Which means we can avoid doing another http request to get the metadata *again*, when we are coming from a REST Catalog table.

PR #255, opened 2025-05-22
[Dev] Share metadata information, reduce redundant HTTP requests

Tishj:
We split up the `IcebergMultiFileReader` and the `IcebergMultiFileList` into separate files, just to make the code a little easier to navigate.

The real change is the introduction of the `IcebergScanInfo`, this derives from `TableFunctionInfo` and is used to share metadata between the `ICTableEntry` and the `iceberg_scan` table function.

Because of this, we no longer have to perform an extra HTTP request to fetch the metadata *again* inside the `iceberg_scan`.

On top of this, we also remove the extra call to the `/v1/{prefix}/namespaces/{namespace}/tables/{table}` endpoint that we did inside `PrepareIcebergScanFromEntry` to get the table credentials.


src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -0,0 +1,752 @@
...
+
+		auto &filter = *filter_it->second;
+		if (!IcebergPredicate::MatchBounds(filter, stats, field.transform)) {
+			return false;
+		}
+	}
+	return true;
+}
+
+void IcebergMultiFileList::InitializeFiles(lock_guard<mutex> &guard) {
```
dentiny, line 449: `guard` seems unused

PR #259, opened 2025-05-23
[Dev][Caching] Move the `IRCSchemaSet` to the `IRCTransaction`, so it gets refreshed with every new transaction

### LINKED ISSUE #247: can not read new data in iceberg rest catalog after insert new data to iceberg
1. create an iceberg secret `CREATE OR REPLACE SECRET (
     TYPE ICEBERG,
     CLIENT_ID 'xxx',
     CLIENT_SECRET 'xxx',
     ENDPOINT 'http://xxxx'
);`
2. ATTACH '' AS iceberg (TYPE ICEBERG);
3. select * from iceberg.mydb.mytable
4. insert into `mydb.mytable` by spark
5. duckdb can not read the new data inserted by spark

Is there any configuration available?

Tishj:
This PR fixes #247 

Previously, the schemas and tables would only hit the REST endpoints once, to get the state of the Catalog at that point in time.
No refresh would ever be made.

This PR changes this behavior to hit the REST endpoints once **per transaction**
We realize this is a little too eager, and we will make further improvements to this to cache existing tables/schemas preventing unnecessary http requests.

But for now we make the choice to value correctness over performance in this case.

### DIFF
--- a/src/catalog_api.cpp
+++ b/src/catalog_api.cpp
@@ -141,18 +141,6 @@ vector<IRCAPISchema> IRCAPI::GetSchemas(ClientContext &context, IRCatalog &catal
 	return result;
 }
 
-IRCAPISchema IRCAPI::CreateSchema(ClientContext &context, IRCatalog &catalog, const string &schema) {
-	throw NotImplementedException("IRCAPI::Create Schema not Implemented");
-}
-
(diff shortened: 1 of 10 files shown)


Tmonster (approved):
Thanks!

PR #268, opened 2025-05-28
Pass extended file info to parquet scan & extended file info to Avro Scan to avoid unnecessary HEAD requests

Tmonster:
requires https://github.com/duckdb/duckdb-avro/pull/28
requires https://github.com/duckdb/duckdb/pull/17698 

By passing an extended open file info to the avro scan, we can pass `validate_external_file_cache = false` to the FileReader, which means we won't make unnecessary HEAD requests on secondary reads.

### DIFF
--- a/extension_config.cmake
+++ b/extension_config.cmake
@@ -4,7 +4,7 @@ duckdb_extension_load(icu)
 duckdb_extension_load(avro
         LOAD_TESTS
         GIT_URL https://github.com/duckdb/duckdb_avro
-        GIT_TAG ec4d3c007be073bfb30ab38367285daf621c4f11
+        GIT_TAG 0e30295c80ccd7cdc8f9163d83e7ee46fdc86b50
 )
 
 # Extension from this repo


vcpkg.json
```diff
@@ -18,7 +18,7 @@
 			{
 				"kind": "git",
 				"repository": "https://github.com/duckdb/vcpkg-duckdb-ports",
-				"baseline": "0f9bf648ba1ee29291890a1ca9a49a80bba017eb",
+				"baseline": "d3a5de6bd45b5907ed189054dd23eff6ec869e14",
```
Tishj, line 21: This is reverting the registry version:
<https://github.com/duckdb/vcpkg-duckdb-ports/commits/main/>
0f9bf648ba1ee29291890a1ca9a49a80bba017eb - https://github.com/duckdb/vcpkg-duckdb-ports/commit/0f9bf648ba1ee29291890a1ca9a49a80bba017eb
d3aa25a569e0eb35e01c59e3e62c9bc7406f0ed9 - https://github.com/duckdb/vcpkg-duckdb-ports/commit/d3aa25a569e0eb35e01c59e3e62c9bc7406f0ed9

This version is older than the one that's currently there, is there a reason for that?

I plan to bump this to add avro writes, I'd like to not run into any unrelated issues due to this PR causing issues with a newer patch version

Tishj: I don't really see why this change is here, but I also don't think it matters, so we can merge it regardless

Tmonster: The new commit is the most recent commit to vcpkg-duckdb-ports
[d3a5de6bd45b5907ed189054dd23eff6ec869e14](https://github.com/duckdb/vcpkg-duckdb-ports/commit/d3a5de6bd45b5907ed189054dd23eff6ec869e14) I don't know where d3aa25 is coming from

Issue #263, opened 2025-05-26
Missing HTTP logs to REST endpoints for AWS catalogs


Tmonster:
Repro. For R2 catalog I do see the http requests to the catalog endpoints. Most likely happened when we moved to the general http client interface

```
pragma enable_logging('HTTP');
CREATE SECRET s3table_secret (
      TYPE s3,
      PROVIDER credential_chain,
      CHAIN 'sts',
      ASSUME_ROLE_ARN 'arn:aws:iam::840140254803:role/pyiceberg-etl-role'
  );
 attach 'arn:aws:s3tables:us-east-2:840140254803:bucket/iceberg-testing' as s3_catalog (
      TYPE ICEBERG, 
       secret s3table_secret.
      ENDPOINT_TYPE 'S3_TABLES'
  );
select * from s3_catalog.tpch_sf1.region;
select request.url from duckdb_logs_parsed('HTTP')
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                                                                                          url                                                                                                                          │
│                                                                                                                        varchar                                                                                                                        │
├───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ https://63af955f-0f27-49e9-n4xkifqwkj15a3yuwijk1gbbup4shuse2b--table-s3.s3.us-east-2.amazonaws.com/metadata/00001-2a06fdcf-0f30-47a0-b8ee-a836e3b89cf7.metadata.json                                                                                  │
│ https://63af955f-0f27-49e9-n4xkifqwkj15a3yuwijk1gbbup4shuse2b--table-s3.s3.us-east-2.amazonaws.com/metadata/00001-2a06fdcf-0f30-47a0-b8ee-a836e3b89cf7.metadata.json                                                                                  │
│ https://63af955f-0f27-49e9-n4xkifqwkj15a3yuwijk1gbbup4shuse2b--table-s3.s3.us-east-2.amazonaws.com/metadata/snap-1104850901315573594-0-0a9675d4-5232-458a-8d64-538c05e208a9.avro                                                                      │
[...]

PR #269, opened 2025-05-28
Reenable logging of IRC requests to AWS catalogs

Tmonster:
Very crude implementation, but we need to create a GetRequestInfo from an AWS request object so we could log the request info. This gets that working in a way 

fixes https://github.com/duckdb/duckdb-iceberg/issues/263

### DIFF
--- a/src/aws.cpp
+++ b/src/aws.cpp
@@ -1,6 +1,7 @@
 #include "iceberg_logging.hpp"
 
 #include "aws.hpp"
+#include "duckdb/common/http_util.hpp"
 #include "duckdb/common/string_util.hpp"
 #include "duckdb/common/exception/http_exception.hpp"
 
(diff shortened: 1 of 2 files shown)


src/aws.cpp
```diff
@@ -55,6 +56,28 @@ static void InitAWSAPI() {
 	}
 }
 
+static void LogAWSRequest(ClientContext &context, std::shared_ptr<Aws::Http::HttpRequest> &req) {
+	if (context.db) {
```
Tishj, line 60: I think this should always be true? We also expect this in:
```c++
unique_ptr<HTTPResponse> APIUtils::GetRequest(ClientContext &context, const IRCEndpointBuilder &endpoint_builder,
                                              const string &token) {
	auto &db = DatabaseInstance::GetDatabase(context);

```

PR #273, opened 2025-06-02
Iceberg should build faster, fix minio/mc image id

Tmonster:


scripts/docker-compose.yml
```diff
@@ -35,7 +35,7 @@ services:
   mc:
     depends_on:
       - minio
-    image: minio/mc
+    image: minio/mc:RELEASE.2025-04-16T18-13-26Z-cpuv1
```
Tishj, line 38: Thanks, I think we just ran into this as well with @carlopi , he needed to change the `mc config` line to use `mc alias` instead, as the `mc config` command was deprecated+removed on the version of `minio/mc` that he pulled

.github/workflows/CloudTesting.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: Setup vcpkg
         uses: lukka/run-vcpkg@v11.1
         with:
-          vcpkgGitCommitId: 5e5d0e1cd7785623065e77eff011afdeec1a3574
+          vcpkgGitCommitId: ce613c41372b23b1f51333815feb3edd87ef8a8b
```
Tishj, line 40: @samansmink do you have an opinion on this?
I thought the `vcpkgGitCommitId` had to be kept in sync between core and out-of-tree extensions, am I wrong in that assumption?

Tmonster: hmm, I just pulled the one in `duckdb/extension-ci-tools` but it could be wrong. It's just weird to me that it takes so long to build Iceberg in our CI for server testing, but doesn't take that long in the extension-ci-tools.

Tishj: if `extension-ci-tools` has that commit id then it's probably fine, so disregard my comment 👍

samansmink: i had to bump the vcpkg commit to make things work with the manylinux build, so this looks correct

PR #274, opened 2025-06-02
Move to use "mc alias" instead of "mc config"

carlopi:


carlopi:
This should also unlock https://github.com/duckdb/duckdb-iceberg/pull/267

kevinjqliu:
similar fix in pyiceberg https://github.com/apache/iceberg-python/pull/2049

carlopi:
I think this can go in, failure looks also connected to `mc`, but it's in a different workflow and can be fixed on the side.

Tmonster:
Thanks! I can look into fixing the failure in the Lakekeeper workflow

PR #276, opened 2025-06-03
Wasm compatible fixes

carlopi:
This is on top of #274, that in some version I will assume will go in.

3 changes are:
* fix build_loadable_extension (currently borked / unusable, even in native)
* feature select src/aws.cpp depending on EMSCRIPTEN define
* feature select [extension_config.cmake](https://github.com/duckdb/duckdb-iceberg/compare/main...carlopi:wasm_compatible_fixes?expand=1#diff-b33f69cb0e87de181f64fad3b3f2f19cc48a67665e0605c1884843634ec24b69) to skip other extensions in wasm

First makes sense in any case, other 2 should have no influence on native compilation.

### DIFF
--- a/.github/workflows/MainDistributionPipeline.yml
+++ b/.github/workflows/MainDistributionPipeline.yml
@@ -14,22 +14,22 @@ concurrency:
 jobs:
   duckdb-stable-build:
     name: Build extension binaries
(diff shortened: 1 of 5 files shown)


carlopi:
This is ready on my side, not really urgent, whenever.

Tmonster (approved):
lgtm

PR #279, opened 2025-06-03
use stable lakekeeper image

Tmonster:
Hoping this works, unsure about the rest of the lake keeper setup if that passes


Tmonster:
closing, going to disable lakekeeper or now

PR #280, opened 2025-06-04
Disable Lakekeeper CI

Tmonster:


Tishj:
I think we should just rewrite this run to use their own docker-compose file (`lakekeeper/examples/access-control-simple/docker-compose.yaml`) at some point

PR #291, opened 2025-06-06
fix: Lakekeeper CI test

c-thiel:
Dear DuckDB Team, Author of Lakekeeper here.
I am working on fixing the integration tests in this PR. The original problem was that we renamed our binary from `iceberg-catalog` to `lakekeeper` - and the name of the binary was used excplictly in the health probe.
While on it, I am also removing OpenFGA because I don't think there is a good reason for it here. Also Keycloak is optional, but as all other tests are without IdP, I thought it is probably good to keep some tests with Authentication.

I want to get the tests back running with this PR, then follow up in the next weeks with a second PR, moving to a docker compose setup that requires much less manual effort to get running.

### DIFF
--- a/.github/workflows/LakekeeperTesting.yml
+++ b/.github/workflows/LakekeeperTesting.yml
@@ -14,7 +14,6 @@ jobs:
   rest:
     name: Test against Lakekeeper Catalog
     runs-on: ubuntu-latest
-    if: false # disabled until lakekeeper setup process is more stable
     env:
       VCPKG_TARGET_TRIPLET: 'x64-linux-release'
       GEN: ninja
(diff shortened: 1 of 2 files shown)


Tishj:
Thanks for putting in the time and effort, I think we should just run the docker-compose from your examples directory

It was a mistake on our part to break down the docker-compose file into separate github actions steps I think

c-thiel:
The problem with the docker-compose example in our examples is that that the storage / minio is reachable under different hostnames for pods inside the docker network and duckdb outside of it. I'll have to test how well network host works with them.
Instead of using the examples, which spin up a lot of pods, we should probably opt for what we have in the `docker-compose` folder, which is much more lightweight.

Issue #37, opened 2024-01-19
Write Support


ibotty:
Is there write support on the roadmap?  If so, is there an ETA?  I'll have to plan a future project and would like to keep using duckdb.

samansmink:
Hi @ibotty! It's something we want to add but we can not give an ETA currently

ibotty:
Thank you for the answer.  I might consider duckdb when time comes.

aaronsteers:
Hi, @samansmink . I'm interested also - and I might have cycles to contribute. Is it correct that PyIceberg is not a blocker here? I think I recall we previously were waiting for write support from PyIceberg - but it looks like they support writing now. Sure that sound right?

Thanks!

rustyconover:
Incrementally, this is getting closer with the COPY function in DuckDB adding support for writing field id metadata for the columns.

There should be an `iceberg_schema()` function that will return the schema of a table and the appropriate field ids.  Then you could write a Parquet file with the right field ids to be added to the Iceberg table.

NielsKorschinsky:
Hi @rustyconover ,

that are great news!
Do I understand you correctly, that this `iceberg_schema()` function is already available (in a pre-release)?
What is the plan with writing the associated parquet file, could this also be done in sync with writing the schema file via duckdb?

rustyconover:
Hi @NielsKorschinsky,

At this current time the function `iceberg_schema()` doesn't exist in the extension, but I can see its usefulness to be along the road of writing changes to Iceberg tables.

Rusty

NicolasPA:
I'd like to test DuckDB on Iceberg for my ELT SQL processing engine on cloud storage for our DWH migration, and I bet many more people do. So here's a more precise use case to motivate the prioritization of this feature.

NicolasPA:
Hi @samansmink, I believe this request is the most voted of all the DuckDB repos.  
Would you have an updated ETA motivated by this demand?
We want to know if we can count on DuckDB for our new data platform.

samansmink:
@NicolasPA I can assure you that the interest for this feature has not gone unnoticed. However, at this time we do not have an ETA for this feature yet.

I hope you understand that as a relatively small [non vc-backed](https://www.theregister.com/2023/10/05/duckdb_labs_puts_limit_on_vc_funds/) company, we don't always have the resources to build everything we want right away. Especially in the case of a feature with the complexity of this issue. 

[...]

PR #305, opened 2025-06-14
[REST] Iceberg INSERT support

Tishj:
This PR marks the start of addressing #37 
We've added support for an INSERT statement targeting an Iceberg (REST Catalog backed) table

An INSERT is fully transactional, does not get committed to the REST catalog until the transaction is committed, and cleans up any parquet files created if the transaction aborts.
The inserted data can be read in this same transaction before it's committed

We make use of the `transactions/commit` endpoint which supports multi-table transactions, which means changes can be made to any number of tables within the same transaction and the same guarantees apply.
Because of the `operation` for a Snapshot differentiating between `append` and `delete`, we create a new Snapshot for every INSERT, this can later be optimized to group sequential inserts, but this isn't done yet.

### Implementation details

Parquet files are written directly as part of the `INSERT`, they feed into the (new) `IcebergInsert` operator, which adds these new data files to the transaction.
When we're scanning we check these new data files when we're out of "real" data to scan, so the scan path behaves as if a real manifest+manifest list exists for this snapshot.

On ABORT of the transaction, the newly written parquet files are cleaned up, leaving no trace of the aborted transaction on the storage.
On COMMIT, we write the `manifest_file` and `manifest_list` for every Snapshot that was created, and finally submit a new AddSnapshotUpdate to the `/v1/{prefix}/transactions/commit` endpoint.

### Limitations

Writing to a table with a partition are not supported yet in this initial PR.
Create-Table-As is not supported yet.
Due to limitations in `avro-c`, we don't yet write the required data into the Avro file's key-value metadata for a manifest file.
Various partition related fields of the `manifest_file` and `manifest_entry` are missing or substituted with NULL for now.


src/iceberg_manifest_list.cpp
```diff
@@ -0,0 +1,196 @@
...
+	field_ids.emplace_back("deleted_rows_count", Value::INTEGER(DELETED_ROWS_COUNT));
+
+	// partitions: list<508: field_summary> - 507
+	names.push_back("partitions");
+	types.push_back(FieldSummaryType());
+	field_ids.emplace_back("partitions", FieldSummaryFieldIds());
+
+	//! Populate the DataChunk with the manifests
+
+	DataChunk data;
```
Tishj, line 118: Future optimization: we can probably construct the data chunk once per transaction, even if we have multiple added snapshots.

The final manifest list is just a combination of diffs of the previous alters, by setting the cardinality lower we can use this chunk for previous alters (in the same transaction)

src/storage/irc_transaction.cpp
```diff
@@ -1,34 +1,191 @@
...
+		auto &metadata = table->table_info.table_metadata;
+		auto current_snapshot = metadata.GetLatestSnapshot();
+		if (current_snapshot) {
+			auto &manifest_list_path = current_snapshot->manifest_list;
+			//! Read the manifest list
+			auto manifest_list_reader = make_uniq<manifest_list::ManifestListReader>(metadata.iceberg_version);
+			auto scan = make_uniq<AvroScan>("IcebergManifestList", *context, manifest_list_path);
+			manifest_list_reader->Initialize(std::move(scan));
+			while (!manifest_list_reader->Finished()) {
+				manifest_list_reader->Read(STANDARD_VECTOR_SIZE, commit_state.manifests);
```
Tishj: This feels kind of wasteful.

Because every manifest list has the complete list of manifests, we need to scan the old one to have the entries
We only scan it once per transaction (commit), but still this could be cached if a SELECT has already required reading this

src/storage/table_update/iceberg_add_snapshot.cpp
```diff
@@ -0,0 +1,68 @@
...
+	manifest.manifest_length = manifest_length;
+
+	D_ASSERT(manifest_list.manifests.empty());
+	manifest_list.manifests = std::move(commit_state.manifests);
+	manifest_list.manifests.push_back(std::move(manifest));
+	manifest_list::WriteToFile(manifest_list, avro_copy, db, context);
+	commit_state.manifests = std::move(manifest_list.manifests);
+
+	commit_state.table_change.updates.push_back(CreateAddSnapshotUpdate(snapshot));
+	commit_state.table_change.updates.push_back(CreateSetSnapshotRefUpdate(snapshot));
```
Tishj, line 65: Future optimization: we only need a single set-snapshot-ref update, pointing at the last alter

Not one for every alter

src/iceberg_manifest.cpp
```diff
@@ -1,72 +1,169 @@
...
-		finished = true;
-		return false;
+	data.SetCardinality(manifest_file.data_files.size());
+
+	CopyInfo copy_info;
+	copy_info.is_from = false;
+	copy_info.options["root_name"].push_back(Value("manifest_entry"));
+	copy_info.options["field_ids"].push_back(Value::STRUCT(field_ids));
+
+	Printer::PrintF("field_ids: %s", copy_info.options["field_ids"][0].ToString());
```
Tmonster: Rogue PrintF?

test/sql/local/irc/insert/test_basic_insert.test
```diff
@@ -0,0 +1,213 @@
...
+	'inserted by con2' c
+;
+
+statement ok con1
+commit
+
+# Invalid Configuration Error: Request to 'http://127.0.0.1:8181/v1/transactions/commit' returned a non-200 status code
+statement error
+commit
+----
```
Tmonster, line 126: Any reason why we don't want to verify what the error message is here? Would be nice in case we somehow start throwing internal exceptions or exceptions because certain write paths no longer work

test/sql/local/irc/insert/test_basic_insert.test
```diff
@@ -0,0 +1,213 @@
...
+
```
Tmonster, line 128: Is the two separate transaction case now done after the error message? Are we no longer in a transaction?

test/sql/local/irc/insert/test_basic_insert.test
```diff
@@ -0,0 +1,213 @@
...
+	9223372036854775807::BIGINT, -- long_col (max for BIGINT)
+	3.14::FLOAT, -- float_col (typical float)
+	-1.7976931348623157::DOUBLE, -- double_col (min for DOUBLE)
+	1234567890.12345::DECIMAL(15, 5), -- decimal_col
+	DATE '2023-12-31' -- date_col
+;
+
+statement ok
+commit
+
```
Tmonster, line 200: can we also have some test cases where we verify we can read from iceberg_metadata and iceberg_snapshots for these tests?

src/iceberg_manifest_list.cpp
```diff
@@ -0,0 +1,198 @@
...
+		data.SetValue(col_idx++, i, manifest.partitions.ToValue());
+	}
+	data.SetCardinality(manifest_list.manifests.size());
+
+	CopyInfo copy_info;
+	copy_info.is_from = false;
+	copy_info.options["root_name"].push_back(Value("manifest_file"));
+	copy_info.options["field_ids"].push_back(Value::STRUCT(field_ids));
+
+	Printer::PrintF("field_ids: %s", copy_info.options["field_ids"][0].ToString());
```
Tmonster: nit: Also here?

src/iceberg_manifest.cpp
```diff
@@ -1,72 +1,169 @@
...
+
+	{
+		ThreadContext thread_context(context);
+		ExecutionContext execution_context(context, thread_context, nullptr);
+		auto bind_data = copy.copy_to_bind(context, input, names, types);
+
+		auto global_state = copy.copy_to_initialize_global(context, *bind_data, manifest_file.path);
+		auto local_state = copy.copy_to_initialize_local(execution_context, *bind_data);
+
+		copy.copy_to_sink(execution_context, *bind_data, *global_state, *local_state, data);
```
Tmonster, line 271: The write to avro happens here right? Should we maybe add some try catch for remote writes in case there is a permissions issue or the server responds with some other funky response code? Do you know if we catch that somewhere else?

Tmonster: Maybe some logging for writing the avro/parquet files would be nice too

test/sql/local/irc/insert/test_basic_insert.test
Tishj: Yea, definitely a fair point
We throw the exception in the comment above, but we don't end up returning that, we hit:
`TransactionContext Error: cannot commit - no transaction is active`

src/iceberg_manifest.cpp
Tishj: I think that would throw an error in the s3fs (or other relevant filesystem) when trying to open the file
We don't report this clearly to the user, because of the `TransactionContext Error: cannot commit - no transaction is active` error being thrown instead

This might also be a place where we want to try refreshing credentials, but ideally that is handled by the filesystem, refresh+retry (which I don't think is currently implemented)

src/storage/table_update/iceberg_add_snapshot.cpp
kevinjqliu, line 65: might need to also add a `AssertRefSnapshotId` here, similar to pyiceberg
https://github.com/apache/iceberg-python/blob/cbbffc5df22385c37419168efcdc2af110ac4c95/pyiceberg/table/update/snapshot.py#L282-L288

src/storage/table_update/iceberg_add_snapshot.cpp
```diff
@@ -0,0 +1,68 @@
...
+	auto &avro_copy = avro_copy_p->Cast<CopyFunctionCatalogEntry>().function;
+
+	auto manifest_length = manifest_file::WriteToFile(table_info, manifest_file, avro_copy, db, context);
+	manifest.manifest_length = manifest_length;
+
+	D_ASSERT(manifest_list.manifests.empty());
+	manifest_list.manifests = std::move(commit_state.manifests);
+	manifest_list.manifests.push_back(std::move(manifest));
+	manifest_list::WriteToFile(manifest_list, avro_copy, db, context);
+	commit_state.manifests = std::move(manifest_list.manifests);
```
kevinjqliu, line 62: nit: this seems to be doing "fast append", perhaps we can rename it as such? 

> Appends can add a new manifest to minimize the amount of data written, instead of adding new records by rewriting and appending to an existing manifest. (This is called a “fast append”.)

https://iceberg.apache.org/spec/#snapshots

src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -314,27 +323,37 @@ optional_ptr<const IcebergManifestEntry> IcebergMultiFileList::GetDataFile(idx_t
...
+
+				data_manifest_reader->Initialize(std::move(scan));
+				data_manifest_reader->SetSequenceNumber(manifest.sequence_number);
+				data_manifest_reader->SetPartitionSpecID(manifest.partition_spec_id);
+
+				while (!data_manifest_reader->Finished()) {
+					data_manifest_reader->Read(STANDARD_VECTOR_SIZE, current_data_files);
+				}
+				current_data_manifest++;
+			} else if (HasTransactionData()) {
```
Tishj: This does check the filter pushdown at the `data_file` (or `manifest_entry`) level, but not at the manifest list (or `manifest_file`)
This should probably be checked at the `InitializeFiles` level, so we can filter (uncommitted) manifests that don't apply to the query

src/storage/table_update/iceberg_add_snapshot.cpp
Tishj: I have to admit that I wasn't even aware of that being an option, "rewriting and appending to an existing manifest"
That implies that manifests aren't immutable, which I thought they were?

How does that work with older manifest lists that contain the manifest we're rewriting, it's deleted until it's rewritten? I guess this has to be done with an atomic swap so there is no down-time for other readers that might try to read the old manifest file?
What happens if multiple writers try to rewrite the same manifest? - doesn't that cause silent data loss?

Not to mention possibly failing the transaction, which I guess would need to undo the edit it did to the existing manifest?

This PR doesn't currently clean up written manifest lists/files on transaction failure, which should be done, but isn't a *huge* deal I feel, since it's not reachable if the transaction fails

kevinjqliu: > "rewriting and appending to an existing manifest"

the new manifest file is appended to the original manifest list. i think we might need to reword this in the spec. heres another way to put it, https://github.com/apache/iceberg-python/blob/cbbffc5df22385c37419168efcdc2af110ac4c95/pyiceberg/table/update/snapshot.py#L462-L465

> What happens if multiple writers try to rewrite the same manifest? - doesn't that cause silent data loss?

no because this is append only

> This PR doesn't currently clean up written manifest lists/files on transaction failure, which should be done, but isn't a huge deal I feel, since it's not reachable if the transaction fails

agreed


also feel free to ignore "fast append", its more of an optimization

Tishj: Ah I see, I think this is talking about inserts done in the same transaction?
I put this in the description as an optimization we will most likely add later:
> Because of the operation for a Snapshot differentiating between append and delete, we create a new Snapshot for every INSERT, this can later be optimized to group sequential inserts, but this isn't done yet

Tishj:
> Added a few comments.
> 
> I think it would be useful to test roundtrip; writing in duckdb and reading with something like pyiceberg.

Thanks for the review, I agree, we should add a CI run after running the test suite to read with pyiceberg

I got pyiceberg to correctly read the equivalent of "select * from my_datalake.default.test_insert" yesterday

src/storage/table_update/iceberg_add_snapshot.cpp
Tishj: Hmm no, I'm mistaken
You look at the existing manifests in the manifest list and decide based on stats and population and other distributions whether to ditch the old (written, committed) manifest in favor of a rewritten manifest containing the old data + (a subset of) the new data, for all the data we're inserting

That is for sure interesting and important to create writes that make for efficient reads

test/sql/local/irc/insert/test_basic_insert.test
Tishj: It was a stupid oversight on my part, missing the `con2` parameter in the `statement error`
Should be properly propagated now 👍

test/sql/local/irc/insert/test_basic_insert.test
```diff
@@ -121,11 +121,10 @@ statement ok con1
 commit
 
 # The second commit failed because both transactions tried to add sequence number: 2
-# Invalid Configuration Error: Request to 'http://127.0.0.1:8181/v1/transactions/commit' returned a non-200 status code
-statement error
+statement error con2
 commit
 ----
-TransactionContext Error
+<REGEX>:.*Requirement failed: branch main has changed.*
```
Tmonster, line 127: Is there a way to check the status code of the HTTP response here? Or in a test of the logs?

src/storage/irc_transaction.cpp
```diff
@@ -1,34 +1,225 @@
...
+
+		auto response = authentication.PostRequest(*context, url_builder, transaction_json);
+		if (response->status != HTTPStatusCode::OK_200) {
+			throw InvalidConfigurationException(
+			    "Request to '%s' returned a non-200 status code (%s), with reason: %s, body: %s", url_builder.GetURL(),
+			    EnumUtil::ToString(response->status), response->reason, response->body);
+		}
+		temp_con.Commit();
+	} catch (std::exception &ex) {
+		ErrorData error(ex);
```
Tmonster, line 192: Should we also abort or commit the temp_con() here?

src/storage/irc_transaction.cpp
```diff
@@ -1,34 +1,225 @@
...
-		//		connection.Execute("COMMIT");
+	if (dirty_tables.empty()) {
+		return;
+	}
+
+	try {
+		Connection temp_con(db);
+		//! This automatically starts a transaction
+		temp_con.BeginTransaction();
+		auto &context = temp_con.context;
```
Tmonster: For my understanding, do we need temp_con just for the ClientContext? Or is there another reason for it?

Tmonster (commented):
Awesome! Thanks for the extra large file insert test. 
Just the one question about the temp_con, otherwise this is good to go

src/storage/irc_transaction.cpp
Tishj: Just for the client context that is properly associated with a transaction, which is a requirement that the ClientContext in the Transaction that we are committing does not satisfy.

src/storage/irc_transaction.cpp
Tishj: It was created in the `try` body, so it doesn't exist here

test/sql/local/irc/insert/test_basic_insert.test
Tishj: That falls under logging / retry logic in my opinion
The main goal currently is forwarding the error correctly, we can intercept based on the response code later

Tishj:
Just stumbled on this issue: https://github.com/apache/iceberg-go/pull/435

Which states that Spark might complain about the names given to list elements and map keys and values, in the avro file

test/sql/local/irc/insert/test_basic_insert.test
Tishj: We can use structured logging to be able to query the status code

The status code is also in this error message, it's just not in the verified portion

data/persistent/big_query_error/rewrite_manifests.test
```diff
@@ -0,0 +1,91 @@
+require avro
+
+# This file contains sql queries used to rewrite the paths of received test data to live inside 'data/persistent/big_query_error'
+mode skip
```
samansmink, line 4: Is this skip intentional?

src/iceberg_manifest.cpp
samansmink: Refresh does exists, but currently needs to be enabled explicitly, see https://github.com/duckdb/duckdb-httpfs/blob/39779f89b16d0a35f04d3cfaf868e6366a2102f0/extension/httpfs/s3fs.cpp#L718

src/include/metadata/iceberg_manifest_list.hpp
```diff
@@ -82,4 +132,33 @@ struct IcebergManifestList {
 	vector<IcebergManifest> manifests;
 };
 
+namespace manifest_list {
+
+static constexpr const int32_t MANIFEST_PATH = 500;
```
samansmink, line 137: nit: Can/should this be an enum? 
nitty nit: Also should this be sorted?

src/storage/iceberg_insert.cpp
```diff
@@ -0,0 +1,386 @@
...
+	info->options["field_ids"] = std::move(field_input);
+
+	// Get Parquet Copy function
+	auto copy_fun = TryGetCopyFunction(*context.db, "parquet");
+	if (!copy_fun) {
+		throw MissingExtensionException("Did not find parquet copy function required to write to iceberg table");
+	}
+
+	vector<idx_t> partition_columns;
+	//! TODO: support partitions
```
samansmink, line 320: nit: there's quite a lot of commented out code here

.github/workflows/LocalTesting.yml
```diff
@@ -95,4 +99,10 @@ jobs:
...
         run: |
-          make test_release
\ No newline at end of file
+          make test_release
+
+      - name: Test reads with PyIceberg
+        env:
+          ICEBERG_SERVER_AVAILABLE: 1
+        run: |
+          python3 -m pytest test/python
```
samansmink, line 108: great idea to have CI test reading the written data with other engines

I think it would be nice to make this flow a bit more explicit where we write separate tests to generate the data for testing in pyiceberg and spark. Currently it would we really easy to break ci by refactoring a test in `test/sql` which you would only catch when running CI.

For now I would say this is fine, but as more of these tests would get added i think its nice to at least group all tests that generate data for reading through pyiceberg or spark in a single folder with a readme explaining how they're tested

samansmink (approved):
Amazing work!! looks great

Added a few nitpicky comments

data/persistent/big_query_error/rewrite_manifests.test
Tishj: It is, this "test" shows how the test files were rewritten
Since the avro files contain absolute paths this has to be done when receiving table data from users

src/include/metadata/iceberg_manifest_list.hpp
Tishj: I tried making it an enum, but then I needed `static_cast<int32_t>(...)` everywhere they are used, and I felt it was ugly

The order they're declared matches the order they are defined in the spec

I'll make a diff for changing it into an enum, I do agree it's nicer to make it an enum

src/storage/iceberg_insert.cpp
Tishj: There is, because it's relevant for supporting partitions, which this PR doesn't do yet, but I think it's logically one of the next steps to support

.github/workflows/LocalTesting.yml
Tishj: Hmm I agree, but I don't see a nice way of grouping these

What we *could* do is not have a `.test` for the insert related tests but implement these in `test/python` using duckdb-python, then they're bundled and we can potentially even do a clean up after the tests are ran to make it rerunnable (which currently requires manual intervention to clean up the files:
```bash
python3 -m scripts.data_generators.generate_data spark-rest --test insert_test
python3 -m scripts.data_generators.generate_data spark-rest --test insert_all_types
```
)

src/storage/iceberg_insert.cpp
samansmink: sounds good!

src/include/metadata/iceberg_manifest_list.hpp
Tishj:
```patch
diff --git a/src/iceberg_manifest.cpp b/src/iceberg_manifest.cpp
index 71d2007e..9d6318e3 100644
--- a/src/iceberg_manifest.cpp
+++ b/src/iceberg_manifest.cpp
@@ -57,10 +57,10 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
 		// status: int - 0
 		names.push_back("status");
 		types.push_back(LogicalType::INTEGER);
-		field_ids.emplace_back("status", Value::INTEGER(STATUS));
+		field_ids.emplace_back("status", Value::INTEGER(static_cast<int32_t>(FieldIds::STATUS)));
 
 		auto field_obj = yyjson_mut_arr_add_obj(doc, fields_arr);
-		yyjson_mut_obj_add_uint(doc, field_obj, "id", STATUS);
+		yyjson_mut_obj_add_uint(doc, field_obj, "id", static_cast<int32_t>(FieldIds::STATUS));
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "name", "status");
 		yyjson_mut_obj_add_bool(doc, field_obj, "required", true);
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "type", "int");
@@ -70,10 +70,10 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
 		// snapshot_id: long - 1
 		names.push_back("snapshot_id");
 		types.push_back(LogicalType::BIGINT);
-		field_ids.emplace_back("snapshot_id", Value::INTEGER(SNAPSHOT_ID));
+		field_ids.emplace_back("snapshot_id", Value::INTEGER(static_cast<int32_t>(FieldIds::SNAPSHOT_ID)));
 
 		auto field_obj = yyjson_mut_arr_add_obj(doc, fields_arr);
-		yyjson_mut_obj_add_uint(doc, field_obj, "id", SNAPSHOT_ID);
+		yyjson_mut_obj_add_uint(doc, field_obj, "id", static_cast<int32_t>(FieldIds::SNAPSHOT_ID));
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "name", "snapshot_id");
 		yyjson_mut_obj_add_bool(doc, field_obj, "required", false);
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "type", "long");
@@ -83,10 +83,10 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
 		// sequence_number: long - 3
 		names.push_back("sequence_number");
 		types.push_back(LogicalType::BIGINT);
-		field_ids.emplace_back("sequence_number", Value::INTEGER(SEQUENCE_NUMBER));
+		field_ids.emplace_back("sequence_number", Value::INTEGER(static_cast<int32_t>(FieldIds::SEQUENCE_NUMBER)));
 
 		auto field_obj = yyjson_mut_arr_add_obj(doc, fields_arr);
-		yyjson_mut_obj_add_uint(doc, field_obj, "id", SEQUENCE_NUMBER);
+		yyjson_mut_obj_add_uint(doc, field_obj, "id", static_cast<int32_t>(FieldIds::SEQUENCE_NUMBER));
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "name", "sequence_number");
 		yyjson_mut_obj_add_bool(doc, field_obj, "required", false);
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "type", "long");
@@ -96,10 +96,11 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
 		// file_sequence_number: long - 4
 		names.push_back("file_sequence_number");
 		types.push_back(LogicalType::BIGINT);
-		field_ids.emplace_back("file_sequence_number", Value::INTEGER(FILE_SEQUENCE_NUMBER));
+		field_ids.emplace_back("file_sequence_number",
+		                       Value::INTEGER(static_cast<int32_t>(FieldIds::FILE_SEQUENCE_NUMBER)));
 
 		auto field_obj = yyjson_mut_arr_add_obj(doc, fields_arr);
-		yyjson_mut_obj_add_uint(doc, field_obj, "id", FILE_SEQUENCE_NUMBER);
+		yyjson_mut_obj_add_uint(doc, field_obj, "id", static_cast<int32_t>(FieldIds::FILE_SEQUENCE_NUMBER));
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "name", "file_sequence_number");
 		yyjson_mut_obj_add_bool(doc, field_obj, "required", false);
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "type", "long");
@@ -114,10 +115,10 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
 	{
 		// content: int - 134
 		children.emplace_back("content", LogicalType::INTEGER);
-		data_file_field_ids.emplace_back("content", Value::INTEGER(CONTENT));
+		data_file_field_ids.emplace_back("content", Value::INTEGER(static_cast<int32_t>(FieldIds::CONTENT)));
 
 		auto field_obj = yyjson_mut_arr_add_obj(doc, child_fields_arr);
-		yyjson_mut_obj_add_uint(doc, field_obj, "id", CONTENT);
+		yyjson_mut_obj_add_uint(doc, field_obj, "id", static_cast<int32_t>(FieldIds::CONTENT));
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "name", "content");
 		yyjson_mut_obj_add_bool(doc, field_obj, "required", true);
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "type", "int");
@@ -126,10 +127,10 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
 	{
 		// file_path: string - 100
 		children.emplace_back("file_path", LogicalType::VARCHAR);
-		data_file_field_ids.emplace_back("file_path", Value::INTEGER(FILE_PATH));
+		data_file_field_ids.emplace_back("file_path", Value::INTEGER(static_cast<int32_t>(FieldIds::FILE_PATH)));
 
 		auto field_obj = yyjson_mut_arr_add_obj(doc, child_fields_arr);
-		yyjson_mut_obj_add_uint(doc, field_obj, "id", FILE_PATH);
+		yyjson_mut_obj_add_uint(doc, field_obj, "id", static_cast<int32_t>(FieldIds::FILE_PATH));
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "name", "file_path");
 		yyjson_mut_obj_add_bool(doc, field_obj, "required", true);
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "type", "string");
@@ -138,10 +139,10 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
 	{
 		// file_format: string - 101
 		children.emplace_back("file_format", LogicalType::VARCHAR);
-		data_file_field_ids.emplace_back("file_format", Value::INTEGER(FILE_FORMAT));
+		data_file_field_ids.emplace_back("file_format", Value::INTEGER(static_cast<int32_t>(FieldIds::FILE_FORMAT)));
 
 		auto field_obj = yyjson_mut_arr_add_obj(doc, child_fields_arr);
-		yyjson_mut_obj_add_uint(doc, field_obj, "id", FILE_FORMAT);
+		yyjson_mut_obj_add_uint(doc, field_obj, "id", static_cast<int32_t>(FieldIds::FILE_FORMAT));
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "name", "file_format");
 		yyjson_mut_obj_add_bool(doc, field_obj, "required", true);
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "type", "string");
@@ -150,10 +151,10 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
 	{
 		// partition: struct(...) - 102
 		children.emplace_back("partition", PartitionStructType(table_info));
-		data_file_field_ids.emplace_back("partition", Value::INTEGER(PARTITION));
+		data_file_field_ids.emplace_back("partition", Value::INTEGER(static_cast<int32_t>(FieldIds::PARTITION)));
 
 		auto field_obj = yyjson_mut_arr_add_obj(doc, child_fields_arr);
-		yyjson_mut_obj_add_uint(doc, field_obj, "id", PARTITION);
+		yyjson_mut_obj_add_uint(doc, field_obj, "id", static_cast<int32_t>(FieldIds::PARTITION));
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "name", "partition");
 		yyjson_mut_obj_add_bool(doc, field_obj, "required", true);
 
@@ -167,10 +168,10 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
 	{
 		// record_count: long - 103
 		children.emplace_back("record_count", LogicalType::BIGINT);
-		data_file_field_ids.emplace_back("record_count", Value::INTEGER(RECORD_COUNT));
+		data_file_field_ids.emplace_back("record_count", Value::INTEGER(static_cast<int32_t>(FieldIds::RECORD_COUNT)));
 
 		auto field_obj = yyjson_mut_arr_add_obj(doc, child_fields_arr);
-		yyjson_mut_obj_add_uint(doc, field_obj, "id", RECORD_COUNT);
+		yyjson_mut_obj_add_uint(doc, field_obj, "id", static_cast<int32_t>(FieldIds::RECORD_COUNT));
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "name", "record_count");
 		yyjson_mut_obj_add_bool(doc, field_obj, "required", true);
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "type", "long");
@@ -179,10 +180,11 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
 	{
 		// file_size_in_bytes: long - 104
 		children.emplace_back("file_size_in_bytes", LogicalType::BIGINT);
-		data_file_field_ids.emplace_back("file_size_in_bytes", Value::INTEGER(FILE_SIZE_IN_BYTES));
+		data_file_field_ids.emplace_back("file_size_in_bytes",
+		                                 Value::INTEGER(static_cast<int32_t>(FieldIds::FILE_SIZE_IN_BYTES)));
 
 		auto field_obj = yyjson_mut_arr_add_obj(doc, child_fields_arr);
-		yyjson_mut_obj_add_uint(doc, field_obj, "id", FILE_SIZE_IN_BYTES);
+		yyjson_mut_obj_add_uint(doc, field_obj, "id", static_cast<int32_t>(FieldIds::FILE_SIZE_IN_BYTES));
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "name", "file_size_in_bytes");
 		yyjson_mut_obj_add_bool(doc, field_obj, "required", true);
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "type", "long");
@@ -200,11 +202,12 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
 		// data_file: struct(...) - 2
 		names.push_back("data_file");
 		types.push_back(LogicalType::STRUCT(std::move(children)));
-		data_file_field_ids.emplace_back("__duckdb_field_id", Value::INTEGER(DATA_FILE));
+		data_file_field_ids.emplace_back("__duckdb_field_id",
+		                                 Value::INTEGER(static_cast<int32_t>(FieldIds::DATA_FILE)));
 		field_ids.emplace_back("data_file", Value::STRUCT(data_file_field_ids));
 
 		auto field_obj = yyjson_mut_arr_add_obj(doc, fields_arr);
-		yyjson_mut_obj_add_uint(doc, field_obj, "id", DATA_FILE);
+		yyjson_mut_obj_add_uint(doc, field_obj, "id", static_cast<int32_t>(FieldIds::DATA_FILE));
 		yyjson_mut_obj_add_strcpy(doc, field_obj, "name", "data_file");
 		yyjson_mut_obj_add_bool(doc, field_obj, "required", true);
 
diff --git a/src/include/metadata/iceberg_manifest.hpp b/src/include/metadata/iceberg_manifest.hpp
index bd8b0665..5395f6a9 100644
--- a/src/include/metadata/iceberg_manifest.hpp
+++ b/src/include/metadata/iceberg_manifest.hpp
@@ -95,50 +95,52 @@ public:
 
 namespace manifest_file {
 
-static constexpr const int32_t STATUS = 0;
-static constexpr const int32_t SNAPSHOT_ID = 1;
-static constexpr const int32_t SEQUENCE_NUMBER = 3;
-static constexpr const int32_t FILE_SEQUENCE_NUMBER = 4;
-static constexpr const int32_t DATA_FILE = 2;
-static constexpr const int32_t CONTENT = 134;
-static constexpr const int32_t FILE_PATH = 100;
-static constexpr const int32_t FILE_FORMAT = 101;
-static constexpr const int32_t PARTITION = 102;
-static constexpr const int32_t RECORD_COUNT = 103;
-static constexpr const int32_t FILE_SIZE_IN_BYTES = 104;
-// static constexpr const int32_t BLOCK_SIZE_IN_BYTES = 105; // (deprecated)
-// static constexpr const int32_t FILE_ORDINAL = 106; // (deprecated)
-// static constexpr const int32_t SORT_COLUMNS = 107; // (deprecated)
-// static constexpr const int32_t SORT_COLUMNS_ELEMENT = 112; // (deprecated)
-static constexpr const int32_t COLUMN_SIZES = 108;
-static constexpr const int32_t COLUMN_SIZES_KEY = 117;
-static constexpr const int32_t COLUMN_SIZES_VALUE = 118;
-static constexpr const int32_t VALUE_COUNTS = 109;
-static constexpr const int32_t VALUE_COUNTS_KEY = 119;
-static constexpr const int32_t VALUE_COUNTS_VALUE = 120;
-static constexpr const int32_t NULL_VALUE_COUNTS = 110;
-static constexpr const int32_t NULL_VALUE_COUNTS_KEY = 121;
-static constexpr const int32_t NULL_VALUE_COUNTS_VALUE = 122;
-static constexpr const int32_t NAN_VALUE_COUNTS = 137;
-static constexpr const int32_t NAN_VALUE_COUNTS_KEY = 138;
-static constexpr const int32_t NAN_VALUE_COUNTS_VALUE = 139;
-// static constexpr const int32_t DISTINCT_COUNTS = 111; // (deprecated)
-static constexpr const int32_t LOWER_BOUNDS = 125;
-static constexpr const int32_t LOWER_BOUNDS_KEY = 126;
-static constexpr const int32_t LOWER_BOUNDS_VALUE = 127;
-static constexpr const int32_t UPPER_BOUNDS = 128;
-static constexpr const int32_t UPPER_BOUNDS_KEY = 129;
-static constexpr const int32_t UPPER_BOUNDS_VALUE = 130;
-// static constexpr const int32_t KEY_METADATA = 131; // (optional)
-static constexpr const int32_t SPLIT_OFFSETS = 132;
-static constexpr const int32_t SPLIT_OFFSETS_ELEMENT = 133;
-static constexpr const int32_t EQUALITY_IDS = 135;
-static constexpr const int32_t EQUALITY_IDS_ELEMENT = 136;
-static constexpr const int32_t SORT_ORDER_ID = 140;
-static constexpr const int32_t FIRST_ROW_ID = 142;
-static constexpr const int32_t REFERENCED_DATA_FILE = 143;
-static constexpr const int32_t CONTENT_OFFSET = 144;
-static constexpr const int32_t CONTENT_SIZE_IN_BYTES = 145;
+enum class FieldIds : int32_t {
+	STATUS = 0,
+	SNAPSHOT_ID = 1,
+	SEQUENCE_NUMBER = 3,
+	FILE_SEQUENCE_NUMBER = 4,
+	DATA_FILE = 2,
+	CONTENT = 134,
+	FILE_PATH = 100,
+	FILE_FORMAT = 101,
+	PARTITION = 102,
+	RECORD_COUNT = 103,
+	FILE_SIZE_IN_BYTES = 104,
+	BLOCK_SIZE_IN_BYTES = 105,  // (deprecated)
+	FILE_ORDINAL = 106,         // (deprecated)
+	SORT_COLUMNS = 107,         // (deprecated)
+	SORT_COLUMNS_ELEMENT = 112, // (deprecated)
+	COLUMN_SIZES = 108,
+	COLUMN_SIZES_KEY = 117,
+	COLUMN_SIZES_VALUE = 118,
+	VALUE_COUNTS = 109,
+	VALUE_COUNTS_KEY = 119,
+	VALUE_COUNTS_VALUE = 120,
+	NULL_VALUE_COUNTS = 110,
+	NULL_VALUE_COUNTS_KEY = 121,
+	NULL_VALUE_COUNTS_VALUE = 122,
+	NAN_VALUE_COUNTS = 137,
+	NAN_VALUE_COUNTS_KEY = 138,
+	NAN_VALUE_COUNTS_VALUE = 139,
+	DISTINCT_COUNTS = 111, // (deprecated)
+	LOWER_BOUNDS = 125,
+	LOWER_BOUNDS_KEY = 126,
+	LOWER_BOUNDS_VALUE = 127,
+	UPPER_BOUNDS = 128,
+	UPPER_BOUNDS_KEY = 129,
+	UPPER_BOUNDS_VALUE = 130,
+	KEY_METADATA = 131, // (optional)
+	SPLIT_OFFSETS = 132,
+	SPLIT_OFFSETS_ELEMENT = 133,
+	EQUALITY_IDS = 135,
+	EQUALITY_IDS_ELEMENT = 136,
+	SORT_ORDER_ID = 140,
+	FIRST_ROW_ID = 142,
+	REFERENCED_DATA_FILE = 143,
+	CONTENT_OFFSET = 144,
+	CONTENT_SIZE_IN_BYTES = 145
+};
 
 idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile &manifest_file,
                   CopyFunction &copy_function, DatabaseInstance &db, ClientContext &context);
diff --git a/src/manifest_file_reader.cpp b/src/manifest_file_reader.cpp
index dab4f336..09218f18 100644
--- a/src/manifest_file_reader.cpp
+++ b/src/manifest_file_reader.cpp
@@ -42,7 +42,7 @@ void ManifestFileReader::CreateVectorMapping(idx_t column_id, MultiFileColumnDef
 	D_ASSERT(column.identifier.type().id() == LogicalTypeId::INTEGER);
 
 	auto field_id = column.identifier.GetValue<int32_t>();
-	if (field_id != DATA_FILE) {
+	if (field_id != static_cast<int32_t>(FieldIds::DATA_FILE)) {
 		vector_mapping.emplace(field_id, ColumnIndex(column_id));
 		return;
 	}
@@ -66,7 +66,9 @@ void ManifestFileReader::CreateVectorMapping(idx_t column_id, MultiFileColumnDef
 }
 
 bool ManifestFileReader::ValidateVectorMapping() {
-	static const int32_t V1_REQUIRED_FIELDS[] = {STATUS, FILE_PATH, FILE_FORMAT, RECORD_COUNT};
+	static const int32_t V1_REQUIRED_FIELDS[] = {
+	    static_cast<int32_t>(FieldIds::STATUS), static_cast<int32_t>(FieldIds::FILE_PATH),
+	    static_cast<int32_t>(FieldIds::FILE_FORMAT), static_cast<int32_t>(FieldIds::RECORD_COUNT)};
 	static const idx_t V1_REQUIRED_FIELDS_SIZE = sizeof(V1_REQUIRED_FIELDS) / sizeof(int32_t);
 	for (idx_t i = 0; i < V1_REQUIRED_FIELDS_SIZE; i++) {
 		if (!vector_mapping.count(V1_REQUIRED_FIELDS[i])) {
@@ -74,7 +76,7 @@ bool ManifestFileReader::ValidateVectorMapping() {
 		}
 	}
 
-	static const int32_t V2_REQUIRED_FIELDS[] = {CONTENT};
+	static const int32_t V2_REQUIRED_FIELDS[] = {static_cast<int32_t>(FieldIds::CONTENT)};
 	static const idx_t V2_REQUIRED_FIELDS_SIZE = sizeof(V2_REQUIRED_FIELDS) / sizeof(int32_t);
 	if (iceberg_version >= 2) {
 		for (idx_t i = 0; i < V2_REQUIRED_FIELDS_SIZE; i++) {
@@ -151,64 +153,69 @@ idx_t ManifestFileReader::ReadChunk(idx_t offset, idx_t count, vector<IcebergMan
 	D_ASSERT(offset < chunk.size());
 	D_ASSERT(offset + count <= chunk.size());
 
-	auto status = FlatVector::GetData<int32_t>(chunk.data[vector_mapping.at(STATUS).GetPrimaryIndex()]);
+	auto status = FlatVector::GetData<int32_t>(
+	    chunk.data[vector_mapping.at(static_cast<int32_t>(FieldIds::STATUS)).GetPrimaryIndex()]);
 
-	auto file_path_idx = vector_mapping.at(FILE_PATH);
+	auto file_path_idx = vector_mapping.at(static_cast<int32_t>(FieldIds::FILE_PATH));
 	auto data_file_idx = file_path_idx.GetPrimaryIndex();
 	auto &child_entries = StructVector::GetEntries(chunk.data[data_file_idx]);
-	D_ASSERT(vector_mapping.at(FILE_FORMAT).GetPrimaryIndex() == data_file_idx);
-	D_ASSERT(vector_mapping.at(RECORD_COUNT).GetPrimaryIndex() == data_file_idx);
+	D_ASSERT(vector_mapping.at(static_cast<int32_t>(FieldIds::FILE_FORMAT)).GetPrimaryIndex() == data_file_idx);
+	D_ASSERT(vector_mapping.at(static_cast<int32_t>(FieldIds::RECORD_COUNT)).GetPrimaryIndex() == data_file_idx);
 	if (iceberg_version > 1) {
-		D_ASSERT(vector_mapping.at(CONTENT).GetPrimaryIndex() == data_file_idx);
+		D_ASSERT(vector_mapping.at(static_cast<int32_t>(FieldIds::CONTENT)).GetPrimaryIndex() == data_file_idx);
 	}
 	optional_ptr<Vector> equality_ids;
 	optional_ptr<Vector> sequence_number;
 	int32_t *content;
 
-	auto partition_idx = vector_mapping.at(PARTITION);
+	auto partition_idx = vector_mapping.at(static_cast<int32_t>(FieldIds::PARTITION));
 	if (iceberg_version > 1) {
-		auto equality_ids_it = vector_mapping.find(EQUALITY_IDS);
+		auto equality_ids_it = vector_mapping.find(static_cast<int32_t>(FieldIds::EQUALITY_IDS));
 		if (equality_ids_it != vector_mapping.end()) {
 			equality_ids = *child_entries[equality_ids_it->second.GetChildIndex(0).GetPrimaryIndex()];
 		}
-		auto sequence_number_it = vector_mapping.find(SEQUENCE_NUMBER);
+		auto sequence_number_it = vector_mapping.find(static_cast<int32_t>(FieldIds::SEQUENCE_NUMBER));
 		if (sequence_number_it != vector_mapping.end()) {
 			sequence_number = chunk.data[sequence_number_it->second.GetPrimaryIndex()];
 		}
-		content =
-		    FlatVector::GetData<int32_t>(*child_entries[vector_mapping.at(CONTENT).GetChildIndex(0).GetPrimaryIndex()]);
+		content = FlatVector::GetData<int32_t>(
+		    *child_entries
+		        [vector_mapping.at(static_cast<int32_t>(FieldIds::CONTENT)).GetChildIndex(0).GetPrimaryIndex()]);
 	}
 
 	auto file_path = FlatVector::GetData<string_t>(*child_entries[file_path_idx.GetChildIndex(0).GetPrimaryIndex()]);
 	auto file_format = FlatVector::GetData<string_t>(
-	    *child_entries[vector_mapping.at(FILE_FORMAT).GetChildIndex(0).GetPrimaryIndex()]);
+	    *child_entries
+	        [vector_mapping.at(static_cast<int32_t>(FieldIds::FILE_FORMAT)).GetChildIndex(0).GetPrimaryIndex()]);
 	auto record_count = FlatVector::GetData<int64_t>(
-	    *child_entries[vector_mapping.at(RECORD_COUNT).GetChildIndex(0).GetPrimaryIndex()]);
+	    *child_entries
+	        [vector_mapping.at(static_cast<int32_t>(FieldIds::RECORD_COUNT)).GetChildIndex(0).GetPrimaryIndex()]);
 	auto file_size_in_bytes = FlatVector::GetData<int64_t>(
-	    *child_entries[vector_mapping.at(FILE_SIZE_IN_BYTES).GetChildIndex(0).GetPrimaryIndex()]);
+	    *child_entries
+	        [vector_mapping.at(static_cast<int32_t>(FieldIds::FILE_SIZE_IN_BYTES)).GetChildIndex(0).GetPrimaryIndex()]);
 	optional_ptr<Vector> lower_bounds;
 	optional_ptr<Vector> upper_bounds;
 	optional_ptr<Vector> value_counts;
 	optional_ptr<Vector> null_value_counts;
 	optional_ptr<Vector> nan_value_counts;
 
-	auto lower_bounds_it = vector_mapping.find(LOWER_BOUNDS);
+	auto lower_bounds_it = vector_mapping.find(static_cast<int32_t>(FieldIds::LOWER_BOUNDS));
 	if (lower_bounds_it != vector_mapping.end()) {
 		lower_bounds = *child_entries[lower_bounds_it->second.GetChildIndex(0).GetPrimaryIndex()];
 	}
-	auto upper_bounds_it = vector_mapping.find(UPPER_BOUNDS);
+	auto upper_bounds_it = vector_mapping.find(static_cast<int32_t>(FieldIds::UPPER_BOUNDS));
 	if (upper_bounds_it != vector_mapping.end()) {
 		upper_bounds = *child_entries[upper_bounds_it->second.GetChildIndex(0).GetPrimaryIndex()];
 	}
-	auto value_counts_it = vector_mapping.find(VALUE_COUNTS);
+	auto value_counts_it = vector_mapping.find(static_cast<int32_t>(FieldIds::VALUE_COUNTS));
 	if (value_counts_it != vector_mapping.end()) {
 		value_counts = *child_entries[value_counts_it->second.GetChildIndex(0).GetPrimaryIndex()];
 	}
-	auto null_value_counts_it = vector_mapping.find(NULL_VALUE_COUNTS);
+	auto null_value_counts_it = vector_mapping.find(static_cast<int32_t>(FieldIds::NULL_VALUE_COUNTS));
 	if (null_value_counts_it != vector_mapping.end()) {
 		null_value_counts = *child_entries[null_value_counts_it->second.GetChildIndex(0).GetPrimaryIndex()];
 	}
-	auto nan_value_counts_it = vector_mapping.find(NAN_VALUE_COUNTS);
+	auto nan_value_counts_it = vector_mapping.find(static_cast<int32_t>(FieldIds::NAN_VALUE_COUNTS));
 	if (nan_value_counts_it != vector_mapping.end()) {
 		nan_value_counts = *child_entries[nan_value_counts_it->second.GetChildIndex(0).GetPrimaryIndex()];
 	}

```

🤔

Tishj: It just looks unreadable, is my only problem..
```c++
	auto lower_bounds_it = vector_mapping.find(static_cast<int32_t>(FieldIds::LOWER_BOUNDS));
	if (lower_bounds_it != vector_mapping.end()) {
		lower_bounds = *child_entries[lower_bounds_it->second.GetChildIndex(0).GetPrimaryIndex()];
	}
	auto upper_bounds_it = vector_mapping.find(static_cast<int32_t>(FieldIds::UPPER_BOUNDS));
	if (upper_bounds_it != vector_mapping.end()) {
		upper_bounds = *child_entries[upper_bounds_it->second.GetChildIndex(0).GetPrimaryIndex()];
	}
	auto value_counts_it = vector_mapping.find(static_cast<int32_t>(FieldIds::VALUE_COUNTS));
	if (value_counts_it != vector_mapping.end()) {
		value_counts = *child_entries[value_counts_it->second.GetChildIndex(0).GetPrimaryIndex()];
	}
	auto null_value_counts_it = vector_mapping.find(static_cast<int32_t>(FieldIds::NULL_VALUE_COUNTS));
	if (null_value_counts_it != vector_mapping.end()) {
		null_value_counts = *child_entries[null_value_counts_it->second.GetChildIndex(0).GetPrimaryIndex()];
	}
	auto nan_value_counts_it = vector_mapping.find(static_cast<int32_t>(FieldIds::NAN_VALUE_COUNTS));
	if (nan_value_counts_it != vector_mapping.end()) {
		nan_value_counts = *child_entries[nan_value_counts_it->second.GetChildIndex(0).GetPrimaryIndex()];
	}
```

.github/workflows/LocalTesting.yml
samansmink: Let's take a look at making this cleaner in a future PR

kevinjqliu:
> CI run has been added 👍

awesome, thanks @Tishj! I can help with getting https://github.com/apache/iceberg-python/pull/2109 merged on the pyiceberg side 

> Just stumbled on this issue: https://github.com/apache/iceberg-go/pull/435
Which states that Spark might complain about the names given to list elements and map keys and values, in the avro file

For this one, we found that the Java Iceberg reader requires the avro field name to be in the form of `f'r{field_id}'`
https://github.com/apache/iceberg-go/pull/435/files#diff-d2ec619bc6d70ff07eca83b594465dce194eeec39a146a8afccb0dd4b6f02f3aR95
This differs from the spec and should be fixed on the java side. This is on my todo list #251 

> Another issue we will probably run into as well:
https://github.com/apache/iceberg-python/issues/2123
Since the avro file has to contain the partition fields in the root metadata, and those fields contain the name

This ones on pyiceberg. We allow schema names that are invalid according to the avro spec. Will be fixed in pyiceberg

PR #327, opened 2025-06-23
[V3 Reads] Add support for V3 deletes (Puffin deletion vectors)

Tishj:
This PR implements https://github.com/duckdb/duckdb-iceberg/discussions/203


src/deletes/deletion_vector.cpp
```diff
@@ -0,0 +1,119 @@
...
+unique_ptr<IcebergDeletionVector> IcebergDeletionVector::FromBlob(data_ptr_t blob_start, idx_t blob_length) {
+	//! https://iceberg.apache.org/puffin-spec/#deletion-vector-v1-blob-type
+
+	auto vector_size = Load<uint32_t>(blob_start);
+	vector_size = BSwap(vector_size);
+	blob_start += sizeof(uint32_t);
+
+	constexpr char DELETION_VECTOR_MAGIC[] = {'\xD1', '\xD3', '\x39', '\x64'};
+	char magic_bytes[4];
+	memcpy(magic_bytes, blob_start, 4);
```
dentiny, line 24: qq: should we check `blob_length` is no smaller than 4, otherwise possible to invalid memory access?

src/deletes/deletion_vector.cpp
```diff
@@ -0,0 +1,123 @@
...
+	return result_p;
+}
+
+void IcebergMultiFileList::ScanPuffinFile(const IcebergManifestEntry &entry) const {
+	auto file_path = entry.file_path;
+	D_ASSERT(!entry.referenced_data_file.empty());
+
+	auto caching_file_system = CachingFileSystem::Get(context);
+
+	auto caching_file_handle = caching_file_system.OpenFile(file_path, FileOpenFlags::FILE_FLAGS_READ);
```
Tmonster, line 66: If you know the file won't change you can call the following
```
OpenFileInfo file_info(file_path);
file_info.extended_info->options["validate_external_file_cache"] = Value::BOOL(false);
```
This will prevent any extra network calls to verify the etag.
I also added a "force_full_download" option, which will download the file completely without making a head request, maybe that's also a good idea to add? Is there a case where we only read part of the deletion vector?

scripts/data_generators/tests/deletion_vectors/q02.sql
```diff
@@ -0,0 +1 @@
+delete from default.deletion_vectors where col % 2 == 0;
```
Tmonster: nit: is there a q01.sql?

src/deletes/deletion_vector.cpp
```diff
@@ -0,0 +1,123 @@
...
+	auto &result = *result_p;
+	result.bitmaps.reserve(amount_of_bitmaps);
+	for (int64_t i = 0; i < amount_of_bitmaps; i++) {
+		auto key = Load<int32_t>(blob_start);
+		blob_start += sizeof(int32_t);
+		vector_size -= sizeof(int32_t);
+
+		size_t bitmap_size =
+		    roaring::api::roaring_bitmap_portable_deserialize_size((const char *)blob_start, vector_size);
+		auto bitmap = roaring::Roaring::readSafe((const char *)blob_start, bitmap_size);
```
Tmonster, line 49: nit: maybe we should assert or check that bitmap_size != 0? Although very unlikely

Tmonster (commented):
Thanks! There is some logic for the bitmaps I don't quite understand, but I have to read up on the spec and step through the code a bit to get it. 
Otherwise I just have some minor comments

src/deletes/deletion_vector.cpp
Tishj: Ah, good suggestion

One file can have multiple deletion vectors, so it could be that we only need a part of the file, if the other deletion vectors are no longer relevant, or aren't relevant for this particular query

Maybe we only want to cache the `content_offset + content_size_in_bytes` ranges, I'm not sure how that works in the external file cache?

EDIT:
I read the PR, I think that's the default behavior, caching byte ranges when they're read from the CachingFileHandle ?
But I do need to change how we call `Read`

src/deletes/deletion_vector.cpp
Tishj: Then `readSafe` will just error, which is fine in my opinion
The deletion vector is just corrupt at that point

src/deletes/deletion_vector.cpp
Tmonster: I think the external_file_cache can store chunks of a file. If a follow up request requests other chunks, the external_file_cache can merge a chunk from the next request and a chunk from the file cache (assuming the file hasn't changed)

src/deletes/deletion_vector.cpp
Tishj: This API is horrible in my opinion, I only need the `bitmap_size` to know how much to forward the pointer *after* deserializing this bitmap.

The second parameter to `readSafe` is a "max bytes" parameter, but the bitmap already marks an end, so it's useless, unless you're not sure you're reading correct data and don't want to exceed your buffer.

What `read` and `readSafe` **should** be returning is the serialized size of the bitmap, so we don't need this `roaring_bitmap_portable_deserialize_size`

src/deletes/deletion_vector.cpp
Tishj: I think it'll only merge if they overlap

PR #332, opened 2025-06-23
Add ability to POST/Write to catalogs with Sigv4 authorization

Tmonster:
A couple of notes here.

This PR
- Adds the ability for catalogs that authorize with sigv4 execute post requests.
- reworks some of the cloud tests so that they happen monday/wednesday/friday and create an issue if something fails
- Removes secrets after every transaction to avoid polluting the catalog

Some notes/Things for future work.
- S3Tables currently doesn't return supported urls, and they do not support all of the defaults from the spec, so there is some special logic currently to handle this case.
- S3tables also does not allow deletes, we store this as a part of our attach_options, but eventually we can read it from `defaults` or `overrides` since it is passed as s3.delete-enabled. [link](https://iceberg.apache.org/docs/latest/aws/#s3-tags)


.github/workflows/CloudTesting.yml
```diff
@@ -51,11 +47,31 @@ jobs:
...
+      - name: install pythong packages
+        run: |
+          python3 -m pip install scripts/requirements.txt
+
+      - name: Set up cloud testing
+        env:
+          AWS_ACCESS_KEY_ID: ${{secrets.S3_ICEBERG_TEST_USER_KEY_ID}}
+          AWS_SECRET_ACCESS_KEY: ${{secrets.S3_ICEBERG_TEST_USER_SECRET}}
+          AWS_DEFAULT_REGION: ${{secrets.S3_ICEBERG_TEST_USER_REGION}}
+          r2_token: ${{secrets.r2_token}}
```
Tmonster, line 59: still need to add an r2 token to the repo

src/include/storage/irc_authorization.hpp
```diff
@@ -7,13 +7,16 @@
...
+enum class IcebergEndpointType : uint8_t { AWS_S3TABLES, AWS_GLUE, INVALID };
+
 enum class IRCAuthorizationType : uint8_t { OAUTH2, SIGV4, NONE, INVALID };
 
 struct IcebergAttachOptions {
 	string endpoint;
 	string warehouse;
 	string secret;
 	string name;
+	bool allows_deletes = true;
```
Tmonster, line 19: this could actually be stored in the overrides/defaults struct of the catalog I think

src/storage/authorization/sigv4.cpp
```diff
@@ -66,7 +66,40 @@ static string GetAwsService(const string &host) {
 unique_ptr<HTTPResponse> SIGV4Authorization::PostRequest(ClientContext &context,
                                                          const IRCEndpointBuilder &endpoint_builder,
                                                          const string &body) {
-	throw NotImplementedException("POST request for SIGV4 Authorization");
+	AWSInput aws_input;
+	aws_input.cert_path = APIUtils::GetCURLCertPath();
```
Tmonster, line 70: This actually shares a ton of code with the GET path, I can probably unify them

test/sql/local/irc/insert/test_basic_insert.test
```diff
@@ -120,12 +120,22 @@ insert into my_datalake.default.insert_test select
...
+2020-08-12	45345	inserted by con2
+
+
 # The second commit failed because both transactions tried to add sequence number: 2
 statement error con2
 commit
 ----
 <REGEX>:.*Requirement failed: branch main has changed.*
 
+mode skip
```
Tishj: intended?

Tmonster:
> Looks great, I'm not really sure I understand the purpose of create_s3_insert_table.py, since it's using pyiceberg, but I guess it makes sure the tables exist for the test, rather than expecting that the tables already exist ?

Yea, it's just to get the remote files back to a state where the tests are valid. Similar to running ` python3 -m scripts.data_generators.generate_data spark-rest --test insert_test`

I'll fix the mode skip 👍

src/storage/irc_transaction.cpp
```diff
@@ -134,68 +200,61 @@ void IRCTransaction::Commit() {
...
+				url_builder.AddPathComponent(catalog.prefix);
+				url_builder.AddPathComponent("namespaces");
+				url_builder.AddPathComponent(table_namespace);
+				url_builder.AddPathComponent("tables");
+				url_builder.AddPathComponent(table_change.identifier.name);
 
-		CommitTransactionToJSON(doc, root_object, transaction);
-		auto transaction_json = JsonDocToString(std::move(doc_p));
+				auto transaction_json = ConstructTableUpdateJSON(table_change);
+				Printer::Print("posting one update: " + transaction_json);
```
Tishj: shouldn't we delete this?

test/sql/local/irc/insert/test_always_ok_insert.test
```diff
@@ -0,0 +1,75 @@
...
+
+require avro
+
+require parquet
+
+require iceberg
+
+require httpfs
+
+mode skip
```
Tishj: this test is skipped, what's the reason?

test/sql/cloud/glue/test_insert_glue.test
```diff
@@ -0,0 +1,174 @@
...
+
+require parquet
+
+require iceberg
+
+require httpfs
+
+require aws
+
+mode skip
```
Tishj: This is also skipped?

test/sql/local/irc/insert/test_always_ok_insert.test
Tmonster: I thought it might be nice to have an insert test that doesn't require modifying the state of the db again, but once we get create table working, this becomes unnecessary. I've removed it

src/storage/irc_transaction.cpp
Tmonster: yessssir

PR #335, opened 2025-06-24
Add UUID value support

patricklucas:
I was getting bounds deserialization errors on a table with a UUID column and traced it here—"UUID not implemented yet". I'm a bit surprised to be the first one to comment on this; I'm writing Iceberg tables using the Java API, but maybe other systems like Spark don't use the native UUID type in Iceberg?

I am not a C++ developer, so I'd appreciate some expert eyes on this. I'm also not familiar with the testing setup—I'm happy to add a test for this, but would like some direction before making something up.


Tishj:
Hey, thanks for raising this, I would like to implement this but at the time I couldn't get a system to properly produce tables with UUID columns

If you could add a table in `data/persistent/<table name>` that has a UUID column that would be greatly appreciated, I can take over the C++ side from there if you like 👍

patricklucas:
Thanks, I'll look into adding the test table.

The main stumbling block I had was at first unpacking into a `hugeint_t` rather than a `uhugeint_t`—literally a single bit out of place!

PR #343, opened 2025-07-01
Add `iceberg_to_ducklake` function, used to do a metadata-only conversion of an Iceberg REST catalog to DuckLake.

Tishj:
### Details

We use the latest metadata version of the Iceberg table to base the conversion on.

Conversion is done in a way that preserves the table history.

We create a DuckLake snapshot based on the `timestamp_ms` of the Iceberg table snapshot, this allows multiple tables to be part of the same DuckLake snapshot (as DuckLake snapshots cover the entire catalog, rather than a single table).
Another notable difference between the two snapshots is that DuckLake snapshots don't only cover data changes, a snapshot can consist only of a schema change.

Because of this, we can't convert **empty** tables, as no snapshot exists for those tables, so they are skipped in the conversion.
A change to a partition/schema that is not followed by a new Iceberg snapshot will not show up in the conversion, as we can't pinpoint those changes to a snapshot.

To correctly convert the data/delete file metadata, we look at the diff between each snapshot to determine when data/delete files are added and deleted.

The same approach is taken for schema changes, the DuckLake table schema consists of separate columns, every add/alter becomes a new entry, which is updated when the column version is dropped (either by a drop or by an alter of the column).

### Parameters

The `skip_table: VARCHAR[]` named parameter can be used to provide a list of tables to skip in the conversion.
This is useful in case the catalog contains tables that do not meet the conversion requirements. (see Requirements)

### Implementation Details

We iterate through the REST catalogs top-level namespaces and their tables, for every table we iterate through its snapshots.

Using the schema_id of the snapshot we use the first snapshot that references it as  the `begin_snapshot` of a new schema.

The same is done for partition specs referenced by a `manifest_file` struct, whatever snapshot these first appear in will be used as the `begin_snapshot` of the partition information.

The following ids are dictated by the `begin_snapshot` that is assigned to them:
- `table_id` (`next_catalog_id`)
- `schema_id` (`next_catalog_id`)
- `data_file_id` (`next_file_id`)
- `delete_file_id` (`next_file_id`)
- `partition_id` (no explicit reference id)

In the `ducklake_snapshot` table, the `next_catalog_id` and `next_file_id` have to be bigger than the ids that are used in the snapshot, and the ids belonging to the next snapshot have to **start** with these `next_*` ids.

Because of this, all assignment of these ids are delayed, until we've processed all tables (and thus created all snapshots)

Thankfully, these references can't be cyclic and there is a defined order we can assign the ids in that ensures all dependent ids are assigned by the time we have to finalize the entry (assign its id)

### Requirements

- Equality Deletes are not supported.
- Deletion Vectors are not supported [1].
- Only `parquet` is supported for the data files.
- Only 1 active positional delete can exist for a data file.
- A positional delete can only contains deletes for a single data file.
- `bucket`, `truncate` and `void` transforms are not supported.

1. There are plans to support this in DuckLake, at which point we can add support to the function for it.

#### Use `metadata-log` to timestamp orphan schema changes.

We can perhaps substitute some of the shortcomings of the Iceberg snapshot by making use of the `metadata-log`:
```json
  "metadata-log" : [ {
    "timestamp-ms" : 1746881394503,
    "metadata-file" : "data/persistent/partition_float/metadata/v1.metadata.json"
  } ]
```
To pinpoint changes made to a schema (namespace), table, table schema or partition information that isn't part of a snapshot to a timestamp, in order to create a suitable DuckLake snapshot for them.

#### Support incremental changes with depth

The function should support incremental conversion, currently the function expects the DuckLake catalog to be entirely empty when the function is ran.

#### Support conversion without depth

This would allow incremental changes to be made more easily, at the cost of sacrificing history.
The latest (or a selected) snapshot would be used as the source of truth, dating this ducklake snapshot to the current time (ignoring the `timestamp_ms` of the iceberg snapshot)

### DIFF
--- a/src/iceberg_functions/iceberg_to_ducklake.cpp
+++ b/src/iceberg_functions/iceberg_to_ducklake.cpp
@@ -70,7 +70,7 @@ public:
 		base_file_id = serializer.next_file_id;
 
 		//! Update the serializer to point to the next id starts
-		serializer.schema_version += catalog_changes;
+		serializer.schema_version += !!catalog_changes;
 		serializer.next_catalog_id += catalog_additions;
 		serializer.next_file_id += files_added;
 


src/iceberg_functions/iceberg_to_ducklake.cpp
```diff
@@ -0,0 +1,1546 @@
...
+				throw InvalidInputException("'skip_tables' has to be provided as a list of strings");
+			}
+			auto &tables = ListValue::GetChildren(kv.second);
+			for (auto &table : tables) {
+				ret->table_names_to_skip.insert(table.GetValue<string>());
+			}
+		}
+	}
+
+	schema_set.LoadEntries(context);
```
Tmonster, line 1394: Why are we Loading all the entries and adding this information in the BindStage? I would like type of functionality should go during function execution?

src/iceberg_functions/iceberg_to_ducklake.cpp
```diff
@@ -0,0 +1,1546 @@
...
+
+public:
+	string FinalizeEntry(DuckLakeMetadataSerializer &serializer) {
+		//! Set these, so we can use these to create the correct ids for the catalog/file entries added by this snapshot
+		base_schema_version = serializer.schema_version;
+		base_catalog_id = serializer.next_catalog_id;
+		base_file_id = serializer.next_file_id;
+
+		//! Update the serializer to point to the next id starts
+		serializer.schema_version += catalog_changes;
```
Tmonster: does the schema version need to be increased by the number of catalog changes?

src/iceberg_functions/iceberg_to_ducklake.cpp
```diff
@@ -0,0 +1,1546 @@
...
+			tables.FillEntry(context, table);
+			ret->AddTable(table, context, options);
+		}
+	}
+
+	ret->AssignSchemaBeginSnapshots();
+
+	ret->sql_statements = ret->CreateSQLStatements();
+	for (auto &statement : ret->sql_statements) {
+		Printer::Print(statement);
```
Tmonster: nit: I guess these should be removed?

src/iceberg_functions/iceberg_to_ducklake.cpp
```diff
@@ -0,0 +1,1546 @@
...
+				auto max_value = stats.max_value.IsNull() ? "NULL" : "'" + stats.max_value.ToString() + "'";
+				auto values = StringUtil::Format("VALUES(%d, %d, %s, %s, %s, %s);", table_id, column_id, contains_null,
+				                                 contains_nan, min_value, max_value);
+				sql.push_back(
+				    StringUtil::Format("INSERT INTO {METADATA_CATALOG}.ducklake_table_column_stats %s", values));
+			}
+		}
+
+		//! ducklake_snapshot_changes
+		for (auto &it : snapshots) {
```
Tmonster, line 1228: Looking at the sql statements produced on my local, it looks like the schema was created in the last snapshot, which shouldn't be the case right? Is this the same in your local testing?
```
INSERT INTO __ducklake_metadata_my_ducklake.ducklake_snapshot_changes VALUES(0, 'created_table:"default"."t1",inserted_into_table:0');
INSERT INTO __ducklake_metadata_my_ducklake.ducklake_snapshot_changes VALUES(1, 'created_table:"default"."t2",inserted_into_table:1');
INSERT INTO __ducklake_metadata_my_ducklake.ducklake_snapshot_changes VALUES(2, 'created_table:"default"."create_table_t1_1",inserted_into_table:2');
INSERT INTO __ducklake_metadata_my_ducklake.ducklake_snapshot_changes VALUES(3, 'created_table:"default"."create_table_t2_1",inserted_into_table:3');
INSERT INTO __ducklake_metadata_my_ducklake.ducklake_snapshot_changes VALUES(4, 'created_table:"default"."new_table",inserted_into_table:4');
INSERT INTO __ducklake_metadata_my_ducklake.ducklake_snapshot_changes VALUES(5, 'created_table:"default"."new_table_2",inserted_into_table:5');
INSERT INTO __ducklake_metadata_my_ducklake.ducklake_snapshot_changes VALUES(6, 'created_table:"default"."from_parquet_table",inserted_into_table:6');
INSERT INTO __ducklake_metadata_my_ducklake.ducklake_snapshot_changes VALUES(7, 'created_table:"default"."new_table_3",inserted_into_table:7');
INSERT INTO __ducklake_metadata_my_ducklake.ducklake_snapshot_changes VALUES(8, 'created_table:"default"."new_table_4",inserted_into_table:8');
INSERT INTO __ducklake_metadata_my_ducklake.ducklake_snapshot_changes VALUES(9, 'created_table:"default"."new_table_5",inserted_into_table:9');
INSERT INTO __ducklake_metadata_my_ducklake.ducklake_snapshot_changes VALUES(10, 'created_table:"default"."my_new_table",inserted_into_table:10');
INSERT INTO __ducklake_metadata_my_ducklake.ducklake_snapshot_changes VALUES(11, 'created_table:"default"."new_table_6",inserted_into_table:11');
INSERT INTO __ducklake_metadata_my_ducklake.ducklake_snapshot_changes VALUES(12, 'created_schema:"default",created_table:"default"."new_table_7",inserted_into_table:12');
```

Tmonster (commented):
Looks good. I keep getting an error though when trying to run the test myself.
```
INTERNAL Error:
Calling GetValueInternal on a value that is NULL

Stack Trace:

0        _ZN6duckdb9ExceptionC2ENS_13ExceptionTypeERKNSt3__112basic_stringIcNS2_11char_traitsIcEENS2_9allocatorIcEEEE + 64
1        _ZN6duckdb17InternalExceptionC1ERKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEE + 20
2        _ZNK6duckdb5Value16GetValueInternalIyEET_v + 1896
3        _ZN6duckdb23DuckLakeMetadataManager16GetFilesForTableERNS_18DuckLakeTableEntryENS_16DuckLakeSnapshotERKNSt3__112basic_stringIcNS4_11char_traitsIcEENS4_9allocatorIcEEEE + 1356
4        _ZN6duckdb21DuckLakeMultiFileList16GetFilesForTableEv + 128
5        _ZN6duckdb21DuckLakeMultiFileList8GetFilesEv + 80
6        _ZN6duckdb21DuckLakeMultiFileList17GetTotalFileCountEv + 12
7        duckdb::MultiFileFunction<duckdb::ParquetMultiFileInfo>::MultiFileProgress(duckdb::ClientContext&, duckdb::FunctionData const*, duckdb::GlobalTableFunctionState const*) + 64
8        duckdb::PhysicalTableScan::GetProgress(duckdb::ClientContext&, duckdb::GlobalSourceState&) const + 52
9        duckdb::Pipeline::GetProgress(duckdb::ProgressData&) + 180
10       duckdb::Executor::GetPipelinesProgress(duckdb::ProgressData&) + 144
11       duckdb::ProgressBar::Update(bool) + 60
12       duckdb::ClientContext::ExecuteTaskInternal(duckdb::ClientContextLock&, duckdb::BaseQueryResult&, bool) + 124
13       duckdb::PendingQueryResult::ExecuteInternal(duckdb::ClientContextLock&) + 120
14       duckdb::PendingQueryResult::Execute() + 56
15       duckdb_shell_sqlite3_print_duckbox + 520
16       duckdb_shell::ShellState::ExecutePreparedStatement(sqlite3_stmt*) + 588
17       duckdb_shell::ShellState::ExecuteSQL(char const*, char**) + 388
18       duckdb_shell::ShellState::RunOneSqlLine(char*) + 248
19       duckdb_shell::ShellState::ProcessInput() + 888
20       main + 3996
21       start + 2360
```

Ducklake works by itself (i.e creating tables etc.). Have you encountered this @Tishj ?

Tmonster:
I got my error by just creating all the spark rest tables and running the conversion script in the test
```
python3 -m scripts.data_generators.generate_data spark-rest
```

src/iceberg_functions/iceberg_to_ducklake.cpp
Tishj: It can be increased by 1 instead, so I could add a `!!` in front of this to only increment it once.
The `schema_version` is just used to indicate to DuckLake that the schema has changed in the snapshot, otherwise it will cache the old version

src/iceberg_functions/iceberg_to_ducklake.cpp
Tishj: Hmm now that you mention it, I have the same result
Good catch, I'll see what's up there

src/iceberg_functions/iceberg_to_ducklake.cpp
Tishj: I don't think it matters, we're not relying on any runtime data, so we can already do all this during the bind / initialization of the global state

PR #346, opened 2025-07-03
Fix struct filter issue, implement hour+month transform,

### LINKED ISSUE #339: `Expected to find a StructFilter for child_index: 2, found 0 instead` when querying nested data produced by redpanda
Hi,

I'm testing again DuckDB interoperability with data produced by Redpanda and ran into the following error:

```sql
SELECT *
  FROM iceberg_scan('s3://panda-bucket/redpanda-iceberg-catalog/redpanda/foo/metadata/v1.metadata.json')
  WHERE redpanda.partition = 0;
```

```
Invalid Configuration Error:
Expected to find a StructFilter for child_index: 2, found 0 instead
```

Table structure can be seen in the following query:

```sql
SELECT *
  FROM iceberg_scan('s3://panda-bucket/redpanda-iceberg-catalog/redpanda/foo/metadata/v1.metadata.json');
```

```
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┬─────────────┐
│                                                          redpanda                                                           │    value    │
│ struct("partition" integer, "offset" bigint, "timestamp" timestamp, headers struct("key" blob, "value" blob)[], "key" blob) │    blob     │
├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────┤
│ {'partition': 0, 'offset': 0, 'timestamp': '2025-06-26 19:45:32.478', 'headers': NULL, 'key': NULL}                         │ hello world │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┴─────────────┘
```

Attaching copy of s3 table contents:  [issue.tar.gz](https://github.com/user-attachments/files/20932394/issue.tar.gz).

Any help appreciated. Seems to work with Spark Iceberg nested data but then Spark has no problem querying redpanda data either. Not sure where exactly the issue lies.

Tishj:
This PR fixes #339 

I've also added the HourTransform, as it's used by the test table.
And while I was at it I figured I'd also add the missing `MonthTransform`, there aren't tests for it, but the change is straight forward enough that I'm confident it's fine without it.

I've also fixed an issue in the `DayTransform`, I don't think we were correctly transforming the constant to days since epoch.

Last, but definitely not least, I've added the `relativize-iceberg-table` script made by @patricklucas (I've transformed it to python though)
This will be a huge help going forward in making reproductions created by users transformable to be ran and submitted as tests to the repo 🥳


patricklucas:
I wondered if you'd spot the script. :) Mostly threw an LLM at it so quite possible there's some redundancy/edge cases in those jq scripts.

Also FYI, noticed that the test tables you added do reference absolute paths.

Tishj:
Whoops, those should be relative to `data/persistent` 😅 
I threw an LLM at converting it to python, then needed to fix a couple small things
Looks like I need to fix another small thing in the paths used

scripts/relativize_iceberg_table.py
```diff
@@ -0,0 +1,203 @@
...
+def backup_and_replace_metadata(table_path: Path, temp_metadata_dir: Path):
+    backup_dir = table_path / f"metadata.backup.{datetime.now().strftime('%Y%m%d_%H%M%S')}"
+    print(f"Creating backup at: {backup_dir}")
+    shutil.move(str(table_path / "metadata"), str(backup_dir))
+    shutil.move(str(temp_metadata_dir), str(table_path / "metadata"))
+
+
+def main():
+    args = parse_args()
+    table_path = Path(args.table_path).resolve()
```
Tishj: This is the culprit for producing absolute paths

PR #354, opened 2025-07-04
Only revert distribution pipeline

Tmonster:


carlopi:
I think solved via https://github.com/duckdb/duckdb-iceberg/pull/356

PR #356, opened 2025-07-07
Make `main` branch wasm compatible

carlopi:
This is supposed to pass CI, and make so that all platforms build from the same branch: `main`. If this is merged, `wasm` branch can then be deleted (or left orphan).

I guess the bad trick/hack is in the Makefile changes, where I perform a Wasm only step to remove both anchoring `openssl` to 3.0.8 and remove aws (as it's done also at the CMake level). I would say this is OK, expecially given changes are isolated to a single file and only for WebAssembly.

The fact that this complies do not really means this is properly functional, also considering tests are not actually enabled for Wasm at the moment, so I would still consider this experimental, and will iterate / test more on a side.

Also this moves submodules and workflow references to 1.3.2, that should allow to close https://github.com/duckdb/duckdb-iceberg/pull/353 and https://github.com/duckdb/duckdb-iceberg/pull/354


carlopi:
This on my side is ready to go.

PR #357, opened 2025-07-08
Implement create table support

Tmonster:
This PR implements support for a create table statement against an Iceberg rest catalog. To implement this, the following was done

- During the PlanCreateTable, we send the request to the catalog to create the table. If the catalog supports stage_create, we set `stage-create=true`. We send the request in the plan stage so we can get the correct field ids. Sometimes catalogs change them. If a catalog does not support `stage-create`, we go ahead and create a table. 
- During commit we hit the table create endpoint again, but with `stage-create=false`. 
- During commit, we recreate temporary secrets since we are in another "fake" transaction that doesn't have access to them.

```
create table t1 (a int);
create table t2 as select * from read_csv('path/to/csv');
create table t3 as select * from t1, t2 where t1.a = t2.b;
```

TODO:
- Fixed array types are not supported. An error message is thrown to cast to list. This error can be more informative
- ~~Find a way to not eagerly hit the create-table endpoint with stage-create = true.~~ not really possible
- If a catalog does not support `stage-create`, we eagerly create the table during a transaction. If the transaction is aborted, we need to hit the drop table endpoint. 
- Support create or replace statements 
- Support Create if not exists statements
- Create a TableUpdateToJSON function for table updates. Currently they pollute the irc_transaction.cpp file


extension_config.cmake
```diff
@@ -9,6 +9,11 @@ duckdb_extension_load(iceberg
 
 duckdb_extension_load(tpch)
 duckdb_extension_load(icu)
+duckdb_extension_load(ducklake
```
Tishj, line 12: This should probably be removed ?

src/iceberg_functions/iceberg_to_ducklake.cpp
```diff
@@ -819,9 +819,11 @@ struct IcebergToDuckLakeBindData : public TableFunctionData {
 	}
 
 public:
-	void AddTable(IcebergTableInformation &table_info, ClientContext &context, const IcebergOptions &options) {
-		auto &metadata = table_info.table_metadata;
-		if (table_names_to_skip.count(table_info.name)) {
+	void AddTable(optional_ptr<IcebergTableInformation> table_info, ClientContext &context,
```
Tishj: I don't see the need for this change, we can just add `*` add the call site instead, if we're asserting that it's always non-null anyways

src/include/storage/iceberg_transaction_data.hpp
```diff
@@ -9,6 +9,7 @@
 #include "duckdb/function/copy_function.hpp"
 #include "storage/iceberg_table_update.hpp"
 #include "storage/table_update/iceberg_add_snapshot.hpp"
+#include "storage/table_create/iceberg_create_table_request.hpp"
```
Tishj, line 13: `table_update` was chosen because that aligns with the spec, `table_create` does not
This should live in `storage/table_update` instead

```yml
    TableUpdate:
      anyOf:
        - $ref: '#/components/schemas/AssignUUIDUpdate'
        - $ref: '#/components/schemas/UpgradeFormatVersionUpdate'
        - $ref: '#/components/schemas/AddSchemaUpdate'
        - $ref: '#/components/schemas/SetCurrentSchemaUpdate'
        - $ref: '#/components/schemas/AddPartitionSpecUpdate'
        - $ref: '#/components/schemas/SetDefaultSpecUpdate'
```

src/include/storage/irc_table_entry.hpp
```diff
@@ -29,15 +29,17 @@ struct ICTableInfo {
 
 class ICTableEntry : public TableCatalogEntry {
 public:
-	ICTableEntry(IcebergTableInformation &table_info, Catalog &catalog, SchemaCatalogEntry &schema,
+	ICTableEntry(shared_ptr<IcebergTableInformation> table_info, Catalog &catalog, SchemaCatalogEntry &schema,
+	             CreateTableInfo &info);
+	ICTableEntry(shared_ptr<IcebergTableInformation> table_info, IRCatalog &catalog, SchemaCatalogEntry &schema,
```
Tishj: We don't need both of these

src/include/utils/json_utils.hpp
```diff
@@ -0,0 +1,23 @@
...
+#include "yyjson.hpp"
+#include "duckdb/common/printer.hpp"
+#include "duckdb/common/file_system.hpp"
+
+using namespace duckdb_yyjson;
+namespace duckdb {
+
+class JSONUtils {
+public:
+	static string JsonDocToString(yyjson_mut_doc *doc);
```
Tishj: This exists: `ICUtils::JsonToString`
Please either remove this one or the old one

src/metadata/iceberg_column_definition.cpp
```diff
@@ -172,4 +172,32 @@ unique_ptr<IcebergColumnDefinition> IcebergColumnDefinition::ParseStructField(re
...
+	case LogicalTypeId::VARCHAR:
+	case LogicalTypeId::DATE:
+	case LogicalTypeId::UINTEGER:
+	case LogicalTypeId::HUGEINT:
+	case LogicalTypeId::BIGINT:
+	case LogicalTypeId::FLOAT:
+	case LogicalTypeId::DOUBLE:
+	case LogicalTypeId::DECIMAL:
+	case LogicalTypeId::UUID:
+	case LogicalTypeId::BLOB: // TODO: check BLOB and BIT 'binary' primitive type ?
```
Tishj: BIT is not supported, it doesn't map to an Iceberg type
There are also other types here that don't map

Supported types are listed here:
https://iceberg.apache.org/spec/#primitive-types

For example there are no unsigned types in the spec

src/storage/iceberg_insert.cpp
```diff
@@ -229,7 +241,8 @@ SinkFinalizeType IcebergInsert::Finalize(Pipeline &pipeline, Event &event, Clien
 // Helpers
 //===--------------------------------------------------------------------===//
 string IcebergInsert::GetName() const {
-	return table ? "ICEBERG_INSERT" : "ICEBERG_CREATE_TABLE_AS";
+	D_ASSERT(table);
+	return "ICEBERG_INSERT";
```
Tishj, line 246: I'm confused, don't we reuse the same logic for Create table as?
equivalent logic in ducklake:
```c++
unique_ptr<GlobalSinkState> DuckLakeInsert::GetGlobalSinkState(ClientContext &context) const {
	optional_ptr<DuckLakeTableEntry> table_ptr;
	if (info) {
		// CREATE TABLE AS - create the table
		auto &catalog = schema->catalog;
		auto &ducklake_schema = schema.get_mutable()->Cast<DuckLakeSchemaEntry>();
		auto transaction = catalog.GetCatalogTransaction(context);
		table_ptr = &ducklake_schema.CreateTableExtended(transaction, *info, table_uuid, table_data_path)
		                 ->Cast<DuckLakeTableEntry>();
	} else {
		// INSERT INTO
		table_ptr = table;
	}
	return make_uniq<DuckLakeInsertGlobalState>(*table_ptr);
}
```

src/storage/iceberg_type.cpp
```diff
@@ -0,0 +1,128 @@
...
+	rest_api_objects::Type rest_type;
+
+	switch (type.id()) {
+	case LogicalTypeId::MAP: {
+		rest_type.has_primitive_type = false;
+		rest_type.has_map_type = true;
+		rest_type.map_type = rest_api_objects::MapType();
+		auto aux = type.AuxInfo()->Cast<ListTypeInfo>();
+		auto child_type = aux.child_type.AuxInfo()->Cast<StructTypeInfo>();
+		// How do I get the child types for a map type?
```
Tishj: I think this is solved and can be removed 👍

src/storage/iceberg_type.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+rest_api_objects::Type IcebergTypeHelper::CreateIcebergRestType(LogicalType &type, std::function<idx_t()> get_next_id) {
+	rest_api_objects::Type rest_type;
+
+	switch (type.id()) {
+	case LogicalTypeId::MAP: {
+		rest_type.has_primitive_type = false;
+		rest_type.has_map_type = true;
+		rest_type.map_type = rest_api_objects::MapType();
+		auto aux = type.AuxInfo()->Cast<ListTypeInfo>();
```
Tishj: This is not necessary though

You can use `MapType::KeyType(type);` and `MapType::ValueType(type);`

src/storage/iceberg_type.cpp
```diff
@@ -0,0 +1,128 @@
...
+		rest_type.map_type.value = make_uniq<rest_api_objects::Type>(
+		    IcebergTypeHelper::CreateIcebergRestType(child_type.child_types.back().second, get_next_id));
+		rest_type.map_type.value_required = false;
+		return rest_type;
+	}
+	case LogicalTypeId::STRUCT: {
+		rest_type.has_primitive_type = false;
+		rest_type.has_struct_type = true;
+		rest_type.struct_type = rest_api_objects::StructType();
+		auto stuct_aux = type.AuxInfo()->Cast<StructTypeInfo>();
```
Tishj: It's never required to access the aux directly, we have helper methods
`auto &children = StructType::GetChildTypes(type);`

src/storage/iceberg_type.cpp
```diff
@@ -0,0 +1,128 @@
...
+			rest_type.struct_type.fields.push_back(std::move(struct_child));
+		}
+		return rest_type;
+	}
+	case LogicalTypeId::LIST: {
+		rest_type.has_primitive_type = false;
+		rest_type.has_list_type = true;
+		rest_type.list_type = rest_api_objects::ListType();
+		auto list_aux = type.AuxInfo()->Cast<ListTypeInfo>();
+		rest_type.list_type.type = IcebergTypeHelper::LogicalTypeToIcebergType(list_aux.child_type);
```
Tishj: `ListType::GetChildType(type)`

src/storage/iceberg_insert.cpp
Tishj: ah I think the answer is that we need the field ids when we're preparing the copy

DuckLake does this:
```c++
	// generate the field ids to be written by the parquet writer
	shared_ptr<DuckLakeFieldData> generated_ids;
	if (!copy_input.field_data) {
		// CTAS - generate new ids from columns
		generated_ids = DuckLakeFieldData::FromColumns(copy_input.columns);
	}
	auto &field_ids = copy_input.field_data ? *copy_input.field_data : *generated_ids;
	vector<Value> field_input;
	field_input.push_back(WrittenFieldIds(field_ids, copy_input.virtual_columns));
```
But that's kind of duplicating the work in our case

src/storage/irc_catalog.cpp
```diff
@@ -437,6 +435,11 @@ unique_ptr<Catalog> IRCatalog::Attach(StorageExtensionInfo *storage_info, Client
 		} else if (lower_name == "endpoint") {
 			attach_options.endpoint = StringUtil::Lower(entry.second.ToString());
 			StringUtil::RTrim(attach_options.endpoint, "/");
+		} else if (lower_name == "supports_stage_create") {
+			auto result = entry.second.GetValue<string>();
```
Tishj: `auto result = entry.second.DefaultCastAs(LogicalType::BOOLEAN).GetValue<bool>();`

Doing a find for `false` is error prone

src/include/storage/irc_table_set.hpp
```diff
@@ -21,7 +21,8 @@ struct IcebergTableInformation {
 
 public:
 	optional_ptr<CatalogEntry> GetSchemaVersion(optional_ptr<BoundAtClause> at);
-	optional_ptr<CatalogEntry> CreateSchemaVersion(IcebergTableSchema &table_schema);
+	static optional_ptr<CatalogEntry> CreateSchemaVersion(shared_ptr<IcebergTableInformation> table_info,
```
Tishj: I see no reason for this change, you need an `IcebergTableInformation` regardless, there is no check for `table_info == nullptr` in there, so we only accept this with a non-null shared pointer.

extension_config.cmake
Tmonster: Any particular reason why? I added it so the iceberg to ducklake test could be run in CI

src/include/storage/iceberg_transaction_data.hpp
Tmonster: There is a `CreateTableReqeust` schema though. https://github.com/apache/iceberg/blob/9faf4311d71ce313e81dea07290b3aea9298422c/open-api/rest-catalog-open-api.yaml#L3503
 I feel putting this in `TableUpdate` is also not correct since we are not updating an existing table. I could put it into `storage/create_table_request.cpp` but the `storage` directory seems a bit polluted as it is

src/metadata/iceberg_column_definition.cpp
Tmonster: True, I was thinking we could cast unsigned to long or something, but forgot about implementing the rest of that. It can be done later. 
Although I should probably add a test to make sure we error out when trying to write a UINT column

src/storage/iceberg_insert.cpp
Tmonster: Yea, I think I can make it work, but I do like getting the correct field Ids before execution

extension_config.cmake
Tishj: Ah yea maybe its fine? I think this is only a development requirement

src/metadata/iceberg_column_definition.cpp
Tmonster: Hmm, already covered by the `create table as select ... from test_all_types()`

src/storage/iceberg_insert.cpp
Tishj: Yea agreed, fine to leave it as is 👍

src/include/storage/iceberg_transaction_data.hpp
Tmonster: I think we can leave the `storage/table_update` folder for creating update and or requirement objects for the table update endpoint. `storage/table_create` can still be left as is to create and populate a CreateTableRequest object

src/storage/create_table/iceberg_create_table_request.cpp
```diff
@@ -0,0 +1,175 @@
...
+	auto column_iterator = table_entry->GetColumns().Logical();
+	idx_t field_id = 1;
+
+	auto next_field_id = [&field_id]() -> idx_t {
+		return field_id++;
+	};
+
+	for (auto column = column_iterator.begin(); column != column_iterator.end(); ++column) {
+		auto name = (*column).Name();
+		// TODO: is this correct?
```
Tishj: I think so, the ColumnDefinition doesn't have constraints, and `required=true` would only be set if there was a `NOT NULL` constraint on the column.

src/storage/create_table/iceberg_create_table_request.cpp
```diff
@@ -0,0 +1,175 @@
...
+		bool required = false;
+		auto logical_type = (*column).GetType();
```
Tishj, line 111: nit (feel free to ignore if there is no other reason to touch the PR):
`auto &`

src/include/storage/table_create/iceberg_create_table_request.hpp
```diff
@@ -0,0 +1,35 @@
...
+struct YyjsonDocDeleter;
+struct IcebergTableInformation;
+class ICTableEntry;
+
+struct IcebergCreateTableRequest {
+
+	IcebergCreateTableRequest(shared_ptr<IcebergTableSchema> schema, string table_name);
+
+public:
+	void CreateManifest(DatabaseInstance &db, ClientContext &context, IcebergCommitState &commit_state);
```
Tishj: This method does not have a definition

src/storage/iceberg_insert.cpp
```diff
@@ -373,10 +362,79 @@ PhysicalOperator &IRCatalog::PlanInsert(ClientContext &context, PhysicalPlanGene
...
 	insert.children.push_back(physical_copy);
 
 	return insert;
 }
 
+PhysicalOperator &IRCatalog::PlanCreateTableAs(ClientContext &context, PhysicalPlanGenerator &planner,
+                                               LogicalCreateTable &op, PhysicalOperator &plan) {
+	auto &create_info = op.info->Base();
+
+	// TODO: check if create_info contains partitioned information, if yes, error
```
Tishj: Dead code?
I don't think we (DuckDB) can create a table directly with partition information, so we shouldn't have to worry about this

src/storage/iceberg_insert.cpp
```diff
@@ -373,10 +362,79 @@ PhysicalOperator &IRCatalog::PlanInsert(ClientContext &context, PhysicalPlanGene
...
+	auto &irc_transaction = IRCTransaction::Get(context, catalog);
+
+	// create the table. Takes care of committing to rest catalog and getting the metadata location etc.
+	// setting the schema
+	auto table = ic_schema_entry.CreateTable(irc_transaction, context, *op.info);
+	auto &ic_table = table->Cast<ICTableEntry>();
+	// we've created the table, since we are running plan create table as, we also need to load
+	// credentials into our secrets for when we copy files
+	ic_table.PrepareIcebergScanFromEntry(context);
+	// ic_table.table_info.GetVendedCredentials(context);
```
Tishj: Dead code?

src/storage/iceberg_insert.cpp
```diff
@@ -373,10 +362,79 @@ PhysicalOperator &IRCatalog::PlanInsert(ClientContext &context, PhysicalPlanGene
...
+	IcebergCopyInput copy_input(context, ic_table);
+	vector<Value> field_input;
+	field_input.push_back(WrittenFieldIds(table_schema));
+	copy_input.options["field_ids"] = std::move(field_input);
+
+	auto &physical_copy = IcebergInsert::PlanCopyForInsert(context, planner, copy_input, plan);
+	physical_index_vector_t<idx_t> column_index_map;
+	auto &insert = planner.Make<IcebergInsert>(op, ic_table, column_index_map);
+
+	//	return planner.Make<IcebergInsert>(op, the_schema.get(), std::move(create_info));
```
Tishj: Dead code?

src/storage/irc_catalog.cpp
```diff
@@ -12,7 +12,9 @@
 #include "duckdb/parser/parsed_data/create_schema_info.hpp"
 #include "duckdb/main/attached_database.hpp"
 #include "rest_catalog/objects/catalog_config.hpp"
+#include "duckdb/planner/operator/logical_create_table.hpp"
 #include "storage/irc_catalog.hpp"
+// #include "storage/iceberg_insert.hpp"
```
Tishj: Dead code?

src/storage/irc_table_entry.cpp
```diff
@@ -27,6 +27,11 @@ ICTableEntry::ICTableEntry(IcebergTableInformation &table_info, Catalog &catalog
 	this->internal = false;
 }
 
+/*ICTableEntry::ICTableEntry(shared_ptr<IcebergTableInformation> table_info, IRCatalog &catalog,
```
Tishj: Dead code?

src/storage/irc_transaction.cpp
```diff
@@ -140,6 +150,55 @@ static rest_api_objects::TableRequirement CreateAssertRefSnapshotIdRequirement(I
...
+                                                                 bool stage_create) {
+	auto &ic_catalog = table->catalog.Cast<IRCatalog>();
+	// Stupid hack for GLUE catalog
+	if (!stage_create && !ic_catalog.attach_options.supports_stage_create) {
+		// if the catalog does not support stage create, and stage_create = false
+		// it means we are trying to commit the table. We have already hit the endpoint with stage_create = true
+		// so this second request will result in an error. so we early out
+		return rest_api_objects::LoadTableResult();
+	}
+	// TODO: add D_ASSERT for the post table url
```
Tishj: Unfinished todo?

src/storage/irc_transaction.cpp
```diff
@@ -251,6 +327,9 @@ void IRCTransaction::Commit() {
 		}
 		DropSecrets(*context);
 	} catch (std::exception &ex) {
+		if (!new_tables.empty()) {
+			// hit drop enpoints.
```
Tishj: Unfinished todo ?

src/storage/irc_transaction.cpp
```diff
@@ -192,13 +251,32 @@ rest_api_objects::CommitTransactionRequest IRCTransaction::GetTransactionRequest
...
+
+	if (dirty_tables.empty() && new_tables.empty()) {
 		return;
 	}
 
 	Connection temp_con(db);
 	temp_con.BeginTransaction();
 	auto &context = temp_con.context;
+
+	if (!new_tables.empty()) {
```
Tishj: I'm not entirely sure of the purpose of `new_tables`, we can tell from the ICTableEntry whether it was new, from the  `unique_ptr<IcebergCreateTableRequest> create;` in its transaction data, right?

So we can always just perform these steps if we see that, when we iterate through our dirty tables

src/include/metadata/iceberg_table_metadata.hpp
```diff
@@ -48,13 +48,16 @@ struct IcebergTableMetadata {
...
 	int32_t default_spec_id;
+	// int32_t default_sort_order_id;
 
 	bool has_current_snapshot = false;
 	int64_t current_snapshot_id;
 	int64_t last_sequence_number;
 
 	//! partition_spec_id -> partition spec
 	unordered_map<int32_t, IcebergPartitionSpec> partition_specs;
+	//! sort_order_id -> sort order
```
Tishj: Is this supposed to be commented out?

Tmonster: I'll delete it. Eventually I think we will need it, but not until we support altering tables/partitions/sort orders

src/storage/create_table/iceberg_create_table_request.cpp
Tmonster: Since column is a column_iterator object, I have to dereference column. Would also rather use `auto &logical_type`

Tishj: Ah I see:
```c++
LogicalType GetType() const;
```
There is no reference method, that's fine then 👍 
Thanks for checking

src/storage/irc_transaction.cpp
```diff
@@ -134,15 +218,54 @@ static rest_api_objects::TableRequirement CreateAssertRefSnapshotIdRequirement(I
 	return req;
 }
 
+rest_api_objects::LoadTableResult IRCTransaction::CommitNewTable(ClientContext &context, const ICTableEntry *table) {
+	auto &ic_catalog = table->catalog.Cast<IRCatalog>();
+	auto table_namespace = table->schema.name;
```
Tishj: Can we move this to `catalog_api.cpp` ?
I'd prefer all the interaction with the REST API to live there.

Also because the `namespace` parameter has to be generated, to support nested namespaces.
The logic for that is in `catalog_api.cpp` (`GetEncodedSchemaName`)

We can generate the body here, then forward that to `IRCAPI::CommitNewTable` (or something similar)

Tmonster: done. For now I'm still generating the body in `CommitNewTable` and passing a table entry. Maybe later we can revisit if this is the best place to have it be generated

Issue #348, opened 2025-07-04
Nested namespaces in REST catalogs


ThomasDsantos:
Hello guys ! First, thanks for your amazing work on iceberg support !

I was doing some tests with iceberg REST catalog and `duckdb==1.3.1` in a python environment.

I have a Lakekeeper + Minio catalog. I created a namespace `ns` with a nested namespace `nested`, which contains my final table `table`.

When I try to access it with following commands

``` python
import duckdb

conn = duckdb.connect(database=":memory:")
conn.execute("""
    INSTALL httpfs; LOAD httpfs;
    INSTALL iceberg; LOAD iceberg;

    SET http_keep_alive=false;
    SET memory_limit = '4GB';
    SET unsafe_enable_version_guessing = true;

    CREATE SECRET lakekeeper_secret (
        TYPE ICEBERG,
        TOKEN 'dummy_token'
    );

    ATTACH 'catalog' as catalog (
        TYPE ICEBERG,
        ENDPOINT 'http://localhost:8181/catalog/'
    );    
)
```

When I run following command I got

```python
conn.execute("PRAGMA show_tables;").fetchall()
# []
conn.execute("SELECT * FROM information_schema.tables;").fetchall()
# []
conn.execute('SELECT * FROM catalog.ns.nested."table";').fetchall()
# *** duckdb.duckdb.ParserException: Parser Error: NameListToString NOT IMPLEMENTED

```

When I copied this table inside top namespace `ns`, I was able to query and interact with the table. I faced the same problem in the duckdb ui, so I think this is not a python specific error.

Is it planned to support nested namespaces ? Do you have a public roadmap or somewhere we can get in touch with what you're doing and what you're planning in the future ? 

Thanks ! Have a nice day

PR #358, opened 2025-07-08
[Catalog] Scan without listing, replaced with HEAD requests

Tishj:
When we're looking up a catalog entry with the purpose of finding a specific entry (table scan or a USE statement), we'll no longer hit the `v1/namespaces` endpoint to list *all* namespaces, or the `v1/namespaces/{namespace}/tables` endpoint to list *all* tables.

Instead we'll use the HEAD request to `v1/namespaces/{namespace}` and `v1/namespace/{namespace}/tables/{table}` respectively to verify the existence of the entry that's being looked up.

This is a required change for #348 because we can't afford to recursively list all namespaces

### Summary of changes

Removed dead code from the IRCatalog:
`OptionalGetCachedValue` and `SetCachedValue`

Added support for `HeadRequest` to all the `IRCAuthorization` implementations.

Removed dead code from `IRCSchemaEntry`:
`IRCAPISchema`

The creation of the `IRCSchemaEntry` is now done lazily in `IRCSchemaSet::GetEntry`, we check if it exists, if it doesn't, we perform the HEAD request to confirm its existence, then create it.
Same goes for `IcebergTableInformation` in `IRCTableSet::GetEntry`


src/storage/irc_table_set.cpp
```diff
@@ -246,22 +247,29 @@ void ICTableSet::Scan(ClientContext &context, const std::function<void(CatalogEn
 		auto &table_info = entry.second;
 		FillEntry(context, table_info);
 		auto schema_id = table_info.table_metadata.current_schema_id;
+		if (table_info.schema_versions.empty()) {
```
Tishj: This is not necessary

src/storage/irc_table_set.cpp
```diff
@@ -204,7 +204,9 @@ optional_ptr<CatalogEntry> IcebergTableInformation::CreateSchemaVersion(IcebergT
 }
 
 optional_ptr<CatalogEntry> IcebergTableInformation::GetSchemaVersion(optional_ptr<BoundAtClause> at) {
-	D_ASSERT(!schema_versions.empty());
+	if (schema_versions.empty()) {
```
Tishj: This is not necessary.

duckdb
Tmonster, line 1: I don't think this change is necessary?

src/aws.cpp
```diff
@@ -128,6 +128,55 @@ unique_ptr<HTTPResponse> AWSInput::GetRequest(ClientContext &context) {
...
+	signer->SignRequest(*request);
+
+	std::shared_ptr<Aws::Http::HttpClient> MyHttpClient;
+	MyHttpClient = Aws::Http::CreateHttpClient(*clientConfig);
+
+	LogAWSRequest(context, request);
+	std::shared_ptr<Aws::Http::HttpResponse> res = MyHttpClient->MakeRequest(request);
+	Aws::Http::HttpResponseCode resCode = res->GetResponseCode();
+	DUCKDB_LOG(context, IcebergLogType,
+	           "GET %s (response %d) (signed with key_id '%s' for service '%s', in region '%s')", uri.GetURIString(),
```
Tmonster, line 169: nit: `HEAD %s` ...

test/sql/local/irc/test_minimal_head_requests.test
```diff
@@ -54,11 +54,11 @@ select * from my_datalake.default.table_unpartitioned order by all;
 2023-03-11	11	k
 2023-03-12	12	l
 
-# 0 head requests to all files
+# 2 head requests to scan all files (1 for the schema, 1 for the table)
 query I
 select count(*) from duckdb_logs_parsed('HTTP') where request.type = 'HEAD';
```
Tmonster, line 59: Thanks for updating the comment. Could you also add a test that makes sure there are no head requests to `.parquet` or `.avro` files? That was the original idea behind "minimal" head requests

Tmonster (commented):
LGTM! 
I tested this also with non-exiting tables. It seems like we send two HEAD requests if the table doesn't exits. 
```
requesting http://127.0.0.1:8181/v1/namespaces/default
requesting http://127.0.0.1:8181/v1/namespaces/default/tables/insert_test_5
requesting http://127.0.0.1:8181/v1/namespaces/default/tables/insert_test_5
```
is what I saw with some print statements. The double request comes from Bind code in duckdb/duckdb, but I wonder if maybe there is a way to prevent the second request? 

We can also improve it in another PR, this is already a huge improvement

PR #361, opened 2025-07-09
Support nested namespaces through an ATTACH option

Tishj:
This PR implements #348 

Listing all namespaces of a catalog requires making a request for *every* namespace that's returned, at all levels.
For this reason we put this behavior behind a `support_nested_namespaces` parameter in the ATTACH statement.


Tishj:
Seems deduplication is not needed, because Spark doesn't accept a schema with a `.` in it

PR #379, opened 2025-07-21
add create table and insert testing for polaris

Tmonster:


src/storage/irc_schema_entry.cpp
```diff
@@ -39,9 +40,12 @@ optional_ptr<CatalogEntry> IRCSchemaEntry::CreateTable(IRCTransaction &irc_trans
 	tables.CreateNewEntry(context, catalog, *this, base_info);
 	auto lookup_info = EntryLookupInfo(CatalogType::TABLE_ENTRY, base_info.table);
 	auto entry = tables.GetEntry(context, lookup_info);
-	// mark table as dirty so at the end of the transaction updates/creates are commited
 	auto &ic_entry = entry->Cast<ICTableEntry>();
-	irc_transaction.MarkTableAsDirty(ic_entry);
+	// mark table as dirty so at the end of the transaction updates/creates are commited
+	// if catalog does not support stage create, then it does not need to be marked dirty
+	if (catalog.attach_options.supports_stage_create) {
+		irc_transaction.MarkTableAsDirty(ic_entry);
```
Tishj, line 49: Just a note for future reference:
We've decided to not mark an eagerly created table (if `stage_create` isn't supported by Catalog) as dirty, because it requires no action on commit/abort.

We can't drop it on abort because that isn't transactionally safe (no guarantees a different transaction didn't also tried to create it and relies on the table existing)

Tishj (approved):
Thanks, good to go from my side!

PR #380, opened 2025-07-21
Add create namespace and delete table requests

Tmonster:
This PR implements the following functionality 

- Drop table requests (purge_requested defaults to true)
- Create a namespace
- Drop a namespace
- Modify PurgeRequested based on Attach option

Also adds some error handling for statements like `CREATE OR REPLACE`.

TODO:
- handle dropping and creating of nested namespaces


src/storage/irc_catalog.cpp
```diff
@@ -66,11 +66,41 @@ optional_ptr<SchemaCatalogEntry> IRCatalog::LookupSchema(CatalogTransaction tran
...
+	auto &schemas = irc_transaction.GetSchemas();
+	auto new_schema = make_uniq<IRCSchemaEntry>(*this, info);
+	schemas.entries.insert(make_pair(new_schema->name, std::move(new_schema)));
+	auto ret = schemas.entries.find(info.schema);
+	return ret->second.get();
 }
 
 void IRCatalog::DropSchema(ClientContext &context, DropInfo &info) {
-	throw NotImplementedException("IRCatalog::DropSchema not implemented");
+	// TODO: how to handle nested namespaces
```
Tishj: Unhandled todo?
I think we just need to split on `.`

Issue #265, opened 2025-05-27
Failing to ATTACH to AWS Glue Catalog + Vanilla S3


phitoduck:
Hey! I'm doing a POC for our DS team, trying to query AWS Glue Catalog with iceberg in vanilla S3 (not s3 tables) via duckdb.

So I did

```bash
aws sso login --profile sandbox
```

I had previously created a table in my AWS Glue Catalog by inserting the NYC taxi dataset's parquet files into an Iceberg table via pyiceberg . I verified that the table shows up in AWS Glue catalog and can be successfully queried with Athena.

Then I tried connecting to my glue catalog via the ["REST Catalogs" connection type](https://duckdb.org/docs/stable/core_extensions/iceberg/iceberg_rest_catalogs):

```python
import duckdb
import boto3

# Get AWS account ID for the ATTACH statement
sts = boto3.client("sts")
account_id = sts.get_caller_identity()["Account"]

con = duckdb.connect()

# Attach to AWS Glue Iceberg REST catalog using explicit REST endpoint and SigV4 options
con.sql(f"""
INSTALL iceberg;
LOAD iceberg;
CREATE SECRET glue_iceberg_secret (
    TYPE s3,
    PROVIDER credential_chain,
    CHAIN sso,
    REGION '{AWS_REGION}',
    PROFILE '{AWS_PROFILE}'
);
ATTACH '{account_id}:nyc_taxi' AS glue_catalog (
    TYPE iceberg,
    ENDPOINT_TYPE 'GLUE',
    ENDPOINT 'https://glue/.{AWS_REGION}.amazonaws.com/iceberg',
    SECRET glue_iceberg_secret
);
""")
```

And I get this error:

```
HTTPException: HTTP Error: Failed to query https://glue.us-east-1.amazonaws.com/iceberg/v1/config?warehouse=<account id>%3Anyc_taxi, http error 403 thrown. Message: {"message":"The security token included in the request is invalid."}
```

You might ask "why don't you just use the SageMaker Lakehouse Catalog support?"

I actually did. It looked like this:

```sql
ATTACH '<account id>:s3tablescatalog/nyc_taxi' AS glue_catalog (
    TYPE iceberg,
    ENDPOINT_TYPE glue
);
```

It actually attaches! But then when I run

```sql
SHOW ALL TABLES
```

I get this:

```
[...]

Issue #372, opened 2025-07-12
Attaching Glue catalog fails when catalog contains non-Iceberg tables


mihirsamdarshi:
I have an AWS environment that uses the root Glue catalog with Iceberg and non-Iceberg tables that are divided by [Database](https://docs.aws.amazon.com/glue/latest/dg/define-database.html) in AWS Glue.

When I run 

```shell
D ATTACH '12345678901' AS glue_catalog (
    TYPE ICEBERG,
    ENDPOINT_TYPE 'GLUE'
  );
D show all tables;
Invalid Input Error:
Input table is not an iceberg table: my_non_iceberg_table_name (type=null)
```

The logs for that look like the following:

```csv
type,log_level,message
QueryLog,INFO,"ATTACH '123456789012' AS glue_catalog (
  TYPE ICEBERG,
  ENDPOINT_TYPE 'GLUE'
);"
HTTP,DEBUG,"{'request': {'type': GET, 'url': 'https://glue.us-east-2.amazonaws.com/iceberg/v1/config??warehouse=123456789012'}, 'response': NULL}"
Iceberg,INFO,"GET https://glue.us-east-2.amazonaws.com/iceberg/v1/config?warehouse=123456789012 (response 200) (signed with key_id 'AKEYID' for service 'glue', in region 'us-east-2')"
QueryLog,INFO,show all tables;
HTTP,DEBUG,"{'request': {'type': GET, 'url': 'https://glue.us-east-2.amazonaws.com/iceberg/v1/catalogs/123456789012/namespaces'}, 'response': NULL}"
Iceberg,INFO,"GET https://glue.us-east-2.amazonaws.com/iceberg/v1/catalogs%2F123456789012/namespaces (response 200) (signed with key_id 'AKEYID' for service 'glue', in region 'us-east-2')"
HTTP,DEBUG,"{'request': {'type': GET, 'url': 'https://glue.us-east-2.amazonaws.com/iceberg/v1/catalogs/123456789012/namespaces/default/tables',  'response': NULL}"
Iceberg,INFO,"GET https://glue.us-east-2.amazonaws.com/iceberg/v1/catalogs%2F123456789012/namespaces/default/tables (response 200) (signed with key_id 'AKEYID' for service 'glue', in region 'us-east-2')"
[...]

PR #381, opened 2025-07-22
Wrap LoadTableInformation response so we can ignore non-iceberg tables

Tmonster:
Potentially fixes https://github.com/duckdb/duckdb-iceberg/issues/265 and https://github.com/duckdb/duckdb-iceberg/issues/372

Wrapping the LoadTableDataResult in an IRCAPI result object so we don't need to place a ton of try_catch blocks. This wil l also allow us to immediately parse the iceberg error when we receive it instead of having to parse it in an error message body.


goober:
I can confirm that with this PR I get a more detailed error response when executing `show all tables` than I described in #372 
image

ken5scal:
@Tmonster Hi, is the PR still waiting for approval and merge from the maintainer by any chance?

ken5scal:
@Tmonster awesome. thanks for your reply

PR #383, opened 2025-07-24
[BUGFIX] Update cloud tests

Tmonster:
Fix testing infrastructure for Glue, s3tables, and the r2 catalog.

The cloud tests workflow should run Monday/wednesday/friday early in the morning


Tishj (approved):
Thanks, looks great 👍

PR #385, opened 2025-07-24
[BUGFIX] use https on r2 gets. Path style access does not mean do not use ssl

Tmonster:
if the provided endpoint starts with https:// set use_ssl to true reg…ardless of path_style_access trye

fixes https://github.com/duckdblabs/duckdb-internal/issues/5407


carlopi:
Thanks fro the fast turnaround, I think this makes sense from my side.

Issue #401, opened 2025-08-02
Support pagination in Iceberg REST Catalog list tables


sean-eyre:
The ListTables Iceberg REST Catalog API operation returns a limit of 100 tables, and a next page token if there are more in the namespace.  As far as I can tell, duckdb-iceberg does not support this.  This leads to errors like:

```
Catalog Error:
Table with name [table number 101 in namespace] does not exist!
```

To recreate:

I'm using the AWS Glue Iceberg REST API, with a namespace that has over 100 tables. (E.g. ListTables: `GET /iceberg/v1/catalogs/{catalog}/namespaces/{ns}/tables` - https://docs.aws.amazon.com/glue/latest/dg/iceberg-rest-apis.html)

Install/Load Iceberg `INSTALL iceberg; LOAD iceberg;`

Create s3 secret
```
CREATE OR REPLACE SECRET secret (
    TYPE s3,
    PROVIDER credential_chain,
    REGION 'us-east-1'
);
```

Attach catalog
```
ATTACH 'glueaccountid' AS awsdatacatalog (
      TYPE 'iceberg',
      endpoint 'glue.us-east-1.amazonaws.com/iceberg',
      AUTHORIZATION_TYPE SigV4
  );
```

Then try querying a table that is in the 2nd or later page from the ListTables API operation.  You should get the error presented above.

Suggested fix:

Implement pagination in `GetTables` in `src/catalog_api.cpp`

PR #404, opened 2025-08-05
add pagination to GetTables

sean-eyre:


Tmonster:
Hi thanks for the PR! Can you tell me again what IRC you tested against? It seems the `apache/iceberg-rest-fixture` docker image we test against doesn't support pagination (at least even for 1000 tables, I get all tables back in 1 call).

Just wondering because it would be nice to add a test

sean-eyre:
Hey no problem! - this was a quick fix and PR, I built and tried the extension locally and the issue (described in https://github.com/duckdb/duckdb-iceberg/issues/401) was no longer happening.

As for test cases, I was querying against the AWS Glue Iceberg REST Endpoint, which may behave differently (i.e. setting a 100 table limit server side).  May need to add a custom test for this?

Tmonster:
Ah yea, we have cloud tests for glue, I can probably add a test there myself and verify the logic a bit better. Will do that tomorrow

Tmonster:
tried testing with GLUE, I have a schema with ~100 tables and am hitting a REQUEST_NOT_MADE error because of a timeout. Can you let me know how many tables you have in your catalog? 
I'll have to fix this timeout issue separately.

sean-eyre:
I'm not familiar with that Glue error but in case it helps, this is how I query it in Python (not sure about C/C++):

```
import boto3
from botocore.auth import SigV4Auth
from botocore.awsrequest import AWSRequest
import requests

session = boto3.Session(
    region_name=region
)
request = AWSRequest(
    "GET",
    url, # e.g. https://glue.us-east-1.amazonaws.com/iceberg/v1/catalogs/{catalog}/namespaces/{ns}/tables/{table}
    headers={'Host': host} # e.g. glue.us-east-1.amazonaws.com
)
SigV4Auth(session.get_credentials(), service, region).add_auth(request)
response = requests.request("GET", url, headers=dict(request.headers), data={})
```

corleyma:
Lakekeeper implements pagination, with 100 table default page size as well iirc, though it is configurable.

Tmonster:
Hi @sean-eyre thanks again for the PR! Sorry it took so long to merge. It's in now, and I also added pagination for schemas 👍

PR #411, opened 2025-08-08
Support GCS OAuth tokens in IRC vended credentials

danklynn:
Apache Polaris returns an OAuth2 bearer token for GCS tables like so:

```json
{
  "metadata-location": "gs://my-bucket/warehouse/table/metadata.json",
  "config": {
    "gcs.oauth2.token": "ya29.a0AfH6SMBx..."
  }
}
```

This PR adds support for passing these bearer tokens through to GCS. This commit depends upon functionality added to the httpfs extension here:

https://github.com/danklynn/duckdb-httpfs/commit/d5a7a2b06d4b6dc5e98cd19748d1dbe3cd9c5aa6

It will compile if you comment out https loading from `out_of_tree_extensions.cmake` in the duckdb submodule, but this can be resolved once this PR is merged:

https://github.com/duckdb/duckdb-httpfs/pull/93

### DIFF
--- a/extension_config.cmake
+++ b/extension_config.cmake
@@ -33,7 +33,7 @@ endif ()
 endif()
 
 duckdb_extension_load(httpfs
-        GIT_URL https://github.com/duckdb/duckdb-httpfs
-        GIT_TAG da2821906eb42f7255d969be3e073bc1b45a71a8
-        INCLUDE_DIR extension/httpfs/include
+        GIT_URL https://github.com/danklynn/duckdb-httpfs
+        GIT_TAG d5a7a2b06d4b6dc5e98cd19748d1dbe3cd9c5aa6
 )
+
--- a/src/storage/iceberg_table_information.cpp
+++ b/src/storage/iceberg_table_information.cpp
@@ -15,8 +15,46 @@ const string &IcebergTableInformation::BaseFilePath() const {
 	return load_table_result.metadata.location;
 }
 
-static void ParseConfigOptions(const case_insensitive_map_t<string> &config, case_insensitive_map_t<Value> &options) {
-	//! Set of recognized config parameters and the duckdb secret option that matches it.
+static string DetectStorageType(const string &location) {
+	// Detect storage type from the location URL
+	if (StringUtil::StartsWith(location, "gs://") || 
+	    StringUtil::Contains(location, "storage.googleapis.com")) {
+		return "gcs";
+	} else if (StringUtil::StartsWith(location, "s3://") || 
+	           StringUtil::StartsWith(location, "s3a://")) {
+		return "s3";
+	} else if (StringUtil::StartsWith(location, "abfs://") || 
+	           StringUtil::StartsWith(location, "az://")) {
+		return "azure";
+	}
+	// Default to s3 for backward compatibility
+	return "s3";
+}
+
+static void ParseGCSConfigOptions(const case_insensitive_map_t<string> &config, case_insensitive_map_t<Value> &options) {
+	//! Set of recognized GCS config parameters and the duckdb secret option that matches it.
+	static const case_insensitive_map_t<string> gcs_config_to_option = {
+		{"gcs.oauth2.token", "bearer_token"},
+		{"gcs.bearer-token", "bearer_token"},
+		{"gcs.access-token", "bearer_token"},
+		{"gcs.endpoint", "endpoint"},
+		{"gcs.project-id", "project_id"},
+		// Map common OAuth2 patterns
+		{"oauth2.token", "bearer_token"},
+		{"bearer_token", "bearer_token"},
+		{"access_token", "bearer_token"}
+	};
+	
+	for (auto &entry : config) {
+		auto it = gcs_config_to_option.find(entry.first);
+		if (it != gcs_config_to_option.end()) {
+			options[it->second] = entry.second;
+		}
+	}
+}
+
+static void ParseS3ConfigOptions(const case_insensitive_map_t<string> &config, case_insensitive_map_t<Value> &options) {
+	//! Set of recognized S3 config parameters and the duckdb secret option that matches it.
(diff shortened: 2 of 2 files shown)


src/storage/iceberg_table_information.cpp
```diff
@@ -25,15 +63,27 @@ static void ParseConfigOptions(const case_insensitive_map_t<string> &config, cas
...
+}
+
+static void ParseConfigOptions(const case_insensitive_map_t<string> &config, case_insensitive_map_t<Value> &options, 
+                               const string &storage_type = "s3") {
+	if (config.empty()) {
+		return;
+	}
+	
+	// Parse storage-specific config options
+	if (storage_type == "gcs") {
```
Tmonster, line 67: Does GCS not have any other options like `s3.path-style-access`

src/storage/iceberg_table_information.cpp
```diff
@@ -105,19 +155,29 @@ IRCAPITableCredentials IcebergTableInformation::GetVendedCredentials(ClientConte
...
+	// Detect storage type from metadata location
+	const auto &metadata_location = load_table_result.metadata.location;
+	string storage_type = DetectStorageType(metadata_location);
+	
+	// If storage is GCS and we have a bearer token from catalog OAuth2 auth
+	if (storage_type == "gcs" && catalog.auth_handler->type == IRCAuthorizationType::OAUTH2) {
+		auto &oauth2_auth = catalog.auth_handler->Cast<OAuth2Authorization>();
+		if (!oauth2_auth.token.empty()) {
+			// Use catalog's OAuth2 token for storage access
+			user_defaults["bearer_token"] = oauth2_auth.token;
```
Tmonster: Is this the same token used for the catalog? I thought Polaris would vend the credentials. In which case they would be in `load_table_result.storage_credentials` I think

src/storage/iceberg_table_information.cpp
```diff
@@ -15,8 +15,46 @@ const string &IcebergTableInformation::BaseFilePath() const {
...
+		return "azure";
+	}
+	// Default to s3 for backward compatibility
+	return "s3";
+}
+
+static void ParseGCSConfigOptions(const case_insensitive_map_t<string> &config, case_insensitive_map_t<Value> &options) {
+	//! Set of recognized GCS config parameters and the duckdb secret option that matches it.
+	static const case_insensitive_map_t<string> gcs_config_to_option = {
+		{"gcs.oauth2.token", "bearer_token"},
```
Tmonster: Is it possible that GCS would send the token with 3 possible names? Is there any documentation on this?

Tmonster (commented):
Thanks! Just a couple of questions. 
I'll try to get the httpfs PR merged as well. Don't want the httpfs extension pointing to a non duckdb repo

src/storage/iceberg_table_information.cpp
danklynn: The GCS XML API defaults to path-style, which is how the . It does support vhost style, but [Polaris doesn't return any alternate configs](https://github.com/apache/polaris/blob/main/polaris-core/src/main/java/org/apache/polaris/core/storage/gcp/GcpCredentialsStorageIntegration.java#L110-L115) for now.

src/storage/iceberg_table_information.cpp
danklynn: Looks like older versions of Polaris vend credentials this in the `config` section only, while newer versions use the `storage-credentials` section. I'll modify this to prefer `storage-credentials` and fall back to `config`.

danklynn: Good catch, this wasn't needed

src/storage/iceberg_table_information.cpp
```diff
@@ -15,8 +15,32 @@ const string &IcebergTableInformation::BaseFilePath() const {
...
+	// Detect storage type from the location URL
+	if (StringUtil::StartsWith(location, "gs://") || 
+	    StringUtil::Contains(location, "storage.googleapis.com")) {
+		return "gcs";
+	} else if (StringUtil::StartsWith(location, "s3://") || 
+	           StringUtil::StartsWith(location, "s3a://")) {
+		return "s3";
+	} else if (StringUtil::StartsWith(location, "abfs://") || 
+	           StringUtil::StartsWith(location, "az://")) {
+		return "azure";
```
Tmonster, line 28: I think here we can just `throw NotImplementedException("Azure storage not yet supported for iceberg catalogs");`

extension_config.cmake
```diff
@@ -33,7 +33,7 @@ endif ()
 endif()
 
 duckdb_extension_load(httpfs
-        GIT_URL https://github.com/duckdb/duckdb-httpfs
```
Tmonster, line 36: You can actually remove this. The extension will still build and pass tests without this. I can update the pointer later

Tmonster (approved):
Thanks again! Managed to test it with a snowflake catalog and it looks good!
Hopefully I can find a way to test with HMAC keys as well before v1.4

extension_config.cmake
danklynn: Hmm, that probably means we need a test. Access-token-based credential vending shouldn't work for GCS without [this pr](https://github.com/duckdb/duckdb-httpfs/pull/93). I can try to find some time to put a test together (it will needs a GCS bucket and Polaris server to test properly).

Tmonster: I'll try to add a test myself with proper credentials. I created a snowflake open catalog with GCS storage to test this PR and  everything seemed to work. We first need to update the httpfs pointer on main though since extensions do not pick up the httpfs pointer they define in `extension_config.cmake` (Don't know why yet, but hopefully this will also get fixed)

Tmonster: I've uploaded a branch here https://github.com/Tmonster/duckdb-iceberg/commits/add_snowflake_gcs_test/
 if you just cherry-pick 432107e54594de4cde9c0e303b23b0296f287c49 then it will add a cloud test

PR #415, opened 2025-08-18
Add Support for positional deletes

Tmonster:
This only adds support for merge on read. Once this is merged, we can look at update statements. Once that is finished we can handle copy on write deletes. 

Works by projecting the file name and row if of all of the deleted rows and writing them to a deletes file. Then the delete file is adding to a manifest list.

### DIFF
--- a/src/iceberg_functions/iceberg_multi_file_list.cpp
+++ b/src/iceberg_functions/iceberg_multi_file_list.cpp
@@ -666,36 +666,6 @@ void IcebergMultiFileList::ProcessDeletes(const vector<MultiFileColumnDefinition
 	D_ASSERT(current_delete_manifest == delete_manifests.end());
 }
 
-vector<IcebergFileListExtendedEntry> IcebergMultiFileList::GetFilesExtended() {
-	lock_guard<mutex> l(lock);
-	InitializeFiles(l);
-
-	vector<IcebergFileListExtendedEntry> result;
-
-	for (auto &file : data_files) {
-		IcebergFileListExtendedEntry file_entry;
-		file_entry.file.path = file.file_path;
-		file_entry.file.file_size_bytes = file.file_size_in_bytes;
-		result.push_back(file_entry);
-	}
-
-	if (HasTransactionData()) {
-		auto &transaction_data = GetTransactionData();
-		for (auto &alter_p : transaction_data.alters) {
-			auto &alter = alter_p.get();
-			for (auto &file : alter.manifest_file.data_files) {
-				// if this is transaction local data, we can be positive every delete file only
-				/// references one data file.
-				IcebergFileListExtendedEntry file_entry;
-				file_entry.file.path = file.file_path;
-				file_entry.file.file_size_bytes = file.file_size_in_bytes;
-				result.push_back(file_entry);
-			}
-		}
-	}
-	return result;
-}
-
 void IcebergMultiFileList::ScanDeleteFile(const IcebergManifestEntry &entry,
                                           const vector<MultiFileColumnDefinition> &global_columns,
                                           const vector<ColumnIndex> &column_indexes) const {
--- a/src/include/storage/iceberg_delete.hpp
+++ b/src/include/storage/iceberg_delete.hpp
@@ -20,19 +20,6 @@ namespace duckdb {
 
 struct IcebergDeleteMap {
 
-	void AddExtendedFileInfo(IcebergFileListExtendedEntry file_entry) {
-		auto filename = file_entry.file.path;
-		file_map.emplace(std::move(filename), std::move(file_entry));
-	}
-
-	IcebergFileListExtendedEntry GetExtendedFileInfo(const string &filename) {
-		auto delete_entry = file_map.find(filename);
-		if (delete_entry == file_map.end()) {
-			throw InternalException("Could not find matching file for written delete file");
-		}
-		return delete_entry->second;
-	}
-
 	optional_ptr<IcebergDeleteData> GetDeleteData(const string &filename) {
 		lock_guard<mutex> guard(lock);
 		auto entry = delete_data_map.find(filename);
@@ -44,7 +31,6 @@ struct IcebergDeleteMap {
 
 private:
 	mutex lock;
-	unordered_map<string, IcebergFileListExtendedEntry> file_map;
 	unordered_map<string, shared_ptr<IcebergDeleteData>> delete_data_map;
 };
 
@@ -61,7 +47,6 @@ class IcebergDeleteLocalState : public LocalSinkState {
 public:
 	string current_file_name;
 	vector<idx_t> file_row_numbers;
-	unordered_set<string> filenames;
 };
 
 class IcebergDeleteGlobalState : public GlobalSinkState {
@@ -94,9 +79,7 @@ public:
 		Flush(local_state);
 		// flush the file names to the global state
 		lock_guard<mutex> guard(lock);
-		for (auto &entry : local_state.filenames) {
-			filenames.emplace(entry);
-		}
+		filenames.emplace(local_state.current_file_name);
 	}
 };
 
@@ -144,6 +127,9 @@ public:
 	InsertionOrderPreservingMap<string> ParamsToString() const override;
 
 private:
+	void WritePositionalDeleteFile(ClientContext &context, IcebergDeleteGlobalState &global_state,
+	                               const string &filename, IcebergDeleteFileInfo delete_file,
+	                               set<idx_t> sorted_deletes) const;
 	void FlushDelete(IRCTransaction &transaction, ClientContext &context, IcebergDeleteGlobalState &global_state,
 	                 const string &filename, vector<idx_t> &deleted_rows) const;
 };
--- a/src/include/storage/iceberg_metadata_info.hpp
+++ b/src/include/storage/iceberg_metadata_info.hpp
@@ -32,13 +32,4 @@ struct IcebergFileData {
 	optional_idx footer_size;
 };
 
-struct IcebergFileListExtendedEntry {
-	IcebergFileData file;
-	IcebergFileData delete_file;
-	optional_idx row_id_start;
-	optional_idx snapshot_id;
-	idx_t row_count;
-	idx_t delete_count = 0;
-};
-
 } // namespace duckdb
--- a/src/storage/iceberg_delete.cpp
+++ b/src/storage/iceberg_delete.cpp
@@ -66,11 +66,6 @@ SinkResultType IcebergDelete::Sink(ExecutionContext &context, DataChunk &chunk, 
 			// file has changed - flush
 			global_state.Flush(local_state);
 			local_state.current_file_name = file_name;
-			// insert the file name for the file if it has not yet been inserted
-			auto entry = local_state.filenames.find(file_name);
-			if (entry == local_state.filenames.end()) {
-				local_state.filenames.emplace(file_name);
-			}
 		}
 		auto row_number = file_row_data[row_idx];
 		local_state.file_row_numbers.push_back(row_number);
@@ -99,35 +94,10 @@ static optional_ptr<CopyFunctionCatalogEntry> TryGetCopyFunction(DatabaseInstanc
 //===--------------------------------------------------------------------===//
 // Finalize
 //===--------------------------------------------------------------------===//
-
-void IcebergDelete::FlushDelete(IRCTransaction &transaction, ClientContext &context,
-                                IcebergDeleteGlobalState &global_state, const string &filename,
-                                vector<idx_t> &deleted_rows) const {
-	// find the matching data file for the deletion
-	auto data_file_info = delete_map->GetExtendedFileInfo(filename);
-
-	// sort and duplicate eliminate the deletes
-	set<idx_t> sorted_deletes;
-	for (auto &row_idx : deleted_rows) {
-		sorted_deletes.insert(row_idx);
-	}
-	if (sorted_deletes.size() != deleted_rows.size()) {
-		throw NotImplementedException("The same row was updated multiple times - this is not (yet) supported in "
-		                              "Iceberg. Eliminate duplicate matches prior to running the UPDATE");
-	}
-
-	IcebergDeleteFileInfo delete_file;
-	delete_file.data_file_path = filename;
-	// check if the file already has deletes
-	auto existing_delete_data = delete_map->GetDeleteData(filename);
-	// FIXME: merge with existing delete data for the same data file for faster reads.
-	D_ASSERT(!existing_delete_data);
-
-	auto &fs = FileSystem::GetFileSystem(context);
-	string delete_file_uuid = UUID::ToString(UUID::GenerateRandomUUID()) + "-deletes.parquet";
-	string delete_file_path =
-	    fs.JoinPath(table.table_info.table_metadata.location, fs.JoinPath("data", delete_file_uuid));
-
+void IcebergDelete::WritePositionalDeleteFile(ClientContext &context, IcebergDeleteGlobalState &global_state,
+                                              const string &filename, IcebergDeleteFileInfo delete_file,
+                                              set<idx_t> sorted_deletes) const {
+	auto delete_file_path = delete_file.data_file_path;
 	auto info = make_uniq<CopyInfo>();
 	info->file_path = delete_file_path;
 	info->format = "parquet";
@@ -222,12 +192,40 @@ void IcebergDelete::FlushDelete(IRCTransaction &transaction, ClientContext &cont
 	if (stats_chunk.size() != 1) {
 		throw InternalException("Expected a single delete file to be written here");
 	}
-	idx_t r = 0;
-	delete_file.file_name = stats_chunk.GetValue(0, r).GetValue<string>();
-	delete_file.delete_count = stats_chunk.GetValue(1, r).GetValue<idx_t>();
-	delete_file.file_size_bytes = stats_chunk.GetValue(2, r).GetValue<idx_t>();
-	delete_file.footer_size = stats_chunk.GetValue(3, r).GetValue<idx_t>();
+	delete_file.file_name = stats_chunk.GetValue(0, 0).GetValue<string>();
+	delete_file.delete_count = stats_chunk.GetValue(1, 0).GetValue<idx_t>();
+	delete_file.file_size_bytes = stats_chunk.GetValue(2, 0).GetValue<idx_t>();
+	delete_file.footer_size = stats_chunk.GetValue(3, 0).GetValue<idx_t>();
 	global_state.written_files.emplace(filename, std::move(delete_file));
+}
+
+void IcebergDelete::FlushDelete(IRCTransaction &transaction, ClientContext &context,
+                                IcebergDeleteGlobalState &global_state, const string &filename,
+                                vector<idx_t> &deleted_rows) const {
+
+	// sort and duplicate eliminate the deletes
+	set<idx_t> sorted_deletes;
+	for (auto &row_idx : deleted_rows) {
+		sorted_deletes.insert(row_idx);
+	}
+	if (sorted_deletes.size() != deleted_rows.size()) {
+		throw NotImplementedException("The same row was updated multiple times - this is not (yet) supported in "
+		                              "Iceberg. Eliminate duplicate matches prior to running the UPDATE");
+	}
+
+	IcebergDeleteFileInfo delete_file;
+	delete_file.data_file_path = filename;
+	// check if the file already has deletes
+	auto existing_delete_data = delete_map->GetDeleteData(filename);
+	// FIXME: merge with existing delete data for the same data file for faster reads.
+	D_ASSERT(!existing_delete_data);
+
+	auto &fs = FileSystem::GetFileSystem(context);
+	string delete_file_uuid = UUID::ToString(UUID::GenerateRandomUUID()) + "-deletes.parquet";
+	string delete_file_path =
+	    fs.JoinPath(table.table_info.table_metadata.location, fs.JoinPath("data", delete_file_uuid));
+
+	WritePositionalDeleteFile(context, global_state, filename, delete_file, sorted_deletes);
 }
 
 SinkFinalizeType IcebergDelete::Finalize(Pipeline &pipeline, Event &event, ClientContext &context,
@@ -330,11 +328,6 @@ PhysicalOperator &IcebergDelete::PlanDelete(ClientContext &context, PhysicalPlan
 	if (delete_source) {
 		auto &bind_data = delete_source->bind_data->Cast<MultiFileBindData>();
 		auto &reader = bind_data.multi_file_reader->Cast<IcebergMultiFileReader>();
-		auto &file_list = bind_data.file_list->Cast<IcebergMultiFileList>();
-		auto files = file_list.GetFilesExtended();
-		for (auto &file_entry : files) {
-			delete_map->AddExtendedFileInfo(std::move(file_entry));
-		}
 		reader.delete_map = delete_map;
 	}
 	return planner.Make<IcebergDelete>(table, child_plan, std::move(delete_map), std::move(row_id_indexes));
--- a/src/include/iceberg_multi_file_list.hpp
+++ b/src/include/iceberg_multi_file_list.hpp
@@ -54,7 +54,6 @@ public:
 	unique_ptr<DeleteFilter> GetPositionalDeletesForFile(const string &file_path) const;
 	void ProcessDeletes(const vector<MultiFileColumnDefinition> &global_columns,
 	                    const vector<ColumnIndex> &column_indexes) const;
-	vector<IcebergFileListExtendedEntry> GetFilesExtended();
 
 public:
 	//! MultiFileList API


src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -645,9 +657,52 @@ void IcebergMultiFileList::ProcessDeletes(const vector<MultiFileColumnDefinition
 		}
 	}
 
+	while (current_transaction_delete_manifest != transaction_delete_manifests.end()) {
```
Tishj, line 651: This feels quite heavily copied, I wonder if we can do the same thing as we did for transaction-local data file discovery?
But it's mostly fine for now

src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -645,9 +657,52 @@ void IcebergMultiFileList::ProcessDeletes(const vector<MultiFileColumnDefinition
...
+		file_entry.file.file_size_bytes = file.file_size_in_bytes;
+		result.push_back(file_entry);
+	}
+
+	if (HasTransactionData()) {
+		auto &transaction_data = GetTransactionData();
+		for (auto &alter_p : transaction_data.alters) {
+			auto &alter = alter_p.get();
+			for (auto &file : alter.manifest_file.data_files) {
+				IcebergFileListExtendedEntry file_entry;
```
Tishj: I imagine this is copied from DuckLake?
I'd be careful about copying the positional delete logic from DuckLake because there are a couple major differences between DuckLake and Iceberg when it comes to positional deletes

(all of these are mostly relevant for < V3, deletion vectors are more sane)
Iceberg can reference multiple data files in a delete, that's why the positional deletes are sorted by filename, then position.
Iceberg can have multiple deletes for a single data file.

DuckLake will only ever reference 1 data file in a delete file.
DuckLake will always have only a single delete file per data file. (rewriting the existing delete file when additional deletes are added)

src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -268,9 +269,9 @@ void IcebergMultiFileReader::FinalizeBind(MultiFileReaderData &reader_data, cons
 		const auto &data_file = multi_file_list.data_files[file_id];
 		// The path of the data file where this chunk was read from
 		const auto &file_path = data_file.file_path;
-
 		lock_guard<mutex> delete_guard(multi_file_list.delete_lock);
-		if (multi_file_list.current_delete_manifest != multi_file_list.delete_manifests.end()) {
+		if (multi_file_list.current_delete_manifest != multi_file_list.delete_manifests.end() ||
```
Tishj, line 273: I'm not sure this is correct, the existing condition checked that deletes are already processed, hence `it != end`, which becomes false once `ProcessDeletes` has ran, whereas `multi_file_list.transaction_delete_manifests.empty()` is static, and not changed by the `ProcessDeletes` call

src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -458,4 +459,44 @@ bool IcebergMultiFileReader::ParseOption(const string &key, const Value &val, Mu
...
+		// this is computed as row_id_start + file_row_number OR read from the file
+		// first check if the row id is explicitly defined in this file
+		for (auto &col : local_columns) {
+			if (col.identifier.IsNull()) {
+				continue;
+			}
+			if (col.identifier.GetValue<int32_t>() == MultiFileReader::ROW_ID_FIELD_ID) {
+				throw InternalException("Should not be here. Iceberg doesn't support global row ids");
+				// it is! return a reference to the global row id column so we can read it from the file directly
+				global_column_reference = row_id_column.get();
```
Tishj: We've already thrown, this will never be reached

src/iceberg_manifest.cpp
```diff
@@ -35,6 +35,31 @@ Value IcebergManifestEntry::ToDataFileStruct(const LogicalType &type) const {
...
+
+	vector<Value> lower_bounds_values;
+	// lower bounds: map<126: int, 127: binary> - 104
+	for (auto &child : lower_bounds) {
+		lower_bounds_values.push_back(Value::STRUCT({{"key", child.first}, {"value", child.second}}));
+	}
+	children.push_back(Value::MAP(LogicalType::STRUCT(bounds_types), lower_bounds_values));
+
+	vector<Value> upper_bounds_values;
+	// lower bounds: map<126: int, 127: binary> - 104
```
Tishj: upper bounds*
And the field ids are also not correct in the comment

src/include/common/index.hpp
```diff
@@ -0,0 +1,138 @@
...
+	};
+	inline bool operator<(const DataFileIndex &rhs) const {
+		return index < rhs.index;
+	};
+	bool IsValid() const {
+		return index != DConstants::INVALID_INDEX;
+	}
+};
+
+struct MappingIndex {
```
Tishj: unused

src/include/common/index.hpp
```diff
@@ -0,0 +1,138 @@
...
+	bool IsValid() const {
+		return index != DConstants::INVALID_INDEX;
+	}
+	bool IsTransactionLocal() const {
+		D_ASSERT(IsValid());
+		return index >= IcebergConstants::TRANSACTION_LOCAL_ID_START;
+	}
+};
+
+struct FieldIndex {
```
Tishj: unused

src/include/common/index.hpp
```diff
@@ -0,0 +1,138 @@
...
+	bool IsValid() const {
+		return index != DConstants::INVALID_INDEX;
+	}
+	bool IsTransactionLocal() const {
+		D_ASSERT(IsValid());
+		return index >= IcebergConstants::TRANSACTION_LOCAL_ID_START;
+	}
+};
+
+struct TableIndex {
```
Tishj: unused

src/include/common/index.hpp
```diff
@@ -0,0 +1,138 @@
...
+#include "duckdb/common/common.hpp"
+#include "duckdb/common/unordered_set.hpp"
+
+namespace duckdb {
+
+struct IcebergConstants {
+	static constexpr const idx_t TRANSACTION_LOCAL_ID_START = 9223372036854775808ULL;
+};
+
+struct SchemaIndex {
```
Tishj: unused

src/storage/iceberg_delete.cpp
```diff
@@ -0,0 +1,386 @@
...
+		manifest_entry.content = IcebergManifestEntryContentType::POSITION_DELETES;
+		manifest_entry.file_path = delete_file.file_name;
+		manifest_entry.file_format = "parquet";
+		manifest_entry.record_count = delete_file.delete_count;
+		manifest_entry.file_size_in_bytes = delete_file.file_size_bytes;
+
+		// set lower and upper bound for the filename column
+		manifest_entry.lower_bounds[2147483546] = Value::BLOB(data_file_name);
+		manifest_entry.upper_bounds[2147483546] = Value::BLOB(data_file_name);
+		// set referenced_data_file
```
Tishj, line 265: Be aware that this is a V3 addition, V2 readers will ignore this

src/include/metadata/iceberg_manifest.hpp
```diff
@@ -41,6 +41,7 @@ struct IcebergManifestEntry {
 	vector<pair<int32_t, Value>> partition_values;
 	//! Inherited from the 'manifest_file' if NULL and 'status == EXISTING'
 	sequence_number_t sequence_number;
+	int64_t snapshot_id;
```
Tishj, line 44: This looks like it's set but never used?

src/iceberg_functions/iceberg_multi_file_list.cpp
Tishj: It's probably fine to introduce this behavior to our Iceberg writing logic though, but do be careful to not assume there is always 0 or 1 delete files referencing a data file, there could be multiple.

Fine to keep the behavior that rewrites the existing delete(s) into a single delete (per partition?), that's more optimized for readers

src/include/storage/iceberg_metadata_info.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+struct IcebergFileData {
+	string path;
+	idx_t file_size_bytes = 0;
+	optional_idx footer_size;
+};
+
+struct IcebergFileListExtendedEntry {
+	DataFileIndex file_id;
+	DataFileIndex delete_file_id;
```
Tishj: This looks like it assumes there's only every 1 delete per data file?

src/storage/iceberg_delete.cpp
```diff
@@ -0,0 +1,386 @@
...
+	// TODO: for V3 deletes, we need to be able to find existing delete files and merge with them
+	D_ASSERT(!existing_delete_data);
+	if (sorted_deletes.size() == data_file_info.row_count) {
+		// ALL rows in this file are deleted - we don't need to write the deletes out to a file
+		// we can just invalidate the source data file directly
+		if (delete_file.data_file_id.IsValid()) {
+			// persistent file - drop the file as part of the transaction
+			// TODO: transaction does not have a DropFile command
+			//  but the concept still stands that we can rewrite the manifest to not include a completely deleted file.
+			// transaction.DropFile(table.GetTableId(), delete_file.data_file_id, data_file_info.file.path);
```
Tishj: unresolved TODO?

src/storage/iceberg_delete.cpp
```diff
@@ -0,0 +1,386 @@
...
+		}
+		return;
+	}
+
+	auto &fs = FileSystem::GetFileSystem(context);
+	// TODO: figure out how to get a UUID filepattern here.
+	// auto delete_file_uuid = transaction.GenerateUUID() + "-delete.parquet";
```
Tishj: TODO looks to be resolved, comment should be removed

src/storage/iceberg_delete.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+	auto info = make_uniq<CopyInfo>();
+	info->file_path = delete_file_path;
+	info->format = "parquet";
+	info->is_from = false;
+
+	// generate the field ids to be written by the parquet writer
+	// these field ids follow icebergs' ids and names for the delete files
+	child_list_t<Value> values;
+	// TODO: once https://github.com/duckdb/duckdb/pull/18617 is merged, used values
```
Tishj: resolved or not?

src/storage/iceberg_delete.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+	if (stats_chunk.size() != 1) {
+		throw InternalException("Expected a single delete file to be written here");
+	}
+	idx_t r = 0;
+	delete_file.file_name = stats_chunk.GetValue(0, r).GetValue<string>();
+	delete_file.delete_count = stats_chunk.GetValue(1, r).GetValue<idx_t>();
+	delete_file.file_size_bytes = stats_chunk.GetValue(2, r).GetValue<idx_t>();
+	delete_file.footer_size = stats_chunk.GetValue(3, r).GetValue<idx_t>();
+	// TODO: add this back in
```
Tishj: resolved ?

src/storage/iceberg_delete.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+		// set lower and upper bound for the filename column
+		manifest_entry.lower_bounds[2147483546] = Value::BLOB(data_file_name);
+		manifest_entry.upper_bounds[2147483546] = Value::BLOB(data_file_name);
+		// set referenced_data_file
+		manifest_entry.referenced_data_file = data_file_name;
+
+		iceberg_delete_files.push_back(std::move(manifest_entry));
+	}
+	// TODO: add a Delete update to the transaction.
```
Tishj: resolved ?

src/storage/irc_catalog.cpp
```diff
@@ -103,13 +111,18 @@ void IRCatalog::DropSchema(ClientContext &context, DropInfo &info) {
 	vector<string> namespace_items;
 	auto namespace_identifier = IRCAPI::ParseSchemaName(info.name);
 	namespace_items.push_back(IRCAPI::GetEncodedSchemaName(namespace_identifier));
+	if (info.if_not_found == OnEntryNotFound::RETURN_NULL) {
+		auto schema_lookup = EntryLookupInfo(CatalogType::SCHEMA_ENTRY, info.name);
+		// auto &irc_transaction = CatalogTran::Get(context, *this);
```
Tishj: commented out code

src/storage/iceberg_delete.cpp
Tmonster: It was added while trying to figure out how to get spark to read our delete files. I may as well remove it

src/include/metadata/iceberg_manifest.hpp
Tmonster: It's used in `iceberg_manifest.cpp:349` `data.SetValue(col_idx++, i, Value::BIGINT(data_file.snapshot_id));`. I had to write the snapshot id, otherwise pyiceberg would not read the delete file.

src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -360,6 +361,14 @@ optional_ptr<const IcebergManifestEntry> IcebergMultiFileList::GetDataFile(idx_t
 				auto &data_files = manifest_file.data_files;
 				current_data_files.insert(current_data_files.end(), data_files.begin(), data_files.end());
 				transaction_data_idx++;
+			} else if (!transaction_delete_manifests.empty()) {
+				if (transaction_delete_idx >= transaction_delete_manifests.size()) {
+					//! Exhausted all the transaction-local data
+					return nullptr;
+				}
+				// add to the current data files? I don't think so.
+				auto &manifest_file = transaction_delete_manifests[transaction_delete_idx].get();
```
Tishj: This is a local variable, what is happening in this body that persists outside of this scope? Besides increasing `transaction_delete_idx`

src/include/metadata/iceberg_manifest.hpp
Tishj: Alright 👍 
I am a little confused by why it's necessary, because they *should* be inherited
> Manifest list files are required in v2, so that the sequence_number and snapshot_id to inherit are always available.

src/storage/irc_table_entry.cpp
```diff
@@ -147,19 +148,29 @@ TableFunction ICTableEntry::GetScanFunction(ClientContext &context, unique_ptr<F
...
+	return VirtualColumns();
+}
+
+virtual_column_map_t ICTableEntry::VirtualColumns() {
 	virtual_column_map_t result;
+	result.insert(
+	    make_pair(MultiFileReader::COLUMN_IDENTIFIER_FILENAME, TableColumn("filename", LogicalType::VARCHAR)));
 	result.insert(make_pair(MultiFileReader::COLUMN_IDENTIFIER_FILE_ROW_NUMBER,
 	                        TableColumn("file_row_number", LogicalType::BIGINT)));
+	// TODO: understand what the file index is for, I don't understand
```
Tishj: Can we remove it then?
I don't think we have tests for it either, I think I might have just copied this from DuckLake at some point

Actually, looks like we use it?
```c++
auto &file_index_vector = chunk.data[row_id_indexes[1]];
```

If that's the same one
These are the `GetRowIdColumns` that we're accessing here (in the snippet)

test/sql/local/irc/delete/delete_from_table.test
```diff
@@ -0,0 +1,62 @@
...
+    CLIENT_ID 'admin',
+    CLIENT_SECRET 'password',
+    ENDPOINT 'http://127.0.0.1:8181'
+);
+
+statement ok
+drop table if exists my_datalake.default.test_delete_table;
+
+statement ok
+create table my_datalake.default.test_delete_table as select range a from range(10);
```
Tishj: Can we spread this out over 5 different inserts?
I just want to verify that we can do a delete that touches multiple different data files in the same transaction

src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -360,6 +361,12 @@ optional_ptr<const IcebergManifestEntry> IcebergMultiFileList::GetDataFile(idx_t
 				auto &data_files = manifest_file.data_files;
 				current_data_files.insert(current_data_files.end(), data_files.begin(), data_files.end());
 				transaction_data_idx++;
+			} else if (!transaction_delete_manifests.empty()) {
```
Tishj: I feel like this is not necessary here, we're looking for the next data file, deletes are processed differently
The deletes are processed in `ProcessDeletes`, which uses `current_transaction_delete_manifest`

We can remove this block and `transaction_delete_idx` as far as I can see

src/storage/iceberg_delete.cpp
```diff
@@ -0,0 +1,358 @@
...
+}
+
+//===--------------------------------------------------------------------===//
+// Finalize
+//===--------------------------------------------------------------------===//
+void IcebergDelete::FlushDelete(IRCTransaction &transaction, ClientContext &context,
+                                IcebergDeleteGlobalState &global_state, const string &filename,
+                                vector<idx_t> &deleted_rows) const {
+	// find the matching data file for the deletion
+	auto data_file_info = delete_map->GetExtendedFileInfo(filename);
```
Tishj: `data_file_info` is unused here
This is the only place that `GetExtendedFileInfo` is used, so I think that method can also be removed

Which also means that `AddExtendedFileInfo` is not necessary ?
And by association the `file_map`

```patch
diff --git a/src/iceberg_functions/iceberg_multi_file_list.cpp b/src/iceberg_functions/iceberg_multi_file_list.cpp
index bf9fe66a..f84d84cc 100644
--- a/src/iceberg_functions/iceberg_multi_file_list.cpp
+++ b/src/iceberg_functions/iceberg_multi_file_list.cpp
@@ -666,36 +666,6 @@ void IcebergMultiFileList::ProcessDeletes(const vector<MultiFileColumnDefinition
 	D_ASSERT(current_delete_manifest == delete_manifests.end());
 }
 
-vector<IcebergFileListExtendedEntry> IcebergMultiFileList::GetFilesExtended() {
-	lock_guard<mutex> l(lock);
-	InitializeFiles(l);
-
-	vector<IcebergFileListExtendedEntry> result;
-
-	for (auto &file : data_files) {
-		IcebergFileListExtendedEntry file_entry;
-		file_entry.file.path = file.file_path;
-		file_entry.file.file_size_bytes = file.file_size_in_bytes;
-		result.push_back(file_entry);
-	}
-
-	if (HasTransactionData()) {
-		auto &transaction_data = GetTransactionData();
-		for (auto &alter_p : transaction_data.alters) {
-			auto &alter = alter_p.get();
-			for (auto &file : alter.manifest_file.data_files) {
-				// if this is transaction local data, we can be positive every delete file only
-				/// references one data file.
-				IcebergFileListExtendedEntry file_entry;
-				file_entry.file.path = file.file_path;
-				file_entry.file.file_size_bytes = file.file_size_in_bytes;
-				result.push_back(file_entry);
-			}
-		}
-	}
-	return result;
-}
-
 void IcebergMultiFileList::ScanDeleteFile(const IcebergManifestEntry &entry,
                                           const vector<MultiFileColumnDefinition> &global_columns,
                                           const vector<ColumnIndex> &column_indexes) const {
diff --git a/src/include/iceberg_multi_file_list.hpp b/src/include/iceberg_multi_file_list.hpp
index 516a28d5..fbb9754e 100644
--- a/src/include/iceberg_multi_file_list.hpp
+++ b/src/include/iceberg_multi_file_list.hpp
@@ -54,7 +54,6 @@ public:
 	unique_ptr<DeleteFilter> GetPositionalDeletesForFile(const string &file_path) const;
 	void ProcessDeletes(const vector<MultiFileColumnDefinition> &global_columns,
 	                    const vector<ColumnIndex> &column_indexes) const;
-	vector<IcebergFileListExtendedEntry> GetFilesExtended();
 
 public:
 	//! MultiFileList API
diff --git a/src/include/storage/iceberg_delete.hpp b/src/include/storage/iceberg_delete.hpp
index 978e7cf8..f467fb9c 100644
--- a/src/include/storage/iceberg_delete.hpp
+++ b/src/include/storage/iceberg_delete.hpp
@@ -19,20 +19,6 @@
 namespace duckdb {
 
 struct IcebergDeleteMap {
-
-	void AddExtendedFileInfo(IcebergFileListExtendedEntry file_entry) {
-		auto filename = file_entry.file.path;
-		file_map.emplace(std::move(filename), std::move(file_entry));
-	}
-
-	IcebergFileListExtendedEntry GetExtendedFileInfo(const string &filename) {
-		auto delete_entry = file_map.find(filename);
-		if (delete_entry == file_map.end()) {
-			throw InternalException("Could not find matching file for written delete file");
-		}
-		return delete_entry->second;
-	}
-
 	optional_ptr<IcebergDeleteData> GetDeleteData(const string &filename) {
 		lock_guard<mutex> guard(lock);
 		auto entry = delete_data_map.find(filename);
@@ -54,7 +40,6 @@ struct IcebergDeleteMap {
 
 private:
 	mutex lock;
-	unordered_map<string, IcebergFileListExtendedEntry> file_map;
 	unordered_map<string, shared_ptr<IcebergDeleteData>> delete_data_map;
 };
 
diff --git a/src/include/storage/iceberg_metadata_info.hpp b/src/include/storage/iceberg_metadata_info.hpp
index 798a1226..538bc04c 100644
--- a/src/include/storage/iceberg_metadata_info.hpp
+++ b/src/include/storage/iceberg_metadata_info.hpp
@@ -32,13 +32,4 @@ struct IcebergFileData {
 	optional_idx footer_size;
 };
 
-struct IcebergFileListExtendedEntry {
-	IcebergFileData file;
-	IcebergFileData delete_file;
-	optional_idx row_id_start;
-	optional_idx snapshot_id;
-	idx_t row_count;
-	idx_t delete_count = 0;
-};
-
 } // namespace duckdb
diff --git a/src/storage/iceberg_delete.cpp b/src/storage/iceberg_delete.cpp
index de0f2351..57ab340f 100644
--- a/src/storage/iceberg_delete.cpp
+++ b/src/storage/iceberg_delete.cpp
@@ -102,9 +102,6 @@ static optional_ptr<CopyFunctionCatalogEntry> TryGetCopyFunction(DatabaseInstanc
 void IcebergDelete::FlushDelete(IRCTransaction &transaction, ClientContext &context,
                                 IcebergDeleteGlobalState &global_state, const string &filename,
                                 vector<idx_t> &deleted_rows) const {
-	// find the matching data file for the deletion
-	auto data_file_info = delete_map->GetExtendedFileInfo(filename);
-
 	// sort and duplicate eliminate the deletes
 	set<idx_t> sorted_deletes;
 	for (auto &row_idx : deleted_rows) {
@@ -329,11 +326,6 @@ PhysicalOperator &IcebergDelete::PlanDelete(ClientContext &context, PhysicalPlan
 	if (delete_source) {
 		auto &bind_data = delete_source->bind_data->Cast<MultiFileBindData>();
 		auto &reader = bind_data.multi_file_reader->Cast<IcebergMultiFileReader>();
-		auto &file_list = bind_data.file_list->Cast<IcebergMultiFileList>();
-		auto files = file_list.GetFilesExtended();
-		for (auto &file_entry : files) {
-			delete_map->AddExtendedFileInfo(std::move(file_entry));
-		}
 		reader.delete_map = delete_map;
 	}
 	return planner.Make<IcebergDelete>(table, child_plan, std::move(delete_map), std::move(row_id_indexes));
```

src/include/storage/iceberg_delete.hpp
```diff
@@ -0,0 +1,161 @@
...
+	optional_ptr<IcebergDeleteData> GetDeleteData(const string &filename) {
+		lock_guard<mutex> guard(lock);
+		auto entry = delete_data_map.find(filename);
+		if (entry == delete_data_map.end()) {
+			return nullptr;
+		}
+		return entry->second.get();
+	}
+
+	void ClearDeletes(const string &filename) {
```
Tishj: This is unused

src/include/storage/iceberg_delete.hpp
```diff
@@ -0,0 +1,161 @@
...
+		lock_guard<mutex> guard(lock);
+		delete_data_map.erase(filename);
+	}
+
+	void AddDeleteData(const string &filename, shared_ptr<IcebergDeleteData> delete_data) {
```
Tishj: This is unused

src/include/storage/iceberg_delete.hpp
```diff
@@ -0,0 +1,161 @@
...
+#include "duckdb/planner/operator/logical_delete.hpp"
+#include "duckdb/common/multi_file/multi_file_reader.hpp"
+#include "storage/iceberg_metadata_info.hpp"
+#include "storage/iceberg_delete_filter.hpp"
+#include "storage/irc_table_entry.hpp"
+#include "storage/irc_schema_entry.hpp"
+
+namespace duckdb {
+
+struct IcebergDeleteMap {
```
Tishj, line 21: It looks like this struct can be removed

src/storage/iceberg_delete.cpp
```diff
@@ -0,0 +1,358 @@
...
+		IcebergManifestEntry manifest_entry;
+		manifest_entry.status = IcebergManifestEntryStatusType::ADDED;
+		manifest_entry.content = IcebergManifestEntryContentType::POSITION_DELETES;
+		manifest_entry.file_path = delete_file.file_name;
+		manifest_entry.file_format = "parquet";
+		manifest_entry.record_count = delete_file.delete_count;
+		manifest_entry.file_size_in_bytes = delete_file.file_size_bytes;
+
+		// set lower and upper bound for the filename column
+		manifest_entry.lower_bounds[2147483546] = Value::BLOB(data_file_name);
```
Tishj: Please use the right constants here, to avoid magic numbers

src/storage/iceberg_delete.cpp
```diff
@@ -0,0 +1,358 @@
...
+	child_list_t<Value> values;
+	values.emplace_back("file_path", Value::INTEGER(MultiFileReader::DELETE_FILE_PATH_FIELD_ID));
+	values.emplace_back("pos", Value::INTEGER(MultiFileReader::DELETE_POS_FIELD_ID));
+	auto field_ids = Value::STRUCT(std::move(values));
+	vector<Value> field_input;
+	field_input.push_back(std::move(field_ids));
+	info->options["field_ids"] = std::move(field_input);
+
+	// get the actual copy function and bind it
+	auto copy_fun = TryGetCopyFunction(*context.db, "parquet");
```
Tishj, line 117: Is it possible to branch this out into a function? `WriteIcebergPositionalDeleteFile` ?

src/storage/iceberg_delete.cpp
```diff
@@ -0,0 +1,358 @@
...
+}
+
+//===--------------------------------------------------------------------===//
+// Sink
+//===--------------------------------------------------------------------===//
+SinkResultType IcebergDelete::Sink(ExecutionContext &context, DataChunk &chunk, OperatorSinkInput &input) const {
+	auto &global_state = input.global_state.Cast<IcebergDeleteGlobalState>();
+	auto &local_state = input.local_state.Cast<IcebergDeleteLocalState>();
+
+	auto &file_name_vector = chunk.data[row_id_indexes[0]];
```
Tishj, line 47: This is being accessed with the selection vector from `file_row_number`, which aren't guaranteed to match
You should create a UnifiedVectorFormat for this Vector as well

src/storage/iceberg_delete.cpp
```diff
@@ -0,0 +1,358 @@
...
+		}
+		auto file_name_data = UnifiedVectorFormat::GetData<string_t>(file_name_vdata);
+		auto file_name = file_name_data[file_name_idx].GetString();
+
+		if (local_state.current_file_name.empty() || local_state.current_file_name != file_name) {
+			// file has changed - flush
+			global_state.Flush(local_state);
+			local_state.current_file_name = file_name;
+			// insert the file name for the file if it has not yet been inserted
+			auto entry = local_state.filenames.find(file_name);
```
Tishj: I don't understand why this is an unordered set, it looks like this can only work if the set has a size of 1?
Since we're not creating any mapping between a `filenames` entry and the `file_row_numbers` set

This `filenames` set is ignored by `Flush`, but `FinalFlush` does copy it, I'm not sure I understand what the reason for this is

src/storage/iceberg_delete.cpp
```diff
@@ -0,0 +1,358 @@
...
+	}
+
+	auto &irc_transaction = IRCTransaction::Get(context, table.catalog);
+	// write out the delete rows
+	for (auto &entry : global_state.deleted_rows) {
+		auto filename_entry = global_state.filenames.find(entry.first);
+		if (filename_entry == global_state.filenames.end()) {
+			throw InternalException("Filename not found for file index");
+		}
+		FlushDelete(irc_transaction, context, global_state, *filename_entry, entry.second);
```
Tishj: Can't `*filename_entry` be replaced with `entry.first` ?
I'm not sure I understand the reason for having `global_state.filenames`

src/storage/iceberg_delete.cpp
Tmonster: I'm not sure what you mean here, the file_row_number and file_name_vector are both converted to the UnifiedVectorFormat in lines after this, into `row_data` and `file_name_vdata`. Once those are created `file_name_vector` and `file_row_number` are no longer used

Tishj: Ah you're right, I missed that, apologies
I thought we were using `row_idx` to access into the `file_name_data`
Apologies :pray:

src/include/storage/iceberg_delete.hpp
```diff
@@ -0,0 +1,137 @@
...
+		global_entry.insert(global_entry.end(), local_entry.begin(), local_entry.end());
+		total_deleted_count += local_entry.size();
+		local_entry.clear();
+	}
+
+	void FinalFlush(IcebergDeleteLocalState &local_state) {
+		Flush(local_state);
+		// flush the file names to the global state
+		lock_guard<mutex> guard(lock);
+		filenames.emplace(local_state.current_file_name);
```
Tishj, line 82: Sorry to make yet another comment about this, but `filenames` can be removed from the global state entirely I think

Now I'm not sure the behavior is correct anymore, because we only write the *last* filename to the global state, but we do use `filenames` to verify, in `IcebergDelete::Finalize`

I'm pretty sure `filenames` is not necessary to have, the `deleted_rows` already serves this purpose

src/include/storage/iceberg_delete.hpp
Tishj:
```patch
diff --git a/src/include/iceberg_multi_file_reader.hpp b/src/include/iceberg_multi_file_reader.hpp
index 3e8406e3..42c5a54f 100644
--- a/src/include/iceberg_multi_file_reader.hpp
+++ b/src/include/iceberg_multi_file_reader.hpp
@@ -24,8 +24,6 @@
 
 namespace duckdb {
 
-struct IcebergDeleteMap;
-
 struct IcebergMultiFileReaderGlobalState : public MultiFileReaderGlobalState {
 public:
 	IcebergMultiFileReaderGlobalState(vector<LogicalType> extra_columns_p, const MultiFileList &file_list_p)
@@ -79,7 +77,6 @@ public:
 
 public:
 	shared_ptr<TableFunctionInfo> function_info;
-	shared_ptr<IcebergDeleteMap> delete_map;
 	IcebergOptions options;
 
 private:
diff --git a/src/include/storage/iceberg_delete.hpp b/src/include/storage/iceberg_delete.hpp
index 6ea07e88..35c806a8 100644
--- a/src/include/storage/iceberg_delete.hpp
+++ b/src/include/storage/iceberg_delete.hpp
@@ -18,22 +18,6 @@
 
 namespace duckdb {
 
-struct IcebergDeleteMap {
-
-	optional_ptr<IcebergDeleteData> GetDeleteData(const string &filename) {
-		lock_guard<mutex> guard(lock);
-		auto entry = delete_data_map.find(filename);
-		if (entry == delete_data_map.end()) {
-			return nullptr;
-		}
-		return entry->second.get();
-	}
-
-private:
-	mutex lock;
-	unordered_map<string, shared_ptr<IcebergDeleteData>> delete_data_map;
-};
-
 struct WrittenColumnInfo {
 	WrittenColumnInfo() = default;
 	WrittenColumnInfo(LogicalType type_p, int32_t field_id) : type(std::move(type_p)), field_id(field_id) {
@@ -85,13 +69,10 @@ public:
 
 class IcebergDelete : public PhysicalOperator {
 public:
-	IcebergDelete(PhysicalPlan &physical_plan, ICTableEntry &table, PhysicalOperator &child,
-	              shared_ptr<IcebergDeleteMap> delete_map_p, vector<idx_t> row_id_indexes);
+	IcebergDelete(PhysicalPlan &physical_plan, ICTableEntry &table, PhysicalOperator &child, vector<idx_t> row_id_indexes);
 
 	//! The table to delete from
 	ICTableEntry &table;
-	// A map of filename -> data file index and filename -> delete data
-	shared_ptr<IcebergDeleteMap> delete_map;
 	//! The column indexes for the relevant row-id columns
 	vector<idx_t> row_id_indexes;
 
diff --git a/src/storage/iceberg_delete.cpp b/src/storage/iceberg_delete.cpp
index 4a3e8295..51abca98 100644
--- a/src/storage/iceberg_delete.cpp
+++ b/src/storage/iceberg_delete.cpp
@@ -23,9 +23,9 @@ class IcebergDeleteGlobalState;
 class ICTableEntry;
 
 IcebergDelete::IcebergDelete(PhysicalPlan &physical_plan, ICTableEntry &table, PhysicalOperator &child,
-                             shared_ptr<IcebergDeleteMap> delete_map_p, vector<idx_t> row_id_indexes)
+                             vector<idx_t> row_id_indexes)
     : PhysicalOperator(physical_plan, PhysicalOperatorType::EXTENSION, {LogicalType::BIGINT}, 1), table(table),
-      delete_map(std::move(delete_map_p)), row_id_indexes(std::move(row_id_indexes)) {
+      row_id_indexes(std::move(row_id_indexes)) {
 	children.push_back(child);
 }
 
@@ -215,10 +215,6 @@ void IcebergDelete::FlushDelete(IRCTransaction &transaction, ClientContext &cont
 
 	IcebergDeleteFileInfo delete_file;
 	delete_file.data_file_path = filename;
-	// check if the file already has deletes
-	auto existing_delete_data = delete_map->GetDeleteData(filename);
-	// FIXME: merge with existing delete data for the same data file for faster reads.
-	D_ASSERT(!existing_delete_data);
 
 	auto &fs = FileSystem::GetFileSystem(context);
 	string delete_file_uuid = UUID::ToString(UUID::GenerateRandomUUID()) + "-deletes.parquet";
@@ -323,14 +319,7 @@ optional_ptr<PhysicalTableScan> FindIcebergDeleteSource(PhysicalOperator &plan)
 
 PhysicalOperator &IcebergDelete::PlanDelete(ClientContext &context, PhysicalPlanGenerator &planner, ICTableEntry &table,
                                             PhysicalOperator &child_plan, vector<idx_t> row_id_indexes) {
-	auto delete_source = FindIcebergDeleteSource(child_plan);
-	auto delete_map = make_shared_ptr<IcebergDeleteMap>();
-	if (delete_source) {
-		auto &bind_data = delete_source->bind_data->Cast<MultiFileBindData>();
-		auto &reader = bind_data.multi_file_reader->Cast<IcebergMultiFileReader>();
-		reader.delete_map = delete_map;
-	}
-	return planner.Make<IcebergDelete>(table, child_plan, std::move(delete_map), std::move(row_id_indexes));
+	return planner.Make<IcebergDelete>(table, child_plan, std::move(row_id_indexes));
 }
 
 PhysicalOperator &IRCatalog::PlanDelete(ClientContext &context, PhysicalPlanGenerator &planner, LogicalDelete &op,

```

PR #417, opened 2025-08-20
Add support for Polaris with GCS backed storage

Tmonster:
Piggy back from https://github.com/duckdb/duckdb-iceberg/pull/411


Tishj (approved):
Thanks, looks good to me 👍

PR #434, opened 2025-08-29
Properly pass down content-type AND implement POST requests for AWS interactions

carlopi:
This is, at the moment, only touching wasm clients, so should be OK-ish to merge.


Tmonster:
awesome thanks!

PR #440, opened 2025-09-02
Replace listing namespaces with HEAD requests when scanning

Tishj:
This PR is an attempt to fix performance issues caused by nested namespaces.

As the title says, it replaces the current logic of running "LoadEntries" when first hitting the schema entry set in a transaction, which does a "list namespaces" request.

The replacement being to instead do a HEAD request to check the existence of the namespace in the lookup.
This is done for both `IRCSchemaSet::GetEntry` and `IRCTableSet::GetEntry`


Tmonster (approved):
LGTM now. 
@Tishj feel free to look at my last commit, otherwise I think we can merge

PR #441, opened 2025-09-02
Support `CREATE IF NOT EXISTS`, and fix an issue causing empty snapshots to be created

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/5755

We had an assertion that WriteToFile was never called with a snapshot with no datafiles, but this was never ensured. Now it is. Because of this I've changed how the "SetSnapshotRefUpdate" update is made.

As part of testing this, I noticed `IF NOT EXISTS` was being ignored, so I fixed that and added a test for it.

### DIFF
--- a/test/sql/local/irc/test_create_if_not_exists.test
+++ b/test/sql/local/irc/test_create_if_not_exists.test
@@ -62,6 +62,14 @@ query III
 select * from my_datalake.default.a;
 ----
 
+# Now the table already exists, making the create a no-op
+statement ok
+CREATE TABLE IF NOT EXISTS my_datalake.default.a AS FROM my_datalake.default.empty_table WITH NO DATA;
+
+query III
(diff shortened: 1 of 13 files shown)


test/sql/local/irc/test_create_if_not_exists.test
```diff
@@ -0,0 +1,75 @@
...
+# Now the table already exists, making the create a no-op
+statement ok
+CREATE TABLE IF NOT EXISTS my_datalake.default.a AS FROM my_datalake.default.empty_table WITH NO DATA;
+
+query III
+select * from my_datalake.default.a;
+----
+
+# FIXME:
+# This needs a DROP 'my_datalake.default.a' to make this test rerunnable.
```
Tmonster, line 74: Pretty sure we have this functionality, but we can add it later.

Issue #444, opened 2025-09-03
Iceberg Table insert can't handle supplied columns


Tishj:
```sql
require-env ICEBERG_SERVER_AVAILABLE

require avro

require parquet

require iceberg

require httpfs

require aws

# Do not ignore 'HTTP' error messages!
set ignore_error_messages

statement ok
CREATE SECRET local_catalog_secret (
    TYPE S3,
    KEY_ID 'admin',
    SECRET 'password',
    ENDPOINT '127.0.0.1:9000',
    URL_STYLE 'path',
    USE_SSL 0
);

statement ok
ATTACH '' AS my_datalake (
    TYPE ICEBERG,
    CLIENT_ID 'admin',
    CLIENT_SECRET 'password',
    ENDPOINT 'http://127.0.0.1:8181'
);

statement ok
use my_datalake.default;

statement ok
insert into my_datalake.default.empty_table(col2) VALUES (42)
```

The last statement is the interesting bit, the rest is just boilerplate.
Note the use of `my_datalake.default.empty_table(col2)`

Tmonster:
Closing since fix has been merged

Tishj:
I wouldnt say the fix has been merged
The exception/crash is resolved, the functiinality is still missing

PR #445, opened 2025-09-03
Disable non-direct insert column orderings

Tishj:
This PR applies a temporary bandaid for #444

If a column is missing or not given in the right order, a NotImplementedException is thrown.


Tmonster (approved):
LGTM, maybe we can also get a test for this? 
I understand it requires more CI though, so maybe a test can come later

PR #446, opened 2025-09-03
Disable inserting NULL into columns marked NOT NULL

Tmonster:
Tests structs and primitive types. Have not yet figured out how create a map type with a `NOT NULL` requirement for lists and maps. I've stepped through the code though, and we should still throw an error in these cases

### DIFF
--- a/src/storage/create_table/iceberg_create_table_request.cpp
+++ b/src/storage/create_table/iceberg_create_table_request.cpp
@@ -2,6 +2,7 @@
 #include "storage/table_create/iceberg_create_table_request.hpp"
 #include "storage/irc_table_set.hpp"
 #include "storage/iceberg_table_information.hpp"
+#include "duckdb/parser/constraints/not_null_constraint.hpp"
 #include "utils/iceberg_type.hpp"
 #include "duckdb/common/enums/catalog_type.hpp"
 #include "duckdb/catalog/catalog_entry/copy_function_catalog_entry.hpp"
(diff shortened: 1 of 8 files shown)


src/storage/create_table/iceberg_create_table_request.cpp
```diff
@@ -105,9 +106,23 @@ shared_ptr<IcebergTableSchema> IcebergCreateTableRequest::CreateIcebergSchema(co
...
+		// check if there is a not null constraint
 		bool required = false;
+		if (!constraints.empty()) {
+			for (auto &constraint : constraints) {
+				if (constraint->type != ConstraintType::NOT_NULL) {
+					continue;
+				}
+				auto &not_null_constraint = constraint->Cast<NotNullConstraint>();
+				if (not_null_constraint.index.IsValid() && not_null_constraint.index.index == column.pos) {
+					required = true;
```
Tishj, line 121: Bit of a nitpick, so if there's no other reason to touch the PR, feel free to ignore:
this can use a `break` to skip the other constraints

src/storage/iceberg_insert.cpp
```diff
@@ -159,8 +179,8 @@ static void AddWrittenFiles(IcebergInsertGlobalState &global_state, DataChunk &c
 		data_file.status = IcebergManifestEntryStatusType::ADDED;
 		data_file.file_format = "parquet";
 
-		if (partition_id.IsValid()) {
-			data_file.partition_spec_id = static_cast<int32_t>(partition_id.GetIndex());
+		if (partition_id) {
```
Tishj, line 182: This if/else is redundant now

Also, I am weary of using `ic_table.table_info.table_metadata.default_spec_id` here, because that is taken from the `rest_api_objects::TableMetadata`, which has an accompanying `has_default_spec_id`, which I'm not sure is checked.

Though that bool could only be used for stage-create, when a spec_id is not assigned to the table yet ?
Not sure

Tmonster: So I'm pretty sure the default spec id is required in the table metadata response, [since the `optional` field is missing here](https://github.com/apache/iceberg/blob/c64c89fbb5390c7c796e022b939118539c7195e1/open-api/rest-catalog-open-api.yaml#L2545) 

That was also the experience I had when creating a table, the catalog would error if no spec id was set. 
We set it in `IcebergCreateTableRequest::CreateTableToJSON`

Tmonster: but it is redundant yes

PR #447, opened 2025-09-03
[Dev] Disable IcebergScan serialization

Tishj:
Throw an exception, instead of setting the method to null.

If we the method is null, the system tries to use a fallback for serialization, which fails because the `parameters` ends up empty (somehow? I couldn't find a logical explanation for that)

This takes after DuckLake, which uses the same logic.


Tmonster (approved):
Thanks!

PR #448, opened 2025-09-03
Disable drops of all types with `CASCADE` for now

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/5754

The actual functionality can be implemented later, for now I think it's important to block off all avenues that are not behaving correctly

### DIFF
--- a/src/storage/irc_catalog.cpp
+++ b/src/storage/irc_catalog.cpp
@@ -109,6 +109,9 @@ optional_ptr<CatalogEntry> IRCatalog::CreateSchema(CatalogTransaction transactio
 }
 
 void IRCatalog::DropSchema(ClientContext &context, DropInfo &info) {
+	if (info.cascade) {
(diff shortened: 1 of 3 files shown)


src/storage/irc_catalog.cpp
```diff
@@ -109,6 +109,9 @@ optional_ptr<CatalogEntry> IRCatalog::CreateSchema(CatalogTransaction transactio
 }
 
 void IRCatalog::DropSchema(ClientContext &context, DropInfo &info) {
+	if (info.cascade) {
+		throw NotImplementedException("DROP SCHEMA <schema_name> CASCADE is not supported currently");
```
Tmonster: Can we add `not support for Iceberg schemas`?  Prevents confusion if a user has multiple catalogs attached.

Tmonster (approved):
Looks good! Just a small comment about the error message.

PR #458, opened 2025-09-09
Fix create table sigv4

Tmonster:
fixes https://github.com/duckdblabs/duckdb-internal/issues/5833

Requests through aws and httplib are not consistent with what is a success and what is a failure. Using the httplib only 200 and 304_not_modified are considered "success", meanwhile a number of other 200 responses would be considered a failures (i.e 204_NoContent for head requests). I will make a separate PR to address this.

This PR just fixes the AWS requests so that anything in the range [200, 300) is a success, and anything >=400 is a failure. I'm pretty sure aws handles 300 responses internally. Also unsure how to get 100 responses.

Also fixes an issue with validating table existence. If a request fails because of a 400, we should make sure it's 404 and not 401_Unauthorized, or 402_BadRequest, or 403_Forbidden. Those are errors we should surface to the user and not return as a failed head attempt. That will return a "Table/Schema does not exist".

Only a 404 response should signify the schema/table has failed because the object doesn't exist. 

I'm working on cloud testing this again, cloud tests have been failing since there was a ton of data put on to R2 that I want to delete

### DIFF
--- a/src/aws.cpp
+++ b/src/aws.cpp
@@ -149,11 +149,14 @@ unique_ptr<HTTPResponse> AWSInput::ExecuteRequest(ClientContext &context, Aws::H
 	if (resCode == Aws::Http::HttpResponseCode::REQUEST_NOT_MADE) {
 		D_ASSERT(response->HasClientError());
 		result->reason = response->GetClientErrorMessage();
+		throw HTTPException(*result, result->reason);
+	}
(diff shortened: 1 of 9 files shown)


Tishj (approved):
Thanks, I don't have all the context on the AWS/SIGV4 side of things, but it looks like the changes align with the intention of the PR at least 👍

PR #459, opened 2025-09-09
Keep track of known fake schemas

Tmonster:
To avoid making a request to check schema existence multiple times in a transaction, we keep track of what schemas we have already sent a HEAD request to. 

Addresses https://github.com/duckdblabs/duckdb-internal/issues/5842 but doesn't fix it entirely

### DIFF
--- a/src/storage/irc_schema_set.cpp
+++ b/src/storage/irc_schema_set.cpp
@@ -11,11 +11,6 @@ namespace duckdb {
 IRCSchemaSet::IRCSchemaSet(Catalog &catalog) : catalog(catalog) {
 }
 
-// we are sure the schema does not exist, we performed the head request and got a 404 response.
-static bool SchemaDoesNotExist(case_insensitive_set_t &known_non_existent_schemas, const string &name) {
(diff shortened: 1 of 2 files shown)


src/storage/irc_schema_set.cpp
```diff
@@ -11,12 +11,26 @@ namespace duckdb {
...
 
+	auto &irc_transaction = IRCTransaction::Get(context, catalog);
+
+	auto verify_existence = irc_transaction.looked_up_entries.insert(name).second;
 	auto entry = entries.find(name);
+	if (entry != entries.end()) {
+		return entry->second.get();
+	}
+	if (!verify_existence) {
+		return nullptr;
```
Tishj: One thing wrong with my suggestion: this needs to take `if_not_found` into account, if `if_not_found != OnEntryNotFound::RETURN_NULL` this should probably throw

Tmonster: Wouldn't the first request of the transaction already throw? Ah, maybe IF_NOT_FOUND has different values in the transaction.

PR #461, opened 2025-09-10
[Dev] Fix test that was silently being skipped due to `require` problems

Tishj:
`avro` has to be required first, `iceberg` will fail if that's not the case
This is a bug in the unittester (of core, duckdb/duckdb), but we're working around it here first


Tmonster (approved):
hopefully CI passes

PR #466, opened 2025-09-11
Reduce calls to GetTableMetadata endpoint; Never search catalog for Function entries

Tmonster:
We want to avoid calling the GetTableMetata endpoint when listing tables in the iceberg catalog. In this PR we just create a dummy schema with one column of unknown type, and return that to the duckdb_columns init. Users can call `Describe {{table}}` to get the actual table information.


src/storage/irc_table_set.cpp
```diff
@@ -67,13 +67,25 @@ void ICTableSet::Scan(ClientContext &context, const std::function<void(CatalogEn
...
+		auto table_entry = make_uniq<ICTableEntry>(table_info, catalog, schema, info);
+
+		if (!table_entry->internal) {
+			table_entry->internal = schema.internal;
 		}
+		auto result = table_entry.get();
+		if (result->name.empty()) {
+			throw InternalException("ICTableSet::CreateEntry called with empty name");
+		}
+		table_info.schema_versions.emplace(0, std::move(table_entry));
```
Tishj: I'm not a huge fan of this change, also because we don't clean it up anywhere
I think we also perform conditional logic on `schema_versions.empty()` in a couple places

I think we can just create a `table_info.dummy` for this?
And do some better "lazy loading" for this, i.e only create it if it doesn't exist yet, which this doesn't do currently

src/include/storage/iceberg_table_information.hpp
```diff
@@ -45,11 +45,13 @@ struct IcebergTableInformation {
 	IRCSchemaEntry &schema;
 	string name;
 	string table_id;
-	// bool deleted;
+	bool filled = false;
```
Tishj: We don't need this, since we don't use `schema_versions` for the dummy anymore
`filled == false` is equal to `schema_versions.empty()`

PR #471, opened 2025-09-15
Bump Wasm to v1.4-andium

carlopi:
Merges `origin/main` (prepared for 1.4-andium) into `wasm` branch


carlopi:
@Tishj / @Tmonster: this influences only the wasm publication, would it be OK to merge?

PR #474, opened 2025-09-15
Add support for reading VARIANT columns

Tishj:
With the merging of https://github.com/duckdb/duckdb/pull/18609 in core, this opens up the ability to support VARIANT in Iceberg.

We add read support for VARIANT with this PR, and upgrade the Spark test dependency to 4.0 so we can generate test data.

NOTE: Bounds deserializing is not implemented in this PR, as I haven't been able to generate any test data with Spark 4.0 that has bounds data written.

### DIFF
--- a/scripts/data_generators/connections/spark_rest/__init__.py
+++ b/scripts/data_generators/connections/spark_rest/__init__.py
@@ -28,7 +28,7 @@ import sys
 import os
 
 CONNECTION_KEY = 'spark-rest'
-SPARK_RUNTIME_PATH = os.path.join(os.path.dirname(__file__), '..', '..', 'iceberg-spark-runtime-3.5_2.12-1.9.0.jar')
+SPARK_RUNTIME_PATH = os.path.join(os.path.dirname(__file__), '..', '..', 'iceberg-spark-runtime-4.0_2.13-1.10.0.jar')
 
 
 @IcebergConnection.register(CONNECTION_KEY)
@@ -39,7 +39,7 @@ class IcebergSparkRest(IcebergConnection):
 
     def get_connection(self):
         os.environ["PYSPARK_SUBMIT_ARGS"] = (
-            "--packages org.apache.iceberg:iceberg-spark-runtime-3.5_2.12:1.9.0,org.apache.iceberg:iceberg-aws-bundle:1.9.0 pyspark-shell"
+            "--packages org.apache.iceberg:iceberg-spark-runtime-4.0_2.13:1.10.0,org.apache.iceberg:iceberg-aws-bundle:1.10.0 pyspark-shell"
         )
         os.environ["AWS_REGION"] = "us-east-1"
         os.environ["AWS_ACCESS_KEY_ID"] = "admin"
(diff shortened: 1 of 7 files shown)


src/storage/iceberg_insert.cpp
```diff
@@ -180,50 +187,62 @@ void IcebergInsert::AddWrittenFiles(IcebergInsertGlobalState &global_state, Data
...
+			auto normalized_column_path = StringUtil::Join(column_names, ".");
+
+			auto it = column_info_map.find(column_names[0]);
+			if (it == column_info_map.end()) {
+				throw InternalException("Root column '%s' can not be found in the schema, but returned by RETURN_STATS",
+				                        column_names[0]);
+			}
+			auto &root_entry = it->second.get();
+			if (root_entry.type.id() == LogicalTypeId::VARIANT) {
+				//! No stats for VARIANT columns yet
```
Tishj: NOTE: We explicitly skip writing stats for VARIANT for now

This currently parses the stats 1 entry at a time (in the shape of `column_path -> map(varchar, varchar)`)
this will need to be delayed for VARIANT, we'll first need to gather all stat items belonging to the VARIANT by building a tree of the various path components and once we've seen everything - iterate through that tree.

At the leafs we can perform the logic that's now here, for a VARIANT we'll need to ~~steal~~borrow some code from the parquet extension to write VARIANT values with the unshredded binary encoding that represents the max/min of the VARIANT.

That's where the `variant_type` comes in that we now skip, that contains the internal variant layout (metadata BLOB, value BLOB, typed_value ...) which has the type information we need to make sense of the stats.

src/storage/irc_transaction.cpp
```diff
@@ -76,6 +76,10 @@ void CommitTableToJSON(yyjson_mut_doc *doc, yyjson_mut_val *root_object,
 			auto summary_json = yyjson_mut_obj_add_obj(doc, snapshot_json, "summary");
 			yyjson_mut_obj_add_strcpy(doc, summary_json, "operation", snapshot.summary.operation.c_str());
 			yyjson_mut_obj_add_uint(doc, snapshot_json, "schema-id", snapshot.schema_id);
+			yyjson_mut_obj_add_uint(doc, snapshot_json, "first-row-id", snapshot.first_row_id);
+			//! TODO: write the actual added rows here
```
Tishj: This needs to be properly handled, I imagine we can get this information from the RETURN_STATS result

src/storage/create_table/iceberg_create_table_request.cpp
```diff
@@ -171,6 +171,7 @@ string IcebergCreateTableRequest::CreateTableToJSON(std::unique_ptr<yyjson_mut_d
 	// unused, but we want to add teh objects
 	auto write_order_fields = yyjson_mut_obj_add_arr(doc, write_order, "fields");
 	auto properties = yyjson_mut_obj_add_obj(doc, root_object, "properties");
+	yyjson_mut_obj_add_strcpy(doc, properties, "format-version", "3");
```
Tishj: NOTE: this is required to create a VARIANT table, as the default `format-version` is 2
This should be handled through extra properties on the CREATE statement (https://github.com/duckdb/duckdb/pull/20431)

PR #480, opened 2025-09-16
bump to v1.4.0

c-herrewijn:
bump to v1.4.0


carlopi:
Hi @c-herrewijn, this PR is wrong, since it doesn't bump https://github.com/duckdb/duckdb-iceberg/blob/main/.github/workflows/MainDistributionPipeline.yml to use 1.4.0.

Could you check the other ones as well ?

c-herrewijn:
The PR is created by a script that bumps `1.3.2` to `1.4.0`.
If `MainDistributionPipeline.yml` targets main, it's left unchanged.
I'll discuss with Sam to what extent this can be automated, considering that different extensions are set up differently.

PR #482, opened 2025-09-16
Update `iceberg_to_ducklake` for DuckLake v0.3, add/improve tests

Tishj:
Added a test where `iceberg_to_ducklake` is used on the entire generated catalog of tests, and (pretty much) all tables are compared against each other:

```sql
query I rowsort expected_res
select * from my_datalake.default.${table_name}
----

query I rowsort expected_res
select * from my_ducklake.default.${table_name}
----
```

### DIFF
--- a/.github/workflows/MainDistributionPipeline.yml
+++ b/.github/workflows/MainDistributionPipeline.yml
@@ -17,7 +17,7 @@ jobs:
     uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@main
     with:
       extension_name: iceberg
-      duckdb_version: main
+      duckdb_version: v1.4-andium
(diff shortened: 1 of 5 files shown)


carlopi:
v1.4-andium is at the moment left behind, but that we can fix tomorrow.

Tishj:
Sure, my CI got canceled for some reason though?

Tishj:
Btw this test is **very** slow, like, several minutes on a `release` build
I'm not sure if the pipeline here (or `extension-ci-tools`) is built to treat `test_slow` differently, like core does
If it doesn't, this is going to increase CI time substantially.

Tmonster (commented):
looks good to me

Issue #158, opened 2025-04-07
Can not read Iceberg Equal Delete file


zhangjun0x01:
Now, duckdb iceberg extension can not read iceberg equal delete file, only can read position delete files

Tishj:
This should be fixed by #201 , as stated there, it would be great if the community could provide test cases, because we're having trouble finding a reliable way to generate equality deletes (spark refuses to do it for me)

PR #485, opened 2025-09-17
Fix query without equality delete fields

zhangjun0x01:
when we do a query without the equality delete fields, it will throw an exception  `Equality deletes need the relevant columns to be selected`, this pr fix this issue.

steps:
1. add the equality delete fields to projection columns when the query sql do not contain relevant columns .
2. execute `ApplyEqualityDeletes`
3. delete the equality delete fields for output data chunk for query result.


src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -359,8 +382,26 @@ void IcebergMultiFileReader::ApplyEqualityDeletes(ClientContext &context, DataCh
...
 					equalities.push_back(make_uniq<BoundConstantExpression>(Value::BOOLEAN(true)));
 				}
-			} else {
+				continue;
+			}
+			if (field_id_to_result_id.empty()) {
 				equalities.push_back(expression->Copy());
+				continue;
+			}
+			idx_t index = field_id_to_result_id.at(field_id);
```
Tishj: I don't understand what's happening here, can you explain and add a clarifying comment?
Especially the fact that there's an if, else if but no else, meaning we could have a situation where an equality delete condition is essentially skipped?

src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -399,17 +440,48 @@ void IcebergMultiFileReader::FinalizeChunk(ClientContext &context, const MultiFi
...
+		auto &col = local_columns[i];
+		column_index_to_field_id[i] = col.identifier.GetValue<int32_t>();
+	}
+
+	//! Map from field_id -> index in 'output_chunk'
+	unordered_map<idx_t, idx_t> field_id_to_result_id;
+	auto &column_indexes = reader.column_indexes;
+	auto result_id = executor.expressions.size() - column_indexes.size();
+	for (auto &column_index : column_indexes) {
+		field_id_to_result_id[column_index_to_field_id[column_index.GetPrimaryIndex()]] = result_id++;
```
Tishj: I reworked this logic because I couldn't find a good reason for iterating backwards, and the code was quite dense because of it
Can you double-check that this behaves the same as your original logic?

src/iceberg_functions/iceberg_multi_file_reader.cpp
```diff
@@ -291,34 +339,9 @@ void IcebergMultiFileReader::FinalizeBind(MultiFileReaderData &reader_data, cons
...
-						continue;
-					}
-				}
-			}
-			delete_rows.insert(delete_rows.end(), file.rows.begin(), file.rows.end());
-		}
-	}
+                                                  const vector<MultiFileColumnDefinition> &local_columns,
+                                                  const unordered_map<idx_t, idx_t> &field_id_to_result_id) {
+	auto delete_rows = multi_file_list.GetEqualityDeletesForFile(data_file);
```
Tishj: I deduplicated your iteration, with the logic that was in `ApplyEqualityDeletes`, which was more complete.
Your iteration didn't take into account the partition spec and partitioning information, which I imagine was not missing on purpose (?)

src/iceberg_functions/iceberg_multi_file_reader.cpp
Tishj: Hmm I think I understand, you're rewriting the expression, so they match with the modifications made to the output_chunk made earlier on

Can't this be done directly when constructing the expressions?
This feels rather error-prone because when additional expressions would be added, they would also need to be added here

In any case, it looks like this *is* actually exhaustive, so can we change the `else if` to just an `else` and turn the condition into an assertion instead?

zhangjun0x01: Consider the following case: there is a datafile with the following data:

```
1,'a', 2025-01-01 
2,'b', 2025-01-02
```

then we add two delete snapshots:
- Snapshot 1: delete where id = 1
- Snapshot 2: delete where name = 'b'

Both of these delete snapshots are valid for the datafile. When we execute a query:

`
// without equality delete field id and name 
select count(*) from mytable 
`


In the `IcebergMultiFileReader::InitializeReader` method, we add all equality delete fields (id and name) to the scan columns.


In the function `IcebergMultiFileReader::FinalizeChunk`, the `input_chunk` parameter will contain the additional equality delete fields (id and name). We copy these two fields to the `output_chunk`. At this point, in the `output_chunk`, the index of the field 'id' should be 1, and the index of the field 'name' should be 2. However, all the equality delete filters we obtain here have an index of 1 for both 'id' and 'name'. Therefore, we rewrite the expression here to be compatible with the above logic.


The reason for this situation is that the delete filters based on the 'id' and 'name' fields come from different snapshots (in `IcebergMultiFileList::ScanEqualityDeleteFile`). At this point, we do not know the order of these two delete snapshots, so we cannot pass in the correct index when constructing the `BoundReferenceExpression` expression.

src/iceberg_functions/iceberg_multi_file_reader.cpp
zhangjun0x01: thanks， I did forget to consider the partitioned table

src/iceberg_functions/iceberg_multi_file_reader.cpp
zhangjun0x01: thank for review， I am a beginner in C++

zhangjun0x01:
@Tishj ,By the way, how do I configure Iceberg to use relative paths for test cases in the data directory? I checked the configuration of Iceberg and it seems that there is no such configuration ,thanks

src/iceberg_functions/iceberg_multi_file_reader.cpp
Tishj: Hmm, I'm not entirely following yet
It's been a while since I implemented this logic, so excuse me while I recollect some thoughts.

In `ScanEqualityDeleteFile` we create the expressions, referencing the index in `output_chunk`, in order to make it projection-agnostic, so expressions don't have to be rewritten later.

You're adding to the `column_indexes` (now `new_column_indexes`) for all the columns that aren't selected but required by the equality deletes.

What I'm hearing is that you're losing the context of the order you're adding these in, to correctly replicate later, here:
```c++
	//! Add the extra equality delete fields to output chunk.
	int32_t diff = 0;
	if (executor.expressions.size() != output_chunk.ColumnCount()) {
		diff = executor.expressions.size() - output_chunk.ColumnCount();
		for (int32_t i = diff; i > 0; i--) {
			int32_t index = input_chunk.ColumnCount() - i;
			output_chunk.data.emplace_back(input_chunk.data[index]);
		}
	}
```

Tishj: I've made a commit to the branch to remove the need to rewrite expressions
Can you take some time to double check my logic?
I left a FIXME in there because I'm not entirely confident that piece of logic will work in all cases

It assumes a direct relation between `new_global_column_ids` and the `output_chunk` sizes, that might break if there is projection pushdown into parquet? I'm not entirely sure

Tishj: For future reference, it's about the connection between the logic in `ScanEqualityDeleteFile`, `InitializeReader` and `FinalizeChunk`, those are now using `equality_id_to_result_id` to synchronize the mapping that need to be done in those methods

zhangjun0x01:
> @zhangjun0x01 check out some of the scripts, like `scripts/persistent/partition_bool.py`, those are the scripts used to generate the persistent data.
> 
> I imagine it's because the `DATA_GENERATION_DIR` is a relative path

I test it , when we create iceberg hadoop catalog with relative path, the table will use relative path.

zhangjun0x01:
> The `mytable_partitioned` is the problem, you'll need to regenerate that table
> 
> For reference, this is what was used to generate `mytable`
> 
> ```java
> // src/main/java/com/example/IcebergEqualDeleteExample.java
> // ------------------------------------------------------
> // Place this file under src/main/java/com/example/IcebergEqualDeleteExample.java
> package com.example;
> 
> import org.apache.hadoop.conf.Configuration;
> import org.apache.iceberg.*;
> import org.apache.iceberg.io.FileAppender;
> 
> import org.apache.iceberg.CatalogUtil;
> import org.apache.iceberg.catalog.Catalog;
> import org.apache.iceberg.PartitionSpec;
> 
> import org.apache.iceberg.catalog.TableIdentifier;
> import org.apache.iceberg.data.GenericRecord;
> import org.apache.iceberg.data.Record;
> import org.apache.iceberg.types.Types;
> import org.apache.iceberg.data.parquet.GenericParquetWriter;
> import com.google.common.collect.ImmutableMap;
> import org.apache.iceberg.deletes.EqualityDeleteWriter;
> import org.apache.iceberg.io.OutputFile;
> import org.apache.iceberg.parquet.Parquet;
> 
> import java.util.*;
> import java.util.stream.Collectors;
> 
> import static org.apache.iceberg.TableProperties.FORMAT_VERSION;
> 
> public class IcebergEqualDeleteExample {
> 
>     public static void main(String[] args) {
>         Map<String, String> props = new HashMap<>();
>         props.put(CatalogProperties.WAREHOUSE_LOCATION, "data/persistent/equality_deletes/warehouse");
>         props.put(CatalogUtil.ICEBERG_CATALOG_TYPE, CatalogUtil.ICEBERG_CATALOG_TYPE_HADOOP);
> 
>         Catalog catalog = CatalogUtil.buildIcebergCatalog("hadoop_catalog", props, new Configuration());
>         TableIdentifier tableId = TableIdentifier.of("mydb", "mytable");
> 
>         Schema schema = new Schema(Types.NestedField.required(1, "id", Types.IntegerType.get()),
>                                    Types.NestedField.optional(2, "name", Types.StringType.get()));
> 
>         if (!catalog.tableExists(tableId)) {
>             catalog.createTable(tableId, schema, PartitionSpec.unpartitioned(), ImmutableMap.of(FORMAT_VERSION, "2"));
>             System.out.println("Created table mydb.mytable");
>         }
> 
>         Table table = catalog.loadTable(tableId);
> 
>         // Insert some records
>         List<Record> records_1 =
>             Arrays.asList(createRecord(table, 1, "a"), createRecord(table, 2, "b"), createRecord(table, 3, "b"));
>         insertRecords(table, records_1);
> 
>         // Delete records where name="b"
>         Map<String, Object> deleteConditions_1 = new HashMap<>();
>         deleteConditions_1.put("name", "b");
>         deleteRecordsByEquality(table, deleteConditions_1);
> 
>         // Insert some records
>         List<Record> records_2 =
>             Arrays.asList(createRecord(table, 1, "a"), createRecord(table, 2, "b"), createRecord(table, 1, "b"));
>         insertRecords(table, records_2);
> 
>         // Delete records where id=1 AND name="a"
>         Map<String, Object> deleteConditions_2 = new HashMap<>();
>         deleteConditions_2.put("id", 1);
>         deleteConditions_2.put("name", "a");
>         deleteRecordsByEquality(table, deleteConditions_2);
>     }
> 
>     private static Record createRecord(Table table, int id, String name) {
>         Record record = GenericRecord.create(table.schema());
>         record.setField("id", id);
>         record.setField("name", name);
>         return record;
>     }
> 
>     private static void insertRecords(Table table, List<Record> records) {
>         Transaction insertTxn = table.newTransaction();
>         try {
>             String dataPath = String.format("%s/data/data-%s.parquet", table.location(), UUID.randomUUID());
>             OutputFile dataOut = table.io().newOutputFile(dataPath);
> 
>             FileAppender<Record> writer = Parquet.write(dataOut)
>                                               .schema(table.schema())
>                                               .createWriterFunc(GenericParquetWriter::buildWriter)
>                                               .build();
> 
>             for (Record record : records) {
>                 writer.add(record);
>             }
> 
>             writer.close();
>             DataFile dataFile = DataFiles.builder(table.spec())
>                                     .withPath(dataPath)
>                                     .withFormat(FileFormat.PARQUET)
>                                     .withFileSizeInBytes(dataOut.toInputFile().getLength())
>                                     .withMetrics(writer.metrics())
>                                     .withSplitOffsets(writer.splitOffsets())
>                                     .build();
> 
>             insertTxn.newAppend().appendFile(dataFile).commit();
>             insertTxn.commitTransaction();
>             System.out.println("Data inserted successfully.");
>         } catch (Exception e) {
>             throw new RuntimeException("Failed to insert data", e);
>         }
>     }
> 
>     private static void deleteRecordsByEquality(Table table, Map<String, Object> fieldValues) {
>         Transaction txn = table.newTransaction();
>         try {
>             List<String> eqFields = new ArrayList<>(fieldValues.keySet());
>             Record deleteRec = GenericRecord.create(table.schema().select(eqFields));
> 
>             // Set all field values in the delete record
>             fieldValues.forEach(deleteRec::setField);
> 
>             String deletePath = String.format("%s/data/delete-%s.parquet", table.location(), UUID.randomUUID());
> 
>             OutputFile out = table.io().newOutputFile(deletePath);
> 
>             EqualityDeleteWriter<Record> writer = Parquet.writeDeletes(out)
>                                                       .forTable(table)
>                                                       .withSpec(table.spec())
>                                                       .rowSchema(table.schema().select(eqFields))
>                                                       .createWriterFunc(GenericParquetWriter::buildWriter)
>                                                       .equalityFieldIds(getFieldIds(table, eqFields))
>                                                       .buildEqualityWriter();
> 
>             writer.write(deleteRec);
>             writer.close();
>             DeleteFile df = writer.toDeleteFile();
> 
>             txn.newRowDelta().addDeletes(df).commit();
>             txn.commitTransaction();
>             System.out.println("Equality delete file written successfully.");
>         } catch (Exception e) {
>             throw new RuntimeException("Failure writing equality deletes", e);
>         }
>     }
> 
>     private static List<Integer> getFieldIds(Table table, List<String> names) {
>         return names.stream().map(n -> table.schema().findField(n).fieldId()).collect(Collectors.toList());
>     }
> }
> ```


I create  the  partitioned table  and insert data by spark , and generate equality delete files by java api use [this case ](https://github.com/duckdb/duckdb-iceberg/issues/158#issuecomment-2832871079) , and I query the  table `mytable_partitioned` by spark, it is correct, I will carefully check again what differences exist between this macOS test environment and my computer.

Issue #479, opened 2025-09-16
Attempting to dereference an optional pointer that is not set when using transactions


djouallah:
was working fine just a couple of days ago, using now the new stable release 1.4.0

```
Stack Trace:

/home/trusted-service-user/.duckdb/extensions/v1.4.0/linux_amd64/iceberg.duckdb_extension(+0xdb1386) [0x7faddf35e386]
/home/trusted-service-user/.duckdb/extensions/v1.4.0/linux_amd64/iceberg.duckdb_extension(+0xdb1434) [0x7faddf35e434]
/home/trusted-service-user/.duckdb/extensions/v1.4.0/linux_amd64/iceberg.duckdb_extension(+0xdb3491) [0x7faddf360491]
/home/trusted-service-user/.duckdb/extensions/v1.4.0/linux_amd64/iceberg.duckdb_extension(+0x1be4ff) [0x7fadde76b4ff]
/home/trusted-service-user/jupyter-env/python3.11/lib/python3.11/site-packages/_duckdb.cpython-311-x86_64-linux-gnu.so(_ZN6duckdb21PhysicalPlanGenerator12PlanInternalERNS_15LogicalOperatorE+0x58) [0x7fadf9b0c8a8]
/home/trusted-service-user/jupyter-env/python3.11/lib/python3.11/site-packages/_duckdb.cpython-311-x86_64-linux-gnu.so(_ZN6duckdb21PhysicalPlanGenerator14ResolveAndPlanENS_10unique_ptrINS_15LogicalOperatorESt14default_deleteIS2_ELb1EEE+0xc7) [0x7fadf9b14897]
/home/trusted-service-user/jupyter-env/python3.11/lib/python3.11/site-packages/_duckdb.cpython-311-x86_64-linux-gnu.so(_ZN6duckdb21PhysicalPlanGenerator4PlanENS_10unique_ptrINS_15LogicalOperatorESt14default_deleteIS2_ELb1EEE+0x2a) [0x7fadf9b14a1a]
[...]

PR #498, opened 2025-09-22
Do not allow replacing a table within a transaction

Tmonster:
fixes https://github.com/duckdb/duckdb-iceberg/issues/479

It's possible to delete and recreate a table with the same name in a transaction. There is currently no easy way to allow this in Iceberg since dropping a table is not reversible, and you cannot stage create a table with the name of a current table. 

In this case, we just throw an error and say this is not allowed.

### DIFF
--- a/src/storage/irc_table_set.cpp
+++ b/src/storage/irc_table_set.cpp
@@ -143,7 +143,7 @@ bool ICTableSet::CreateNewEntry(ClientContext &context, IRCatalog &catalog, IRCS
 		case OnCreateConflict::ALTER_ON_CONFLICT:
 			throw NotImplementedException("Alter on conflict");
 		default:
-			break;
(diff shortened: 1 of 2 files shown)


Tishj:
I actually think there *is* a way to drop and recreate a table in the same transaction, but you might not like it:

For a drop, the table could be set to an empty snapshot (essentially a drop of all data)
Then a recreate will need to change the schema (if the schema is different) and add a new snapshot for the new data.

Tmonster:
> For a drop, the table could be set to an empty snapshot (essentially a drop of all data)
Then a recreate will need to change the schema (if the schema is different) and add a new snapshot for the new data.

I thought of something like this as well, but for the bug mentioned, it feels like the scope is too big. I do think in the long term we should do something like this

Tishj (approved):
LGTM

Issue #494, opened 2025-09-22
Issue when writing to Iceberg Table using duckdb 1.4


heman026:
Hi 
I tried writing to Iceberg table using the example given in duckdb.org/2025/09/16/announcing-duckdb-140.html#iceberg-writes using the below code. I use Nessie catalog and Minio storage for the Iceberg Setup.

```
import duckdb

con = duckdb.connect(database=':memory:')
con.sql("INSTALL iceberg; LOAD iceberg;")
con.sql("""ATTACH '' AS iceberg_catalog (
TYPE iceberg, ENDPOINT 'http://10.xx.xx.xx:19120/iceberg/',
AUTHORIZATION_TYPE 'none'); 
""")
con.sql("""
CREATE SECRET (TYPE S3, KEY_ID 'minioadmin', SECRET 'minioadmin' ENDPOINT 'http://10.xx.xx.xx:9000',
URL_STYLE 'path', USE_SSL false );
""")

con.sql("""
USE iceberg_catalog.test;
ATTACH 'duckdb.db' AS duckdb_db;
CREATE TABLE duck_db.t AS SELECT a FROM range(4) t(a);
CREATE TABLE t AS SELECT * FROM duck_db.t;
FROM iceberg_catalog.test.t;
""")
```

I got the below error:

> _duckdb.InvalidInputException: Invalid Input Error: Substring not found

Please help me resolve this issue.

And also, I tried inserting into already available table in Iceberg using "INSERT INTO iceberg_catalog.default.test, SELECT * FROM i was getting the below error,

> _duckdb.TransactionException: Transaction Error: Failed to commit: Failed to commit Iceberg Transaction: Unable to connect to URL "http://10.xx.xx.xx: 9000/bucket/default/test_435lkfd-sdfs2-gs23-dsfs-sdshgagf/metadata/12jkdlsj-sfsg-3efs-b34f-jsklajf234bb-m0.avro?uploads=: Forbidden (HTTP code 403)

Tmonster:
Hi, thank you for filing an issue!

Do you have a link or a gist I can follow to replicate your setup? This will make it easier for me to track down the issue on the DuckDB side.

Tmonster:
Nevermind, I have managed to reproduce these issues 👍

Issue #450, opened 2025-09-04
REST catalog without vended-credentials


razorkam:
With every request to REST catalog there is always a `X-Iceberg-Access-Delegation=vended-credentials` header attached. Looks like it's hardcoded

The problem is I have a catalog without vended-credentials mechanism support, so I get HTTP 400 response always

Could we avoid relying on vended credentials, and use separate S3 secret instead?

Tishj:
> Optional signal to the server that the client supports delegated access
via a comma-separated list of access mechanisms.  The server may choose
to supply access via any or none of the requested mechanisms.

Can you please raise this problem on Nessie that they're not following the spec if they're failing to accept a query with `X-Iceberg-Access-Delegation=vended-credentials` attached?

They're not required to honor it, but failing outright is not ok

PR #503, opened 2025-09-24
Unify request logic and enable no vended credentials

Tmonster:
This PR tackles a couple of things

fixes https://github.com/duckdb/duckdb-iceberg/issues/494
fixes https://github.com/duckdb/duckdb-iceberg/issues/450

Nessie seems to support vending credentials, but doesn't actually return a key-id and secret, which is annoying. Instead of implementing a workaround in the attach options by adding a `storage_key_id` and `storage_secret` I decided to implement disabling of vended credentials. This lead me down a rabbit hole of realizing we do not have fine grain access to headers when making requests to an Iceberg catalog. All header/authorization logic was scattered throughout multiple api_util/aws/none.cpp files and multiple Http request functions with the names GetRequest,HeadRequest,DeleteRequest, and PostRequest. This has all now been unified into `Request` and you can pass the request type. 

In addition you can pass the request headers from catalog_api.cpp, so if you know exactly what url you are hitting, you can also selectively choose what http headers to include in the request. With this change, we no longer pass a Vended-Credentials header for every single request, now we only pass it for GetRequests.

This will also make it easier to implement PUT requests if we ever need to.

To get Nessie working locally now, you can attach without vended credentials, and create your own storage secret.

See the test file or here below
```
CREATE SECRET storage_secret (
    TYPE S3,
    KEY_ID 'minioadmin',
    SECRET 'minioadmin',
    ENDPOINT '127.0.0.1:9002',
    URL_STYLE 'path',
    USE_SSL 0
);
create secret my_secret (
    TYPE ICEBERG,
    CLIENT_ID 'client1',
    CLIENT_SECRET 's3cr3t',
    OAUTH2_SCOPE 'catalog sign',
    OAUTH2_SERVER_URI 'http://127.0.0.1:8080/realms/iceberg/protocol/openid-connect/token'
);
attach '' as my_datalake (
    type ICEBERG,
    ENDPOINT 'http://127.0.0.1:19120/iceberg/',
    ACCESS_DELEGATION_MODE 'none',
    SECRET 'my_secret'
);
```

### DIFF
--- a/src/catalog_api.cpp
+++ b/src/catalog_api.cpp
@@ -76,7 +76,7 @@ bool IRCAPI::VerifySchemaExistence(ClientContext &context, IRCatalog &catalog, c
 	url_builder.AddPathComponent(schema_name);
 
 	HTTPHeaders headers(*context.db);
-	auto response = catalog.auth_handler->Request(HTTPRequestType::HTTP_HEAD, context, url_builder, headers);
+	auto response = catalog.auth_handler->Request(RequestType::HEAD_REQUEST, context, url_builder, headers);
 	// the httputil currently only sets 200 and 304 response to success
 	// for AWS all responses < 400 are successful
 	if (response->Success() || response->status == HTTPStatusCode::NoContent_204) {
@@ -121,7 +121,7 @@ bool IRCAPI::VerifyTableExistence(ClientContext &context, IRCatalog &catalog, co
 
 	HTTPHeaders headers(*context.db);
 	string body = "";
-	auto response = catalog.auth_handler->Request(HTTPRequestType::HTTP_HEAD, context, url_builder, headers, body);
+	auto response = catalog.auth_handler->Request(RequestType::HEAD_REQUEST, context, url_builder, headers, body);
 	// the httputil currently only sets 200 and 304 response to success
 	// for AWS all responses < 400 are successful
 	if (response->Success() || response->status == HTTPStatusCode::NoContent_204) {
@@ -149,7 +149,7 @@ static unique_ptr<HTTPResponse> GetTableMetadata(ClientContext &context, IRCatal
 	if (catalog.attach_options.access_mode == IRCAccessDelegationMode::VENDED_CREDENTIALS) {
 		headers.Insert("X-Iceberg-Access-Delegation", "vended-credentials");
 	}
-	return catalog.auth_handler->Request(HTTPRequestType::HTTP_GET, context, url_builder, headers);
+	return catalog.auth_handler->Request(RequestType::GET_REQUEST, context, url_builder, headers);
 }
 
 APIResult<rest_api_objects::LoadTableResult> IRCAPI::GetTable(ClientContext &context, IRCatalog &catalog,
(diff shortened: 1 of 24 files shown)


src/include/storage/irc_authorization.hpp
```diff
@@ -11,6 +11,8 @@ enum class IcebergEndpointType : uint8_t { AWS_S3TABLES, AWS_GLUE, INVALID };
 
 enum class IRCAuthorizationType : uint8_t { OAUTH2, SIGV4, NONE, INVALID };
 
+enum class IRCAccessDelegationMode : uint8_t { NONE, VENDED_CREDENTIALS };
```
Tishj, line 14: I think we want to introduce a `IRCAccessDelegationMode::DEFAULT` setting, and solidify this once we have the response from `v1/config`, which I believe tells us whether the catalog supports vended credentials.

When they do, `DEFAULT` becomes `VENDED_CREDENTIALS`, when they don't, it becomes `NONE`
This way we reduce the need to reach for ATTACH options, only when you're going against the grain of the Catalog (setting to NONE when the catalog *does* accept vended credentials)

We also want to verify whether the preference can be honored after we have the `v1/config` response

Tmonster: So I checked with our cloud tests on s3tables/glue/snowflake and r2 and none of them return any kind of indication for what access_delegation_mode they support. I think maybe we keep vended credentials as a default (to stay with current functionality) and then users can specify `none` if they want. 

Eventually `remote_signing` will be an option, and it seems like this can be detected because there is a remote-signing url in the config response

Tishj: > data-access:
      name: X-Iceberg-Access-Delegation
      in: header
      description: >
        Optional signal to the server that the client supports delegated access
        via a comma-separated list of access mechanisms.  The server may choose
        to supply access via any or none of the requested mechanisms.

When I read this, there shouldn't even be a need to omit this, the catalog is free to not honor the request..
It's merely saying that the client (duckdb-iceberg) supports it..

src/catalog_api.cpp
```diff
@@ -172,8 +174,11 @@ static unique_ptr<HTTPResponse> GetTableMetadata(ClientContext &context, IRCatal
 	url_builder.AddPathComponent("tables");
 	url_builder.AddPathComponent(table);
 
-	auto url = url_builder.GetURL();
-	return catalog.auth_handler->GetRequest(context, url_builder);
+	HTTPHeaders headers(*context.db);
+	if (catalog.attach_options.access_mode == IRCAccessDelegationMode::VENDED_CREDENTIALS) {
+		headers.Insert("X-Iceberg-Access-Delegation", "vended-credentials");
```
Tishj, line 179: I still think this really shouldn't be required to be omitted, given the spec:
> Optional signal to the server that the client supports delegated access
via a comma-separated list of access mechanisms.  The server may choose
to supply access via any or none of the requested mechanisms.

But fine, the real world is different from the spec

Tishj (approved):
Thanks, LGTM

PR #506, opened 2025-09-28
Add prefix for rest catalog

zhangjun0x01:
In our business, we provide a unified REST catalog service, different business lines use  catalog `prefix` to distinguish themselves.

so I want to add 'prefix' option for attach options, after  add  prefix , we use the following sql to attach iceberg
```
 ATTACH '' AS iceberg (TYPE ICEBERG, prefix 'xxx');
```
if the iceberg rest catalog do not support `prefix`, we use the attach as before

```
 ATTACH '' AS iceberg (TYPE ICEBERG);
```

### DIFF
--- a/src/include/storage/irc_authorization.hpp
+++ b/src/include/storage/irc_authorization.hpp
@@ -15,6 +15,8 @@ struct IcebergAttachOptions {
 	string endpoint;
 	string warehouse;
 	string secret;
+	// optional prefix, if the catalog do not support prefix, we do not need to add this option.
+	string prefix;
 	string name;
 	// some catalogs do not yet support stage create
 	bool supports_stage_create = true;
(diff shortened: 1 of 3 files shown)


carlopi:
Can you add an example of what's the intended usage / behaviour?
In the PR explanation and ideally also in a test?

zhangjun0x01:
> Can you add an example of what's the intended usage / behaviour? In the PR explanation and ideally also in a test?

ok, maybe we need to start other rest catalog server  with prefix for test case

zhangjun0x01:
I find in [iceberg test case](https://github.com/apache/iceberg/blob/5ea78e3fbe5d8c9c846a1b86bcd2b77d13a31acd/core/src/test/java/org/apache/iceberg/rest/TestResourcePaths.java#L30)  , there are indeed cases where the prefix is specified manually.  so we may have multiple ways to add the prefix.

zhangjun0x01:
It seems that the official Iceberg documentation does not clearly specify the usage guidelines for this `prefix`

PR #511, opened 2025-10-02
Add nullable=false to data_file struct in manifest lists.

Tmonster:
This wil fail tests. Once the [avro PR merges](https://github.com/duckdb/duckdb-avro/pull/58), I can update this PR to point to that and tests should pass. 
Opening this to validate that this is an issue with later versions of spark


src/iceberg_manifest.cpp
```diff
@@ -226,6 +226,7 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
 		names.push_back("data_file");
 		types.push_back(LogicalType::STRUCT(std::move(children)));
 		data_file_field_ids.emplace_back("__duckdb_field_id", Value::INTEGER(DATA_FILE));
+		// data_file_field_ids.emplace_back("__duckdb_nullable", Value::BOOLEAN(false));
```
Tishj: Don't you need to change `extension_config.cmake` to point to https://github.com/duckdb/duckdb-avro/pull/58 to test this?

Tmonster: yea, this is just a test run to make sure he CI fails with the new iceberg runtime config

Tmonster: have just updated python tests to test for the error we have been seeing. Hoping this fails in CI, then can update

PR #521, opened 2025-10-07
Bump nightly extension so it is built for v1.4.1

Tmonster:
Duckdb pointer is already at v1.4.1 hash


carlopi (approved):
Thanks!

Arguably we could also just merge, given CI will in any case be run when actual building this. @samansmink ?

samansmink:
inshallah

PR #527, opened 2025-10-11
Basic support for IRC Azure vended credentials

smaheshwar-pltr:
Addresses https://github.com/duckdb/duckdb-iceberg/issues/528.

Initial support for Azure vended credentials. 

## Notes

- Writes not supported due to the Azure extension's lack of support (see https://github.com/duckdb/duckdb-azure/issues/50). Example error:

```shell
D INSERT INTO c.s.duckdbtable VALUES ('jeans', 20.0, 1), ('hammer', 42.2, 2);
Not implemented Error:
AzureDfsStorageFileSystem: DirectoryExists is not implemented!
```

- Users are prompted to load / install the `azure` extension that this PR requires to work. (See https://github.com/duckdb/duckdb-iceberg/pull/527#discussion_r2423120471).

- REST catalogs often return one vended SAS token in their `LoadTableResponse` config - this PR then extracts the account name and SAS token from the config and configures the `connection_string` property on the Azure secret using this.

### DIFF
--- a/.github/workflows/LocalTesting.yml
+++ b/.github/workflows/LocalTesting.yml
@@ -55,10 +55,6 @@ jobs:
           sudo curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
           sudo chmod +x /usr/local/bin/docker-compose
 
-      - name: Install required python packages
-        run: |
-          python3 -m pip install pyiceberg pytest pyarrow
-
       - name: Install CMake 3.x
         run: |
           sudo apt-get remove -y cmake cmake-data
@@ -68,6 +64,14 @@ jobs:
         with:
           fetch-depth: 0
           submodules: 'true'
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.13' 
+
+      - name: Install required python packages
+        run: |
+          python3 -m pip install -r scripts/requirements.txt 
 
       - name: Check installed versions
         run: |
--- a/.github/workflows/MainDistributionPipeline.yml
+++ b/.github/workflows/MainDistributionPipeline.yml
@@ -17,8 +17,8 @@ jobs:
     uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@main
     with:
       extension_name: iceberg
-      duckdb_version: v1.4.0
-      ci_tools_version: main
+      duckdb_version: v1.4.1
+      ci_tools_version: v1.4.1
       exclude_archs: 'windows_amd64_mingw;wasm_mvp;wasm_eh;wasm_threads'
       extra_toolchains: 'python3'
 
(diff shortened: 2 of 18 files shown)


src/storage/iceberg_table_information.cpp
```diff
@@ -37,6 +38,30 @@ static void ParseGCSConfigOptions(const case_insensitive_map_t<string> &config,
...
+	for (const auto &entry : config) {
+		// SAS token config format is e.g. {adls.sas-token.<account-name>.dfs.core.windows.net, <token>}
+		if (StringUtil::StartsWith(entry.first, ADLS_SAS_TOKEN_PREFIX)) {
+			string token_key = entry.first.substr(ADLS_SAS_TOKEN_PREFIX.length());
+
+			// Extract account name
+			auto dot_pos = StringUtil::Find(token_key, ".");
+			string account_name = dot_pos.IsValid() ? token_key.substr(0, dot_pos.GetIndex()) : token_key;
+
+			if (!account_name.empty() && !entry.second.empty()) {
```
smaheshwar-pltr, line 53: The value emptiness check here mirrors https://github.com/apache/iceberg/blob/77bc1d08ecd46b1d62a3fb51ad7d1935788f30be/azure/src/main/java/org/apache/iceberg/azure/AzureProperties.java#L160

src/storage/iceberg_table_information.cpp
```diff
@@ -64,6 +89,8 @@ static void ParseConfigOptions(const case_insensitive_map_t<string> &config, cas
 	// Parse storage-specific config options
 	if (storage_type == "gcs") {
 		ParseGCSConfigOptions(config, options);
+	} else if (storage_type == "azure") {
+		ParseAzureConfigOptions(config, options);
```
smaheshwar-pltr, line 93: Pointing out - if a user tries to read an Azure table without having installed `azure`, they receive:

```
Invalid Input Error:
Secret provider 'config' for type 'azure' does not exist, but it exists in the azure extension.
Please try installing and loading the azure extension by running:
INSTALL azure;
LOAD azure;
```

which feels OK to me?

src/storage/iceberg_table_information.cpp
```diff
@@ -37,6 +38,31 @@ static void ParseGCSConfigOptions(const case_insensitive_map_t<string> &config,
...
+			// Extract account name
+			auto dot_pos = StringUtil::Find(token_key, ".");
+			string account_name = dot_pos.IsValid() ? token_key.substr(0, dot_pos.GetIndex()) : token_key;
+
+			if (!account_name.empty() && !entry.second.empty()) {
+				options["account_name"] = account_name;
+				options["connection_string"] =
+				    StringUtil::Format("AccountName=%s;SharedAccessSignature=%s", account_name, entry.second);
+
+				// For now, only process the first {storage account, token} pair we find in the config
```
smaheshwar-pltr, line 58: Curious for thoughts here - handle the single-token-in-config case feels simplest initially and I think would largely unblock Azure.

The multiple `storage_credentials` case is handled by setting each secret's scope to the prefix of the credential - that's fine because the IRC returns us the prefix.

What we don't handle is when a single config has multiple credentials - I think that it's hard to find the appropriate scope for each storage account because we don't know the exact prefix given just the host we receive; we'd need to also know the file system / container name in some cases (https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction-abfs-uri#uri-syntax) to get an absolute scope.

(Playing around with `which_secret` locally makes me think wiring a given storage account to a given secret with Azure is hard, I think regex scope is not supported)

smaheshwar-pltr: For context, the way that the Java client works is: `newInputFile(String path)` for example configures the `DataLakeFileClient` for that path - the account name is _parsed_ from the path and the token for *that account* is used to configure the client: https://github.com/apache/iceberg/blob/77bc1d08ecd46b1d62a3fb51ad7d1935788f30be/azure/src/main/java/org/apache/iceberg/azure/AzureProperties.java#L159.

smaheshwar-pltr: (IMHO IRCs wishing to support multi-location tables should probably prefer the `storage-credentials` approach which is why I'm happy with this PR's approach for now)

PR #529, opened 2025-10-11
Initial snapshot summary metrics support

smaheshwar-pltr:
Addresses #530.

Iceberg writers tend to include metrics as part of snapshot summaries. This PR begins snapshot metrics support by the DuckDB writer.

Concretely, this PR supports

- added-data-files
- added-records
- deleted-data-files
- deleted-records
- total-data-files
- total-records

for writes.

With this PR, DuckDB-written snapshots have: e.g.

```json
"summary" : {
      "operation" : "append",
      "added-data-files" : "1",
      "deleted-records" : "0",
      "deleted-data-files" : "0",
      "total-data-files" : "1",
      "added-records" : "2",
      "total-records" : "2"
    },
```

Maintainer note:
This PR was made before UPDATE support was added, I've added a FIXME to `AddUpdateSnapshot` for the metrics there, and we are not writing metrics for UPDATE statements currently.

### DIFF
--- a/src/storage/iceberg_transaction_data.cpp
+++ b/src/storage/iceberg_transaction_data.cpp
@@ -45,8 +45,9 @@ static IcebergSnapshot::metrics_map_t GetSnapshotMetrics(const IcebergManifestFi
 	if (previous_total_files != previous_metrics.end()) {
 		int64_t total_files =
 		    previous_total_files->second + manifest_file.added_files_count - manifest_file.deleted_files_count;
-		if (total_files >= 0)
+		if (total_files >= 0) {
 			metrics[SnapshotMetricType::TOTAL_DATA_FILES] = total_files;
+		}
 	}
 
 	auto previous_total_records = previous_metrics.find(SnapshotMetricType::TOTAL_RECORDS);


src/metadata/iceberg_snapshot.cpp
```diff
@@ -18,6 +18,40 @@ static string OperationTypeToString(IcebergSnapshotOperationType type) {
 	}
 }
 
+static const std::unordered_map<SnapshotMetricType, string> kSnapshotMetricKeys = {
```
smaheshwar-pltr: See spec or https://github.com/apache/iceberg/blob/06ebf07e671c3ee2b0d0add1c4dc849ab9cbf630/core/src/main/java/org/apache/iceberg/SnapshotSummary.java#L31 for constants

src/metadata/iceberg_snapshot.cpp
```diff
@@ -18,6 +18,40 @@ static string OperationTypeToString(IcebergSnapshotOperationType type) {
...
+static IcebergSnapshot::metrics_map_t MetricsFromSummary(const case_insensitive_map_t<string> &snapshot_summary) {
+	IcebergSnapshot::metrics_map_t metrics;
+	for (auto &entry : kSnapshotMetricKeys) {
+		auto it = snapshot_summary.find(entry.second);
+		if (it != snapshot_summary.end()) {
+			int64_t value;
+			try {
+				value = std::stoll(it->second);
+			} catch (...) {
+				// Skip invalid metrics
```
smaheshwar-pltr, line 61: Somewhat matches https://github.com/apache/iceberg/blob/06ebf07e671c3ee2b0d0add1c4dc849ab9cbf630/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L818-L820, except we discard invalid metrics at earlier stage

src/metadata/iceberg_snapshot.cpp
```diff
@@ -18,6 +18,40 @@ static string OperationTypeToString(IcebergSnapshotOperationType type) {
...
+}
+
+static IcebergSnapshot::metrics_map_t MetricsFromSummary(const case_insensitive_map_t<string> &snapshot_summary) {
+	IcebergSnapshot::metrics_map_t metrics;
+	for (auto &entry : kSnapshotMetricKeys) {
+		auto it = snapshot_summary.find(entry.second);
+		if (it != snapshot_summary.end()) {
+			int64_t value;
+			try {
+				value = std::stoll(it->second);
```
smaheshwar-pltr, line 59: Java-side does `ParseLong` - https://github.com/apache/iceberg/blob/06ebf07e671c3ee2b0d0add1c4dc849ab9cbf630/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L802

src/storage/iceberg_transaction_data.cpp
```diff
@@ -17,6 +17,35 @@ static int64_t NewSnapshotId() {
 	return random_number;
 }
 
+static const IcebergSnapshot::metrics_map_t kMetricsWithEmptyTotals {{SnapshotMetricType::TOTAL_DATA_FILES, 0},
+                                                                     {SnapshotMetricType::TOTAL_RECORDS, 0}};
+
+static IcebergSnapshot::metrics_map_t GetSnapshotMetrics(const IcebergManifest &manifest,
```
smaheshwar-pltr: Because DuckDB fast appends a manifest for new snapshot, this implementation is simpler than the Java side

src/storage/iceberg_transaction_data.cpp
```diff
@@ -17,6 +17,35 @@ static int64_t NewSnapshotId() {
...
+                                                         const IcebergSnapshot::metrics_map_t &previous_metrics) {
+	IcebergSnapshot::metrics_map_t metrics {{SnapshotMetricType::ADDED_DATA_FILES, manifest.added_files_count},
```
smaheshwar-pltr: See https://github.com/apache/iceberg/blob/06ebf07e671c3ee2b0d0add1c4dc849ab9cbf630/core/src/main/java/org/apache/iceberg/SnapshotSummary.java#L350-L353

src/storage/iceberg_transaction_data.cpp
```diff
@@ -17,6 +17,35 @@ static int64_t NewSnapshotId() {
...
+	                                        {SnapshotMetricType::ADDED_RECORDS, manifest.added_rows_count},
+	                                        {SnapshotMetricType::DELETED_DATA_FILES, manifest.deleted_files_count},
+	                                        {SnapshotMetricType::DELETED_RECORDS, manifest.deleted_rows_count}};
+
+	auto previous_total_files = previous_metrics.find(SnapshotMetricType::TOTAL_DATA_FILES);
```
smaheshwar-pltr, line 44: See https://github.com/apache/iceberg/blob/06ebf07e671c3ee2b0d0add1c4dc849ab9cbf630/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L792 for this combining process with previous snapshot summary (again simpler for us since we can just use the manifest)

src/storage/iceberg_transaction_data.cpp
```diff
@@ -43,6 +72,32 @@ void IcebergTransactionData::AddSnapshot(IcebergSnapshotOperationType operation,
 	auto manifest_list_path = table_info.BaseFilePath() + "/metadata/snap-" + std::to_string(snapshot_id) + "-" +
 	                          manifest_list_uuid + ".avro";
 
+	//! Construct the manifest
```
smaheshwar-pltr: Rearranging - wanted to construct manifest fully first so it can be then used to construct the snapshot (the snapshot now depends on it for metrics). Just thought this read nicer now

src/storage/iceberg_transaction_data.cpp
```diff
@@ -57,43 +112,24 @@ void IcebergTransactionData::AddSnapshot(IcebergSnapshotOperationType operation,
...
 			new_snapshot.parent_snapshot_id = last_alter.snapshot.snapshot_id;
+			new_snapshot.metrics = GetSnapshotMetrics(manifest, last_alter.snapshot.metrics);
 		} else {
 			D_ASSERT(table_info.table_metadata.has_current_snapshot);
 			new_snapshot.parent_snapshot_id = table_info.table_metadata.current_snapshot_id;
+			new_snapshot.metrics = GetSnapshotMetrics(
+			    manifest, table_info.table_metadata.GetSnapshotById(new_snapshot.parent_snapshot_id)->metrics);
 		}
+	} else {
+		// If there was no previous snapshot, default the metrics to start totals at 0
```
smaheshwar-pltr: https://github.com/apache/iceberg/blob/06ebf07e671c3ee2b0d0add1c4dc849ab9cbf630/core/src/main/java/org/apache/iceberg/SnapshotProducer.java#L350

src/storage/iceberg_transaction.cpp
```diff
@@ -77,6 +77,9 @@ void CommitTableToJSON(yyjson_mut_doc *doc, yyjson_mut_val *root_object,
 			yyjson_mut_obj_add_strcpy(doc, snapshot_json, "manifest-list", snapshot.manifest_list.c_str());
 			auto summary_json = yyjson_mut_obj_add_obj(doc, snapshot_json, "summary");
 			yyjson_mut_obj_add_strcpy(doc, summary_json, "operation", snapshot.summary.operation.c_str());
+			for (auto &prop : snapshot.summary.additional_properties) {
```
smaheshwar-pltr, line 84: This is similar to what's done to serialize `set_properties_update`

src/include/metadata/iceberg_snapshot.hpp
```diff
@@ -9,13 +9,23 @@ struct IcebergTableMetadata;
...
+};
+
 //! An Iceberg snapshot https://iceberg.apache.org/spec/#snapshots
 class IcebergSnapshot {
 public:
 	IcebergSnapshot() {
 	}
 	static IcebergSnapshot ParseSnapshot(rest_api_objects::Snapshot &snapshot, IcebergTableMetadata &metadata);
 	rest_api_objects::Snapshot ToRESTObject() const;
+	using metrics_map_t = std::unordered_map<SnapshotMetricType, int64_t>;
```
Tishj: I believe this needs to be `uint8_t`  for the key, as the enum class isn't hashable
for reference: <https://github.com/duckdb/duckdb/pull/19046>

Actually, the fix is different (?)
That PR seems to suggest it should be a `map<SnapshotMetricType, ...>` instead

src/storage/iceberg_transaction.cpp
Tishj: Keep in mind that the `key` (`prop.first.c_str()`) is not copied, and thus has to stay alive until the document gets stringified. I think that contract holds, but I figured I'd mention it just in case

src/include/metadata/iceberg_snapshot.hpp
smaheshwar-pltr: Ah thanks for pointing this out! It looks like it was introduced in C++14 but switched to `std::map` instead to be more lenient.

src/storage/iceberg_transaction.cpp
smaheshwar-pltr: Thanks, yeah looking at usages I also think this holds. This contract was also relied on by the `set_properties_update` serialization before, as in both cases, the keys' lifetimes are tied to the the `CommitTableRequest &table` parameter (I believe).

src/metadata/iceberg_snapshot.cpp
```diff
@@ -18,6 +18,40 @@ static string OperationTypeToString(IcebergSnapshotOperationType type) {
 	}
 }
 
+static const std::map<SnapshotMetricType, string> kSnapshotMetricKeys = {
```
Tmonster: nit: why the `k`?

src/storage/iceberg_transaction_data.cpp
```diff
@@ -22,68 +29,101 @@ static int64_t NewSnapshotId() {
...
+	                                        {SnapshotMetricType::ADDED_RECORDS, manifest_file.added_rows_count},
+	                                        {SnapshotMetricType::DELETED_DATA_FILES, manifest_file.deleted_files_count},
+	                                        {SnapshotMetricType::DELETED_RECORDS, manifest_file.deleted_rows_count}};
+
+	auto previous_total_files = previous_metrics.find(SnapshotMetricType::TOTAL_DATA_FILES);
+	if (previous_total_files != previous_metrics.end()) {
+		int64_t total_files =
+		    previous_total_files->second + manifest_file.added_files_count - manifest_file.deleted_files_count;
+		if (total_files >= 0)
+			metrics[SnapshotMetricType::TOTAL_DATA_FILES] = total_files;
```
Tishj, line 49: {}

src/storage/iceberg_transaction_data.cpp
```diff
@@ -22,68 +29,101 @@ static int64_t NewSnapshotId() {
...
+static IcebergSnapshot::metrics_map_t GetSnapshotMetrics(const IcebergManifestFile &manifest_file,
+                                                         const IcebergSnapshot::metrics_map_t &previous_metrics) {
+	IcebergSnapshot::metrics_map_t metrics {{SnapshotMetricType::ADDED_DATA_FILES, manifest_file.added_files_count},
+	                                        {SnapshotMetricType::ADDED_RECORDS, manifest_file.added_rows_count},
+	                                        {SnapshotMetricType::DELETED_DATA_FILES, manifest_file.deleted_files_count},
+	                                        {SnapshotMetricType::DELETED_RECORDS, manifest_file.deleted_rows_count}};
+
+	auto previous_total_files = previous_metrics.find(SnapshotMetricType::TOTAL_DATA_FILES);
+	if (previous_total_files != previous_metrics.end()) {
+		int64_t total_files =
```
Tishj, line 46: Where is `existing` in this?

Tishj: Nvm it only calculates add - delete, for every snapshot added

PR #539, opened 2025-10-16
Support `write.data.path` and `write.metadata.path`

### LINKED ISSUE #534: Support for `write.data.path` and `write.metadata.path`
By default, data and metadata are stored under the `location + /{metadata,data}`, but the catalog can give a hint to store it at a different location by passing in properties:

Image

I did a pass on the codebase, but I believe these are not taken into account. It would be a good addition 👍

Fokko:
Resolves #534

@samansmink I'm not a C++ expert, does this make sense? :D


src/include/metadata/iceberg_table_metadata.hpp
```diff
@@ -60,6 +64,10 @@ struct IcebergTableMetadata {
 	//! schema_id -> schema
 	unordered_map<int32_t, shared_ptr<IcebergTableSchema>> schemas;
 	vector<IcebergFieldMapping> mappings;
+
+	//! Custom write paths from table properties
+	string write_data_path;
+	string write_metadata_path;
```
smaheshwar-pltr, line 70: (Jumping in with a nit 😄:) wonder if we can avoid having these as public fields with getters having different logic. Could make them private - but WDYT about ditching the methods and setting the defaults appropriately during `IcebergTableMetadata` initialisation instead? (So `IcebergTransactionData::AddSnapshot` would just use `table_metadata.write_metadata_path`)

Fokko: I don't have a strong opinion here, but that's mostly because I'm not a C++ native :)

Issue #397, opened 2025-08-01
Query slowly when the iceberg table has lots of files


zhangjun0x01:
I have a iceberg table ,which has about 13000 data files , I exec the query  `select count(*) from mytable`  by duckdb iceberg extension ,it couse about  **4 minutes** ,while the spark only need  **4 seconds**

zhangjun0x01:
In my case, there are more that 10000 data files in some manifest, 
I debug the code , and find that when read this manifest, this code cost lots of times.

```
MultiFileReader::FinalizeChunk
```

PR #540, opened 2025-10-18
Optimizer for query statement `select count(*) from iceberg_scan`

zhangjun0x01:
Add optimizer for query `select count(*) from iceberg_scan('xx')` , fix https://github.com/duckdb/duckdb-iceberg/issues/397


src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -233,6 +233,29 @@ unique_ptr<NodeStatistics> IcebergMultiFileList::GetCardinality(ClientContext &c
 	return make_uniq<NodeStatistics>(cardinality, cardinality);
 }
 
+void IcebergMultiFileList::GetStatistics(vector<PartitionStatistics> &result) const {
+	if (GetMetadata().iceberg_version == 1) {
+		//! We collect no statistics information from manifests for V1 tables.
+		return;
+	}
+
+	if (!delete_manifests.empty()) {
```
zhangjun0x01: refer to spark , if exists delete file, continue
https://github.com/apache/iceberg/blob/36bb82675ff68ac0ed059d4db62550d30aa35760/spark/v4.0/spark/src/main/java/org/apache/iceberg/spark/source/SparkScanBuilder.java#L242

Tishj:
This shouldn't need a persistent test
Generated tests are prefered

See for example `empty_table` https://github.com/duckdb/duckdb-iceberg/tree/main/scripts/data_generators/tests/empty_table

zhangjun0x01:
> This shouldn't need a persistent test Generated tests are prefered
> 
> See for example `empty_table` https://github.com/duckdb/duckdb-iceberg/tree/main/scripts/data_generators/tests/empty_table


I refer to other test case ,  it seems that there is no difference. but the test case execute failed, while I run the script in my computer ,it is correct. so  I don't know why the "actions" failed to run.

```
(dbt-env) zhangjun@zhangjundeMacBook-Air duckdb-iceberg % python -m scripts.data_generators.generate_data local --test count_star_optimizer
25/10/21 16:53:34 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Setting default log level to "WARN".
To adjust logging level use sc.setLogLevel(newLevel). For SparkR, use setLogLevel(newLevel).
25/10/21 16:53:34 WARN Utils: Service 'SparkUI' could not bind on port 4040. Attempting port 4041.
25/10/21 16:53:34 WARN Utils: Service 'SparkUI' could not bind on port 4041. Attempting port 4042.
25/10/21 16:53:34 WARN Utils: Service 'SparkUI' could not bind on port 4042. Attempting port 4043.
Generating for 'local'
Generating test 'count_star_optimizer'
(dbt-env) zhangjun@zhangjundeMacBook-Air duckdb-iceberg % ll data/generated/iceberg/spark-local/default/                    
total 0
drwxr-xr-x  4 zhangjun  staff  128 Oct 21 16:53 count_star_optimizer
drwxr-xr-x  4 zhangjun  staff  128 Oct 21 16:53 count_star_optimizer_with_deletes
(dbt-env) zhangjun@zhangjundeMacBook-Air duckdb-iceberg % ./build/release/test/unittest test/sql/local/count_star_optimizer.test           
Filters: test/sql/local/count_star_optimizer.test
[1/1] (100%): test/sql/local/count_star_optimizer.test                                                                                                                 
===============================================================================
All tests passed (4 assertions in 1 test case)


```

zhangjun0x01:
I have  fixed the test case

zhangjun0x01:
Hi @Tishj could you help me review again ? thanks

test/sql/local/irc/insert/test_insert_into_sorted_table_fails.test
```diff
@@ -48,9 +48,6 @@ insert into my_datalake.default.table_sort_order select * from lineitem limit 5;
 ----
 <REGEX>:.*Not implemented.*Insert into a sorted iceberg table is not supported yet.*
 
-statement ok
```
zhangjun0x01, line 51: The transaction will cause the deletion operation to not be committed to Iceberg, and subsequent queries will return the initial value.

Tishj: "subsequent queries will return the initial value"
Please explain

The goal of this test is to check transactionality. Even though the delete doesn't get committed, within this transaction we will read the table as if the delete was committed already.

Removing the transactionality introduces unwanted side-effects to this test

Issue #454, opened 2025-09-06
Add support for Glue Catalog without S3 Tables


phitoduck:
The duckdb docs have [a page on accessing SageMaker Catalog/Lakehouse](https://duckdb.org/docs/stable/core_extensions/iceberg/amazon_sagemaker_lakehouse.html), but the examples given use S3 tables. I attempted in several ways in [this notebook](https://github.com/mlops-club/duckdb-iceberg-glue-catalog/blob/main/notebooks/iceberg/iceberg-glue.ipynb).

I tried several more variations outside this notebook as well and came to the conclusion that it probably isn't supported. Everything can be reproduced by running the commands described [in the README](https://github.com/mlops-club/duckdb-iceberg-glue-catalog/blob/main/README.md).

Could work be done to add the ability to connect to non-S3tables-based Iceberg tables in AWS Glue Catalog?

phitoduck:
@nicornk exactly. I did try to use the REST endpoint, passing the SigV4 endpoint in there, although I don't think I included it in the notebook. 

I can't remember the exact error I got, but I essentially followed [this approach](https://aws.amazon.com/blogs/big-data/accelerate-lightweight-analytics-using-pyiceberg-with-aws-lambda-and-an-aws-glue-iceberg-rest-endpoint/) from an AWS blog on Pyicebeg as closely as I could to form the connection:

```python
# Configure the catalog
catalog_properties = {
   "type": "rest",
   "uri": f"https://glue.{region}.amazonaws.com/iceberg",
   "s3.region": region,
   "rest.sigv4-enabled": "true",
   "rest.signing-name": "glue",
   "rest.signing-region": region
}
catalog = load_catalog(**catalog_properties)
```

Tmonster:
Hi @phitoduck ,

We had a similar issue filed a while ago. Have you tried some of the attach methods described here?  https://github.com/duckdb/duckdb-iceberg/issues/265

Please use the latest nightly when trying, since fixes for this have only been merged recently

nicornk:
We had issues with Glue and Lakeformation permissions in the past. Which lakeformation permissions does your role have?

shadercode:
We finally got the Glue Iceberg REST endpoint working in DuckDB 1.4 just a few minutes ago.  After days of troubleshooting and multiple AWS engineers invovled, I finally pieced together the missing permissions.  The main permission we were lacking was this high-level setting:

[...]

PR #545, opened 2025-10-22
catalog prefix can contain / and must not be encoded

alisoleimani:
Addresses #454


Tmonster:
Hi Thanks for the PR! In my experience some catalogs already return encoded prefixes and others not. This PR makes it seem like all prefixes will be encoded no matter what. I think if this is specific to AWS GLUE (which it seems like), maybe you can piggy-back off of the `ENDPOINT_TYPE` attach option to check if it is glue, and if so, encode the returned prefix

PR #547, opened 2025-10-23
Use GET request instead of HEAD+GET for an ICTableEntry

carlopi:
Note that on queries like `FROM <schema>.<tbl>;` this means a GET on the schema level and a GET at the table level.

I am not sure if there are tests that need changing, but CI will tell.

### DIFF
--- a/test/sql/local/irc/test_minimal_head_requests.test
+++ b/test/sql/local/irc/test_minimal_head_requests.test
@@ -54,11 +54,11 @@ select * from my_datalake.default.table_unpartitioned order by all;
 2023-03-11	11	k
 2023-03-12	12	l
 
-# 2 head requests to scan all files (1 for the schema, 1 for the table)
+# 1 head requests to scan all files (1 for the schema)
 query I
 select count(*) from duckdb_logs_parsed('HTTP') where request.type = 'HEAD';
 ----
-2
+1
 
 mode skip
 
--- a/test/sql/local/irc/test_table_information_requests.test
+++ b/test/sql/local/irc/test_table_information_requests.test
@@ -78,7 +78,7 @@ s_comment	VARCHAR
 query I
 select count(*) from duckdb_logs_parsed('HTTP');
 ----
-4
+3
 
 statement ok
 begin;
@@ -107,4 +107,4 @@ commit;
 query I
 select count(*) from duckdb_logs_parsed('HTTP');
 ----
-11
+10


src/storage/irc_table_set.cpp
```diff
@@ -29,14 +29,21 @@ namespace duckdb {
...
 	}
 
 	auto &ic_catalog = catalog.Cast<IRCatalog>();
-
-	auto get_table_result = IRCAPI::GetTable(context, ic_catalog, schema, table.name);
+	APIResult<rest_api_objects::LoadTableResult> get_table_result;
+	try {
+		get_table_result = IRCAPI::GetTable(context, ic_catalog, schema, table.name);
+	} catch (...) {
+		if (!might_throw) {
```
Tmonster, line 42: I worry if this is ever caught? I don't believe `IRCAPI::GetTable` will throw, I think all the information is in the `APIResult`

carlopi: I tried vs S3Tables, and `FROM <datalake>.<schema>.not_existing_table;` will throw in `IRCAPI::GetTable`.

That felt unexpected to me.

carlopi:
I think if you can at some point have a proper look, by yourself or together, it would be great

src/storage/irc_table_set.cpp
```diff
@@ -194,11 +204,14 @@ optional_ptr<CatalogEntry> ICTableSet::GetEntry(ClientContext &context, const En
 	auto table_name = lookup.GetEntryName();
 	auto entry = entries.find(table_name);
 	if (entry == entries.end()) {
-		if (!IRCAPI::VerifyTableExistence(context, ic_catalog, schema, table_name)) {
```
Tishj, line 197: If you're deleting the only use of this, please delete the dead code as well

Tishj:
Looking at the spec
```yaml
        404:
          description:
            Not Found - NoSuchTableException, table to load does not exist
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/IcebergErrorResponse'
              examples:
                TableToLoadDoesNotExist:
                  $ref: '#/components/examples/NoSuchTableError'
```

Only 404 should be seen as "does not exist"
The equivalent of `!VerifyTableExistence(...)`

Though `VerifyTableExistence` treats these as meaning "not found"
```c++
	case HTTPStatusCode::Forbidden_403:
	case HTTPStatusCode::NotFound_404:
		return IRCEntryLookupStatus::NOT_FOUND;
		break;
```

PR #548, opened 2025-10-23
Avoid needs for explicit casts in interpreting AT clause

carlopi:
Currenty `AT (TIMESTAMP => now())` will not work, requiring explicit cast like `AT (TIMESTAMP => now()::TIMESTAMP)`, and similar for passing integers or doubles to `VERSION`, I think this is just nicer / more user friendly.


Tmonster (approved):
Thanks!

PR #556, opened 2025-10-28
Refactor Manifest Files and Manifest List structs

Tmonster:
refactor manifest files and manifest lists so that a snapshot can cre…ate multiple new manifest files, and add them to a new manifest list for the snapshot

### DIFF
--- a/src/include/metadata/iceberg_manifest_list.hpp
+++ b/src/include/metadata/iceberg_manifest_list.hpp
@@ -101,7 +101,7 @@ public:
 	IcebergManifestFile manifest_file;
 
 public:
-	IcebergManifestListEntry() {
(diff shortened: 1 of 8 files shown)


src/include/metadata/iceberg_manifest_list.hpp
```diff
@@ -127,9 +133,18 @@ struct IcebergManifestList {
 	IcebergManifestList(const string &path) : path(path) {
 	}
 
+	vector<reference<IcebergManifestListEntry>> GetManifestFiles();
```
Tishj: I don't immediately understand why this is not `const vector<IcebergManifestListEntry> &` (with or without the const)

PR #558, opened 2025-10-29
Add Update Support to DuckDB-Iceberg

Tmonster:
Follow up to https://github.com/duckdb/duckdb-iceberg/pull/554

This PR will add support for Updates to DuckDB-Iceberg. 

Updates currently only work using `write.update.mode=merge-on-read` semantics. There is a PR in the works to enforce respecting table properties. 

There are many improvements to be made, but this will round out a lot of the functionality for v2 iceberg tables. Currently new positional delete files are added when a delete occurs, which can cause a lot of pollution and contribute to the small files problem. 

Hopefully in the future we can write an iceberg function to cleanup delete files and produce one delete file per data file, which will help with a conversion to V3.

### DIFF
--- a/test/python/test_pyiceberg_read.py
+++ b/test/python/test_pyiceberg_read.py
@@ -94,3 +94,35 @@ class TestPyIcebergRead:
             {'a': 57},
             {'a': 59},
         ]
+
+@pytest.mark.skipif(
+    os.getenv('ICEBERG_SERVER_AVAILABLE', None) == None, reason="Test data wasn't generated, run 'make data' first"
+)
+class TestPyIcebergRead:
+    def test_pyiceberg_read(self, rest_catalog):
+        table = rest_catalog.load_table("default.duckdb_updates_for_other_engines")
+        arrow_table: pa.Table = table.scan().to_arrow()
+        res = arrow_table.to_pylist()
+        assert len(res) == 20
+        assert res == [
+            {'a': 1},
+            {'a': 3},
+            {'a': 5},
+            {'a': 7},
+            {'a': 9},
+            {'a': 51},
+            {'a': 53},
+            {'a': 55},
+            {'a': 57},
+            {'a': 59},
+            {'a': 100},
+            {'a': 100},
+            {'a': 100},
+            {'a': 100},
+            {'a': 100},
+            {'a': 100},
+            {'a': 100},
+            {'a': 100},
+            {'a': 100},
+            {'a': 100},
+        ]
--- a/test/python/test_spark_read.py
+++ b/test/python/test_spark_read.py
@@ -130,3 +130,38 @@ class TestSparkReadDuckDBTableWithDeletes:
             Row(a=57),
             Row(a=59),
         ]
+
+@pytest.mark.skipif(
+    os.getenv('ICEBERG_SERVER_AVAILABLE', None) == None,
+    reason="Test data wasn't generated, run tests in test/sql/local/irc first",
+)
+class TestSparkReadDuckDBTableWithDeletes:
+    def test_spark_read(self, spark_con):
+        df = spark_con.sql(
+            """
+            select * from default.duckdb_updates_for_other_engines order by a
+            """
+        )
+        res = df.collect()
+        assert res == [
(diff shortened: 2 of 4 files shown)


src/include/storage/iceberg_insert.hpp
```diff
@@ -29,11 +29,21 @@ struct IcebergCopyInput {
 	case_insensitive_map_t<vector<Value>> options;
 };
 
+class IcebergInsertGlobalState : public GlobalSinkState {
+public:
+	explicit IcebergInsertGlobalState() = default;
+	vector<IcebergManifestEntry> written_files;
```
Tishj, line 36: Won't this need a lock?
If multiple threads are appending to the vector?

src/storage/iceberg_update.cpp
```diff
@@ -0,0 +1,330 @@
...
+	return std::move(result);
+}
+
+unique_ptr<LocalSinkState> IcebergUpdate::GetLocalSinkState(ExecutionContext &context) const {
+	auto result = make_uniq<IcebergUpdateLocalState>();
+	result->copy_local_state = copy_op.GetLocalSinkState(context);
+	result->delete_local_state = delete_op.GetLocalSinkState(context);
+
+	vector<LogicalType> delete_types;
+	// TODO: Verify these deletes tupes
```
Tishj: Stray TODO?

src/storage/iceberg_update.cpp
```diff
@@ -0,0 +1,330 @@
...
+	for (idx_t i = 0; i < 2; i++) {
+		if (delete_chunk.data[i].GetVectorType() != chunk.data[delete_idx_start + i].GetVectorType()) {
+			delete_chunk.data[i].Flatten(chunk.size());
+			chunk.data[delete_idx_start + i].Flatten(chunk.size());
+		}
+		delete_chunk.data[i].Reference(chunk.data[delete_idx_start + i]);
+	}
+
+	OperatorSinkInput delete_input {*delete_op.sink_state, *lstate.delete_local_state, input.interrupt_state};
+	delete_op.Sink(context, delete_chunk, delete_input);
```
Tishj, line 98: One question popped up just now:
Updates don't need to care whether the update applies to existing data or transaction-local (uncommitted) data, right?

That complexity is handled by the Delete operator?

test/sql/local/irc/update/update_same_transaction.test
```diff
@@ -0,0 +1,64 @@
...
+statement ok
+DROP TABLE IF EXISTS my_datalake.default.test;
+
+statement ok
+BEGIN TRANSACTION;
+
+statement ok
+CREATE TABLE my_datalake.default.test (id INTEGER, name TEXT);
+
+statement ok
```
Tishj, line 52: Can we have multiple inserts that are all partially touched by an update?

Also a test where we begin, insert, delete parts of it, then try to update the rows we deleted. That should fail I think (?)

src/storage/iceberg_update.cpp
Tmonster: Yea, the update just processes the data, it's up to the iceberg multi file reader to read the transaction local data and pass it to the delete & insert operator.

src/storage/iceberg_update.cpp
Tmonster: yup

test/sql/local/irc/update/update_same_transaction.test
Tmonster: > Also a test where we begin, insert, delete parts of it, then try to update the rows we deleted. That should fail I think (?)

This will just be an empty delete/no new snapshot will be made

Tmonster: But I will add a test case regardless

src/include/storage/iceberg_insert.hpp
Tmonster: yesss, it does. Surprised this didn't fail earlier (Code was just moved here from iceberg_insert.cpp)

Tishj (approved):
Looks great, thanks!

PR #561, opened 2025-10-30
Enable Setting/Checking of Table Properties. Do not allow deletes if `'write.delete.mode' != 'merge-on-delete'`

Tmonster:
Hopefully in the future we can have syntax like 
```
ALTER TABLE my_datalake.default.test SET/ADD/DROP TBLPROPERTIES ('option1'='option2');
```

For now I think table functions are easiest. I don't enjoy having to prefix the properties with `MAP` though, still looking for better ways to fix that.

### DIFF
--- a/test/sql/local/irc/update/test_update_v3_table.test
+++ b/test/sql/local/irc/update/test_update_v3_table.test
@@ -41,5 +41,5 @@ update my_datalake.default.simple_v3_table set c_name = 'new_customer_key' where
 statement error
 update my_datalake.default.pyspark_iceberg_table_v1 set l_suppkey_long = 100000 where l_suppkey_long%2 = 0;
 ----
-<REGEX>:.*Not implemented Error: Update from Iceberg V1 tables.*
+<REGEX>:.*Not implemented Error.*Update Iceberg V1 tables.*
 


src/iceberg_functions/iceberg_table_properties_functions.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+	auto &map_children = MapValue::GetChildren(map);
+	for (idx_t col_idx = 0; col_idx < map_children.size(); col_idx++) {
+		auto &struct_children = StructValue::GetChildren(map_children[col_idx]);
+		auto &key = StringValue::Get(struct_children[0]);
+		auto &val = StringValue::Get(struct_children[1]);
+		ret->properties.emplace(key, val);
+	}
+
+	auto iceberg_table = IcebergUtils::GetICTableEntry(context, input_string);
```
Tishj: This also needs the `catalog_entry->type == CatalogType::TABLE_ENTRY` check, because it could be a view

In fact, that's not even guarantee enough that it's an iceberg table
I think we need to use `GetCatalogType` to verify that the catalog associated with the table is `"iceberg"`

Tishj: Also note that `GetICTableEntry` goes against the purpose it's used for in `GetStorageLocation`
We explicitly do this look up so it gives us a view/table and we can act accordingly, to unify tests:

```sql
statement ok
attach ':memory:' as my_datalake;

statement ok
create schema my_datalake.default;

statement ok
create view my_datalake.default.pyspark_iceberg_table_v1 as select * from ICEBERG_SCAN('__WORKING_DIRECTORY__/data/generated/iceberg/spark-local/default/pyspark_iceberg_table_v1');
```

This view can then be used with `iceberg_snapshots(my_datalake.default.pyspark_iceberg_table_v1)`
And an actual table that comes from an attached iceberg catalog can be used with the exact same path

src/metadata/iceberg_table_metadata.cpp
```diff
@@ -2,6 +2,8 @@
 
 #include "iceberg_utils.hpp"
 #include "catalog_utils.hpp"
+#include "../include/metadata/iceberg_snapshot.hpp"
```
Tishj: No relative paths please

src/include/iceberg_functions/iceberg_table_properties_functions.hpp
```diff
@@ -0,0 +1,129 @@
+//===----------------------------------------------------------------------===//
+//                         DuckDB
+//
+// iceberg_metadata.hpp
```
Tishj: Why is this file called `iceberg_table_properties_functions.hpp` ?
I assume this is separated out from an existing file so it can be included in two places more cleanly?

Can we rename it to something more generic / related to what is in the file

src/iceberg_functions/iceberg_table_properties_functions.cpp
```diff
@@ -0,0 +1,255 @@
...
+		return;
+	}
+	// if we have already returned all properties.
+	if (global_state.property_count >= table_info.load_table_result.metadata.properties.size()) {
+		output.SetCardinality(0);
+		return;
+	}
+
+	idx_t count = global_state.property_count;
+	for (auto &property : table_info.load_table_result.metadata.properties) {
```
Tishj: This is not making sure we don't return more than `STANDARD_VECTOR_SIZE` entries per call
That's going to be impossible to test, but let's at least try to make this resilient

I think all we need is to keep an iterator, check if it's not at `end`
start `count` at 0 and check `count < STANDARD_VECTOR_SIZE` in this loop

src/common/utils.cpp
```diff
@@ -1,6 +1,8 @@
 #include "duckdb.hpp"
 #include "iceberg_utils.hpp"
 #include "fstream"
+#include "../include/storage/irc_table_entry.hpp"
```
Tishj: No relative includes please

src/common/utils.cpp
```diff
@@ -47,22 +49,42 @@ static string ExtractIcebergScanPath(const string &sql) {
...
+	if (!table) {
+		throw InvalidInputException("Could not find Catalog Entry with name '%s'", input);
+	}
+	return table;
+}
+
+optional_ptr<ICTableEntry> IcebergUtils::GetIcebergTableEntry(ClientContext &context, const string &name) {
+	auto catalog_entry = IcebergUtils::GetICTableEntry(context, name);
+	if (catalog_entry->type == CatalogType::TABLE_ENTRY) {
+		//! This is a IRCTableEntry, set up the scan from this
```
Tishj: Correction, this is a `TableCatalogEntry` we need to cast it to that first, then get the `Catalog()` (maybe `ParentCatalog()`?) and check `GetCatalogType() == "iceberg"`

src/iceberg_functions/iceberg_table_properties_functions.cpp
```diff
@@ -0,0 +1,295 @@
...
+	bool all_properties_initialized = false;
+	bool properties_set = false;
+	bool properties_removed = false;
+};
+
+static optional_ptr<ICTableEntry> GetTableEntry(ClientContext &context, string &input_string) {
+
+	auto qualified_name = QualifiedName::ParseComponents(input_string);
+	optional_ptr<ICTableEntry> iceberg_table = nullptr;
+	auto default_db = DatabaseManager::GetDefaultDatabase(context);
```
Tishj: I don't think this is what we want to do, we can check inside GetIcebergTableEntry, see my earlier comment

src/common/utils.cpp
Tmonster: CLion keeps adding these automatically, it's been so annoying. I also don't want these 😭. Have tried a couple of times to fix it, but haven't gotten anywhere

Tishj: Ah I figured, I'll ignore it then 👍

Tmonster: Still call it out though, otherwise they will accidentally end up in there

Issue #555, opened 2025-10-27
Partition pruning not working with month transform in Iceberg table scan via DuckDB


phmu16ab:
Hi all,

I have encountered what appears to be a partition pruning issue when querying Iceberg tables partitioned by month(ts) using DuckDB.
When applying a filter on the timestamp column, < and = filters **result in zero rows returned and no data files scanned**, even though matching records exist. In contrast, > filters behave correctly and return expected results.
This suggests that partition pruning logic may incorrectly exclude relevant partitions when using < or = filters on a month(ts) partition transform.

### Environment
PyIceberg version: 0.10.0
DuckDB version: 1.4.1 (latest iceberg extension via UPDATE EXTENSIONS)
Catalog: SQLite
Warehouse: Local filesystem
Partition spec: month(ts) transform

### Reproduction
Here’s a minimal reproducible example (full code at the end):

- Create a table partitioned by month(ts) using PyIceberg.
- Insert 5 timestamps (YYYY-MM-DD) for each of Jan, Feb, Mar 2025. 
- Use DuckDB’s iceberg_scan() with different timestamp filters 

1. ts < '2025-02-01'
2. ts > '2025-02-01'; 
3. ts = '2025-02-01'.

- Run the queries with filter pushdown once enabled and once disabled --> (SET disabled_optimizers = 'filter_pushdown')

### Observations
When enabling the filter pushdown, it appears the monthly partition pruning breaks the '<' and '=' filters. The '>' filter works correctly. 

**Expected results:**
```
Total rows @version: 15
Rows <  2025-02-01: 5
Rows  = 2025-02-01 exactly: 1
Rows >  2025-02-01: 9
```

**Results obtained with filter pushdown enabled:**
```
Total rows @version: 15
Rows <  2025-02-01: 0
Rows  = 2025-02-01 exactly: 0
Rows >  2025-02-01: 9
```

**Results obtained with filter pushdown disabled:**
```
Total rows @version: 15
Rows <  2025-02-01: 5
Rows  = 2025-02-01 exactly: 1
Rows >  2025-02-01: 9
```

Profiling confirms that no files are scanned for < and = filters, indicating that partition pruning may be overly aggressive or misinterpreting the filter bounds.

### Questions
- Is month(ts) transform currently supported for partition pruning in DuckDB’s Iceberg integration?
- Could this be a limitation in the metadata interpretation or filter pushdown logic?

Please let me know if this is expected behavior or if additional debugging info would help.

[...]

PR #563, opened 2025-10-31
Fix reading from tables partitioned by month

Tmonster:
Need to use GetAge() and not GetDifference().

Closes https://github.com/duckdb/duckdb-iceberg/issues/555


Tmonster:
There is still some weird stuff going on with timezones though. Will need to take another look at that

PR #564, opened 2025-11-03
Handle unsigned types when creating a table

Tmonster:
We don't want to create a table with invalid types using `Create Table As`. For some reason we cannot promote those types easily, maybe if we add a projection to cast them. You can create the table with `Create Table t1 ({name} {type})` and the type will be cast to a valid iceberg type. Then when you insert we cast to the correct type.

### DIFF
--- a/src/include/utils/iceberg_type.hpp
+++ b/src/include/utils/iceberg_type.hpp
@@ -1,10 +1,6 @@
 
 #pragma once
 
-#include "duckdb/common/string_util.hpp"
-#include "duckdb/common/types.hpp"
-#include "duckdb/parser/column_definition.hpp"
-#include "duckdb/parser/parsed_data/create_table_info.hpp"
 #include "rest_catalog/objects/type.hpp"
 #include "functional"
 


src/utils/iceberg_type.cpp
```diff
@@ -9,12 +10,41 @@
...
+			                            LogicalTypeIdToString(column.Type().id()));
+		}
+	}
+}
+
 string IcebergTypeHelper::LogicalTypeToIcebergType(const LogicalType &type) {
 	switch (type.id()) {
-	case LogicalTypeId::TINYINT:
+	case LogicalTypeId::USMALLINT:
 	case LogicalTypeId::UTINYINT:
```
Tishj: I don't understand why we are accepting some of the unsigned types?
Especially because we did remove all tests for unsigned types now

src/include/utils/iceberg_type.hpp
```diff
@@ -12,6 +14,7 @@ class IcebergTypeHelper {
 public:
 	static rest_api_objects::Type CreateIcebergRestType(const LogicalType &type, std::function<idx_t()> get_next_id);
 	static string LogicalTypeToIcebergType(const LogicalType &type);
+	static void IsValidIcebergType(CreateTableInfo &info);
```
Tishj: "Is" signals that it returns a boolean
Common terminology is to use Verify for checks that throw if the condition isn't matched, Check is also fine

Tishj (approved):
LGTM
Thanks for working on this 🙌

PR #569, opened 2025-11-06
Fix aws logging logic

Tmonster:
Currenlty AWS requests are logged before the request (which we shouldn't do). In addition the responses are not logged. This is a PR to fix that.

### DIFF
--- a/src/aws.cpp
+++ b/src/aws.cpp
@@ -160,34 +160,35 @@ unique_ptr<HTTPResponse> AWSInput::ExecuteRequest(ClientContext &context, Aws::H
 	auto request = CreateSignedRequest(method, uri, headers, body);
 
 	auto httpClient = Aws::Http::CreateHttpClient(clientConfig);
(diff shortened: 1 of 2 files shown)


src/aws.cpp
```diff
@@ -180,13 +178,16 @@ unique_ptr<HTTPResponse> AWSInput::ExecuteRequest(ClientContext &context, Aws::H
...
+		result->headers[header.first] = header.second;
+	}
 	Aws::StringStream resBody;
 	resBody << response->GetResponseBody().rdbuf();
 	result->body = resBody.str();
-
 	if (static_cast<uint16_t>(result->status) > 400) {
 		result->success = false;
 	}
+	LogAWSHTTPRequest(context, request, *result, method);
```
samansmink, line 188: I think this is potentially too late, now if an httpexception is thrown we will not log the request

maybe we try catch then log in the catch

PR #576, opened 2025-11-11
Interact with Iceberg REST catalogs always via HTTPUtil (instead of AWS SDK)

carlopi:
This adds a default enabled new codepath that allows all interactions with REST Iceberg catalogs to occur via DuckDB's own HTTPUtil (moving away from using the AWS's SDK).

This requires 2 components:
* adding logic (borrowed for `duckdb-httpfs` via `duckdb-iceberg`'s `wasm` branch) to compute SigV4 signatures
* adding logic to constructing the relevant requests

Note that at this moment computing the credentials is still delegated to AWS's SDK. This seemed simplest from a code perspective.
So flow will now be:
1. AWS's SDK is asked to produce relevant credentials
2. A DuckDB's own HTTPRequest is built adding relevant headers
3. SigV4 signature is computed, and also added as header
4. Request is sent using HTTPUtil

Adding this in moves 1 step closer to properly merging back `wasm` branch into main, but to keep this already non-trivial changes to a minimum I'll do so in a follow up.

An option `iceberg_via_aws_sdk_for_catalog_interactions` is added, allowing opting back in to the previous behaviour. To be consider whether full removal is cleaner vs keeping the options in.

I would consider running the full set of tests vs remote S3Tables server giving CI in this PR is mostly checking for consistency.

### DIFF
--- a/test/sql/cloud/glue/test_insert_glue.test
+++ b/test/sql/cloud/glue/test_insert_glue.test
@@ -121,7 +121,7 @@ commit
 statement error con1
 commit
 ----
-<REGEX>:.*TransactionContext Error.*branch main has changed.*
+<REGEX>:.*TransactionContext Error.*
 
 # we see the con2 update
 query IIII
--- a/test/sql/cloud/s3tables/test_logging_aws.test
+++ b/test/sql/cloud/s3tables/test_logging_aws.test
@@ -42,9 +42,10 @@ select count(*) from s3_catalog.tpch_sf1.region;
 ----
 5
 
+
 # make sure we log calls to IRC
 query II
-select request.url, request.type from duckdb_logs_parsed('HTTP') where request.type in ('GET', 'HEAD') and request.url like '%/iceberg-testing/namespaces/%'
+select request.url, request.type from duckdb_logs_parsed('HTTP') where request.type in ('GET', 'HEAD') and request.url like 'iceberg-testing/namespaces/%'
 ----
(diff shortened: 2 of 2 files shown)


src/include/hash_utils.hpp
```diff
@@ -0,0 +1,13 @@
+#include <string>
```
Tmonster, line 1: `#include "duckdb/common/string.hpp"`? Or maybe I'm missing something?

src/aws.cpp
```diff
@@ -190,6 +193,144 @@ unique_ptr<HTTPResponse> AWSInput::ExecuteRequest(ClientContext &context, Aws::H
...
+	if (!use_httputils) {
+		// Query Iceberg REST catalog via AWS's SDK
+		return ExecuteRequestLegacy(context, method, headers, body);
+	}
+
+	auto uri = BuildURI();
+	auto &db = DatabaseInstance::GetDatabase(context);
+
+	HTTPHeaders res(db);
+	{
```
Tmonster: Any specific reason we have a block here?

src/include/hash_utils.hpp
carlopi: This is meant to mirror https://github.com/duckdb/duckdb-httpfs/blob/main/src/include/hash_functions.hpp, I should likely make that explicit in a comment

src/aws.cpp
carlopi: Right, I'll fix it up.

Tmonster:
Thanks!

PR #582, opened 2025-11-12
Bump duckdb to 68d7555f68bd25c1a251ccca2e6338949c33986a (also in CI)

carlopi:
Test vs tentative 1.4.2

### DIFF
--- a/.github/workflows/MainDistributionPipeline.yml
+++ b/.github/workflows/MainDistributionPipeline.yml
@@ -17,7 +17,7 @@ jobs:
     uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@main
     with:
       extension_name: iceberg
-      duckdb_version: v1.4.1
+      duckdb_version: 68d7555f68bd25c1a251ccca2e6338949c33986a
       ci_tools_version: v1.4.1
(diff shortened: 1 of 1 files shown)


.github/workflows/MainDistributionPipeline.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@main
     with:
       extension_name: iceberg
-      duckdb_version: v1.4.1
+      duckdb_version: 68d7555f68bd25c1a251ccca2e6338949c33986a
```
Tmonster, line 20: I think you need to add [duckdb_tag](https://github.com/duckdb/extension-ci-tools/blob/49436d8e634187eecde1e32a377f5fe3887b5fff/.github/workflows/_extension_distribution.yml#L126) as well

carlopi: ducdkb_tag is a bit brittle, I did review and was not super comfortable. I'll fire a different PR with actual v1.4.2 / v1.4.2 as soon as available.

Tmonster: 👍 ok. sounds good

carlopi:
This is now outdated, thanks

PR #597, opened 2025-11-19
Cherry-pick Tmonster/remove_head_requests_from_positional_deletes

carlopi:
This is due to the fact that `wasm` branch is not built from `v1.4-andium` at the moment. I'll sort that out, but for starters bring back feature parity.


carlopi:
Now irrelevant, wasm-* targets also coming from `main` branch

PR #601, opened 2025-11-24
Update how we run tests

Tmonster:
This is a PR to update how tests are run. It still needs some changes to point branches to the duckdb-iceberg workflows instead of my personal ones. Leaving as is for now to show that this works


Tishj (approved):
I'm leaving this entirely to you

PR #602, opened 2025-11-24
Use partition information in manifest file to perform filter pushdown

Tmonster:
Given a table with many small files and many snapshots, it's possible some kind of compaction is performed. During this time, new manifest lists and manifest files (and potentially parquet files) are written. Somewhere in the Spark Rest catalog, new manifest files that are written during compaction **do not** contain upper and lower bound values for certain columns. What spark will do, however, is write partition information in the `partitions` field of the manifest columns. DuckDB only checks upper and lower bound values of columns in manifest files. Because of this, DuckDB will end up reading waaaaay too many files when there is a filter on a partitioned column.

To fix this problem, DuckDB will now read partition information in manifest lists.

Some other notes.

1. I could not manage a way to consistently reproduce this issue, so I had to add a persistent data. I have stepped through the code and tested that this works
2. I also added a test where a table has multiple partitions, that caused a bug in Iceberg to DuckLake, which I have now also fixed.
3. Also added some helper scripts to make it easier to create a persistent table from one in an IRC catalog somewhere

### DIFF
--- a/.github/workflows/Nessie_Testing.yml
+++ b/.github/workflows/Nessie_Testing.yml
@@ -35,11 +35,11 @@ jobs:
       - name: Install nessie required packages
         shell: bash
         run: |
+          sudo apt-get update -y -qq
           # install docker
           sudo curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
           sudo chmod +x /usr/local/bin/docker-compose
           # install java
-          suto apt-get update -y -qq
           sudo apt install -y -qq openjdk-21-jre-headless
           sudo apt install -y -qq openjdk-21-jdk-headless
           sudo apt-get install -y -qq python3-venv


src/iceberg_functions/iceberg_to_ducklake.cpp
```diff
@@ -384,6 +384,7 @@ struct DuckLakePartition {
 		for (auto &field : partition.fields) {
 			columns.push_back(DuckLakePartitionColumn(field));
 		}
+		partition_id = partition.spec_id;
```
Tishj: I don't think this is correct?
See the comment:
`//! The id is assigned after we've processed all tables`

This already happens later, see `FinalizeEntry`

src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -283,6 +312,72 @@ bool IcebergMultiFileList::FileMatchesFilter(const IcebergManifestEntry &file) {
...
+					continue;
+				}
+
+				auto &column = IcebergTableSchema::GetFromColumnIndex(schema, column_id, 0);
+
+				// initialize dummy stats
+				auto stats = IcebergPredicateStats();
+				bool found_parition_field = false;
+				for (auto &partition_val : file.partition_values) {
+					auto source_id = partition_val.first;
```
Tishj: This shouldn't be called `source_id`, the partition values are matched on the `field_id` of the partition:
> Partition data tuple, schema based on the partition spec output using partition field ids for the struct field ids

src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -283,6 +312,72 @@ bool IcebergMultiFileList::FileMatchesFilter(const IcebergManifestEntry &file) {
...
+				auto table_filter = GetFilterForColumnIndex(table_filters, column_id);
+				if (!table_filter) {
+					continue;
+				}
+
+				auto &column = IcebergTableSchema::GetFromColumnIndex(schema, column_id, 0);
+
+				// initialize dummy stats
+				auto stats = IcebergPredicateStats();
+				bool found_parition_field = false;
```
Tishj: `found_partition_field`

src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -283,6 +312,72 @@ bool IcebergMultiFileList::FileMatchesFilter(const IcebergManifestEntry &file) {
...
+							stats.has_null = true;
+						} else {
+							stats.has_not_null = true;
+						}
+						break;
+					}
+				}
+
+				if (!found_parition_field) {
+					break;
```
Tishj: I'm not sure if breaking or continuing is the best course of action here, can't we still determine if a file should be skipped based on the other partitions that are present?

src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -283,6 +312,72 @@ bool IcebergMultiFileList::FileMatchesFilter(const IcebergManifestEntry &file) {
...
+				}
+
+				auto result_type = field.transform.GetSerializedType(column.type);
```
Tishj: This is unused?

Tishj: I went looking to see if the partition values adhere to the same schema evolution rules as lower/upper bounds of a data file, but they do not:
> Type promotion is not allowed for a field that is referenced by source-id or source-ids of a partition field if the partition transform would produce a different value after promoting the type

So this is correct 👍

Tishj: I went looking at iceberg-python just to check, but this is so abstract..
https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L1929
I can't make out what they're actually doing 🙃 

But then this:
```py
    def _build_partition_evaluator(self, spec_id: int) -> Callable[[DataFile], bool]:
        spec = self.table_metadata.specs()[spec_id]
        partition_type = spec.partition_type(self.table_metadata.schema())
        partition_schema = Schema(*partition_type.fields)
        partition_expr = self.partition_filters[spec_id]

        # The lambda created here is run in multiple threads.
        # So we avoid creating _EvaluatorExpression methods bound to a single
        # shared instance across multiple threads.
        return lambda data_file: expression_evaluator(partition_schema, partition_expr, self.case_sensitive)(data_file.partition)
```
Makes me think that schema evolution is applied, and we need to potentially up-cast the stored partition value to the current type of the partition spec's field ..

Tmonster: > Type promotion is not allowed for a field that is referenced by source-id or source-ids of a partition field if the partition transform would produce a different value after promoting the type

I think I get what is happening here. 
If you promote an int column that is partitioned by `bucket(a)` to string, then the new partition values on strings are different than on ints, (i.e 24 vs "24"). When querying with a filter on the string partitioned  column (`where a = "24"`), DuckDB will assume the constant value is a string, and not an int, and the transform for partition pruning may be incorrect. 

For safe type promotions (ie `int` -> `long`) existing data files remain valid. To figure out what kind of Transform to apply to the filter constant, you can look at the partition_id in each entry of a manifest_list. Every manifest file has only data files from one partition. Using that partition_id, we can find out what transform was used to write the data files, and apply that to the expression constant to see if we can filter out certain data files/manifest files. 

The spec also says that a manifest file can only list data files from the same partition spec. https://iceberg.apache.org/spec/#partitioning

Don't know if this helps at all

Tishj: Yea you're right, so schema evolution *can* happen, as long as it doesn't change the meaning of the partition transform

Then I think this logic needs to up-cast the stored `partition` Value, don't you think?

Tishj: So just to provide an update for future reference:
I **don't** think any upcasting is necessary

The lower/upper bounds in the `data_file` are serialized according to the type of the column at the time of file creation.
To properly deserialize them we need to infer what this type was, in case schema evolution took place. (i.e INTEGER -> BIGINT, the bounds are inferred to be INTEGER when the size of the blob is 4 (sizeof(int32_t))

Since the `partition` values have type information stored, there is no need for any inference.

PR #608, opened 2025-11-28
apply patches, update submodules, bump ducklake

Tmonster:


Tmonster:
Tracked this down. When creating an IcebergMultiFileList list we need to also keep a reference/pointer to the Iceberg Table. Will work on this after v1.4.3 is out

PR #613, opened 2025-12-01
Fix encoding issues v1.4.3

Tmonster:
Fix how we construct our catalog urls. Previously we would not encode table names or warehouse names and now we do. Changed the `GetURL` to `GetEncodedURL` as well to make this distinction clear.

If components are passed to the URLBuilder as already encoded, this will cause problems, but catalogs should be passing warehouse names as not encoded. 
The AWS request logic is not affected by this because DuckDB-Iceberg passes the individual non-encoded components to the AWS uri builder. The AWS uri object then will encode the URL when it is used.

### DIFF
--- a/test/sql/cloud/s3tables/test_create_table_s3tables.test
+++ b/test/sql/cloud/s3tables/test_create_table_s3tables.test
@@ -24,7 +24,6 @@ CREATE SECRET (
     PROVIDER credential_chain
 );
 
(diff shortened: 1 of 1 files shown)


Tmonster:
cloud tests passing here https://github.com/duckdb/duckdb-iceberg/actions/runs/19875146744

PR #614, opened 2025-12-02
Perform delete file pruning

Tmonster:
Delete file pruning is helpful when reading from a partition. Delete manifests can still pass the filter pushdown, but there is no guarantee the individual data files need to be read

Still would like to add a test for this, but again requires a copying a spark created table that was made after many inserts.

### DIFF
--- a/src/iceberg_functions/iceberg_multi_file_list.cpp
+++ b/src/iceberg_functions/iceberg_multi_file_list.cpp
@@ -647,7 +647,8 @@ void IcebergMultiFileList::InitializeFiles(lock_guard<mutex> &guard) {
 
 			if (manifest.content == IcebergManifestContentType::DATA) {
 				data_manifests.push_back(manifest);
-			} else if (manifest.content == IcebergManifestContentType::DELETE) {
+			} else {
+				D_ASSERT(manifest.content == IcebergManifestContentType::DELETE);
 				delete_manifests.push_back(manifest);
 			}
 		}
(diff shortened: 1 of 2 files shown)


src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -648,9 +647,10 @@ void IcebergMultiFileList::InitializeFiles(lock_guard<mutex> &guard) {
 
 			if (manifest.content == IcebergManifestContentType::DATA) {
 				data_manifests.push_back(manifest);
-			} else {
-				D_ASSERT(manifest.content == IcebergManifestContentType::DELETE);
+			} else if (manifest.content == IcebergManifestContentType::DELETE) {
 				delete_manifests.push_back(manifest);
+			} else {
+				auto break_here = 0;
```
Tishj: What's this?

src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -718,7 +718,20 @@ void IcebergMultiFileList::ProcessDeletes(const vector<MultiFileColumnDefinition
 
 		current_delete_manifest++;
 
+		if (manifest_file.path ==
```
Tishj: More debug code

Tishj (changes requested):
Just need to strip the debug code, otherwise looks great 👍

PR #628, opened 2025-12-15
Fix commit endpoint typo to enable transactions

smaheshwar-pltr:
The `commitTransaction` endpoint is currently not used due a typo when DuckDB is used with S3 tables or REST catalogs that don't return an explicit list of endpoints. This PR fixes that


Tmonster:
Thanks @smaheshwar-pltr. I merged the backport to v1.4-andium, commit will show up in main next time I merge v1.4 into main.

PR #638, opened 2025-12-29
Add extended filter pruning optimizations

achille-roussel:
This PR adds support for additional filter types in Iceberg file pruning:

- **CONJUNCTION_OR:** Prune files when ALL child filters would prune
- **COMPARE_NOTEQUAL (!=):** Prune when file contains only the excluded value
- **COMPARE_BETWEEN:** Prune when file range doesn't overlap query range
- **BOUND_FUNCTION (prefix/starts_with):** Prune based on string prefix bounds
- **COMPARE_NOT_IN:** Prune when file's single value is in exclusion list

All optimizations follow the existing conservative approach - unknown or unhandled cases return true (don't prune) to ensure correctness.

I included tests in extended_filter_pruning.test validating both correctness and pruning behavior.

---

The second commit adds support for filter pruning on expression patterns that DuckDB's FilterCombiner drops:
  - Complex nested OR with AND children: (col >= 0 AND col < 500) OR (col >= 4500)
  - NOT BETWEEN expressions
  - NOT (expr) logical negation
  - ends_with/suffix string functions

### DIFF
--- a/src/catalog_api.cpp
+++ b/src/catalog_api.cpp
@@ -144,7 +144,7 @@ bool IRCAPI::VerifySchemaExistence(ClientContext &context, IcebergCatalog &catal
 	auto schema_name = GetEncodedSchemaName(namespace_items);
 
 	auto url_builder = catalog.GetBaseUrl();
-	url_builder.AddPathComponent(catalog.prefix);
+	url_builder.AddPrefixComponent(catalog.prefix, catalog.prefix_is_one_component);
 	url_builder.AddPathComponent("namespaces");
 	url_builder.AddPathComponent(schema_name);
 	bool execute_head =
@@ -157,7 +157,7 @@ bool IRCAPI::VerifyTableExistence(ClientContext &context, IcebergCatalog &catalo
 	auto schema_name = GetEncodedSchemaName(schema.namespace_items);
 
 	auto url_builder = catalog.GetBaseUrl();
-	url_builder.AddPathComponent(catalog.prefix);
+	url_builder.AddPrefixComponent(catalog.prefix, catalog.prefix_is_one_component);
 	url_builder.AddPathComponent("namespaces");
 	url_builder.AddPathComponent(schema_name);
 	url_builder.AddPathComponent("tables");
@@ -172,7 +172,7 @@ static unique_ptr<HTTPResponse> GetTableMetadata(ClientContext &context, Iceberg
 	auto schema_name = IRCAPI::GetEncodedSchemaName(schema.namespace_items);
 
 	auto url_builder = catalog.GetBaseUrl();
-	url_builder.AddPathComponent(catalog.prefix);
+	url_builder.AddPrefixComponent(catalog.prefix, catalog.prefix_is_one_component);
 	url_builder.AddPathComponent("namespaces");
 	url_builder.AddPathComponent(schema_name);
 	url_builder.AddPathComponent("tables");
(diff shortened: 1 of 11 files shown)


achille-roussel:
Hello @Tmonster,

I have a follow up change with more optimizations that I would like to submit but didn't want to mix in this one:
- https://github.com/achille-roussel/duckdb-iceberg/pull/1 

I'll submit it after we got this one merged in.

achille-roussel:
Hello @Tmonster 

Would you be able to take a look at this change in the coming week?

Tmonster:
Hi @achille-roussel,

Yes, I will try to take a look sometime this week 👍

src/iceberg_predicate.cpp
```diff
@@ -203,11 +219,41 @@ bool MatchBoundsTemplated(ClientContext &context, const TableFilter &filter, con
 			D_ASSERT(expr.type == ExpressionType::OPERATOR_IS_NOT_NULL);
 			return MatchBoundsIsNotNullFilter<TRANSFORM>(stats, transform);
 		}
+		case ExpressionType::COMPARE_BETWEEN: {
+			D_ASSERT(expr.GetExpressionClass() == ExpressionClass::BOUND_BETWEEN);
```
Tmonster: I tried looking into this with my own between statement and could never hit this code. To hit this, I think you need to add `BETWEEN` in the `ShouldEvaluateDirectly` function?

Although, in my debugging, I have also found that in the filter combiner, `BETWEEN` is transformed into a bound and conjunction expression with `greater than or equal to` and `less than or equal to` expressions, which seem to achieve the same thing added here

src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -189,6 +196,34 @@ IcebergMultiFileList::DynamicFilterPushdown(ClientContext &context, const MultiF
...
+		for (auto &child : conj.children) {
+			if (child->type == ExpressionType::CONJUNCTION_AND) {
+				return true; //! (AND) OR (AND) pattern
+			}
+		}
+		return false;
+	}
+	case ExpressionType::OPERATOR_NOT:
+		//! NOT expressions wrapping prunable expressions
+		return true;
```
Tmonster: Here we say we should evaluate them, but [here](https://github.com/duckdb/duckdb-iceberg/pull/638/files#diff-4fd79172dd60a414106a77f2e76aa4e4307550c2dfc9723c9ae24ccdf075b222R674) you early out, and say we cannot prune. Is there still reasoning behind this?

Tmonster (changes requested):
So I have looked at this and stepped through it myself, and am a bit confused with what the original problem is, and how this PR is trying to fix it.

One thing I do I agree on, is that we can add a check for `TableFilterType::CONJUNCTION_OR` on line 158 of `src/iceberg_predicate.cpp`. 

The rest though I'm unsure about. With just the `TableFilterType::CONJUNCTION_OR` addition, you test file passes all tests. If you add tests for `suffix/starts_with` functions, then I think some tests will start failing, and we can look into that later.

Simple BETWEEN expressions are also handled with the present logic currently, the filter_combiner transforms them into `AND(greater_than_or_equal_to, less_than_or_equal_to)` table filter expressions, so the addition on line 222 of `iceberg_predicate` either isn't necessary, or it needs another test case that I'm not thinking of

I agree we should try to do more work to prune `TableFilterType::EXPRESSION_FILTER:`s, but I don't think we need an addition of `complex_filters` to do this, most filters can be generated as table_filters now.

Also, I have implemented writing upper and lower bounds to manifest files, so you can write more end-to-end tests in your PR. See https://github.com/duckdb/duckdb-iceberg/blob/main/test/sql/local/irc/insert/test_write_upper_and_lower_bounds.test which tests writing of upper and lower bounds to see what is supported

achille-roussel:
Thanks for the detailed feedback @Tmonster !

I'm still new to contributing to duckdb and it shows it the change I think, your guidance is extremely valuable.

Here's the angle that I was coming from: I have code that current uses a pre-processing step to query iceberg tables and push filters to partitions (parsing the SQL AST and converting this into predicate expressions), then use `read_parquet` on the resulting files. This code handles complex nested expression to get the right set of data files to run the queries against. I would like to port this code to use duckdb-iceberg instead, but it appeared that the current state of partition filtering in the extension was less advanced, hence what I tried to address with this change.

I'm going to digest your feedback and will push an update, thanks again!

PR #640, opened 2026-01-02
[Feature] Write upper_bounds, lower_bounds and null_value counts for multiple data types to manifest files

Tmonster:
### Title
Write column stats (upper_bounds, lower_bounds, null_value_counts) to Iceberg manifests on INSERT

### Summary
This PR extends the Duckdb-Iceberg::Insert writer to write upper_bounds, lower_bounds, and null_value counts to manifest entries when inserting data.

These statistics are stored in the manifest file(s) following the Iceberg spec, improving predicate pushdown, data skipping, and planner pruning.

### Supported Types (current)
- Numeric primitives: `TINYINT/SMALLINT/INT/BIGINT`, `FLOAT/DOUBLE`, `DECIMAL`
- Temporal primitives: `DATE`, `TIME`, `TIMESTAMP` (without/with tz as represented in our writer)
- String-like primitives: `VARCHAR`, `BLOB` where applicable for bounds
- BOOL and UUID must still be implemented.
- `null_value_counts` are tracked for all types

### Not Yet Supported / Future Work
- Boolean bounds:
  - Add explicit `lower_bounds`/`upper_bounds` for `BOOLEAN`. While Iceberg allows min/max on booleans, DuckDB doesn't support them when returning stats yet. The plan is to include them eventually.
- Nested types (Lists, Structs, Maps):
  - writing upper and lower bounds for nested types is not supported in the iceberg specification
- NaN counts and distinct counts:
  - Optionally add `nan_value_counts` for floating-point types and consider additional summaries if beneficial.

### DIFF
--- a/.github/workflows/test-nessie-catalog.yml
+++ b/.github/workflows/test-nessie-catalog.yml
@@ -11,6 +11,7 @@ jobs:
   test-nessie:
     name: Test against Nessie Catalog
     runs-on: ubuntu-latest
+    if: false # skip for now until nessie docker image can work again on github runners
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/test-polaris-catalog.yml
+++ b/.github/workflows/test-polaris-catalog.yml
@@ -11,7 +11,6 @@ jobs:
   test-polaris:
     name: Test against Polaris Catalog
     runs-on: ubuntu-latest
-    if: false # skip for now until nessie docker image can work again on github runners
 
     steps:
       - name: Checkout repository


Tmonster:
Nessie catalog failure unrelated

src/iceberg_manifest.cpp
```diff
@@ -51,8 +52,21 @@ Value IcebergManifestEntry::ToDataFileStruct(const LogicalType &type) const {
 	for (auto &child : upper_bounds) {
 		upper_bounds_values.push_back(Value::STRUCT({{"key", child.first}, {"value", child.second}}));
 	}
+
 	children.push_back(Value::MAP(LogicalType::STRUCT(bounds_types), upper_bounds_values));
 
+	child_list_t<LogicalType> null_value_count_types;
+	null_value_count_types.emplace_back("key", LogicalType::INTEGER);
+	null_value_count_types.emplace_back("value", LogicalType::BIGINT);
```
Tishj, line 60: `LogicalType::BLOB` ?
Ah, no, nvm, the comment is just wrong:
```
// upper bounds: map<129: int, 130: binary> - 128
```
should be:
```
// null_value_counts: map<121: int, 122: long> - 110
```

src/iceberg_manifest.cpp
```diff
@@ -335,6 +349,47 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
...
 	}
 
+	// null_value_counts_struct
+	// lower bounds struct
+	child_list_t<LogicalType> null_value_counts_fields;
+	null_value_counts_fields.emplace_back("key", LogicalType::INTEGER);
+	null_value_counts_fields.emplace_back("value", LogicalType::BIGINT);
+	{
+		// child_list_t<Value> null_value_counts_struct;
+		// upper bounds: map<121: int, 122: binary>
```
Tishj: `upper bounds` -> `null value counts`

src/iceberg_manifest.cpp
```diff
@@ -335,6 +349,47 @@ idx_t WriteToFile(IcebergTableInformation &table_info, const IcebergManifestFile
 		yyjson_mut_obj_add_uint(doc, val_obj, "id", UPPER_BOUNDS_VALUE);
 	}
 
+	// null_value_counts_struct
+	// lower bounds struct
+	child_list_t<LogicalType> null_value_counts_fields;
+	null_value_counts_fields.emplace_back("key", LogicalType::INTEGER);
+	null_value_counts_fields.emplace_back("value", LogicalType::BIGINT);
+	{
+		// child_list_t<Value> null_value_counts_struct;
```
Tishj: delete?

src/iceberg_value.cpp
```diff
@@ -244,4 +250,228 @@ DeserializeResult IcebergValue::DeserializeValue(const string_t &blob, const Log
...
+		truncated_length--;
+	}
+	if (truncated_length == 0 && original_input_size > 0) {
+		throw ConversionException("Could not write upper bounds for string column");
+	}
+	if (truncated_length == 0) {
+		bytes.resize(0);
+	}
+	// Convert back to string
+	return std::string(bytes.begin(), bytes.end());
```
Tishj, line 298: I'm confused, where does `truncated_length` end up being used?

src/iceberg_value.cpp
```diff
@@ -244,4 +250,228 @@ DeserializeResult IcebergValue::DeserializeValue(const string_t &blob, const Log
...
+		string val;
+		if (bound_type == SerializeBound::UPPER_BOUND) {
+			// if we are serializing upper bound, we must truncate and increment
+			val = IcebergValue::TruncateAndIncrementString(input_value.GetValue<string>());
+		} else {
+			// for lower bound truncating is enough
+			val = IcebergValue::TruncateString(input_value.GetValue<string>());
+		}
+		// create blob value of int32
+		// const_data_ptr_t string_data = reinterpret_cast<const_data_ptr_t>(&val);
```
Tishj: Dead code

src/iceberg_value.cpp
```diff
@@ -244,4 +250,228 @@ DeserializeResult IcebergValue::DeserializeValue(const string_t &blob, const Log
...
+		uint8_t byte = std::stoi(hex.substr(i, 2), nullptr, 16);
+		bytes.push_back(byte);
+	}
+	return bytes;
+}
+
+SerializeResult IcebergValue::SerializeValue(Value input_value, LogicalType &column_type, SerializeBound bound_type) {
+	switch (column_type.id()) {
+	case LogicalTypeId::INTEGER: {
+		// get const data ptr for the string value
```
Tishj: Comment copy pasting oversights

src/iceberg_value.cpp
```diff
@@ -244,4 +250,228 @@ DeserializeResult IcebergValue::DeserializeValue(const string_t &blob, const Log
...
+		auto serialized_const_data_ptr = const_data_ptr_cast<int32_t>(&epoch_days);
+		// create blob value of int32
+		auto serialized_val = Value::BLOB(serialized_const_data_ptr, sizeof(int32_t));
+		auto ret = SerializeResult(column_type, serialized_val);
+		return ret;
+	}
+	case LogicalTypeId::TIMESTAMP: {
+		// get const data ptr for the string value
+		timestamp_t val = input_value.GetValue<timestamp_t>();
+		int64_t micros_since_epoch = Timestamp::GetEpochMicroSeconds(val);
```
Tishj: Does this work for infinities?

src/iceberg_value.cpp
```diff
@@ -244,4 +250,228 @@ DeserializeResult IcebergValue::DeserializeValue(const string_t &blob, const Log
...
+
+		auto serialized_const_data_ptr = const_data_ptr_cast<hugeint_t>(&result);
+		// create blob value of int32, using only big_endian_bytes.size() bytes
+		auto ret_val = Value::BLOB(serialized_const_data_ptr, big_endian_bytes.size());
+		auto ret = SerializeResult(column_type, ret_val);
+		return ret;
+	}
+	case LogicalTypeId::BLOB:
+	case LogicalTypeId::BIT: {
+		// get const data ptr for the string value
```
Tishj, line 460: This is leaking the internal bitstring representation, we don't support BIT in Iceberg, right? We can just take it out

src/include/storage/iceberg_insert.hpp
```diff
@@ -37,6 +37,41 @@ class IcebergInsertGlobalState : public GlobalSinkState {
...
+	idx_t column_size_bytes = 0;
+	bool contains_nan = false;
+	bool has_null_count = false;
+	bool has_min = false;
+	bool has_max = false;
+	bool any_valid = true;
+	bool has_contains_nan = false;
+	bool has_column_size_bytes = false;
+
+	// unique_ptr<DuckLakeColumnExtraStats> extra_stats;
```
Tishj: Dead code

src/iceberg_value.cpp
```diff
@@ -1,7 +1,13 @@
 #include "iceberg_value.hpp"
+#include "utf8proc_wrapper.hpp"
 #include "duckdb/common/helper.hpp"
 #include "duckdb/common/types/uuid.hpp"
 #include "duckdb/common/bswap.hpp"
+#include "duckdb/common/exception/conversion_exception.hpp"
+#include "duckdb/common/types/date.hpp"
+#include "duckdb/common/types/timestamp.hpp"
+#include "duckdb/common/types/value.hpp"
+#include "mbedtls/ecp.h"
```
Tishj: Is this necessary?

src/metadata/iceberg_table_schema.cpp
```diff
@@ -135,4 +136,14 @@ void IcebergTableSchema::SchemaToJson(yyjson_mut_doc *doc, yyjson_mut_val *root_
 	yyjson_mut_obj_add_arr(doc, root_object, "identifier-field-ids");
 }
 
+LogicalType IcebergTableSchema::GetColumnTypeFromFieldId(idx_t field_id) const {
+	for (auto &column : columns) {
+		if (column->id == field_id) {
+			return column->type;
+		}
+	}
+	throw InvalidInputException("GetColumnTypeFromFieldId:: field id %dd does not exist in schema with id %d", field_id,
```
Tishj: %dd

src/metadata/iceberg_table_schema.cpp
```diff
@@ -135,4 +136,14 @@ void IcebergTableSchema::SchemaToJson(yyjson_mut_doc *doc, yyjson_mut_val *root_
 	yyjson_mut_obj_add_arr(doc, root_object, "identifier-field-ids");
 }
 
+LogicalType IcebergTableSchema::GetColumnTypeFromFieldId(idx_t field_id) const {
```
Tishj: This can be a `const &`

src/storage/iceberg_insert.cpp
```diff
@@ -99,75 +100,61 @@ static vector<string> ParseQuotedList(const string &input, char list_separator)
...
+			column_stats.column_size_bytes = StringUtil::ToUnsigned(StringValue::Get(stats_children[1]));
 		} else if (stats_name == "has_nan") {
 			column_stats.has_contains_nan = true;
-			column_stats.contains_nan = stats_value == "true";
-		} else {
-			throw NotImplementedException("Unsupported stats type \"%s\" in IcebergInsert::Sink()", stats_name);
+			column_stats.contains_nan = StringValue::Get(stats_children[1]) == "true";
 		}
+		// Ignore other stats types.
+		// TODO: Log that we did not write stats
```
Tishj: Unresolved todo

src/iceberg_value.cpp
Tmonster: `bytes.resize(truncated_length);`, so the length of bytes is `truncated_length`

src/iceberg_value.cpp
Tmonster: `scripts/data_generators/tests/spark_written_upper_lower_bounds/q01.sql`
writes is the max spark can support I think.
`test/sql/local/irc/insert/test_write_upper_and_lower_bounds.test` we test if we have the same upper and lower bound value.

I also added a test where we explicity write `infinity` and `-infinity` and check the deserialized results with `iceberg_column_stats()`

Tmonster (commented):
cleaned up quite a bit

Tmonster:
@Tishj good looking out on those infinities, pyicberg and spark don't support reading them. They throw an error that there is an overflow for their internal representation of date/timestamp/timestamptz. 
I have added a test to throw if we ever write infinity to those types

src/iceberg_value.cpp
```diff
@@ -244,4 +249,230 @@ DeserializeResult IcebergValue::DeserializeValue(const string_t &blob, const Log
...
+		// create blob value of int32
+		auto serialized_val = Value::BLOB(serialized_const_data_ptr, sizeof(int64_t));
+		auto ret = SerializeResult(column_type, serialized_val);
+		return ret;
+	}
+	case LogicalTypeId::VARCHAR: {
+		string val;
+		if (bound_type == SerializeBound::UPPER_BOUND) {
+			// if we are serializing upper bound, we must truncate and increment
+			val = IcebergValue::TruncateAndIncrementString(input_value.GetValue<string>());
```
Tishj, line 335: Just for my understanding: we only have to increment because we truncated right?
If we didn't end up truncating, the increment isn't necessary

src/iceberg_value.cpp
```diff
@@ -244,4 +249,230 @@ DeserializeResult IcebergValue::DeserializeValue(const string_t &blob, const Log
...
+		}
+		// Get the timestamp component (microseconds since epoch in UTC)
+		int64_t micros_since_epoch = val.value;
+		auto serialized_const_data_ptr = const_data_ptr_cast<int64_t>(&micros_since_epoch);
+		// create blob value of int32
+		auto serialized_val = Value::BLOB(serialized_const_data_ptr, sizeof(int64_t));
+		auto ret = SerializeResult(column_type, serialized_val);
+		return ret;
+	}
+	case LogicalTypeId::DECIMAL: {
```
Tishj, line 397: I have to admit I'm not entirely following what's going on here, all I know is that the decoding was also hell
One thing I did want to mention is that the spec says:
> Stores unscaled value as two’s-complement big-endian binary, using the minimum number of bytes for the value

It looks at first glance like we're always storing as `hugeint_t` ? That is probably incorrect and I think there could be implementations that would trip up on that

src/iceberg_value.cpp
Tmonster: yea, if the string length is <= 16, we don't truncate.

src/iceberg_value.cpp
Tmonster: Yea, we first store it in a hugeint_t, but then we push the bytes we need into `big_endian_bytes`. Then the blob we create only has size `big_endian_bytes`. I've also done comparisons with how spark stores DECIMAL values and I haven't seen any differences in upper/lower bound values

PR #643, opened 2026-01-07
Table changes live in transaction. Catalog maintains SchemaSet knowledge

Tmonster:
This PR modifies where transaction updates live for iceberg tables and where the entries for the iceberg schemas & tables live.

### PR Summary
Currently all information for iceberg tables/schemas live in the `IRCTransaction`. This is undesireable since we would like to keep some knowledge of the catalog (like what table exist() around between transactions so that we do not have to request the latest table state over and over again. The simplest way to do this is would be to move the `IRCSchemaSet` to the `IRCatalog`. One problem with that, however, is that all transaction data for a table (i.e schema updates/snapshot updates) live in `ICTableEntry.table_info.transaction_info`. This means if we have two transactions both making updates to a table, they will both see updates from both transactions.

To fix this, I am still moving the `IRCSchemaSet` to the `IRCatalog`, but each `IRCTransaction` now has a set of `updated_tables`, `deleted_tables`, and `looked_up_tables`

- `updated_tables` stores tables that have new snapshots/shema updates etc. When a table gets updated, all TableInformation is copied to `updated_tables` and the updates are then reflected on table metadata in the updated tables. This will allow you to update a schema and add a snapshot in one transaction, and DuckDB-Iceberg will only issue one post request to update the table.*
- `deleted_tables` stores tables that have been deleted in a transaction, if you delete a table in a transaction and select, you will get a `CatalogException`
- `looked_up_tables` stores the latest snapshot information for a table. In every transaction, we request the latest table state when we select from a table, but if it is a long running transaction, we do not re-request the latest state. This will help to maintain the current feature where requests in a transaction are not duplicated.

Whenever a table is requested, first the IRCTransaction is checked to see if the table has been updated, deleted, or requested already. If so, we return the table entry in the transaction. Otherwise we request the latest version, record the version in transaction, and update the catalog.

Notes:
Many tests are updated to do some clean up before running. Other tests are updated because request counts have changed. Namely, since we now store information between queries, we save ourselves a trip to verify if a schema exists. Schema existence is checked once per session, and assumed to exist for the rest of the session unless someone deletes it.

There is also a lot of code to try an separate `IcebergTableInformation` and `IcebergTableMetadata`. Access to table metadata should only happen through the table_information.

### TODO:
* Ideally, when a transaction commits and there are updates, only 1 request is issued to the IRC. If it fails, the IRC transaction fails. It is possible to update a table, and delete another table. This needs to be done with 2 requests. I have left this as TODO work, since this PR is quiet big as it is.

### Future work.
Since we now store table information between transactions, part 2 of this PR will be to add a `MAX_TABLE_STALENESS` to the attach parameters, which we can use to not look up table every time we access a table. Some work was already done to support this, the `MetadataCacheValue` now stores a unique_ptr to a `rest_api_objects::LoadTableResult`, which can be returned instead of fetching from the IRC. This `MetadataCacheValue` is also used to create temporary secrets when reading/writing from/to storage.

### DIFF
--- a/src/storage/iceberg_table_information.cpp
+++ b/src/storage/iceberg_table_information.cpp
@@ -279,14 +279,6 @@ string IcebergTableInformation::GetTableKey() const {
 	return GetTableKey(schema.namespace_items, name);
 }
 
-IcebergTableInformation IcebergTableInformation::Copy() const {
-	auto ret = IcebergTableInformation(catalog, schema, name);
-	auto table_key = ret.GetTableKey();
-	auto &cached_load_table_result = catalog.GetLoadTableResult(table_key);
-	ret.table_metadata = IcebergTableMetadata::FromTableMetadata(cached_load_table_result.load_table_result->metadata);
-	return ret;
-}
-
 IcebergSnapshotLookup IcebergTableInformation::GetSnapshotLookup(IRCTransaction &irc_transaction) const {
 	auto &context = *irc_transaction.context.lock();
 	return GetSnapshotLookup(context);
@@ -322,12 +314,17 @@ bool IcebergTableInformation::HasTransactionUpdates() {
 	return transaction_data && (!transaction_data->updates.empty() || !transaction_data->requirements.empty());
 }
 
-IcebergTableInformation IcebergTableInformation::Copy(IRCTransaction &irc_transaction) const {
+IcebergTableInformation IcebergTableInformation::Copy() const {
 	auto ret = IcebergTableInformation(catalog, schema, name);
 	auto table_key = ret.GetTableKey();
 	auto &cached_load_table_result = catalog.GetLoadTableResult(table_key);
 	ret.table_metadata = IcebergTableMetadata::FromTableMetadata(cached_load_table_result.load_table_result->metadata);
+	ret.table_metadata.latest_metadata_json = cached_load_table_result.load_table_result->metadata_location;
+	return ret;
+}
 
+IcebergTableInformation IcebergTableInformation::Copy(IRCTransaction &irc_transaction) const {
+	auto ret = Copy();
 	// get snapshot from start of transaction
 	// latest_snapshot_id and sequence of copied table information should be asof the transaction start
 	// this is to ensure when the transaction commits, the assert ref snapshot id is the one closest to the start of
--- a/src/storage/iceberg_transaction_data.cpp
+++ b/src/storage/iceberg_transaction_data.cpp
@@ -104,7 +104,6 @@ void IcebergTransactionData::AddSnapshot(IcebergSnapshotOperationType operation,
 	new_snapshot.manifest_list = manifest_list_path;
 	new_snapshot.operation = operation;
 	new_snapshot.timestamp_ms = Timestamp::GetEpochMs(Timestamp::GetCurrentTimestamp());
-
 	new_snapshot.has_parent_snapshot = table_info.table_metadata.has_current_snapshot || !alters.empty();
 	if (new_snapshot.has_parent_snapshot) {
 		if (!alters.empty()) {
@@ -131,6 +130,8 @@ void IcebergTransactionData::AddSnapshot(IcebergSnapshotOperationType operation,
 	CreateManifestListEntry(*add_snapshot, table_metadata, manifest_content_type, std::move(data_files));
 	alters.push_back(*add_snapshot);
 	updates.push_back(std::move(add_snapshot));
+
+	// modify the current table metadata to also have the new snapshot
 }
 
 void IcebergTransactionData::AddUpdateSnapshot(vector<IcebergManifestEntry> &&delete_files,
--- a/test/sql/local/irc/transactions/test_get_transaction_snapshot.test
+++ b/test/sql/local/irc/transactions/test_get_transaction_snapshot.test
@@ -1,4 +1,4 @@
-# name: test/sql/local/irc/transactions/test_transactional_updates.test
+# name: test/sql/local/irc/transactions/test_get_transaction_snapshot.test
 # group: [transactions]
 
 require-env ICEBERG_SERVER_AVAILABLE
@@ -41,38 +41,32 @@ ATTACH '' AS my_datalake (
 
 # Perform an INSERT and COMMIT it so table has a snapshot id other transactions can find
 statement ok
-drop table if exists my_datalake.default.multi_transaction_update;
+drop table if exists my_datalake.default.get_transaction_snapshot;
 
 statement ok
-create table my_datalake.default.multi_transaction_update (a int, b int);
+create table my_datalake.default.get_transaction_snapshot (a int, b int);
 
 # make sure the table has a snapshot id to start
 statement ok
-insert into my_datalake.default.multi_transaction_update select 0, 0;
+insert into my_datalake.default.get_transaction_snapshot select 0, 0;
 
 statement ok con1
 begin
 
-statement ok con2
-begin
+statement ok con1
+insert into my_datalake.default.get_transaction_snapshot select 1, 1;
 
-statement ok con2
-insert into my_datalake.default.multi_transaction_update select 2, 2;
-
-statement ok con2
-commit
-
-# con1 should only see data from con1 and before (from transaction start of con1)
-query II con1
-select * from my_datalake.default.multi_transaction_update;
+# we will not show committed local snapshots for (iceberg_snapshots)
+query I con1
+select count(*) from iceberg_snapshots(my_datalake.default.get_transaction_snapshot);
 ----
-0	0
+1
 
 statement ok con1
-update my_datalake.default.multi_transaction_update set a = 50;
-
-# con1 should error. con1 cannot update the table since the last snapshot it saw was the 1, 1 insert.
-statement error con1
 commit;
+
+query I
+select count(*) from iceberg_snapshots(my_datalake.default.get_transaction_snapshot);
 ----
-<REGEX>:.*TransactionContext Error:.*Conflict.*
+2
+
--- a/test/sql/local/irc/transactions/test_transaction_get_from_transaction_start.test
+++ b/test/sql/local/irc/transactions/test_transaction_get_from_transaction_start.test
@@ -39,9 +39,6 @@ ATTACH '' AS my_datalake (
     ENDPOINT 'http://127.0.0.1:8181'
 );
 
-statement ok
-set threads=1;
-
 # Perform an INSERT and COMMIT it
 statement ok
 drop table if exists my_datalake.default.multi_transaction_read;
--- a/test/sql/local/irc/transactions/test_transactions_reading_empty_tables.test
+++ b/test/sql/local/irc/transactions/test_transactions_reading_empty_tables.test
@@ -44,9 +44,6 @@ ATTACH '' AS my_datalake (
     ENDPOINT 'http://127.0.0.1:8181'
 );
 
-statement ok
-set threads=1;
-
 # Perform an INSERT and COMMIT it
 statement ok
 drop table if exists my_datalake.default.multi_transaction_read_empty;


src/iceberg_functions/iceberg_table_properties_functions.cpp
```diff
@@ -191,17 +192,18 @@ static void RemoveIcebergTablePropertiesFunction(ClientContext &context, TableFu
 	}
 
 	auto iceberg_table = bind_data.iceberg_table;
+	auto &table_info = iceberg_table->table_info;
 	auto &irc_transaction = IRCTransaction::Get(context, iceberg_table->catalog);
-	if (!iceberg_table->table_info.transaction_data) {
-		iceberg_table->table_info.transaction_data =
-		    make_uniq<IcebergTransactionData>(context, iceberg_table->table_info);
+	irc_transaction.updated_tables.emplace(table_info.GetTableKey(), table_info.Copy());
```
Tishj: This should do a `find(table_info.GetTableKey())` and only `table_info.Copy()` if it doesn't exist
also the `at` below is not necessary, `emplace` returns an iterator, we can use that directly

src/include/storage/irc_schema_set.hpp
```diff
@@ -14,17 +18,21 @@ class IRCSchemaSet {
...
 
 public:
 	Catalog &catalog;
-	case_insensitive_map_t<unique_ptr<CatalogEntry>> entries;
-	//! Whether a listing has been done for the catalog
-	bool listed = false;
 
 private:
+	//! Whether a listing has been done for the catalog
+	bool listed = false;
```
Tishj: This can't be in the catalog entry anymore, right?
Or it should be a timestamp perhaps, to indicate what the latest listed state is

Tmonster: True, I've moved the bool to the transaction. Eventually we can make it a timestamp, but that would require another layer of indirection to keep track of what time tables were first made available. This is made extra difficult because not even the Iceberg Rest catalog supports that functionality without requesting table information for every table.

src/metadata/iceberg_table_metadata.cpp
```diff
@@ -52,6 +53,14 @@ optional_ptr<const IcebergSortOrder> IcebergTableMetadata::FindSortOrderById(int
 	return it->second;
 }
 
+const unordered_map<int32_t, IcebergSortOrder> IcebergTableMetadata::GetSortOrderSpecs() const {
```
Tishj: Was this meant to be a reference perhaps?

src/include/storage/irc_schema_set.hpp
```diff
@@ -14,17 +18,19 @@ class IRCSchemaSet {
...
 	optional_ptr<CatalogEntry> CreateEntryInternal(ClientContext &context, unique_ptr<CatalogEntry> entry);
 
 public:
 	Catalog &catalog;
-	case_insensitive_map_t<unique_ptr<CatalogEntry>> entries;
-	//! Whether a listing has been done for the catalog
-	bool listed = false;
 
 private:
+	//! Whether a listing has been done for the catalog
```
Tishj: Remove comment

src/include/storage/irc_table_entry.hpp
```diff
@@ -50,4 +50,13 @@ class ICTableEntry : public TableCatalogEntry {
 	IcebergTableInformation &table_info;
 };
 
+struct ICTableEntryHashFunction {
+	uint64_t operator()(const optional_ptr<ICTableEntry> &entry) const {
+		D_ASSERT(entry);
+		// FIXME: we shuold use a table uuid in case renaming tables to the same name happens
```
Tishj: Should we not address this fixme? the `table_id` of the `IcebergTableInformation` is accessible here

src/metadata/iceberg_table_metadata.cpp
```diff
@@ -52,6 +53,14 @@ optional_ptr<const IcebergSortOrder> IcebergTableMetadata::FindSortOrderById(int
...
+	return sort_specs;
+}
+
+const unordered_map<int32_t, IcebergPartitionSpec> IcebergTableMetadata::GetPartitionSpecs() const {
```
Tishj: Same here

src/metadata/iceberg_table_metadata.cpp
```diff
@@ -321,13 +350,25 @@ IcebergTableMetadata IcebergTableMetadata::FromTableMetadata(rest_api_objects::T
...
+	}
+
 	return res;
 }
 
 const case_insensitive_map_t<string> &IcebergTableMetadata::GetTableProperties() const {
 	return table_properties;
 }
 
+string IcebergTableMetadata::GetLatestMetadataJson() const {
```
Tishj: `const string &`

src/metadata/iceberg_table_metadata.cpp
```diff
@@ -321,13 +350,25 @@ IcebergTableMetadata IcebergTableMetadata::FromTableMetadata(rest_api_objects::T
...
+	return latest_metadata_json;
+}
+
+string IcebergTableMetadata::GetLocation() const {
```
Tishj: `const string &`

src/storage/iceberg_insert.cpp
```diff
@@ -281,11 +280,19 @@ SinkFinalizeType IcebergInsert::Finalize(Pipeline &pipeline, Event &event, Clien
 	auto &irc_table = table->Cast<ICTableEntry>();
 	auto &table_info = irc_table.table_info;
 	auto &transaction = IRCTransaction::Get(context, table->catalog);
+	auto &irc_transaction = transaction.Cast<IRCTransaction>();
 
 	lock_guard<mutex> guard(global_state.lock);
 	if (!global_state.written_files.empty()) {
-		table_info.AddSnapshot(transaction, std::move(global_state.written_files));
-		transaction.MarkTableAsDirty(irc_table);
+		if (table_info.IsTransactionLocalTable(irc_transaction)) {
```
Tishj: This looks the same as the logic in `iceberg_delete.cpp` (aside from the different type of update we're making), can't we deduplicate this logic somehow?

src/storage/iceberg_delete.cpp
```diff
@@ -274,8 +274,16 @@ SinkFinalizeType IcebergDelete::Finalize(Pipeline &pipeline, Event &event, Clien
 	auto &table_info = irc_table.table_info;
 	auto &transaction = IRCTransaction::Get(context, table.catalog);
 	auto iceberg_delete_files = GenerateDeleteManifestEntries(global_state);
-	table_info.AddDeleteSnapshot(transaction, std::move(iceberg_delete_files));
-	transaction.MarkTableAsDirty(irc_table);
+	if (!global_state.written_files.empty()) {
+		if (table_info.IsTransactionLocalTable(irc_transaction)) {
+			table_info.AddDeleteSnapshot(transaction, std::move(iceberg_delete_files));
+		} else {
+			irc_transaction.updated_tables.emplace(table_info.GetTableKey(), table_info.Copy());
```
Tishj: Same comment as before, we can check if the table doesn't already exist in `updated_tables` (with `find`) and only then add it

also `emplace` returns an iterator, making the `at` below unnecessary

src/storage/iceberg_update.cpp
```diff
@@ -177,9 +177,16 @@ SinkFinalizeType IcebergUpdate::Finalize(Pipeline &pipeline, Event &event, Clien
 	auto delete_manifest_entries = IcebergDelete::GenerateDeleteManifestEntries(delete_global_state);
 	auto insert_manifest_entries = IcebergInsert::GetInsertManifestEntries(insert_global_state);
 	if (!insert_manifest_entries.empty()) {
-		table_info.AddUpdateSnapshot(irc_transaction, std::move(delete_manifest_entries),
-		                             std::move(insert_manifest_entries));
-		irc_transaction.MarkTableAsDirty(irc_table);
+		if (table_info.IsTransactionLocalTable(irc_transaction)) {
+			table_info.AddUpdateSnapshot(irc_transaction, std::move(delete_manifest_entries),
```
Tishj: Again the same logic as in `iceberg_delete.cpp` and `iceberg_insert.cpp`
I think this can be callback based instead? where the callback receives a `IcebergTableInformation &` on which we can then call `AddUpdateSnapshot(...)` or `AddSnapshot(...)` or `AddDeleteSnapshot(...)`

src/storage/irc_catalog.cpp
```diff
@@ -66,13 +62,44 @@ optional_ptr<SchemaCatalogEntry> IRCatalog::LookupSchema(CatalogTransaction tran
...
+	// erase load table result if it exists.
+	if (metadata_cache.find(table_key) != metadata_cache.end()) {
+		metadata_cache.erase(table_key);
+	}
+	auto now = system_clock::now() + std::chrono::minutes(10);
+	auto val = make_uniq<MetadataCacheValue>(now, std::move(load_table_result));
+	metadata_cache.emplace(table_key, std::move(val));
+}
+
+MetadataCacheValue &IRCatalog::GetLoadTableResult(string table_key) {
```
Tishj: `const string &`

src/storage/irc_catalog.cpp
```diff
@@ -66,13 +62,44 @@ optional_ptr<SchemaCatalogEntry> IRCatalog::LookupSchema(CatalogTransaction tran
 	return reinterpret_cast<SchemaCatalogEntry *>(entry.get());
 }
 
+void IRCatalog::StoreLoadTableResult(string table_key,
```
Tishj: `const string &`

src/storage/irc_schema_entry.cpp
```diff
@@ -30,64 +34,87 @@ IRCTransaction &GetICTransaction(CatalogTransaction transaction) {
...
-	// get the entry from the catalog.
-	D_ASSERT(entry);
-	D_ASSERT(entry->type == CatalogType::TABLE_ENTRY);
 
-	return entry;
+	if (!ICTableSet::CreateNewEntry(context, ir_catalog, *this, base_info)) {
+		throw InternalException("We should not be here");
+	}
+	auto table_key = IcebergTableInformation::GetTableKey(namespace_items, base_info.table);
+	auto entry = irc_transaction.updated_tables.find(table_key);
```
Tishj, line 91: Don't we need to check if `entry != irc_transaction.updated_tables.end()` ?
Or at least assert it if we're confident that it will always be there

src/storage/irc_schema_entry.cpp
```diff
@@ -30,64 +34,87 @@ IRCTransaction &GetICTransaction(CatalogTransaction transaction) {
...
-	auto &ic_entry = table_entry_actual->Cast<ICTableEntry>();
-	D_ASSERT(table_entry_actual);
-	if (!table_entry.transaction_data) {
-		table_entry.transaction_data = make_uniq<IcebergTransactionData>(context, table_entry);
-	}
-	table_entry.transaction_data->is_deleted = true;
-	transaction.MarkTableAsDeleted(ic_entry);
+	auto &table_info = table_info_it->second;
+	auto table_key = table_info.GetTableKey();
+	transaction.deleted_tables.emplace(table_key, table_info.Copy());
```
Tishj, line 114: Same pattern as before

src/storage/irc_schema_set.cpp
```diff
@@ -57,16 +56,35 @@ void IRCSchemaSet::Scan(ClientContext &context, const std::function<void(Catalog
 	}
 }
 
+void IRCSchemaSet::AddEntry(string name, unique_ptr<IRCSchemaEntry> entry) {
```
Tishj: `const string &`

src/storage/irc_schema_set.cpp
```diff
@@ -57,16 +56,35 @@ void IRCSchemaSet::Scan(ClientContext &context, const std::function<void(Catalog
...
+
 static string GetSchemaName(const vector<string> &items) {
 	return StringUtil::Join(items, ".");
 }
 
 void IRCSchemaSet::LoadEntries(ClientContext &context) {
-	if (listed) {
+	auto &ic_catalog = catalog.Cast<IRCatalog>();
+	auto &irc_transaction = IRCTransaction::Get(context, catalog);
+	// auto &irc_transaction = transaction.Cast<IRCTransaction>();
```
Tishj: dead code?

src/storage/irc_table_set.cpp
```diff
@@ -93,87 +98,69 @@ void ICTableSet::Scan(ClientContext &context, const std::function<void(CatalogEn
...
+}
+
+case_insensitive_map_t<IcebergTableInformation> &ICTableSet::GetEntriesMutable() {
+	return entries;
+}
+
 void ICTableSet::LoadEntries(ClientContext &context) {
-	if (listed) {
+	auto &irc_transaction = IRCTransaction::Get(context, catalog);
+	// auto &irc_transaction = transaction.Cast<IRCTransaction>();
```
Tishj: dead code?

src/storage/irc_table_set.cpp
```diff
@@ -93,87 +98,69 @@ void ICTableSet::Scan(ClientContext &context, const std::function<void(CatalogEn
...
-		default:
-			throw InternalException("Unknown conflict state when creating a table");
-		}
-	}
-
-	entries.emplace(table_name, IcebergTableInformation(catalog, schema, info.table));
-	auto &table_info = entries.find(table_name)->second;
 
+	auto key = IcebergTableInformation::GetTableKey(schema.namespace_items, info.table);
+	irc_transaction.updated_tables.emplace(key, IcebergTableInformation(catalog, schema, info.table));
```
Tishj, line 129: Same pattern as before

src/storage/irc_table_set.cpp
```diff
@@ -185,21 +172,81 @@ unique_ptr<ICTableInfo> ICTableSet::GetTableInfo(ClientContext &context, IRCSche
...
+	auto &irc_transaction = IRCTransaction::Get(context, catalog);
 	auto table_name = lookup.GetEntryName();
-	auto entry = entries.find(table_name);
-	if (entry == entries.end()) {
-		if (!IRCAPI::VerifyTableExistence(context, ic_catalog, schema, table_name)) {
-			return nullptr;
+	// first check transaction entries
+	auto table_key = IcebergTableInformation::GetTableKey(schema.namespace_items, table_name);
+	// Check if table has been deleted within in the transaction.
+	auto deleted_table_entry = irc_transaction.deleted_tables.find(table_key);
```
Tishj: Nit: if we're not interested in the iterator, we can just use `count`:
`if (irc_transaction.deleted_tables.count(table_key)) { return nullptr; }`

src/storage/irc_table_set.cpp
```diff
@@ -185,21 +172,81 @@ unique_ptr<ICTableInfo> ICTableSet::GetTableInfo(ClientContext &context, IRCSche
...
+	auto transaction_entry = irc_transaction.updated_tables.find(table_key);
+	if (transaction_entry != irc_transaction.updated_tables.end()) {
+		return transaction_entry->second.GetSchemaVersion(lookup.GetAtClause());
+	}
+	// Check regular catalog Entries
+	auto previous_requested_snapshot = irc_transaction.requested_tables.find(table_name);
+	if (previous_requested_snapshot != irc_transaction.requested_tables.end()) {
+		// transaction has already looked up this table, find it in entries
+		auto entry = entries.find(table_name);
+		if (entry != entries.end()) {
```
Tishj: early-return please:
`if (entry == entries.end()) { return nullptr; }`

src/storage/iceberg_delete.cpp
Tmonster: The check happens in `table_info.IsTransactionLocalTable(irc_transaction))`, so if the table exists in updated_tables, we won't be `emplace`ing it

src/storage/irc_schema_entry.cpp
Tmonster: If `ICTableSet::CreateNewEntry` returns false then it's not in `updated_tables`, but I will add a D_ASSERT just in case

src/storage/irc_schema_entry.cpp
Tmonster: have added a D_ASSERT below

PR #644, opened 2026-01-08
Apply patches from duckdb/duckdb

maiadegraaf:
This PR applies patches from duckdb/duckdb:
- `autoloading_tests.patch`


Tishj:
I don't really understand the FIXME comment, I assume this is not something to fix in core, since it wouldnt be a patch otherwise?

But what is required in iceberg then?

carlopi:
I wrote the patch, without that the autoloading test fail here, and correctly so since there is currently no way to pre-register some logging types to the extension that will declare them. I am not sure if we want to go that way.

Alternative would be just adding `require no_extension_autoloading` or similar.

PR #661, opened 2026-01-19
Fix Nessie Docker tests by lowering the client timeouts

Tmonster:
Also re-enable wasm builds


Tmonster:
@carlopi does this look good for re-enabling wasm builds? lmk, otherwise I will merge later today

PR #673, opened 2026-01-28
Omit stack trace on error response

yan-alex:
This omits a long Java stack trace which used to be visible on the response.
Also added quotes on the `pip install duckdb pyspark[sql]==3.5.0` command in the README, as in zsh shell, the `[` `]` characters need to either be escaped or surrounded by quotes.

### DIFF
--- a/src/rest_catalog/objects/error_model.cpp
+++ b/src/rest_catalog/objects/error_model.cpp
@@ -1,7 +1,6 @@
 
 #include "rest_catalog/objects/error_model.hpp"
 
-#include "catalog_utils.hpp"
 #include "yyjson.hpp"
 #include "duckdb/common/string.hpp"
 #include "duckdb/common/vector.hpp"
@@ -81,37 +80,6 @@ string ErrorModel::TryFromJSON(yyjson_val *obj) {
 		}
 	}
 	return string();
-}
-
-string ErrorModel::ToString() const {
-	unique_ptr<yyjson_mut_doc, YyjsonDocDeleter> doc(yyjson_mut_doc_new(nullptr));
-
-	auto error_obj = yyjson_mut_obj(doc.get());
-	yyjson_mut_obj_add_str(doc.get(), error_obj, "message", message.c_str());
-	yyjson_mut_obj_add_str(doc.get(), error_obj, "type", type.c_str());
-	yyjson_mut_obj_add_int(doc.get(), error_obj, "code", code);
-
-	if (!stack.empty()) {
-		auto stack_arr = yyjson_mut_arr(doc.get());
-		for (const auto &s : stack) {
-			yyjson_mut_arr_add_str(doc.get(), stack_arr, s.c_str());
-		}
-		yyjson_mut_obj_add_val(doc.get(), error_obj, "stack", stack_arr);
-	}
-
-	auto root = yyjson_mut_obj(doc.get());
-
-	yyjson_mut_obj_add_val(doc.get(), root, "error", error_obj);
-	yyjson_mut_doc_set_root(doc.get(), root);
-
-	yyjson_write_err err;
-	size_t len;
-	const string json_str = yyjson_mut_write_opts(doc.get(), YYJSON_WRITE_PRETTY, nullptr, &len, &err);
-	if (err.code != YYJSON_WRITE_SUCCESS) {
-		throw InternalException("ErrorModel::ToString() failed to write JSON: %s", err.msg);
-	}
-
-	return json_str;
 }
 
 } // namespace rest_api_objects
--- a/src/include/rest_catalog/objects/error_model.hpp
+++ b/src/include/rest_catalog/objects/error_model.hpp
@@ -26,9 +26,6 @@ public:
 	string TryFromJSON(yyjson_val *obj);
 
 public:
-	string ToString() const;
-
-public:
 	string message;
 	string type;
 	int32_t code;
--- a/src/include/rest_catalog/objects/iceberg_error_response.hpp
+++ b/src/include/rest_catalog/objects/iceberg_error_response.hpp
@@ -27,9 +27,6 @@ public:
 	string TryFromJSON(yyjson_val *obj);
 
 public:
-	string ToString() const;
-
-public:
 	ErrorModel _error;
 };
 
--- a/src/rest_catalog/objects/iceberg_error_response.cpp
+++ b/src/rest_catalog/objects/iceberg_error_response.cpp
@@ -38,9 +38,5 @@ string IcebergErrorResponse::TryFromJSON(yyjson_val *obj) {
 	return string();
 }
 
-string IcebergErrorResponse::ToString() const {
-	return _error.ToString();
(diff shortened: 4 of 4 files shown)


yan-alex:
@Tmonster While this is working, I currently have no tests. `test_transactional_updates.test` finds the error by doing
```sql
# con1 should error. con1 cannot update the table since the last snapshot it saw was the 1, 1 insert.
statement error con1
commit;
----
<REGEX>:.*TransactionContext Error:.*Conflict.*
```
I wanted to add a `<!REGEX>:.*"stack"` to this, but not sure how to do another check on this error result. I can also duplicate the whole test and change just this regex at the end, but that's not really clean. What do you suggest?

Tmonster:
I would suggest just copying the whole test. Not very clean, but I can't think of another way quickly

src/catalog_api.cpp
```diff
@@ -327,6 +328,7 @@ void IRCAPI::CommitMultiTableUpdate(ClientContext &context, IRCatalog &catalog,
 	headers.Insert("Content-Type", "application/json");
 	auto response = catalog.auth_handler->Request(RequestType::POST_REQUEST, context, url_builder, headers, body);
```
Tmonster, line 329: So we already have some logic to parse error messages from Iceberg. I was thinking you could put something like this in case of an error.
``
auto result = catalog.auth_handler->Request(RequestType::POST_REQUEST, context, url_builder, headers, body);
	if (result->status != HTTPStatusCode::OK_200 && result->status != HTTPStatusCode::NoContent_204) {
		yyjson_val *error_obj = ICUtils::get_error_message(result->body);
		if (error_obj == nullptr) {
			throw InvalidConfigurationException(result->body);
		}
		auto error = rest_api_objects::IcebergErrorResponse::FromJSON(error_obj);
        error._error.stack = vector<string>(); // empty vector or something
	}
```

What could also be nice, since you have the context, is to log via DUCKDB_LOG_INFO the stack trace. That way if users want to see the stack trace, they can still look at the errors

Comment (yan-alex) [reply src/catalog_api.cpp:329]: Thanks, implemented that. Also tried retrieving the INFO logging in the test by doing something like 
```sql
query I con1
SELECT * FROM duckdb_logs_parsed('Iceberg');
----
```
but just couldn't get it to work

Tmonster: did you try `duckdb_logs('Iceberg')`?

yan-alex: @Tmonster 
```sql
Binder Error: No function matches the given name and argument types 'duckdb_logs(VARCHAR)'. You might need to add explicit type casts.
	Candidate functions:
	duckdb_logs(denormalized_table : BOOLEAN)


LINE 1: SELECT * FROM duckdb_logs('Iceberg');
```

src/catalog_api.cpp
```diff
@@ -327,9 +328,19 @@ void IRCAPI::CommitMultiTableUpdate(ClientContext &context, IRCatalog &catalog,
...
-		throw InvalidConfigurationException(
-		    "Request to '%s' returned a non-200 status code (%s), with reason: %s, body: %s",
-		    url_builder.GetURLEncoded(), EnumUtil::ToString(response->status), response->reason, response->body);
+		yyjson_val *error_obj = ICUtils::get_error_message(response->body);
+		if (error_obj == nullptr) {
+			throw InvalidConfigurationException(response->body);
+		}
+		auto error = rest_api_objects::IcebergErrorResponse::FromJSON(error_obj);
+		for (const auto &str : error._error.stack) {
+			DUCKDB_LOG(context, IcebergLogType, str);
```
Tmonster: I don't think I would log the whole stack in separate logs. I would construct the whole stack and log that once.

src/rest_catalog/objects/error_model.cpp
```diff
@@ -83,5 +83,36 @@ string ErrorModel::TryFromJSON(yyjson_val *obj) {
 	return string();
 }
 
+string ErrorModel::ToString() const {
+	unique_ptr<yyjson_mut_doc, YyjsonDocDeleter> doc(yyjson_mut_doc_new(nullptr));
```
Tmonster: Why are we trying to write this back to JSON? I don't think this code is needed

src/catalog_api.cpp
Tmonster: i think I know what it is now 
```
select * from duckdb_log() where type like 'Iceberg';
```
`type` might be a different column name, but that should work

src/catalog_api.cpp
```diff
@@ -327,9 +328,19 @@ void IRCAPI::CommitMultiTableUpdate(ClientContext &context, IRCatalog &catalog,
...
+		}
+		// Omit stack from error output
+		error._error.stack = vector<string>();
+		throw InvalidConfigurationException("Request to '%s' returned a non-200 status code (%s), with reason: %s, body: %s",
+		                                    url_builder.GetURLEncoded(), EnumUtil::ToString(response->status),
+		                                    response->reason, error.ToString());
```
Tmonster: do here you should be able to just use `error._error.message` instead of `error.ToString()`

src/rest_catalog/objects/error_model.cpp
yan-alex: @Tmonster 
Well the previous functionality throws `InvalidConfigurationException()` , where `response->body` was passed, that way the error has this in it's message:
```json
{
  "error": {
    "message": "Requirement failed: branch main has changed: expected id 227932290685815360 != 1125114298610568469",
    "type": "CommitFailedException",
    "code": 409
  }
}
```
So the way I replicated this behaviour is to simply create a JSON string from ErrorModel. Not sure if you had something else in mind?

src/catalog_api.cpp
yan-alex: Right. Then `ErrorModel.type` will no longer be exposed though.

In this case I see it's fine because our error message already says enough:
```json
TransactionContext Error: Failed to commit: Failed to commit Iceberg transaction: Request to 'http://127.0.0.1:8181/v1/transactions/commit' returned a non-200 status code (Conflict_409), with reason: Conflict, body: {
    "error": {
        "message": "Requirement failed: branch main has changed: expected id 4653786732879561299 != 2933198401030603263",
        "type": "CommitFailedException",
        "code": 409
    }
}
```
So "type" is redundant, but I was wondering if it will be fine for other cases

src/catalog_api.cpp
yan-alex: That works!

src/catalog_api.cpp
Tmonster: You can also format it a bit nicer yourself if you want
```c++
throw InvalidConfigurationException("Request to '%s' returned a non-200 status code (%s). \n"
     "reason: %s\n" +
     "type: %s\n" + 
     "body: %s\n", 
	 url_builder.GetURLEncoded(), EnumUtil::ToString(response->status),
	response->reason, error._error.type, error._error.body);
```

PR #674, opened 2026-01-28
Add support for extra HTTP headers in IRCAuthorization

talatuyarer:
Adds support in DuckDB’s Iceberg extension to attach user-provided “extra headers” to all HTTP requests made to the REST catalog (including the initial /v1/config call and subsequent table/namespace operations).

This aligns DuckDB with how other Iceberg clients treat REST catalogs. Example usage:

```sql
CREATE SECRET iceberg_secret_single (
	TYPE ICEBERG,
	CLIENT_ID 'admin',
	CLIENT_SECRET 'password',
	OAUTH2_SERVER_URI 'http://127.0.0.1:8181/v1/oauth/tokens',
	EXTRA_HTTP_HEADERS MAP {'X-Custom-Header': 'custom-value'}
);
```

I also fixed a bug if catalog return prefix which has slashes in it. Duckdb encode those slashes which it should not encode.

### DIFF
--- a/src/catalog_api.cpp
+++ b/src/catalog_api.cpp
@@ -266,7 +266,7 @@ vector<IRCAPISchema> IRCAPI::GetSchemas(ClientContext &context, IcebergCatalog &
 	string page_token = "";
 	do {
 		auto endpoint_builder = catalog.GetBaseUrl();
-		endpoint_builder.AddPathComponent(catalog.prefix);
+		endpoint_builder.AddPrefixComponent(catalog.prefix);
 		endpoint_builder.AddPathComponent("namespaces");
 		if (!parent.empty()) {
 			auto parent_name = GetSchemaName(parent);
--- a/src/storage/authorization/oauth2.cpp
+++ b/src/storage/authorization/oauth2.cpp
@@ -282,13 +282,13 @@ unique_ptr<BaseSecret> OAuth2Authorization::CreateCatalogSecretFunction(ClientCo
 unique_ptr<HTTPResponse> OAuth2Authorization::Request(RequestType request_type, ClientContext &context,
                                                       const IRCEndpointBuilder &endpoint_builder, HTTPHeaders &headers,
                                                       const string &data) {
-	if (!token.empty()) {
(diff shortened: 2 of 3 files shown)


Tmonster:
Hi @talatuyarer, 

yes, I will take a look. Just have been super busy with some things at the moment.

src/common/url_utils.cpp
```diff
@@ -18,7 +18,19 @@ void IRCEndpointBuilder::AddPathComponent(const string &component) {
 	if (component.empty()) {
 		return;
 	}
-	path_components.push_back(component);
+	
+	// If the component contains slashes, split it into multiple segments
+	if (component.find('/') != string::npos) {
```
Tmonster, line 30: Any specific reason why you split into multiple segments here? 
I have had issues in the past where warehouse names contain slashes and if they are split up into separate components then the attach doesn't work.

talatuyarer: `IRCEndpointBuilder::AddPathComponent()` is treating the entire prefix as a single component and URL-encoding it.  `GetURLEncoded()` calls `StringUtil::URLEncode(component)` for each path component. This is correct for individual components, but when BigLake returns a prefix like `projects/1057666841514/catalogs/biglake-public-nyc-taxi-iceberg`, it contains multiple path segments separated by /.

When we call `AddPathComponent(catalog.prefix)` with this multi-segment prefix, it treats the entire string (including the / characters) as a single component and URL-encodes it, turning / into %2F.

If you want i can create a function just prefix specific. @Tmonster What do you think ?

talatuyarer: I added AddPrefixComponent method. fyi

talatuyarer:
@Tmonster Thank you for pointer I added also tests too. and also run make format to fix format issue fyi

talatuyarer:
> I think the fix here is to write some functionality to detect if the prefix is already encoded or not? If it is encoded, decode it and add it as the prefix.

`AddPrefixComponent` handles this by checking for `/`. If the prefix is URL-encoded, it won't contain `/`.

I've added a unit test to ensure the `Authorization` header isn't overridden by extra headers and also added the missing `AddPrefixComponent`.

PR #682, opened 2026-02-03
Support partition definitions

Tmonster:
Supports partition defintions (not inserts)

You can define a partition definition in either the `CREATE TABLE` statement, or in the `ALTER TABLE` statement.


src/metadata/iceberg_partition_spec.cpp
```diff
@@ -71,4 +71,18 @@ string IcebergPartitionSpec::FieldsToJSON() const {
 	return ICUtils::JsonToString(std::move(doc_p));
 }
 
+void IcebergPartitionSpec::FieldsToJson(yyjson_mut_doc *doc, yyjson_mut_val *root_object,
+                                        const vector<rest_api_objects::PartitionField> &fields) {
+	yyjson_mut_obj_add_strcpy(doc, root_object, "type", "struct");
```
Tishj, line 76: I don't know where this is in the spec
I feel like we're building this object, no?
```yml
    PartitionSpec:
      type: object
      required:
        - fields
      properties:
        spec-id:
          type: integer
          readOnly: true
        fields:
          type: array
          items:
            $ref: '#/components/schemas/PartitionField'
```
This should only have `spec-id` and `fields` if that's the case?

We already added `spec-id` earlier:
```c++
yyjson_mut_obj_add_int(doc, spec_json, "spec-id", ref_update.spec.spec_id);
```

PR #683, opened 2026-02-03
Use the MultiFileReader for scanning manifest files (all data manifests in a single reader)

Tishj:
Instead of reading 1 manifest at a time, we make use of the MultiFileReader to read multiple files.

The schema of manifests evolves throughout different versions, and is also influenced by the `partition-spec-id` of the `manifest_file` entry describing it.
To deal with this, we construct a set of `global_columns` for the MultiFileReader, which will map every read file by the field id of the column/field (in case of nested types).

This simplifies the transformation logic for the `IcebergManifestFile` and `IcebergManifestEntry` structs, as we are now in full control of the order and appearance of columns in the result, we can do away with the handrolled mapping we had in place before.

It should also simplify the `GetDataFile` path, as we won't need to manually keep track of the manifest we're currently scanning, but can instead just pull a DataChunk from the reader every time the current one is depleted (same lazy-loading approach, where we'll process 2048 (STANDARD_VECTOR_SIZE) tuples, and only fetch a new chunk when the current one is fully read)

NOTE:
This also requires changes to `duckdb-avro`, because it wasn't putting out `MultiFileColumnDefinition`s that the ColumnMapper was happy with (for LIST and MAP)

----

This isn't parallelizing anything just yet
We're just creating a multi file reader that *can* be parallelized (by giving each worker a local state)

I'll cover that in a follow up, just to keep the PR somewhat smaller

### DIFF
--- a/src/manifest_reader.cpp
+++ b/src/manifest_reader.cpp
@@ -93,6 +93,8 @@ idx_t ManifestReader::ReadChunk(idx_t offset, idx_t count, vector<IcebergManifes
 	D_ASSERT(offset + count <= chunk.size());
 	auto &scan_info = GetScanInfo().Cast<IcebergManifestFileScanInfo>();
 
+	//! NOTE: the order of these columns is defined by the order that they are produced in BuildManifestSchema
+	//! see `iceberg_avro_multi_file_reader.cpp`
 	idx_t vector_index = 0;
 	auto &status = chunk.data[vector_index++];
 	auto &snapshot_id = chunk.data[vector_index++];
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,7 @@ set(EXTENSION_SOURCES
     src/hash_utils.cpp
     src/base_manifest_reader.cpp
     src/manifest_list_reader.cpp
-    src/manifest_file_reader.cpp
+    src/manifest_reader.cpp
     src/deletes/equality_delete.cpp
     src/deletes/positional_delete.cpp
     src/deletes/deletion_vector.cpp
--- a/src/manifest_list_reader.cpp
+++ b/src/manifest_list_reader.cpp
@@ -31,6 +31,8 @@ idx_t ManifestListReader::ReadChunk(idx_t offset, idx_t count, vector<IcebergMan
 	D_ASSERT(offset < chunk.size());
 	D_ASSERT(offset + count <= chunk.size());
 
+	//! NOTE: the order of these columns is defined by the order that they are produced in BuildManifestListSchema
+	//! see `iceberg_avro_multi_file_reader.cpp`
 	idx_t vector_index = 0;
 	auto &manifest_path = chunk.data[vector_index++];
 	auto &manifest_length = chunk.data[vector_index++];


src/iceberg_functions/iceberg_avro_multi_file_reader.cpp
```diff
@@ -1,24 +1,502 @@
...
 
 unique_ptr<MultiFileReader> IcebergAvroMultiFileReader::CreateInstance(const TableFunction &table) {
-	(void)table;
-	return make_uniq<IcebergAvroMultiFileReader>();
+	return make_uniq<IcebergAvroMultiFileReader>(table.function_info);
+}
+
+static MultiFileColumnDefinition CreateManifestFilePartitionsColumn() {
+	MultiFileColumnDefinition partitions("partitions", IcebergManifestList::FieldSummaryType());
+	partitions.identifier = Value::INTEGER(507);
```
Tishj: This has a lot of repetition with `iceberg_manifest.cpp` and `iceberg_manifest_list.cpp`, perhaps some of it can be unified

src/manifest_reader.cpp
```diff
@@ -165,133 +126,103 @@ idx_t ManifestFileReader::ReadChunk(idx_t offset, idx_t count, vector<IcebergMan
...
+	idx_t entry_index = 0;
+	auto &data_file_entries = StructVector::GetEntries(data_file);
+	optional_ptr<Vector> content;
+	if (iceberg_version >= 2) {
+		content = *data_file_entries[entry_index++];
 	}
-	auto upper_bounds_it = vector_mapping.find(UPPER_BOUNDS);
-	if (upper_bounds_it != vector_mapping.end()) {
-		upper_bounds = *child_entries[upper_bounds_it->second.GetChildIndex(0).GetPrimaryIndex()];
+	auto &file_path = *data_file_entries[entry_index++];
```
Tmonster, line 115: Maybe we should this closer to where we define the schemas of the IcebergManifestFiles?

src/iceberg_functions/iceberg_avro_multi_file_reader.cpp
```diff
@@ -1,24 +1,502 @@
...
+	deleted_rows_count.identifier = Value::INTEGER(514);
+	schema.push_back(deleted_rows_count);
+
+	// partitions (v2+, field-id 507)
+	schema.push_back(CreateManifestFilePartitionsColumn());
+
+	// first_row_id (v3+, field-id 520, default 0)
+	if (iceberg_version >= 3) {
+		MultiFileColumnDefinition first_row_id("first_row_id", LogicalType::BIGINT);
+		first_row_id.identifier = Value::INTEGER(520);
```
Tmonster: Should we use the names defined in `iceberg_manifest_list.hpp`?

src/iceberg_functions/iceberg_avro_multi_file_reader.cpp
```diff
@@ -1,24 +1,502 @@
...
+	if (type == AvroScanInfoType::MANIFEST_LIST) {
+		schema = BuildManifestListSchema(metadata);
+	} else {
+		auto &manifest_file_scan = scan_info.Cast<IcebergManifestFileScanInfo>();
+		auto &manifest_files = manifest_file_scan.manifest_files;
+		unordered_set<int32_t> partition_spec_ids;
+		for (auto &manifest_file : manifest_files) {
+			partition_spec_ids.insert(manifest_file.partition_spec_id);
+		}
+		schema = BuildManifestSchema(snapshot, metadata, partition_spec_ids);
```
Tmonster: can we add a comment here why we need the partition specs? 
i.e the partition specs can vary in a manifest list. Because of this, the partition column of the manifest_entry can be different per file, so they need to be merged.

src/manifest_file_reader.cpp
```diff
@@ -52,51 +43,21 @@ void ManifestFileReader::CreateVectorMapping(idx_t column_id, MultiFileColumnDef
...
-
-		vector_mapping.emplace(child_field_id, ColumnIndex(column_id, child_indexes));
-		if (child_field_id == PARTITION) {
-			for (idx_t partition_idx = 0; partition_idx < child.children.size(); partition_idx++) {
-				auto &partition_field = child.children[partition_idx];
-
-				auto partition_field_id = partition_field.identifier.GetValue<int32_t>();
-				auto partition_child_indexes = child_indexes;
-				partition_child_indexes.emplace_back(partition_idx);
+	idx_t column_index = 2;
```
Tmonster: can we change the order of columns in the global columns so we don't have to do this?

Tishj: Did you one better, fully removed the need for this mapping logic 👍

src/manifest_reader.cpp
Tishj: I didn't move them, but I added a comment explaining where the order comes from, hope that's enough 👍

Issue #669, opened 2026-01-28
Parquet and Avro traffic are not routed through HTTP_PROXY


erikcw:
When settings `HTTP_PROXY`, I would expect all traffic to be routed through the proxy server.  However, only the iceberg catalog metadata HTTP traffic is.

```bash
# Start proxy server
uvx mimtproxy --mode regular:@8878
```

```sql

INSTALL aws;
INSTALL httpfs;
INSTALL iceberg;

-- HTTP_PROXY works, but the parquet calls to S3 are not picked up.

CREATE SECRET http_proxy (
    TYPE http,
    HTTP_PROXY 'http://localhost:8878'
);

CREATE OR REPLACE SECRET sample (
    TYPE s3,
    KEY_ID '*********************',
    SECRET '*********************',
    REGION 'us-east-1',
    HTTP_PROXY 'http://localhost:8878',
    URL_STYLE path
);

FROM duckdb_secrets();

ATTACH 'arn:aws:s3tables:us-east-1:*************:bucket/aws-s3' AS s3_tables_aws_s3 (
   TYPE iceberg,
   ENDPOINT_TYPE s3_tables,
   SECRET 'sample'
);

SHOW ALL TABLES;
SELECT 'test reading from iceberg s3_tables via mitmproxy';
FROM s3_tables_aws_s3.my_bucket.inventory limit 10;
-- I can see traffic to the iceberg metadata catalog but no parquet or avro requests.

SELECT 'test reading parquet file from s3 via mitmproxy';
FROM read_parquet('s3://my_bucket/fd4d8b44-ffd0-4596-8ac7-c856fcae9232/sample.parq') limit 10;
-- I can see this parquet traffic in the proxy.
```

Tmonster:
Thank you for reporting this! We will look into it.
Chances are it is because we create a new secret from the vended credentials, but we do not add the Http proxy during that time. Hopefully we can get a fix in for v1.5

PR #686, opened 2026-02-05
Fix local http proxy

yan-alex:
https://github.com/duckdb/duckdb-iceberg/issues/669

### DIFF
--- a/src/storage/catalog/iceberg_table_entry.cpp
+++ b/src/storage/catalog/iceberg_table_entry.cpp
@@ -35,8 +35,8 @@ unique_ptr<BaseStatistics> IcebergTableEntry::GetStatistics(ClientContext &conte
 	return nullptr;
 }
 
-case_insensitive_map_t<Value> AddHTTPSecretsToOptions(SecretEntry &http_secret_entry,
-                                                      case_insensitive_map_t<Value> options) {
+void AddHTTPSecretsToOptions(SecretEntry &http_secret_entry,
+                                                      case_insensitive_map_t<Value> &options) {
 	auto http_kv_secret = dynamic_cast<const KeyValueSecret &>(*http_secret_entry.secret);
 
 	options["http_proxy"] =
@@ -45,8 +45,8 @@ case_insensitive_map_t<Value> AddHTTPSecretsToOptions(SecretEntry &http_secret_e
 	    http_kv_secret.TryGetValue("verify_ssl").IsNull()
 	        ? Value::BOOLEAN(true)
 	        : http_kv_secret.TryGetValue("verify_ssl").DefaultCastAs(LogicalType::BOOLEAN);
-	return options;
 }
+
 void IcebergTableEntry::PrepareIcebergScanFromEntry(ClientContext &context) const {
 	auto &ic_catalog = catalog.Cast<IcebergCatalog>();
 	auto &secret_manager = SecretManager::Get(context);
@@ -127,8 +127,8 @@ void IcebergTableEntry::PrepareIcebergScanFromEntry(ClientContext &context) cons
 			                {"endpoint", endpoint}};
 		}
 
-		if (http_secret_entry != nullptr) {
-			info.options = AddHTTPSecretsToOptions(*http_secret_entry, info.options);
+		if (http_secret_entry) {
+			AddHTTPSecretsToOptions(*http_secret_entry, info.options);
 		}
 
 		(void)secret_manager.CreateSecret(context, info);
@@ -140,8 +140,8 @@ void IcebergTableEntry::PrepareIcebergScanFromEntry(ClientContext &context) cons
 		}
 	} else {
 		for (auto &info : table_credentials.storage_credentials) {
-			if (http_secret_entry != nullptr) {
-				info.options = AddHTTPSecretsToOptions(*http_secret_entry, info.options);
+			if (http_secret_entry) {
+				AddHTTPSecretsToOptions(*http_secret_entry, info.options);
 			}
 			(void)secret_manager.CreateSecret(context, info);
 		}
--- a/test/sql/local/irc/http_proxy.test
+++ b/test/sql/local/irc/http_proxy.test
@@ -4,7 +4,6 @@
 
 require-env HTTP_PROXY_PUBLIC
 
-# shouldn't run on cloud testing
 require-env ICEBERG_SERVER_AVAILABLE
 
 require avro


src/storage/catalog/iceberg_table_entry.cpp
```diff
@@ -73,14 +73,18 @@ void IcebergTableEntry::PrepareIcebergScanFromEntry(ClientContext &context) cons
...
 			                                      : kv_secret.TryGetValue("session_token").ToString()},
 			                {"region", region},
-			                {"endpoint", endpoint}};
-		}
+						    {"endpoint", endpoint},
+						    {"HTTP_PROXY", kv_secret.TryGetValue("HTTP_PROXY").IsNull() ? "" : kv_secret.TryGetValue("HTTP_PROXY").ToString()},
+			};
+		};
+
 		(void)secret_manager.CreateSecret(context, info);
```
Tmonster, line 134: I acually think we should add this regardless of the key type. In this PR, it looks like you are only adding it for s3tables, so for GLUE and any other general catalog this will not work

yan-alex: It's still in progress, I opened the PR for CI since it didn't run on PRs in my fork for some reason?
So this solution doesn't work actually. When testing locally, it always sets it to `""`, kv_secret never contains  HTTP_PROXY, so not sure where it's being saved. 
Would be nice if we can link up tomorrow to take a look

src/storage/catalog/iceberg_catalog.cpp
```diff
@@ -261,6 +261,21 @@ unique_ptr<SecretEntry> IcebergCatalog::GetIcebergSecret(ClientContext &context,
 	}
 	return secret_entry;
 }
+unique_ptr<SecretEntry> IcebergCatalog::GetHTTPSecret(ClientContext &context, const string &secret_name) {
+	auto transaction = CatalogTransaction::GetSystemCatalogTransaction(context);
+	unique_ptr<SecretEntry> secret_entry = nullptr;
+	if (secret_name.empty()) {
+		//! Try to find any secret with type 'iceberg'
```
Tmonster: Is this comment correct? Should it be any secret with type 'http'?

yan-alex: Oh good catch! Forgot to remove this

.github/workflows/test-spark-catalog.yml
```diff
@@ -6,11 +6,14 @@ on:
         description: "Name of the DuckDB-Iceberg artifact to download"
         required: true
         type: string
+  workflow_dispatch:
```
Tmonster: should this be in the PR?

scripts/requirements.txt
```diff
@@ -4,3 +4,4 @@ pydantic==2.9.0
 pyiceberg==0.9.1
 pyarrow
 pytest
+mitmproxy== 12.2.1
```
Tmonster: nit: you can remove the extra space in here

src/storage/catalog/iceberg_table_entry.cpp
```diff
@@ -57,32 +57,41 @@ void IcebergTableEntry::PrepareIcebergScanFromEntry(ClientContext &context) cons
 			DUCKDB_LOG_INFO(context, "Creating Iceberg Table secret with no scope. Returned metadata location is %s",
 			                lc_storage_location);
 		}
+		auto &sigv4_auth = ic_catalog.auth_handler->Cast<SIGV4Authorization>();
+		auto http_secret_entry = IcebergCatalog::GetHTTPSecret(context, sigv4_auth.secret);
+		auto http_kv_secret = dynamic_cast<const KeyValueSecret &>(*http_secret_entry->secret);
```
Tmonster: Here you need to check that http_secret_entry is not null

src/storage/catalog/iceberg_table_entry.cpp
```diff
@@ -57,32 +57,41 @@ void IcebergTableEntry::PrepareIcebergScanFromEntry(ClientContext &context) cons
 			DUCKDB_LOG_INFO(context, "Creating Iceberg Table secret with no scope. Returned metadata location is %s",
 			                lc_storage_location);
 		}
+		auto &sigv4_auth = ic_catalog.auth_handler->Cast<SIGV4Authorization>();
```
Tmonster: I'm not 100% sure we can guarantee the auth handler is Sigv4 here.

Tmonster (commented):
So I took a look at some of the cloud issues.
We can find the http secret with the following code
```
		unique_ptr<SecretEntry> http_secret_entry;
		switch (ic_catalog.auth_handler->type) {
		case IcebergAuthorizationType::SIGV4: {
			auto &sigv4_auth = ic_catalog.auth_handler->Cast<SIGV4Authorization>();
			http_secret_entry = IcebergCatalog::GetHTTPSecret(context, sigv4_auth.secret);
			break;
		}
		case IcebergAuthorizationType::OAUTH2: {
			auto &oauth2_auth = ic_catalog.auth_handler->Cast<OAuth2Authorization>();
			auto empty_str = "";
			http_secret_entry = IcebergCatalog::GetHTTPSecret(context, empty_str);
			break;
		}
		default:
			break;
		}
```
And then add the necessary information to the secret based on if http_secret_entry is null or not. I have a version of this locally which seemed to pass the cloud tests

.github/workflows/test-spark-catalog.yml
yan-alex: This enables the workflow to be run manually. We should have this in every workflow IMO. Should I make a different PR?

src/storage/catalog/iceberg_table_entry.cpp
yan-alex: So that means I can only retrieve this `http_secret_entry` for glue and s3tables? We wanted to do it for all hosts, no? Is there another way to do it?

src/storage/catalog/iceberg_table_entry.cpp
```diff
@@ -34,6 +35,18 @@ unique_ptr<BaseStatistics> IcebergTableEntry::GetStatistics(ClientContext &conte
 	return nullptr;
 }
 
+case_insensitive_map_t<Value> AddHTTPSecretsToOptions(SecretEntry &http_secret_entry,
+                                                      case_insensitive_map_t<Value> options) {
```
Tmonster: This function can be void, you can pass the options by reference.
```
case_insensitive_map_t<Value> &options
```

.github/workflows/test-spark-catalog.yml
Tmonster: This is a reusable workflow that requires a binary built in the calling workflow, so I'm sure running this manually will even work. I think we should take it out, and we can look at adding it back in another PR

src/storage/catalog/iceberg_table_entry.cpp
```diff
@@ -34,6 +35,18 @@ unique_ptr<BaseStatistics> IcebergTableEntry::GetStatistics(ClientContext &conte
...
+	auto http_kv_secret = dynamic_cast<const KeyValueSecret &>(*http_secret_entry.secret);
+
+	options["http_proxy"] =
+	    http_kv_secret.TryGetValue("http_proxy").IsNull() ? "" : http_kv_secret.TryGetValue("http_proxy").ToString();
+	options["verify_ssl"] =
+	    http_kv_secret.TryGetValue("verify_ssl").IsNull()
+	        ? Value::BOOLEAN(true)
+	        : http_kv_secret.TryGetValue("verify_ssl").DefaultCastAs(LogicalType::BOOLEAN);
+	return options;
+}
```
Tmonster, line 48: nit: can we get one empty line between the two functions?

src/storage/catalog/iceberg_table_entry.cpp
```diff
@@ -83,17 +126,25 @@ void IcebergTableEntry::PrepareIcebergScanFromEntry(ClientContext &context) cons
 			                {"region", region},
 			                {"endpoint", endpoint}};
 		}
+
+		if (http_secret_entry != nullptr) {
```
Tmonster: nit: we don't need `!= nullptr`, we can just have `if (http_secret_entry)`

src/storage/catalog/iceberg_table_entry.cpp
```diff
@@ -83,17 +126,25 @@ void IcebergTableEntry::PrepareIcebergScanFromEntry(ClientContext &context) cons
...
 			DUCKDB_LOG_INFO(context, "Failed to create valid secret from Vendend Credentials for table '%s'",
 			                table_info.name);
 		}
-	}
-
-	for (auto &info : table_credentials.storage_credentials) {
-		(void)secret_manager.CreateSecret(context, info);
+	} else {
+		for (auto &info : table_credentials.storage_credentials) {
+			if (http_secret_entry != nullptr) {
```
Tmonster: same nit here

test/sql/local/irc/http_proxy.test
```diff
@@ -0,0 +1,85 @@
+# name: test/sql/local/irc/http_proxy.test
+# description: Test http proxy
+# group: [irc]
+
+require-env HTTP_PROXY_PUBLIC
+
+# shouldn't run on cloud testing
```
Tmonster: This comment can be removed

PR #687, opened 2026-02-05
Add BigLake support and better encoding/decoding of prefix's in catalog response

Tmonster:
This builds off of https://github.com/duckdb/duckdb-iceberg/pull/674

We detect if a prefix is already encoded by URLDecoding it, if it equals the prefix returned by the catalog, we assume it is not URL encoded, and each `/` is a separate component. If it does not equal the prefix returned by the catalog, we assume it is encoded and that the prefix is meant to be one component in the URL.

S3Tables expects 1 component
Glue can be either 1 component or multiple components

Here is a breakdown of some current catalogs

| catalog  | prefix example                                                           | encoded | is multiple components? |
|----------|--------------------------------------------------------------------------|---------|-------------------------|
| Biglake  | projects/1057666841514/catalogs/biglake-public-nyc-taxi-iceberg          | no      | yes                     |
| Glue     | catalogs/<account_id>:s3tablescatalog:<bucket_name>                      | no      | yes&no                  |
| s3tables | arn%3Aaws%3As3tables%3Aus-east-1%3A<account_id>%3Abucket%2F<bucket_name> | yes     | no                      |


Tishj (approved):
Thanks LGTM 👍

Tmonster:
@talatuyarer 
I have an internal issue to add to the docs, but if you want to take a stab at it feel free. 
Here is a PR I opened to add our current documentation, that should be enough to get started

https://github.com/duckdb/duckdb-web/pull/5410

PR #688, opened 2026-02-05
Implement a Max table staleness option

Tmonster:
This enables a max table staleness, which can greatly improve performance for analytical workloads. If a user knows data will be slow moving, they can add a max_table_staleness attach option which will prevent duckdb-iceberg from making extra GetTableInformation requests until after the max_table_staleness interval has expired.

### DIFF
--- a/src/catalog_api.cpp
+++ b/src/catalog_api.cpp
@@ -231,7 +231,9 @@ vector<rest_api_objects::TableIdentifier> IRCAPI::GetTables(ClientContext &conte
 		}
 		auto response = catalog.auth_handler->Request(RequestType::GET_REQUEST, context, url_builder, headers);
 		if (!response->Success()) {
-			if (response->status == HTTPStatusCode::Forbidden_403 || response->status == HTTPStatusCode::NotFound_404) {
+			if (response->status == HTTPStatusCode::Forbidden_403 ||
+			    response->status == HTTPStatusCode::Unauthorized_401 ||
+			    response->status == HTTPStatusCode::NotFound_404) {
 				// return empty result if user cannot list tables for a schema.
 				vector<rest_api_objects::TableIdentifier> ret;
 				return ret;
@@ -278,7 +280,8 @@ vector<IRCAPISchema> IRCAPI::GetSchemas(ClientContext &context, IcebergCatalog &
 		HTTPHeaders headers(*context.db);
 		auto response = catalog.auth_handler->Request(RequestType::GET_REQUEST, context, endpoint_builder, headers);
 		if (!response->Success()) {
-			if (response->status == HTTPStatusCode::Forbidden_403) {
+			if (response->status == HTTPStatusCode::Forbidden_403 ||
+			    response->status == HTTPStatusCode::Unauthorized_401) {
 				// return empty result if user cannot list schemas.
 				return result;
 			}
--- a/src/storage/catalog/iceberg_table_set.cpp
+++ b/src/storage/catalog/iceberg_table_set.cpp
@@ -58,6 +58,7 @@ bool IcebergTableSet::FillEntry(ClientContext &context, IcebergTableInformation 
 			return false;
 		}
 		if (get_table_result.status_ == HTTPStatusCode::Forbidden_403 ||
+		    get_table_result.status_ == HTTPStatusCode::Unauthorized_401 ||
 		    get_table_result.status_ == HTTPStatusCode::NotFound_404) {
 			return false;
 		}


src/storage/catalog/iceberg_catalog.cpp
```diff
@@ -85,6 +95,20 @@ MetadataCacheValue &IcebergCatalog::GetLoadTableResult(const string &table_key)
...
+	auto it = metadata_cache.find(table_key);
+	if (it == metadata_cache.end()) {
+		return nullptr;
+	}
+	auto &cached_value = *it->second;
+	if (system_clock::now() > cached_value.expires_at) {
+		// cached value has expired
+		return nullptr;
+	}
+	return &cached_value;
```
Tmonster, line 114: Either we need to grab the lock in the calling function or we need to do a callback, or we need to copy the cached value. We cannot return a reference to a function that does not hold the lock

src/storage/catalog/iceberg_catalog.cpp
```diff
@@ -85,6 +95,20 @@ MetadataCacheValue &IcebergCatalog::GetLoadTableResult(const string &table_key)
...
+}
+
 void IcebergCatalog::RemoveLoadTableResult(string table_key) {
```
Tmonster: const string &table_key

src/storage/catalog/iceberg_catalog.cpp
```diff
@@ -101,57 +125,53 @@ optional_ptr<CatalogEntry> IcebergCatalog::CreateSchema(CatalogTransaction trans
...
-	auto namespace_identifier = IRCAPI::ParseSchemaName(info.name);
-	namespace_items.push_back(IRCAPI::GetEncodedSchemaName(namespace_identifier));
-	if (info.if_not_found == OnEntryNotFound::RETURN_NULL) {
-		auto schema_lookup = EntryLookupInfo(CatalogType::SCHEMA_ENTRY, info.name);
-		auto transaction = CatalogTransaction::GetSystemCatalogTransaction(context);
-		auto schema_exists = LookupSchema(transaction, schema_lookup, info.if_not_found);
-		if (!schema_exists) {
+
+	// Verify schema existence on the server first
+	bool schema_exists = IRCAPI::VerifySchemaExistence(context, *this, info.name);
```
Tmonster, line 168: Does this still make a schema request to the first table read? Please double check

src/storage/catalog/iceberg_schema_set.cpp
```diff
@@ -15,9 +15,24 @@ optional_ptr<CatalogEntry> IcebergSchemaSet::GetEntry(ClientContext &context, co
...
+		}
+		throw CatalogException("Schema '%s' does not exist", name);
+	}
+
+	// If the schema was created in this transaction, return the local entry directly
+	if (iceberg_transaction.created_schemas.count(name)) {
+		auto entry = entries.find(name);
+		if (entry != entries.end()) {
+			return entry->second.get();
+		}
```
Tmonster, line 33: maybe throw an error if it was created but it is not in the entries?

src/storage/catalog/iceberg_schema_set.cpp
```diff
@@ -31,16 +46,22 @@ optional_ptr<CatalogEntry> IcebergSchemaSet::GetEntry(ClientContext &context, co
 	}
 	if (entry == entries.end()) {
 		CreateSchemaInfo info;
-		if (!IRCAPI::VerifySchemaExistence(context, ic_catalog, name)) {
-			if (if_not_found == OnEntryNotFound::RETURN_NULL) {
-				return nullptr;
-			} else {
-				throw CatalogException("Iceberg namespace by the name of '%s' does not exist", name);
+		// Look up existence of default schema to avoid lookup of `duckdb_*` tables
```
Tmonster, line 51: Investigate why we are doing this again. Something where if you call `use {{catalog}}` then start querying, you get some crazy behavior where duckdb tables are searched for.

src/storage/catalog/iceberg_table_set.cpp
```diff
@@ -33,19 +33,36 @@ bool IcebergTableSet::FillEntry(ClientContext &context, IcebergTableInformation
 	}
 
 	auto &ic_catalog = catalog.Cast<IcebergCatalog>();
+	auto table_key = table.GetTableKey();
 
+	// Only check cache if MAX_TABLE_STALENESS option is set
+	if (ic_catalog.attach_options.max_table_staleness_micros.IsValid()) {
+		auto cached_result = ic_catalog.TryGetValidCachedLoadTableResult(table_key);
```
Tmonster: Here we should grab the metadata cache lock, and pass it to TryGetValidCachedLoadTableResult

src/storage/catalog/iceberg_table_set.cpp
```diff
@@ -33,19 +33,36 @@ bool IcebergTableSet::FillEntry(ClientContext &context, IcebergTableInformation
...
+		if (cached_result && cached_result->expires_at > system_clock::now()) {
```
Tmonster: `cached_result->expires_at > system_clock::now()` is not needed here. It is already checked in TryGetValidCachedLoadTableResult

src/storage/catalog/iceberg_table_set.cpp
```diff
@@ -33,19 +33,36 @@ bool IcebergTableSet::FillEntry(ClientContext &context, IcebergTableInformation
...
 	auto get_table_result = IRCAPI::GetTable(context, ic_catalog, schema, table.name);
 	if (get_table_result.has_error) {
 		if (get_table_result.error_._error.type == "NoSuchIcebergTableException") {
 			return false;
 		}
 		if (get_table_result.status_ == HTTPStatusCode::Forbidden_403 ||
-		    get_table_result.status_ == HTTPStatusCode::Unauthorized_401) {
+		    get_table_result.status_ == HTTPStatusCode::Unauthorized_401 ||
+		    get_table_result.status_ == HTTPStatusCode::NotFound_404) {
 			return false;
```
Tmonster, line 63: Might be worth it to log the actual response from the server

Tmonster: Or maybe remove the 401 response?

src/catalog_api.cpp
```diff
@@ -232,7 +232,8 @@ vector<rest_api_objects::TableIdentifier> IRCAPI::GetTables(ClientContext &conte
 		auto response = catalog.auth_handler->Request(RequestType::GET_REQUEST, context, url_builder, headers);
 		if (!response->Success()) {
 			if (response->status == HTTPStatusCode::Forbidden_403 ||
-			    response->status == HTTPStatusCode::Unauthorized_401) {
+			    response->status == HTTPStatusCode::Unauthorized_401 ||
```
Tmonster: remove 401

src/storage/catalog/iceberg_schema_set.cpp
Tmonster: If you run the following commands,
```
use my_datalake.default;
select * from non_existent_table;
select request.url, request.type from duckdb_logs_parsed('http');
```
The following requests get made
```
┌───────────────────────────────────────────────────────────────────────┬─────────┐
│                                  url                                  │  type   │
│                                varchar                                │ varchar │
├───────────────────────────────────────────────────────────────────────┼─────────┤
│ http://127.0.0.1:8181/v1/namespaces/main                              │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default                           │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default                           │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/non_existent_table │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/main                              │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/non_existent_table │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_tables      │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_views       │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_indexes     │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/sqlite_master      │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_tables      │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_views       │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_indexes     │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/sqlite_master      │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_tables      │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_views       │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_indexes     │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/sqlite_master      │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_tables      │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_views       │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_indexes     │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables                    │ GET     │
│ http://127.0.0.1:8181/v1/namespaces/pg_catalog                        │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_indexes     │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces                                   │ GET     │
│ http://127.0.0.1:8181/v1/namespaces/information_schema                │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_columns     │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_views       │ HEAD    │
│ http://127.0.0.1:8181/v1/namespaces/level1/tables                     │ GET     │
├───────────────────────────────────────────────────────────────────────┴─────────┤
│ 29 rows                                                               2 columns │
└─────────────────────────────────────────────────────────────────────────────────┘
```

For some reason duckdb is querying all of these random `duckdb_*` tables, I think to look for the tables in a different catalog, I'm not sure. 
The default schema now for iceberg catalogs (if it is not set) is the empty string, so we error when users execute, `use my_datalake;` without a schema, and attach option 'default schema' was not provided.

So when `use my_datalake.default` is executed, and a user queries some non-existent table, for some reason, duckdb also starts to query the schema `main` (which is duckdb's default schema) for tables like `duckdb_*`. Why this happens I don't know, I will need to file an issue for this probably, I feel like we should be able to turn this off.

But the reason why we verify `DEFAULT_SCHEMA` (i.e 'main') is because if we can return that it does not exist, then the queries to `duckdb_*` are not made. 

Instead the requests look as follows 
```
┌──────────────────────────────────────────────────────────────────────────────┬─────────┐
│                                     url                                      │  type   │
│                                   varchar                                    │ varchar │
├──────────────────────────────────────────────────────────────────────────────┼─────────┤
│ http://127.0.0.1:8181/v1/namespaces/default/tables/non_existent_table        │ GET     │
│ http://127.0.0.1:8181/v1/namespaces/pg_catalog/tables/duckdb_indexes         │ GET     │
│ http://127.0.0.1:8181/v1/namespaces/default/tables                           │ GET     │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_tables             │ GET     │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/duckdb_views              │ GET     │
│ http://127.0.0.1:8181/v1/namespaces/default/tables/sqlite_master             │ GET     │
│ http://127.0.0.1:8181/v1/namespaces                                          │ GET     │
│ http://127.0.0.1:8181/v1/namespaces/information_schema/tables/duckdb_columns │ GET     │
│ http://127.0.0.1:8181/v1/namespaces/pg_catalog/tables                        │ GET     │
│ http://127.0.0.1:8181/v1/namespaces/level1/tables                            │ GET     │
└──────────────────────────────────────────────────────────────────────────────┴─────────┘
```
This could be cleaned up even more actually. May investigate this in a follow up PR.

src/storage/catalog/iceberg_catalog.cpp
Tmonster: Oh, no it does not, this is on drop schema

src/storage/catalog/iceberg_catalog.cpp
Tmonster: done

Tmonster:
Going to add a bunch of logging as well for a PR on monday, this will include logging for responses that are 401/404 etc.

djouallah:
is there any chance you backport this to 1.4.4 ?

PR #692, opened 2026-02-07
Parallel manifest scans

Tishj:
This PR is a follow up to #683 

We delegate the actual scanning to the worker threads now.
To facilitate this, we better separate the concerns of the `AvroScan` and the `BaseManifestReader` classes, and also introduce `IcebergManifestReadingState`.

`AvroScan` now only holds the global state of the scan.
`BaseManifestReader` no longer owns the `AvroScan`, instead it only holds a reference and creates a local state of the scan managed by the `AvroScan`.

`IcebergManifestReadingState` now owns the `AvroScan` and the executor that manages the active scanning tasks.
The scan is performed by a `ManifestReadTask`, the execution of which consists of calling `ManifestReader::Read` until it's depleted, which all tasks are doing in parallel.

In `GetDataFile` we simply poll whether there are buffered entries available to process*
^ We could probably simplify this further, offloading the pruning to the tasks as well

### DIFF
--- a/src/iceberg_functions/iceberg_multi_file_list.cpp
+++ b/src/iceberg_functions/iceberg_multi_file_list.cpp
@@ -384,7 +384,8 @@ IcebergPredicateStats IcebergPredicateStats::DeserializeBounds(const Value &lowe
 	return res;
 }
 
-bool IcebergMultiFileList::FileMatchesFilter(const IcebergManifestEntry &manifest_entry) const {
+bool IcebergMultiFileList::FileMatchesFilter(const IcebergManifestEntry &manifest_entry,
+                                             IcebergDataFileType file_type) const {
 	D_ASSERT(!table_filters.filters.empty());
 
 	auto &filters = table_filters.filters;
@@ -401,7 +402,6 @@ bool IcebergMultiFileList::FileMatchesFilter(const IcebergManifestEntry &manifes
 		auto &data_file = manifest_entry.data_file;
 		// First check if there are partitions
 		if (!data_file.partition_values.empty()) {
-			// check if the index is in the parititon value thing.
 			auto partition_spec_it = metadata.partition_specs.find(manifest_entry.partition_spec_id);
 			if (partition_spec_it == metadata.partition_specs.end()) {
 				throw InvalidConfigurationException(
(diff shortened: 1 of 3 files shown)


src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -445,40 +480,67 @@ bool IcebergMultiFileList::FileMatchesFilter(const IcebergManifestEntry &manifes
...
+			lock_guard<mutex> entry_guard(entry_lock);
+			current_manifest_entries.insert(current_manifest_entries.end(),
+			                                std::make_move_iterator(new_entries.begin()),
+			                                std::make_move_iterator(new_entries.end()));
+			if (!current_manifest_entries.empty()) {
+				//! We have entries to read, return
+				return true;
+			}
+		}
+		if (read_state.started == 0) {
```
Tishj: We can probably simplify this a lot, we should be able to determine that there are no worker threads just by checking `scheduler.NumberOfThreads()`.
We only need to do our own scan if there are no worker threads, otherwise we can just spin-lock until we have at least 1 entry.

src/include/iceberg_multi_file_list.hpp
```diff
@@ -25,13 +25,36 @@
...
+	IcebergManifestReadingState(ClientContext &context, unique_ptr<AvroScan> scan, mutex &lock,
+	                            vector<IcebergManifestEntry> &entries)
+	    : executor(context), scan(std::move(scan)), lock(lock), entries(entries), started(0) {
+	}
+
+public:
+	void AddTask();
+
+public:
+	TaskExecutor executor;
```
Tishj, line 43: Not sure if we should upgrade this to an `Executor`
I think it's slightly less lightweight, but it does have better support for canceling

src/include/iceberg_multi_file_list.hpp
```diff
@@ -117,8 +144,12 @@ struct IcebergMultiFileList : public MultiFileList {
 	mutable vector<IcebergManifestFile> data_manifests;
 	mutable vector<reference<IcebergManifest>> transaction_data_manifests;
 	mutable idx_t transaction_data_idx = 0;
+	mutable unique_ptr<IcebergManifestReadingState> manifest_read_state;
+	mutable atomic<bool> finished;
```
Tishj, line 147: I think these are unused

src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -445,40 +480,67 @@ bool IcebergMultiFileList::FileMatchesFilter(const IcebergManifestEntry &manifes
...
+		if (!current_manifest_entries.empty()) {
+			return true;
+		}
+	}
+	if (!manifest_read_state) {
+		return false;
+	}
+	auto &read_state = *manifest_read_state;
+	auto &executor = read_state.executor;
+	while (!executor.IsFinished()) {
```
Tishj: This was patched into core, will need to find a different route
don't really feel like pushing a patch to core for this

Issue #691, opened 2026-02-06
Duplicate rows when filtering for null values in S3Table table


Maegor:
Hello,
I have noticed that when I filter by “where <field> is null,” rows that I had previously deleted appear again. For example, I insert one row with the field “id,” then delete it. If I filter by id is null, the expected result is zero rows, but the deleted row appears.

Step to reproduce, (execute one by one)
```SQL
DROP TABLE if EXISTS s3_tables."default".table_test;

create table s3_tables."default".table_test as 
select * from read_csv('data_test.csv'); 
--> Table is created but no row are inserted due to "SQL Error: Invalid Configuration Error: Table table_test already exists"

select * from s3_tables."default".table_test where invoice_refunded is  null;
--> No rows OK!

insert into s3_tables."default".table_test
select * from read_csv('data_test.csv');
-->1 row inserted
select * from s3_tables."default".table_test where invoice_refunded is  null; --> Return 1 row
--> 1 row is returned OK!

DELETE FROM  s3_tables."default".table_test WHERE month = '2026-01-01 00:00:00.000' 
--> Row delete OK!

select * from s3_tables."default".table_test;
--> No rows. OK!
select * from s3_tables."default".table_test where invoice_refunded is null;
--> 1 row is returned. **NOT OK!!**
```
Thank you!

[data_test.csv](https://github.com/user-attachments/files/25126487/data_test.csv)

PR #698, opened 2026-02-09
Do not filter delete files based on stats, only based on partition information

Tmonster:
Should fix https://github.com/duckdb/duckdb-iceberg/issues/691


Tishj (approved):
Thanks 👍

PR #705, opened 2026-02-11
Add OAuth2 refresh_token support

talatuyarer:
OAuth2's refresh token flow (https://datatracker.ietf.org/doc/html/rfc6749#section-6) is not supported yet in iceberg extension. 

This PR adds RFC 6749 refresh token grant support to DuckDB’s Iceberg extension OAuth2 flow. The Iceberg extension previously supported acquiring access tokens but did not implement the refresh path, causing long-running sessions to fail once the access token expired.

Refresh token rotation is tracked in-memory only. Rotated refresh tokens are not persisted back into DuckDB secrets; after DETACH/re-ATTACH, the original refresh token is used.

### DIFF
--- a/test/sql/local/irc/oauth2/test_incorrect_secret_and_attach.test
+++ b/test/sql/local/irc/oauth2/test_incorrect_secret_and_attach.test
@@ -1,6 +1,6 @@
-# name: test/sql/local/irc/test_incorrect_secret_and_attach.test
+# name: test/sql/local/irc/oauth2/test_incorrect_secret_and_attach.test
 # description: test combinations of create secret and attah
-# group: [irc]
+# group: [oauth2]
 
 require avro
 


talatuyarer:
@Tmonster Do you want to review this ?

talatuyarer:
You can also give a try by using this commands

Extract Your Credentials from gcloud ADC file:

```shell
CLIENT_ID=$(cat ~/.config/gcloud/application_default_credentials.json | jq -r '.client_id')
CLIENT_SECRET=$(cat ~/.config/gcloud/application_default_credentials.json | jq -r '.client_secret')
REFRESH_TOKEN=$(cat ~/.config/gcloud/application_default_credentials.json | jq -r '.refresh_token')
```

```sql
LOAD iceberg;

-- Create secret with refresh_token for automatic refresh
CREATE SECRET biglake_auto_refresh (
    TYPE ICEBERG,
    CLIENT_ID '<your-client-id>',
    CLIENT_SECRET '<your-client-secret>',
    REFRESH_TOKEN '<your-refresh-token>',
    OAUTH2_SERVER_URI 'https://oauth2.googleapis.com/token',
    EXTRA_HTTP_HEADERS MAP {'x-goog-user-project': '<your-project-id>'}
);

-- Attach to BigLake
ATTACH 'gs://biglake-public-nyc-taxi-iceberg' AS biglake_public (
    TYPE ICEBERG,
    SECRET biglake_auto_refresh,
    ENDPOINT 'https://biglake.googleapis.com/iceberg/v1/restcatalog'
);

-- Query data - token will auto-refresh when needed!
SELECT COUNT(*) FROM biglake_public.public_data.nyc_taxicab;
```

test/sql/local/irc/oauth2/test_incorrect_secret_and_attach.test
```diff
@@ -32,7 +32,7 @@ ATTACH '' AS my_datalake (
     ENDPOINT 'http://127.0.0.1:8181'
 );
 ----
```
Tmonster, line 34: small nit: could you put these in `test/sql/local/irc/oauth`?

talatuyarer: i am waiting your initial review. After that i will move all tests under oauth folder. @Tmonster

src/storage/authorization/oauth2.cpp
```diff
@@ -62,49 +67,149 @@ static string XWWWFormUrlEncode(const string &input) {
...
 		auto endpoint_builder = IRCEndpointBuilder::FromURL(uri);
 		unique_ptr<HTTPClient> placeholder_client;
-		auto response = APIUtils::Request(RequestType::POST_REQUEST, context, endpoint_builder, placeholder_client,
-		                                  headers, post_data);
-		doc = std::unique_ptr<yyjson_doc, YyjsonDocDeleter>(ICUtils::api_result_to_doc(response->body));
+		response = APIUtils::Request(RequestType::POST_REQUEST, context, endpoint_builder, placeholder_client, headers,
+		                             post_data);
+	} catch (HTTPException &http_ex) {
+		// APIUtils::Request throws HTTPException for non-2xx responses.
+		throw InvalidConfigurationException("Could not get token from %s: HTTP error: %s", uri, http_ex.what());
```
Tmonster: APIUtils actually doesn't throw for non-2XX responses. 400 responses can be parsed via the response object so you can inspect codes and error messages directly

src/storage/authorization/oauth2.cpp
```diff
@@ -251,52 +403,233 @@ unique_ptr<BaseSecret> OAuth2Authorization::CreateCatalogSecretFunction(ClientCo
...
+		auto expiry_time = now + expiry_duration;
+		token_expires_at = std::chrono::duration_cast<std::chrono::seconds>(expiry_time.time_since_epoch()).count();
+	} else {
+		// No expiry information available at all (shouldn't happen with the logic above)
+		token_expires_at = 0;
+	}
+}
+
+bool OAuth2Authorization::IsTokenExpiredUnlocked(ClientContext &context) const {
+	// Internal method - caller must hold token_mutex
```
Tmonster, line 593: can we have the caller pass the mutex as well if they are forced to hold it? An example of that [here](https://github.com/duckdb/duckdb-iceberg/pull/706/changes)

src/storage/authorization/oauth2.cpp
```diff
@@ -162,6 +268,28 @@ unique_ptr<OAuth2Authorization> OAuth2Authorization::FromAttachOptions(ClientCon
...
+		Value expires_in_val = kv_iceberg_secret.TryGetValue("expires_in");
+		int32_t expires_in_seconds = 0;
+		if (!expires_in_val.IsNull() && expires_in_val.type().id() == LogicalTypeId::INTEGER) {
+			expires_in_seconds = expires_in_val.GetValue<int32_t>();
+		}
+
+		// Compute expiry time if we have both token and expires_in
+		if (!token.IsNull() && expires_in_seconds > 0) {
+			// Pass empty string for refresh_token to preserve the one we just set
+			result->UpdateTokenState(token.ToString(), expires_in_seconds, "");
```
Tmonster, line 316: Accoring to src/storage/authorization/oauth2.cpp:630 we should have a lock when calling UpdateTokenState. Seems Seems like we don't have it here?

src/storage/authorization/oauth2.cpp
```diff
@@ -62,49 +67,149 @@ static string XWWWFormUrlEncode(const string &input) {
 
 } // namespace
 
-string OAuth2Authorization::GetToken(ClientContext &context, const string &grant_type, const string &uri,
-                                     const string &client_id, const string &client_secret, const string &scope) {
+namespace {
```
Tmonster: nit: Can you merge with the other namespace above?

test/sql/local/irc/oauth2/test_oauth2_edge_cases.test
```diff
@@ -0,0 +1,136 @@
...
+    USE_SSL 0
+);
+
+# Test 1: Small expires_in (< 30 seconds) - should not underflow
+# Create a token that expires quickly
+statement ok
+ATTACH '' AS quick_expiry_lake (
+    TYPE ICEBERG,
+    CLIENT_ID 'client1',
+    CLIENT_SECRET 's3cr3t',
```
Tmonster, line 41: Should we also pass an `expires_in` here?

Tmonster (commented):
Thanks! a couple of comments. 
Also, do you know of any cloud based catalogs I could test this against? I'm guessing BigLake supports this?

test/sql/local/irc/oauth2/test_oauth2_edge_cases.test
talatuyarer: expires_in comes from the OAuth2 server response. No need to add expires_in her.  it's automatically retrieved from the OAuth2 server response when using client credentials!

src/storage/authorization/oauth2.cpp
talatuyarer: No lock needed here. we're still constructing this object and haven't shared it with other threads yet. UpdateTokenState is safe to call during construction. Once this object is returned and shared, UpdateTokenState will only be called from RefreshAccessTokenUnlocked which requires holding token_mutex.

test/sql/local/irc/oauth2/test_incorrect_secret_and_attach.test
talatuyarer: Done. I moved all tests under oauth2 folder

src/storage/authorization/oauth2.cpp
talatuyarer: I changed the code accordingly.

PR #712, opened 2026-02-13
 Get lakekeeper running and testing locally again

yan-alex:
https://github.com/duckdblabs/duckdb-internal/issues/7415

### DIFF
--- a/.github/workflows/MainDistributionPipeline.yml
+++ b/.github/workflows/MainDistributionPipeline.yml
@@ -17,7 +17,7 @@ jobs:
     uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@main
     with:
       extension_name: iceberg
-      duckdb_version: cbb5d8b7388d034c83d7fd22467074c00b322254
+      duckdb_version: e10e37bc42eaefb4421a55746a4c6c767db07c6f
       ci_tools_version: main
       exclude_archs: 'windows_amd64_mingw'
       extra_toolchains: 'python3'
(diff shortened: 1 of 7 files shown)


.github/workflows/LakekeeperTesting.yml
```diff
@@ -70,208 +69,52 @@ jobs:
...
         run: |
-          docker run -d \
-            --name keycloak \
-            -p 30080:8080 \
-            -e KC_BOOTSTRAP_ADMIN_USERNAME=admin \
-            -e KC_BOOTSTRAP_ADMIN_PASSWORD=admin \
-            -v $(pwd)/lakekeeper/examples/access-control-simple/keycloak/realm.json:/opt/keycloak/data/import/realm.json \
-            quay.io/keycloak/keycloak:26.0.7 \
-            start-dev --metrics-enabled=true --health-enabled=true --import-realm --verbose --log-level=INFO --features=token-exchange
+          echo "127.0.0.1 minio" | sudo tee -a /etc/hosts
```
yan-alex, line 74: @Tmonster  This is a bit of a ugly workaround I had to do to enable DuckDB to reach minio, otherwise we can't retrieve .avro manifest files...
Not sure why Lakekeeper chose to keep minio unexposed to the outside world on [their docker-compose.yaml](https://github.com/lakekeeper/lakekeeper/blob/f6aaa4570cf62d54943581e7a00b9f882d9c992d/examples/access-control-simple/docker-compose.yaml#L171) , does Lakekeeper expect to handle manifest retrieval it all internally for us? I wonder if we are using it wrong by querying avro files ourselves.
Probably this is just a oversight on their part though, idk

Tmonster: probably an oversight on their part yes. Is there a way to expose it from within their `docker-compose.yaml`?

yan-alex: Yeah that's what I did [here](https://github.com/duckdb/duckdb-iceberg/pull/712/changes/BASE..0720bf6f7dc8271658d63cc0b6c758d5478d96d2#diff-828f657c06154fd2627df6b4a5eaf35ae1a29cf50456e58120676c94d37672d7R42) 
But then (if I remember correctly) lakekeeper would inform us to reach it at minio:9000 which is a docker internal endpoint. That's why I had to let minio resolve to 127.0.0.1 here

PR #702, opened 2026-02-10
[V3][Read] Row lineage support

Tishj:
This PR adds support for reading the row lineage of an Iceberg V3 table: https://iceberg.apache.org/spec/#row-lineage

PR #721, opened 2026-02-17
Row lineage write support

Tishj:
Follow up to #702
this is a bit of a bigger PR

With this PR, we add:
- Support for the `WITH ('format-version': <number>)` syntax on a `CREATE TABLE` statement.
- `INSERT` support for V3 tables
- ~~`UPDATE` support for V3 tables~~
- ~~`DELETE` support for V3 tables~~

Also cleaned up the previous test: `test/sql/local/irc/test_row_lineage_read.test` to be snapshot-based, and verify the *order* of the rows, rather than verify the exact row_id (this is unstable in Spark because they parallelize everything and end up creating arbitrary rewrites/compactions, bumping the `next-row-id` of a table arbitrarily)

Decided to disable UPDATE and DELETE functionality for now, so this can be merged already


src/storage/catalog/iceberg_catalog.cpp
```diff
@@ -192,6 +192,10 @@ DatabaseSize IcebergCatalog::GetDatabaseSize(ClientContext &context) {
 	return size;
 }
 
+ErrorData IcebergCatalog::SupportsCreateTable(BoundCreateTableInfo &info) {
+	return ErrorData();
```
Tishj, line 206: TODO: This needs to be more granular

src/storage/iceberg_delete.cpp
```diff
@@ -306,8 +306,7 @@ PhysicalOperator &IcebergCatalog::PlanDelete(ClientContext &context, PhysicalPla
 		row_id_indexes.push_back(bound_ref.index);
 	}
 	auto &ic_table_entry = op.table.Cast<IcebergTableEntry>();
-	// Verify Iceberg table version is v2
-	if (ic_table_entry.table_info.table_metadata.iceberg_version != 2) {
+	if (ic_table_entry.table_info.table_metadata.iceberg_version < 2) {
```
Tishj: TODO: we're not technically allowed to do this, before we start writing DeletionVectors for positional deletes

src/storage/iceberg_insert.cpp
```diff
@@ -202,44 +188,43 @@ void IcebergInsert::AddWrittenFiles(IcebergInsertGlobalState &global_state, Data
 			auto &col_name = StringValue::Get(struct_children[0]);
 			auto &col_stats = MapValue::GetChildren(struct_children[1]);
 			auto column_names = ParseQuotedList(col_name, '.');
+			if (column_names[0] == "_row_id") {
```
Tishj, line 202: I think they're having issues with this in ducklake currently, if there's an actual column named `_row_id`, we're not writing stats for it now.

src/storage/catalog/iceberg_table_set.cpp
```diff
@@ -159,6 +188,12 @@ bool IcebergTableSet::CreateNewEntry(ClientContext &context, IcebergCatalog &cat
 	table_ptr->table_info.table_metadata.schemas[0] = IcebergCreateTableRequest::CreateIcebergSchema(table_ptr);
 	table_ptr->table_info.table_metadata.current_schema_id = 0;
 	table_ptr->table_info.table_metadata.schemas[0]->schema_id = 0;
+	if (iceberg_version.IsValid()) {
+		table_ptr->table_info.table_metadata.iceberg_version = iceberg_version.GetIndex();
```
Tishj: Maybe we should verify that this is within a valid range?
Though that's honestly more the Catalog's job imo
At least checking that it's positive is a good idea though

src/storage/iceberg_update.cpp
```diff
@@ -260,22 +252,19 @@ PhysicalOperator &IcebergCatalog::PlanUpdate(ClientContext &context, PhysicalPla
...
 		throw NotImplementedException("Update Iceberg V%d tables", table.table_info.table_metadata.iceberg_version);
 	}
 
-	IcebergCopyInput copy_input(context, table);
-	vector<Value> field_input;
-	field_input.push_back(WrittenFieldIds(table_schema));
-	copy_input.options["field_ids"] = std::move(field_input);
-
+	IcebergCopyInput copy_input(context, table, table_schema);
+	copy_input.virtual_columns = IcebergInsertVirtualColumns::WRITE_ROW_ID;
```
Tishj: This should probably be guarded by `version >= 3`

src/include/storage/iceberg_insert.hpp
```diff
@@ -17,16 +17,19 @@
 
 namespace duckdb {
 
+enum class IcebergInsertVirtualColumns { NONE, WRITE_ROW_ID, WRITE_SNAPSHOT_ID, WRITE_ROW_ID_AND_SNAPSHOT_ID };
```
Tmonster: `Snapshot_id` -> `sequence_number`?

src/storage/iceberg_transaction_data.cpp
```diff
@@ -119,11 +134,26 @@ void IcebergTransactionData::AddSnapshot(IcebergSnapshotOperationType operation,
...
-	case IcebergSnapshotOperationType::APPEND:
+	case IcebergSnapshotOperationType::APPEND: {
 		manifest_content_type = IcebergManifestContentType::DATA;
+		if (table_metadata.has_next_row_id) {
+			new_snapshot.has_added_rows = true;
+			new_snapshot.added_rows = 0;
+			for (auto &manifest_entry : data_files) {
+				D_ASSERT(manifest_entry.status != IcebergManifestEntryStatusType::DELETED);
+				auto &data_file = manifest_entry.data_file;
+				new_snapshot.added_rows += data_file.record_count;
```
Tmonster, line 155: Do we write `existing_row_count` for updates correctly?

src/storage/iceberg_transaction_data.cpp
```diff
@@ -162,6 +192,20 @@ void IcebergTransactionData::AddUpdateSnapshot(vector<IcebergManifestEntry> &&de
...
 	new_snapshot.manifest_list = manifest_list_path;
 	new_snapshot.timestamp_ms = Timestamp::GetEpochMs(Timestamp::GetCurrentTimestamp());
+	if (table_metadata.iceberg_version >= 3) {
+		D_ASSERT(table_metadata.has_next_row_id);
+		new_snapshot.has_added_rows = true;
+		new_snapshot.added_rows = 0;
+		for (auto &manifest_entry : data_files) {
+			D_ASSERT(manifest_entry.status != IcebergManifestEntryStatusType::DELETED);
+			auto &data_file = manifest_entry.data_file;
+			new_snapshot.added_rows += data_file.record_count;
```
Tmonster, line 213: should we double check that `existing_rows_count` is 0?

Tishj: We already do `manifest_list_delete_entry.existing_rows_count = 0;` 👍

src/iceberg_functions/iceberg_avro_multi_file_reader.cpp
Tishj: This looks wrong?

Tishj: Apart from this, there are still a couple issues with reading row lineage from upgraded tables.
Those tables won't have any `next-row-id` in the metadata.json, or `first-row-id` in the `snapshot` or `first_row_id` in the `manifest_file`, so those will also need to be inferred.

src/include/metadata/iceberg_manifest_list.hpp
```diff
@@ -155,8 +155,6 @@ struct IcebergManifestList {
 
 public:
 	string path;
-
-private:
 	vector<IcebergManifestFile> manifest_entries;
```
Tmonster, line 158: Does this need to stay public?

src/storage/iceberg_insert.cpp
Tmonster: can we add a comment as to why we skip?

src/storage/iceberg_transaction_data.cpp
```diff
@@ -162,6 +203,19 @@ void IcebergTransactionData::AddUpdateSnapshot(vector<IcebergManifestEntry> &&de
 	new_snapshot.schema_id = table_metadata.current_schema_id;
 	new_snapshot.manifest_list = manifest_list_path;
 	new_snapshot.timestamp_ms = Timestamp::GetEpochMs(Timestamp::GetCurrentTimestamp());
+	if (table_metadata.iceberg_version >= 3) {
+		D_ASSERT(table_metadata.has_next_row_id);
+		new_snapshot.has_added_rows = true;
+		new_snapshot.added_rows = 0;
```
Tmonster, line 209: I thought updates were not yet supported, so this code won't run. Is it just here for future implementation?

Tmonster (commented):
Should we throw an error when rowid is querying on a v2 table? Or a DuckDB_LOG message?

```
create table my_datalake.default.v3_table (a int);
insert into my_datalake.defualt.v3_table select range a from range(1000);
select rowid from my_datalake.default.v3_table;
┌────────────┐
│   rowid    │
│   int64    │
├────────────┤
│       NULL │
│       NULL │
│       NULL │
│       NULL │
...
```

Can we also get a CTAS test?

And maybe a test with a column that has rowid as one of the columns?
Running it locally I get the following. We should be able to distinguish between them via field ids right?

```
create table my_datalake.default.v3_table_3 WITH ('format-version'=3) as select range rowid from range(100, 1100);
memory D select rowid, * from my_datalake.default.v3_table_3;
┌───────┬───────┐
│ rowid │ rowid │
│ int64 │ int64 │
├───────┼───────┤
│   100 │   100 │
│   101 │   101 │
│   102 │   102 │
│   103 │   103 │
│   104 │   104 │
│   105 │   105 │
```

src/storage/iceberg_transaction_data.cpp
Tishj: Future work, it was already implemented but needed DELETE support, which is #728

src/storage/iceberg_transaction_data.cpp
```diff
@@ -119,20 +131,49 @@ void IcebergTransactionData::AddSnapshot(IcebergSnapshotOperationType operation,
...
 		throw NotImplementedException("Cannot have use snapshot operation type REPLACE or OVERWRITE here");
 	}
 	auto add_snapshot = make_uniq<IcebergAddSnapshot>(table_info, manifest_list_path, std::move(new_snapshot));
 	CreateManifestListEntry(*add_snapshot, table_metadata, manifest_content_type, std::move(data_files));
+	if (table_metadata.iceberg_version >= 3) {
+		auto &snapshot = add_snapshot->snapshot;
+		snapshot.has_first_row_id = true;
+		snapshot.first_row_id = next_row_id;
+
+		auto &last_manifest_file = add_snapshot->manifest_list.manifest_entries.back();
```
Tishj, line 170: This is why it's public, but I made a change in #529 that makes it so this won't be needed anymore

Issue #549, opened 2025-10-24
Fails to write an iceberg table to Glue Catalog (`400 InvalidInputException` from AWS)


phitoduck:
Hey all! I took the [new iceberg write support](https://duckdb.org/2025/09/16/announcing-duckdb-140#iceberg-writes) for a spin using iceberg in S3 registered with Glue Catalog. 

I was able to successfully attach to a glue catalog, but not write a trivial table.

My eventual goal is to write into an iceberg table using the new `MERGE INTO` support. But currently stuck on an `400: InvalidInputException` returned from AWS.

Most simple, self-contained example I can come up with:

`iceberg.py`

```python
#!/usr/bin/env python3
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "boto3",
#     "duckdb==1.4.1",
# ]
# ///
import os

import boto3
import duckdb

GLUE_DB_REGION = os.getenv("AWS_REGION", "us-west-2")
AWS_ACCOUNT_ID = boto3.client("sts").get_caller_identity()["Account"]


sql = f"""
-- Install required extensions
INSTALL aws;
INSTALL iceberg;
INSTALL httpfs;

-- Load the extensions
LOAD aws;
LOAD iceberg;
LOAD httpfs;

-- Create AWS credentials secret using credential_chain provider (2-line approach)
CREATE SECRET IF NOT EXISTS aws_credentials (
    TYPE S3,
    PROVIDER credential_chain
);

-- Configure S3 settings for iceberg
CREATE SECRET IF NOT EXISTS s3_settings (
    TYPE S3,
    PROVIDER credential_chain
);

-- Try to attach to Glue catalog
ATTACH '{AWS_ACCOUNT_ID}' AS glue_catalog (
    TYPE iceberg,
    ENDPOINT 'glue.{GLUE_DB_REGION}.amazonaws.com/iceberg',
    AUTHORIZATION_TYPE 'sigv4'
);

USE glue_catalog.health_db;

CREATE TABLE IF NOT EXISTS example AS SELECT a FROM range(4) example(a)    
"""

# create a duckdb connection in a with block and execute the setup_sql
with duckdb.connect() as conn:
    result = conn.execute(sql).fetchall()
    print(result)
```

Run with this command

```bash
AWS_PROFILE=some-profile-with-admin AWS_REGION=us-west-2 uv run iceberg.py
```

Results in this error:

```
Traceback (most recent call last):
[...]

PR #722, opened 2026-02-17
Table Properties during create table, as well as location

Tmonster:
Taking over https://github.com/duckdb/duckdb-iceberg/pull/719 since this adds more cloud tests. 
Turns out the apache/iceberg-rest-fixture image does not save table properties when creating a table, so the cloud tests are a way to check that we are correctly sending table properties on create table 

fixes https://github.com/duckdb/duckdb-iceberg/issues/549

Cloud tests running here 
https://github.com/duckdb/duckdb-iceberg/actions/runs/22100969326

### DIFF
--- a/test/sql/cloud/snowflake/test_create_table_with_properties.test
+++ b/test/sql/cloud/snowflake/test_create_table_with_properties.test
@@ -43,23 +43,6 @@ create schema if not exists my_datalake.test_create_schema;
 statement ok
 use my_datalake.test_create_schema;
 
-statement ok
-drop table if exists test_create_table;
-
-statement ok
-create table test_create_table (a int) 
-WITH (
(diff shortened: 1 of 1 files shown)


src/storage/catalog/iceberg_catalog.cpp
```diff
@@ -192,6 +192,10 @@ DatabaseSize IcebergCatalog::GetDatabaseSize(ClientContext &context) {
 	return size;
 }
 
+ErrorData IcebergCatalog::SupportsCreateTable(BoundCreateTableInfo &info) {
+	return ErrorData();
```
Tishj, line 206: This was WIP, this needs granularity, as we don't support all of the modifiers yet

PR #728, opened 2026-02-19
Add support for V3 UPDATE + DELETE

Tishj:
This PR includes https://github.com/duckdb/duckdb-iceberg/pull/721

To reference the spec: https://iceberg.apache.org/spec/#version-3

> Deletion vectors are added in v3, stored using the Puffin deletion-vector-v1 blob type
> Writers are not allowed to add new position delete files to v3 tables

> referenced_data_file was added and can be used for both deletion vectors (required) and v2 position delete files that contain deletes for only one data file (optional)
content_offset was added and must match the deletion vector blob's offset in a Puffin file
content_size_in_bytes was added and must match the deletion vector blob's length in a Puffin file

> Deletion vectors are maintained synchronously: Writers must merge DVs (and older position delete files) to ensure there is at most one DV per data file
> These position delete files must be merged into the DV for a data file when one is created
> Position delete files that contain deletes for more than one data file need to be kept in table metadata until all deletes are replaced by DVs

> Readers can safely ignore position delete files if there is a DV for a data file

Because of this part:
> Position delete files that contain deletes for more than one data file need to be kept in table metadata until all deletes are replaced by DVs

We will only delete delete-files containing deletion vectors, we will never delete delete-files containing positional deletes (parquet).
These are just ignored when reading, deletion vectors take precedence over them.

### DIFF
--- a/test/configs/nessie.json
+++ b/test/configs/nessie.json
@@ -36,7 +36,10 @@
         "test/sql/local/irc_any_catalog/transactions/test_transactional_updates.test",
         "test/sql/local/irc_any_catalog/transactions/test_transactional_updates_error_omits_stacktrace.test",
         "test/sql/local/irc_any_catalog/update/test_one_update_one_snapshot.test",
-        "test/sql/local/irc_any_catalog/update/test_update_v3_table.test"
+        "test/sql/local/irc_any_catalog/update/test_update_v3_table.test",
+        "test/sql/local/irc_any_catalog/delete/test_delete_consolidation_transactional.test",
+        "test/sql/local/irc_any_catalog/delete/test_delete_consolidation_commit.test",
+        "test/sql/local/irc_any_catalog/update/test_update_deletion_vector_consolidation_commit.test"
       ]
     }
   ]


src/storage/iceberg_transaction_data.cpp
```diff
@@ -119,24 +134,56 @@ void IcebergTransactionData::AddSnapshot(IcebergSnapshotOperationType operation,
...
+	case IcebergSnapshotOperationType::APPEND: {
 		manifest_content_type = IcebergManifestContentType::DATA;
+		if (table_metadata.has_next_row_id) {
+			new_snapshot.has_added_rows = true;
+			new_snapshot.added_rows = 0;
+
+			for (auto &manifest_entry : data_files) {
+				D_ASSERT(manifest_entry.status != IcebergManifestEntryStatusType::DELETED);
+				auto &data_file = manifest_entry.data_file;
+				//! Since we don't create the avro files yet,
```
Tishj: TODO: deal with this, we need to unset this before writing the avro file
Otherwise we're materializing these values ahead of time, they should be NULL here

Or we need to copy the `IcebergManifestEntry` from our transaction data, so we can modify the copy
without disturbing the data we want to actually write

Tishj: That will be a follow up at some point
Not super interesting yet

src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -886,30 +890,49 @@ void IcebergMultiFileList::ProcessDeletes(const vector<MultiFileColumnDefinition
...
-				ScanDeleteFile(manifest_entry, global_columns, column_indexes);
-			} else if (StringUtil::CIEquals(data_file.file_format, "puffin")) {
-				ScanPuffinFile(manifest_entry.data_file);
-			} else {
-				throw NotImplementedException(
-				    "File format '%s' not supported for deletes, only supports 'parquet' and 'puffin' currently",
-				    data_file.file_format);
+			auto &referenced_data_file = data_file.referenced_data_file;
+			if (!referenced_data_file.empty() && transactional_delete_files &&
+			    transactional_delete_files->count(referenced_data_file)) {
```
Tmonster, line 928: What happens in the following case?
```
BEGIN;
-- delete from t1;
-- read from t1 from before the start of the transaction
```
Do we ignore deletion vectors still?

src/storage/table_update/iceberg_add_snapshot.cpp
```diff
@@ -30,25 +33,125 @@ static rest_api_objects::TableUpdate CreateAddSnapshotUpdate(const IcebergTableI
...
+	rewritten_manifest_file.deleted_files_count = 0;
+	rewritten_manifest_file.existing_files_count = 0;
+	rewritten_manifest_file.added_rows_count = 0;
+	rewritten_manifest_file.deleted_rows_count = 0;
+	rewritten_manifest_file.existing_rows_count = 0;
+
+	bool removed_any_entries = false;
+	idx_t handled_entries = 0;
+	for (auto &manifest_entry : rewritten_manifest.entries) {
+		manifest_entry.status = IcebergManifestEntryStatusType::EXISTING;
```
Tmonster: can we double check that these entry status types are not `DELETED`?

src/storage/table_update/iceberg_add_snapshot.cpp
```diff
@@ -30,25 +33,125 @@ static rest_api_objects::TableUpdate CreateAddSnapshotUpdate(const IcebergTableI
...
+			rewritten_manifest_file.existing_rows_count += manifest_entry.data_file.record_count;
+			rewritten_manifest_file.existing_files_count++;
+			continue;
+		}
+		removed_any_entries = true;
+		D_ASSERT(delete_file_state.referenced_data_files.size() == 1);
+		D_ASSERT(manifest_entry.data_file.referenced_data_file == delete_file_state.referenced_data_files[0]);
+		//! Remove the entry
+		manifest_entry.status = IcebergManifestEntryStatusType::DELETED;
+		rewritten_manifest_file.deleted_rows_count += manifest_entry.data_file.record_count;
```
Tmonster, line 97: Can we double check that data_file.record_count should be added to deleted_rows_count?

test/sql/local/irc_any_catalog/delete/test_delete_consolidation_commit.test
```diff
@@ -0,0 +1,113 @@
...
+
+query IIII
+select sequence_number, rowid, id, data from my_datalake.default.write_v3_update_and_delete order by rowid;
+----
+1	0	1	a
+
+statement ok
+commit;
+
+query IIII
```
Tmonster, line 95: can we double check there is only one deletion vector at the end of the test?
```
set variable = select manifest_list from iceberg_snapshots(tbl) order by sequence_number desc
select count(*) from read_avro(variable) where content = 'DELETE`;
```

test/sql/local/irc_any_catalog/delete/test_delete_consolidation_transactional.test
```diff
@@ -0,0 +1,114 @@
...
+WHERE id = 2;
+
+# The first transaction's delete should now be skipped, and only the last delete should be used
+query IIII
+select sequence_number, rowid, id, data from my_datalake.default.write_v3_update_and_delete order by rowid;
+----
+1	0	1	a
+2	3	4	d_u1
+
+statement ok
```
Tmonster, line 98: Maybe another read to make sure we did not commit? That none of the files are on the server? Idk, some catalogs are weird about cleaning up your own files though

i.e check that we delete puffin files

src/iceberg_functions/iceberg_multi_file_list.cpp
Tishj: 😬 I think we are not ignoring transaction data for time travel queries, because the test I added failed
I don't think that's limited to this feature, because all of the transaction data was present, despite the AT clause

Tishj: Pushed a fix in #740

PR #734, opened 2026-02-24
Test nessie extensively

yan-alex:
Part of https://github.com/duckdblabs/duckdb-internal/issues/7414
Made most IRC tests able to be run with a generic setup. Now, Nessie is also tested against an extra 58 tests! 23 failing tests are skipped for now  in the _nessie.json_ config.

To run tests from the folder _test/sql/local/irc_any_catalog_, we now use `--test-config`:

When testing against _apache/iceberg-rest-fixture_ Catalog:
`./build/release/test/unittest my_test.test --test-config test/configs/fixture.json`
For the Nessie Catalog:
`./build/release/test/unittest my_test.test --test-config test/configs/nessie.json`

### DIFF
--- a/test/sql/local/irc_any_catalog/delete/time_travel_after_uncommitted_delete.test
+++ b/test/sql/local/irc_any_catalog/delete/time_travel_after_uncommitted_delete.test
@@ -1,4 +1,4 @@
-# name: test/sql/local/irc/delete/time_travel_after_uncommitted_delete.test
+# name: test/sql/local/irc_any_catalog/delete/time_travel_after_uncommitted_delete.test
 # description: test integration with iceberg catalog read
 # group: [delete]
 


test/sql/local/irc_specific_catalogs/nessie/oauth2/test_oauth2_edge_cases.test
```diff
@@ -1,4 +1,4 @@
-# name: test/sql/local/irc/oauth2/test_oauth2_edge_cases.test
+# name: test/sql/local/irc_specific_catalogs/nessie/oauth2/test_oauth2_edge_cases.test
```
Tishj, line 1: I think this test isn't running, looking at the CI?
```
Run ./build/release/test/unittest "$PWD/test/local/irc_specific_catalogs/nessie/*" --list-test-names-only || true
Filters: /home/runner/work/duckdb-iceberg/duckdb-iceberg/test/local/irc_specific_catalogs/nessie/*
No test cases matched '/home/runner/work/duckdb-iceberg/duckdb-iceberg/test/local/irc_specific_catalogs/nessie/*'
===============================================================================
No tests ran
```

I think the path there is wrong ^
Looking at this one:
`$PWD/test/sql/local/irc_any_catalog/*`
```
Run ./build/release/test/unittest "$PWD/test/sql/local/irc_any_catalog/*" --list-test-names-only || true
  ./build/release/test/unittest "$PWD/test/sql/local/irc_any_catalog/*" --list-test-names-only || true
  ./build/release/test/unittest "$PWD/test/sql/local/irc_any_catalog/*" --test-config test/configs/nessie.json  --autoloading available
  shell: /usr/bin/bash -e {0}
  env:
    JAVA_HOME: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/21.0.10-7/x64
    JAVA_HOME_21_X64: /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/21.0.10-7/x64
    ICEBERG_SERVER_AVAILABLE: 1
    DUCKDB_ICEBERG_HAVE_GENERATED_DATA: 1
    SECRETS_CREATED_AND_CATALOG_ATTACHED: 1
    HAVE_FRESH_GENERATED_DATA: 1
```

.github/workflows/test-fixture-catalog.yml
```diff
@@ -61,24 +61,42 @@ jobs:
...
+          S3_SECRET: password
+          S3_ENDPOINT: http://127.0.0.1:9000
+          ICEBERG_ENDPOINT: http://127.0.0.1:8181
+          WAREHOUSE: s3://warehouse/wh/
         run: |
           source .venv-spark4/bin/activate
-          make data
+          rm -rf data/generated
+          ./scripts/start-fixture-rest-catalog.sh
+          python3 -m scripts.data_generators.generate_data spark-rest local
```
Tishj: This doesn't need `local`
`local` is not using the REST catalog, it just creates iceberg tables directly on the local filesystem

.github/workflows/test-fixture-catalog.yml
```diff
@@ -61,24 +61,42 @@ jobs:
...
         run: |
-          source .venv-spark4/bin/activate
-          make test_release
+          ./build/release/test/unittest "$PWD/test/sql/local/irc_any_catalog/*" --list-test-names-only || true
+          ./build/release/test/unittest "$PWD/test/sql/local/irc_any_catalog/*" --test-config test/configs/fixture.json
+
+      - name: Run other tests
+        env:
+          ICEBERG_SERVER_AVAILABLE: 1
+          DUCKDB_ICEBERG_HAVE_GENERATED_DATA: 1
```
Tishj: I imagine we don't need this either, `DUCKDB_ICEBERG_HAVE_GENERATED_DATA` is used for `local` generated tests and persistent tests

.github/workflows/test-fixture-catalog.yml
yan-alex: I thought local data was used by other tests? Prior to this PR local was already being generated when `make data` was ran

.github/workflows/test-nessie-catalog.yml
```diff
@@ -69,9 +72,40 @@ jobs:
...
+          WAREHOUSE: warehouse
+        run: |
+          source .venv-spark4/bin/activate
+          rm -rf data/generated
+          python3 -m scripts.data_generators.generate_data nessie
+
+      - name: Run Generic IRC tests
+        env:
+          ICEBERG_SERVER_AVAILABLE: 1
+          DUCKDB_ICEBERG_HAVE_GENERATED_DATA: 1
```
Tishj, line 106: Same here, I'm not even sure how this completes successfully if I'm honest, since we don't generate `local`, but tests that have this require expect that to be generated 🤔

.github/workflows/test-nessie-catalog.yml
```diff
@@ -69,9 +72,40 @@ jobs:
...
+          ./build/release/test/unittest "$PWD/test/sql/local/irc_any_catalog/*" --list-test-names-only || true
+          ./build/release/test/unittest "$PWD/test/sql/local/irc_any_catalog/*" --test-config test/configs/nessie.json  --autoloading available
+
+      - name: Run Nessie Specific Tests
         env:
           NESSIE_SERVER_AVAILABLE: 1
         run: |
-          ./build/release/test/unittest "$PWD/test/*" --list-test-names-only || true
-          ./build/release/test/unittest "$PWD/test/*"
+          ./build/release/test/unittest "$PWD/test/local/irc_specific_catalogs/nessie/*" --list-test-names-only || true
```
Tishj: `test/sql/local`

scripts/data_generators/connections/nessie/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+ICEBERG_LIBRARY_VERSION = RUNTIME_CONFIG['iceberg_library_version']
+
+import sys
+import os
+
+CONNECTION_KEY = 'nessie'
+SPARK_RUNTIME_PATH = os.path.join(os.path.dirname(__file__), '..', '..', f'iceberg-spark-runtime-{SPARK_VERSION}_{SCALA_BINARY_VERSION}-{ICEBERG_LIBRARY_VERSION}.jar')
+
+@IcebergConnection.register(CONNECTION_KEY)
+class IcebergSparkRest(IcebergConnection):
```
Tishj, line 40: nit: just noticed we have this on `main` as well, but this shouldn't be called `IcebergSparkRest`, `IcebergNessieRest` is probably what it should be called

scripts/data_generators/connections/spark_rest/__init__.py
```diff
@@ -47,8 +47,8 @@ def get_connection(self):
             f"--packages org.apache.iceberg:iceberg-spark-runtime-{SPARK_VERSION}_{SCALA_BINARY_VERSION}:{ICEBERG_LIBRARY_VERSION},org.apache.iceberg:iceberg-aws-bundle:{ICEBERG_LIBRARY_VERSION} pyspark-shell"
         )
         os.environ["AWS_REGION"] = "us-east-1"
-        os.environ["AWS_ACCESS_KEY_ID"] = "admin"
-        os.environ["AWS_SECRET_ACCESS_KEY"] = "password"
+        os.environ["AWS_ACCESS_KEY_ID"] = os.environ["S3_KEY_ID"]
```
Tishj: This makes it harder to use for local testing, can we use `os.getenv("S3_KEY_ID", "admin")` instead? Same for the others below

.github/workflows/test-fixture-catalog.yml
yan-alex: Tests like`test/sql/local/irc_custom_setup/memory/test_basic_deletion_vectors.test`  and `test/sql/local/iceberg_scans/filtering_on_bounds.test` need it

.github/workflows/test-fixture-catalog.yml
```diff
@@ -61,24 +61,42 @@ jobs:
...
         run: |
           source .venv-spark4/bin/activate
           mitmdump --mode regular@8878 --flow-detail 2 > mitmproxy.log 2>&1 &
 
-      - name: Test with rest catalog
+      - name: Run Generic IRC tests
         env:
           ICEBERG_SERVER_AVAILABLE: 1
           DUCKDB_ICEBERG_HAVE_GENERATED_DATA: 1
+          SECRETS_CREATED_AND_CATALOG_ATTACHED: 1
```
Tishj, line 86: I'm not entirely sure what this is adding, isn't this already implied by `ICEBERG_SERVER_AVAILABLE` ?

.github/workflows/test-fixture-catalog.yml
```diff
@@ -61,24 +61,42 @@ jobs:
...
+          HAVE_FRESH_GENERATED_DATA: 1
```
Tishj: Same here, it's not unset by anything so it's not really adding anything

.github/workflows/test-fixture-catalog.yml
Tishj: I'm aware, but if we're cleaning up testing anyways, I figured I'd point it out. I hope we test `local` elsewhere as well, if we only want to test nessie/fixture runs, generating `local` takes up unnecessary execution time

.github/workflows/test-fixture-catalog.yml
yan-alex: The idea is that we can use `test/configs/fixture.json` or `test/configs/nessie.json` to create the secret and attach the catalog, making some tests catalog-agnostic.

So now I split IRC tests into the folders
_irc_any_catalog_ and _irc_custom_setup_
They all require `ICEBERG_SERVER_AVAILABLE`. But _irc_any_catalog_ is for catalog-agnostic tests,  and they ` require SECRETS_CREATED_AND_CATALOG_ATTACHED`. 
Tests in the folder _irc_custom_setup_ are not catalog-agnostic, and **don't** require SECRETS_CREATED_AND_CATALOG_ATTACHED.

.github/workflows/test-fixture-catalog.yml
Tishj: Yes, those are `local` tests
The deletion vectors test should probably not live in `irc_custom_setup` now that you bring it up

.github/workflows/test-fixture-catalog.yml
Tishj: Sure, if the `on_init` of `fixture.json` and `nessie.json` would set a environment variable that is then required by the test, that's great.

But that isn't the case, and not something we can do in the tester currently, so I'm not sure what this is adding, other than friction when trying to run locally.
Having to manually set `export ICEBERG_SERVER_AVAILABLE=1` in every session is already annoying, I'd like to not double that

.github/workflows/test-fixture-catalog.yml
yan-alex: Yeah so this `HAVE_FRESH_GENERATED_DATA` was meant to flag tests that can't be run twice like `test/sql/local/irc_any_catalog/insert/test_basic_insert.test`, but I haven't gone through the process of checking all tests that have that yet, as the PR was big enough. I'll remove it for now

Tishj:
Given that there's a lot of environment variable setting happening and requiring the use of `--test-config` to be able to run tests correctly, can we get a description on the PR detailing this?

That way there's at least a trail of bread crumbs through `git blame` to figure out what is required to use this

scripts/data_generators/connections/spark_rest/__init__.py
yan-alex: Oh yeah that's way better indeed. This whole environment variable stuff was to be able to , with Nessie, run tests like `spark_written_upper_lower_bounds` , which would switch to `spark_rest_single_thread/__init__.py`. Without env variables it would not reach the Nessie endpoint but would send requests to the hardcoded fixture endpoint

scripts/data_generators/connections/nessie/__init__.py
yan-alex: We still use spark to generate the data. On `scripts/data_generators/connections/polaris/__init__.py` we also call it spark. So I think its fine no?

.github/workflows/test-fixture-catalog.yml
Tmonster: Yea I don't quite understand the need for `SECRETS_CREATED_AND_CATALOG_ATTACHED`. Tests in both *irc_any_catalog* and *irc_custom_setup* both require `ICEBERG_SERVER_AVAILABLE=1`, I think that's enough already. Users just need to pass the right config and it's fine. 
Hopefulling testing will also eventually just be `make test_spark_fixture` or something.

yan-alex: @Tmonster maybe we can soon review all test "flags", and clean it up a bit.

 I felt `SECRETS_CREATED_AND_CATALOG_ATTACHED` was good to keep in for now because otherwise, steps of the CI that run all tests would end up failing as they don't use the config

yan-alex:
> ^ `on_init` won't be able to do it, but the `test_env` will:

I tried that too, but you can't do `require` with these test envs, so no skipping happens. It works differently if i'm not mistaken.

PR #736, opened 2026-02-25
[Dev] Decouple 'manifest_file' creation from 'snapshot' creation

Tishj:
Some cleanup required by #529

### DIFF
--- a/src/storage/iceberg_transaction_data.cpp
+++ b/src/storage/iceberg_transaction_data.cpp
@@ -54,6 +54,7 @@ IcebergManifestFile IcebergTransactionData::CreateManifestFile(int64_t snapshot_
 	manifest_file.existing_files_count = 0;
 	manifest_file.added_rows_count = 0;
 	manifest_file.existing_rows_count = 0;
+	manifest_file.deleted_rows_count = 0;
 	//! TODO: support partitions
 	manifest_file.partition_spec_id = 0;
 	//! manifest.partitions = CreateManifestPartition();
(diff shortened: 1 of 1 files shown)


src/storage/iceberg_transaction_data.cpp
```diff
@@ -29,61 +29,69 @@ IcebergTransactionData::IcebergTransactionData(ClientContext &context, IcebergTa
...
+				data_file.has_first_row_id = true;
+				data_file.first_row_id = next_row_id;
+				next_row_id += data_file.record_count;
+			}
+			manifest_file.added_rows_count += data_file.record_count;
 			break;
+		}
 		case IcebergManifestContentType::DELETE:
-			manifest_list_delete_entry.deleted_rows_count += data_file.record_count;
+			manifest_file.deleted_rows_count += data_file.record_count;
```
Tishj, line 80: Should initiwlize this to 0 at the top

src/storage/iceberg_transaction_data.cpp
```diff
@@ -116,62 +129,29 @@ void IcebergTransactionData::AddSnapshot(IcebergSnapshotOperationType operation,
...
 
-	auto manifest_content_type = IcebergManifestContentType::DATA;
-	switch (operation) {
-	case IcebergSnapshotOperationType::DELETE:
-		if (table_metadata.has_next_row_id) {
-			//! IRC requires this for a successful commit..
-			new_snapshot.has_added_rows = true;
-			new_snapshot.added_rows = 0;
-		}
-		manifest_content_type = IcebergManifestContentType::DELETE;
```
Tishj, line 139: We need to keep this

Issue #733, opened 2026-02-24
[BUG] Bucket partition predicate pushdown not implemented, causing full table scan


herefree:
When querying an Iceberg table with bucket(N, column) partitioning, the predicate pushdown for the bucket transform is not implemented, resulting in a full table scan instead of scanning only the relevant bucket files. This causes severe performance degradation.
### **Table Schema**

```sql
CREATE TABLE rest_catalog.default.app_bizrpt_cmpgn_detail_iceberg (
    id STRING,
    tenant_id LONG,
    order_id LONG,
    order_no STRING,
    -- ... (其他列)
    dt STRING,
    hour STRING
)
USING iceberg
PARTITIONED BY (bucket(512, tenant_id))
TBLPROPERTIES (
    'write.format.default' = 'parquet',
    'write.parquet.compression-codec' = 'zstd'
);
```
### **Query**

```sql
SELECT * FROM iceberg_scan('s3://bucket/table')
WHERE tenant_id = 10316844 AND order_id = 2005508558472974478;
```
The query should only scan files belonging to the bucket containing tenant_id = 10316844,but it scans all files.
### **Root Cause Analysis**

I've analyzed the code and identified the root cause:

**File**: `src/iceberg_predicate.cpp`, lines 248-249

```cpp
bool IcebergPredicate::MatchBounds(ClientContext &context, const TableFilter &filter,
                                   const IcebergPredicateStats &stats, const IcebergTransform &transform) {
    switch (transform.Type()) {
    case IcebergTransformType::IDENTITY:
        return MatchBoundsTemplated<IdentityTransform>(context, filter, stats, transform);
    case IcebergTransformType::BUCKET:
        return true;  // ⚠️ BUG: Always returns true, no filtering
    case IcebergTransformType::TRUNCATE:
        return true;  // ⚠️ BUG: Same issue
    // ...
    }
}
```
The BUCKET case always returns true, which means no files are filtered out

PR #737, opened 2026-02-25
support bucket pushdown

herefree:
#733 
This commit implements full bucket transform support for Iceberg tables,
enabling predicate pushdown on bucket-partitioned columns.
## Core Implementation
- **Hash Algorithm** (src/iceberg_hash.cpp, src/include/iceberg_hash.hpp):
  - Implement MurmurHash3 32-bit algorithm per Iceberg specification
  - Support all Iceberg data types: integers, strings, dates, timestamps,
    UUIDs, decimals, floats, doubles, booleans, and binary data
  - Handle type-specific encoding (little-endian for integers, big-endian
    for UUIDs, IEEE 754 for floats)
  - Correct NULL value handling (hash to 0)

- **BucketTransform** (src/metadata/iceberg_transform.cpp):
  - Implement ApplyTransform() to compute bucket IDs from filter constants
  - Apply Iceberg formula: bucket_N(x) = (hash(x) & 0x7FFFFFFF) % N
  - Add division-by-zero protection
  - Support CompareEqual for exact match filtering

- **Predicate Pushdown** (src/iceberg_predicate.cpp):
  - Enable bucket predicate pushdown using MatchBoundsTemplated
  - Allow manifest-level and file-level pruning for bucket partitions
  - Support equality filters (WHERE col = value) and IN filters

- **Build System** (CMakeLists.txt):
  - Add src/iceberg_hash.cpp to EXTENSION_SOURCES
## Testing
- **Test Data** (data/persistent/partition_bucket_test/):
  - Generate 100-row test table with bucket(10, partition_col)
  - Data distributed evenly across 10 buckets
- **Test Cases** (test/sql/local/iceberg_scans/test_bucket_predicate_pushdown.test)
-  **Data Generator** (scripts/persistent/partition_bucket.py)


src/metadata/iceberg_transform.cpp
```diff
@@ -102,4 +103,42 @@ LogicalType IcebergTransform::GetSerializedType(const LogicalType &input) const
 	}
 }
 
+Value BucketTransform::ApplyTransform(const Value &constant, const IcebergTransform &transform) {
+	if (constant.IsNull()) {
+		// Iceberg spec: null values hash to 0, then apply modulo
+		return Value::INTEGER(0);
```
Tmonster: Are we sure this is the correct way to handle this? 
The spec also says all `All transforms must return null for a null input value`

https://iceberg.apache.org/spec/#partition-transforms

Tmonster (commented):
Thanks! A couple of extra questions. 
We've been updating a lot of our testing infra, so apologies you have to keep changing things. 

the testing blocks with the following code 
```
statement ok
attach ':memory:' as my_datalake;

statement ok
create schema my_datalake.default;

statement ok
create view my_datalake.default.bucket_partitioned_bigint as
select * from ICEBERG_SCAN('__WORKING_DIRECTORY__/data/generated/iceberg/spark-local/default/bucket_partitioned_bigint');
```
Can be removed now. If you place the tests in `test/sql/local/irc_any_catalog/partitioning` then an `ATTACH` statement is run before the tests that will attach the available catalog in the testing environment.

The start of your test can look something like the following and tests against the rest catalog should work
```
# name: test/sql/local/irc_any_catalog/partitioning/test_paritioning_pushdown_bucket.test
# description: Test bucket partitioning with INTEGER type
# group: [partitioning]

require-env ICEBERG_SERVER_AVAILABLE

require-env SECRETS_CREATED_AND_CATALOG_ATTACHED

require avro

require parquet

require iceberg

require httpfs

require core_functions

# Do not ignore 'HTTP' error messages!
set ignore_error_messages

# turn off external file cache.
statement ok
set enable_external_file_cache=false;

statement ok
CALL enable_logging('HTTP');

# read whole table
statement ok
select * from my_datalake.default.bucket_partitioned_integer;

query I
select count(*) from duckdb_logs_parsed('http') where request.url like '%.parquet';
----
8
```
you can immediately follow it up with queries where you verify the output.

Also have you tested this on your own Iceberg catalogs? We don't have any benchmarking in place (yet...), but as far as I can tell, all of these tests will pass regardless of it there is bucket partitioning pushdown or not. For that reason, could you add some tests where you also query what files are requested from the server? That way if this regresses and we start querying a bunch of unnecessary files not in the partition, we can catch the bug before it lands in main 

something like 
```
statement ok
set enable_external_file_cache=false;

statement ok
CALL enable_logging('HTTP');

# read whole table
statement ok
select * from my_datalake.default.bucket_partitioned_integer;

query I
select count(*) from duckdb_logs_parsed('http') where request.url like '%.parquet';
----
8

statement ok
call truncate_duckdb_logs();

# Test 2: Filter by specific value (should use bucket pushdown)
query I
select count(*) from my_datalake.default.bucket_partitioned_integer where value = 10;
----
2

query I
select count(*) from duckdb_logs_parsed('http') where request.url like '%.parquet';
----
2
```
should be a good start. This checks the number of parquet files queried.

Can we also have a test with one or two other bucket sizings? Maybe one where there are more buckets than values? 

Thanks again! With these changes I think this will be good to go.

PR #742, opened 2026-02-26
[Dev] CreateUpdate should be const so it can be retried safely

Tishj:
This PR fixes an issue found internally, and makes a small step towards retry functionality for failed catalog commits

### DIFF
--- a/src/storage/table_update/iceberg_add_snapshot.cpp
+++ b/src/storage/table_update/iceberg_add_snapshot.cpp
@@ -30,7 +30,18 @@ static rest_api_objects::TableUpdate CreateAddSnapshotUpdate(const IcebergTableI
 	return table_update;
 }
 
-void IcebergAddSnapshot::CreateUpdate(DatabaseInstance &db, ClientContext &context, IcebergCommitState &commit_state) {
+static IcebergManifestFile WriteManifestListEntry(IcebergTableInformation &table_info,
+                                                  const IcebergManifestFile &manifest_file, CopyFunction &avro_copy,
+                                                  DatabaseInstance &db, ClientContext &context) {
+	auto manifest_length =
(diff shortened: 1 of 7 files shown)


src/storage/table_update/iceberg_add_snapshot.cpp
```diff
@@ -30,24 +30,36 @@ static rest_api_objects::TableUpdate CreateAddSnapshotUpdate(const IcebergTableI
...
-	for (idx_t manifest_index = 0; manifest_index < manifest_list_entries_size; manifest_index++) {
-		manifest_list.WriteManifestListEntry(table_info, manifest_index, avro_copy, db, context);
+	auto &uncommitted_manifest_files = manifest_list.GetManifestFilesConst();
+	D_ASSERT(!uncommitted_manifest_files.empty());
+
+	IcebergManifestList new_manifest_list(manifest_list.GetPath());
+	for (auto &manifest_file : uncommitted_manifest_files) {
+		new_manifest_list.AddManifestFile(WriteManifestListEntry(table_info, manifest_file, avro_copy, db, context));
 	}
+	new_manifest_list.AddToManifestEntries(commit_state.manifests);
```
Tmonster, line 59: Hmm, not sure how commit_state is generated, but we move the manifests in commit state. Is it possible that when we start a re-try, the `commit_state` won't have any more manifests?

Tishj: The commit state would need to be recreated for a retry, and we'll likely need to populate it with a new manifest list anyways, if the commit failed the state of the table would need to be refreshed so that means we have a new source snapshot+manifest list.

Tishj: `new_manifest_list.AddToManifestEntries(commit_state.manifests);` is followed by:
`commit_state.manifests = new_manifest_list.GetManifestListEntries();`

So for every update, `commit_state.manifests` is both our input and our output

PR #744, opened 2026-02-26
Support partitioned inserts

Tmonster:
Requires the below PRs to land
https://github.com/duckdb/duckdb-avro/pull/78
and
https://github.com/duckdb/duckdb-avro-c/pull/2

This is a work in progress still. This covers most of the work necessary for partitioned inserts, but there is still work to be done. Partitioned inserts for v2 tables is working so far for all transforms except bucket&truncate

TODOs

Must before merge
- [X] Test partitioned inserts into v3 tables
- [x] Test updates to partitioned tables (also on v3 tables) (not supported)
- [x] Test count(*) read optimization (re: [this line](https://github.com/duckdb/duckdb-iceberg/pull/567/changes#diff-e77015a99249440fa5ce30f700ee64a60c3383bae3caaa66822cb45b07746114R341)) Also make sure the files are initialized

Nice to have before merge.
- [ ] better manifest file skipping. On strict `<` and `>` operations we can do better file skipping. See test_manifest_skip_combined_transforms.test:103, we should be able to skip 3 manifests
- [ ] better logic to handle updating schemas
- [ ] TransformPartitionValueTemplated should be revisited, the avro reader now returns correct logical types. I think we are double transforming some of the types, but the transforms are identical on the physical level, so it's not an issue

### DIFF
--- a/src/include/storage/iceberg_insert.hpp
+++ b/src/include/storage/iceberg_insert.hpp
@@ -22,12 +22,11 @@ namespace duckdb {
 enum class IcebergInsertVirtualColumns { NONE, WRITE_ROW_ID, WRITE_SEQUENCE_NUMBER, WRITE_ROW_ID_AND_SEQUENCE_NUMBER };
 
 struct IcebergCopyInput {
-	explicit IcebergCopyInput(ClientContext &context, IcebergTableEntry &table, const IcebergTableSchema &schema);
+	explicit IcebergCopyInput(ClientContext &context, const IcebergTableMetadata &table_metadata,
+	                          const IcebergTableSchema &schema);
 
-	IcebergCatalog &catalog;
-	//! FIXME: this feels redundant?
-	const ColumnList &columns;
-	const IcebergTableInformation &table_info;
+public:
+	const IcebergTableMetadata &table_metadata;
 	const IcebergTableSchema &schema;
 	string data_path;
 	//! Set of (key, value) options
--- a/src/storage/iceberg_insert.cpp
+++ b/src/storage/iceberg_insert.cpp
@@ -46,17 +46,16 @@ IcebergInsert::IcebergInsert(PhysicalPlan &physical_plan, const vector<LogicalTy
     : PhysicalOperator(physical_plan, PhysicalOperatorType::EXTENSION, types, 1), table(&table), schema(nullptr) {
 }
 
-IcebergCopyInput::IcebergCopyInput(ClientContext &context, IcebergTableEntry &table, const IcebergTableSchema &schema)
-    : catalog(table.catalog.Cast<IcebergCatalog>()), columns(table.GetColumns()), table_info(table.table_info),
-      schema(schema) {
-	data_path = table.table_info.table_metadata.GetDataPath();
+IcebergCopyInput::IcebergCopyInput(ClientContext &context, const IcebergTableMetadata &table_metadata,
+                                   const IcebergTableSchema &schema)
+    : table_metadata(table_metadata), schema(schema) {
+	data_path = table_metadata.GetDataPath();
 
 	// Get partition spec if the table is partitioned
-	auto &metadata = table.table_info.table_metadata;
-	table_schema = table.table_info.table_metadata.GetSchemaFromId(table.table_info.table_metadata.current_schema_id);
+	auto &metadata = table_metadata;
+	table_schema = table_metadata.GetSchemaFromId(table_metadata.current_schema_id);
 	if (metadata.GetLatestPartitionSpec().IsPartitioned()) {
-		partition_spec =
-		    table.table_info.table_metadata.FindPartitionSpecById(table.table_info.table_metadata.default_spec_id);
+		partition_spec = table_metadata.FindPartitionSpecById(table_metadata.default_spec_id);
 	}
 }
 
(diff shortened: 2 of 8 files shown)


src/metadata/iceberg_transform.cpp
```diff
@@ -102,4 +73,17 @@ LogicalType IcebergTransform::GetSerializedType(const LogicalType &input) const
 	}
 }
 
+void IcebergTransform::SetBucketOrModuloValue(idx_t value) {
```
Tishj: Bucket or Truncate

src/rest_catalog/objects/table_metadata.cpp
```diff
@@ -181,9 +181,9 @@ string TableMetadata::TryFromJSON(yyjson_val *obj) {
 	}
 	auto last_partition_id_val = yyjson_obj_get(obj, "last-partition-id");
 	if (last_partition_id_val) {
-		has_last_partition_id = true;
+		has_last_partition_field_id = true;
```
Tishj: You can't do this without adjusting `api.yaml`
In general, can we not do this?
Everything in `rest_catalog/` should be considered immutable

src/storage/table_update/common.cpp
```diff
@@ -74,6 +78,69 @@ void AssertCreateRequirement::CreateRequirement(DatabaseInstance &db, ClientCont
...
+	req.assert_last_assigned_field_id.last_assigned_field_id = last_assigned_column_field_id;
+}
+
+AssertLastAssignedPartitionIdRequirement::AssertLastAssignedPartitionIdRequirement(IcebergTableInformation &table_info)
+    : IcebergTableRequirement(IcebergTableRequirementType::ASSERT_LAST_ASSIGNED_PARTITION_ID, table_info) {
+	if (table_info.table_metadata.HasLastPartitionId()) {
+		last_assigned_partition_id = table_info.table_metadata.GetLastPartitionFieldId();
+	} else {
+		// If no partition field IDs have been assigned, use 999 as the base (partition fields start at 1000)
+		last_assigned_partition_id = 999;
```
Tishj, line 126: Can we link to the spec for justification of using 999?
Because I'm not following why 999 and not 0

src/storage/iceberg_insert.cpp
```diff
@@ -114,6 +126,15 @@ static void AddToColDefMap(case_insensitive_map_t<optional_ptr<IcebergColumnDefi
 	}
 }
 
+//! Get the name for a partition expression (used in hive-style paths)
+static string GetPartitionExpressionName(const IcebergPartitionSpecField &field) {
+	// Use the partition field name directly
+	if (field.transform.Type() == IcebergTransformType::IDENTITY) {
+		return field.name;
```
Tishj: Is there any normalization that needs to be put on this?

src/storage/iceberg_insert.cpp
```diff
@@ -184,7 +205,7 @@ void IcebergInsert::AddWrittenFiles(IcebergInsertGlobalState &global_state, Data
 		auto column_stats = chunk.GetValue(4, r);
 		auto &map_children = MapValue::GetChildren(column_stats);
 
-		global_state.insert_count += data_file.record_count;
+		auto partition_values = chunk.GetValue(5, r);
```
Tishj, line 223: Can you put a comment linking to the place where we define this order?

src/storage/iceberg_insert.cpp
```diff
@@ -194,6 +215,56 @@ void IcebergInsert::AddWrittenFiles(IcebergInsertGlobalState &global_state, Data
...
 
+		auto ic_partition_info = ic_table.table_info.table_metadata.GetLatestPartitionSpec();
+
+		// Build a map from partition column name to its partition spec field
+		// To be used later to add partitioning info to the data file
+		case_insensitive_map_t<reference<const IcebergPartitionSpecField>> partition_colname_to_field;
+		for (auto &partition_field : ic_partition_info.fields) {
+			auto partition_col_name = GetPartitionExpressionName(partition_field);
+			bool partition_col_found = false;
+			for (auto &column : ic_schema->columns) {
```
Tishj: This feels like a duplicate of `GetColumnIndexBySourceId`

src/metadata/iceberg_column_definition.cpp
```diff
@@ -196,4 +196,17 @@ bool IcebergColumnDefinition::IsIcebergPrimitiveType() {
 	}
 }
 
+unique_ptr<IcebergColumnDefinition> IcebergColumnDefinition::Copy() const {
+	auto res = make_uniq<IcebergColumnDefinition>();
+	res->id = id;
+	res->name = name;
+	res->type = type;
+	res->initial_default = initial_default;
```
Tishj: `write_default` was just added, in case it doesn't raise a merge conflict anywhere and we find a bug months later 😅

src/include/storage/iceberg_insert.hpp
```diff
@@ -29,6 +31,12 @@ struct IcebergCopyInput {
 	string data_path;
 	//! Set of (key, value) options
 	case_insensitive_map_t<vector<Value>> options;
+	//! Partition specification for the table (if partitioned)
+	optional_ptr<const IcebergPartitionSpec> partition_spec;
+	//! Table schema for looking up source columns by ID
+	optional_ptr<IcebergTableSchema> table_schema;
```
Tishj, line 37: nit: there's already a `const IcebergTableSchema &schema` ?

vcpkg.json
```diff
@@ -21,8 +21,8 @@
 		"registries": [
 			{
 				"kind": "git",
-				"repository": "https://github.com/duckdb/vcpkg-duckdb-ports",
-				"baseline": "02558971ebafdbaa697a0704a3ed7ba365cd5495",
+				"repository": "https://github.com/Tmonster/vcpkg-duckdb-ports",
```
Tishj, line 24: Doesn't this need to be changed before we merge?

Tishj: though if it passes CI then I'd prefer merging sooner rather than later
we can push fixes in a follow up

src/iceberg_extension.cpp
```diff
@@ -80,6 +81,13 @@ static void LoadInternal(ExtensionLoader &loader) {
 	OAuth2Authorization::SetCatalogSecretParameters(secret_function);
 	loader.RegisterFunction(secret_function);
 
+	// Register DATE -> INTEGER implicit cast.
+	// Spark writes day-transform partition values as Avro 'date' logical type (int32 days since epoch) when it should
+	// be int. DuckDB's date_t is also int32 days since epoch, so this is a physical no-op reinterpretation. Cost 200
+	// keeps it low-priority so it doesn't interfere with standard casts elsewhere.
+	loader.RegisterCastFunction(LogicalType::DATE, LogicalType::INTEGER, BoundCastInfo(DefaultCasts::ReinterpretCast),
```
hannes, line 88: haaaaa

PR #745, opened 2026-02-26
Support row lineage on upgraded table (V2 -> V3)

Tishj:
This PR is a follow up to https://github.com/duckdb/duckdb-iceberg/pull/721

There's a couple of things to keep in mind specifically when dealing with tables that were upgraded to V3, this PR deals with those details.

- `IcebergAvroMultiFileReader::FinalizeChunk` for `manifest_entry` scans should only assign `first_row_id` to records that aren't deleted.
- Save the scanned `IcebergManifestFile` entries in the transaction data, so we can save the modifications made to it for upgraded tables[0]
- ~~Backfill row ids in Scan for V2 snapshots, when the table version is V3~~
- Push a `COALESCE(_row_id, manifest_entry.first_row_id + _pos)` for the `_row_id` virtual column[1].
- Push a `COALESCE(_last_updated_sequence_number, manifest_entry.sequence_number)` for the `_last_updated_sequence_number` virtual column). (no longer handled by `FinalizeChunk` at all any more)
- `manifest_file` is allowed to be missing a `first_row_id` (V2 support)
- `manifest_entry` is allowed to be missing a `first_row_id` (V2 support)

[0] For upgraded tables, we need to assign `manifest_file.first_row_id` if it's NULL (if the snapshot came from V2) on write.
To make this work for transaction-local changes, we save this in the transaction data, this way it's both available in further scans in the transaction and in the commit of the transaction.

[1] This is not entirely supported by the MultiFileReader, on account of requiring 2 local columns to satisfy one global column (real `_row_id` + virtual `COLUMN_IDENTIFIER_FILE_ROW_NUMBER`)
As a result, this has some buggy behavior when it comes to pushed down filters, made a PR to core so we can disable filter pushdown for this: https://github.com/duckdb/duckdb/pull/21119


src/iceberg_functions/iceberg_multi_file_list.cpp
```diff
@@ -801,18 +801,36 @@ void IcebergMultiFileList::InitializeFiles(lock_guard<mutex> &guard) const {
...
+			                                   ? IcebergUtils::GetFullPath(iceberg_path, snapshot.manifest_list, fs)
+			                                   : snapshot.manifest_list;
+			//! Read the manifest list
+			auto scan = AvroScan::ScanManifestList(snapshot, metadata, context, manifest_list_full_path);
+			auto manifest_list_reader = make_uniq<manifest_list::ManifestListReader>(*scan);
+			while (!manifest_list_reader->Finished()) {
+				manifest_list_reader->Read(STANDARD_VECTOR_SIZE, manifest_files);
+			}
+			if (metadata.iceberg_version >= 3 && !snapshot.has_first_row_id) {
+				//! Backfill the row lineage for V2 snapshots, required for correct UPDATE behavior
```
Tishj: This also makes it so that we add row ids when we scan from a snapshot that was made by V2.
Which is technically wrong (should be NULL)
But we need this behavior to make UPDATE work correctly, as we need to assign a row id to the row we want to update, in order to materialize it (by writing the `_row_id` in the parquet file)

src/iceberg_manifest_list.cpp
```diff
@@ -144,10 +144,18 @@ void WriteToFile(const IcebergTableMetadata &table_metadata, const IcebergManife
 	}
 
 	//! Populate the DataChunk with the manifests
-	auto manifest_files = manifest_list.GetManifestFilesConst();
+	auto &manifest_files = manifest_list.GetManifestFilesConst();
 	DataChunk data;
 	data.Initialize(allocator, types, manifest_files.size());
 
+	idx_t next_row_id;
+	auto latest_snapshot = table_metadata.GetLatestSnapshot();
```
Tishj: unused?

src/iceberg_functions/iceberg_multi_file_list.cpp
Tishj: No, we don't need this to make UPDATE work
For an UPDATE, we need to materialize the `_row_id` that gets inferred in the scan path. That happens to be NULL for an upgraded table, but that's still correct.

Because it's NULL, it will receive the `first_row_id` from the `manifest_file` we create as part of the update (or from the `manifest_entry` if we rewrite an existing manifest to add it*)

\* not something we do currently, but in the future we probably will have support for a "smarter" balanced snapshot creation.

PR #765, opened 2026-03-05
Expose parquet copy options

yan-alex:
https://github.com/duckdblabs/duckdb-internal/issues/6713

### DIFF
--- a/test/sql/local/irc_any_catalog/delete/test_deletes_from_multiple_files.test
+++ b/test/sql/local/irc_any_catalog/delete/test_deletes_from_multiple_files.test
@@ -30,7 +30,7 @@ statement ok
 DROP TABLE IF EXISTS my_datalake.default.test;
 
 statement ok
-CREATE TABLE my_datalake.default.test WITH ('write.parquet.compression-codec'='uncompressed') AS SELECT i id FROM range(5000000) t(i);
+CREATE TABLE my_datalake.default.test WITH ('write.parquet.compression-codec'='snappy') AS SELECT i id FROM range(5000000) t(i);
 
 query I
 select count(*) from iceberg_metadata(my_datalake.default.test);
--- a/test/sql/local/irc_any_catalog/delete/test_multiple_deletes_multiple_delete_files.test
+++ b/test/sql/local/irc_any_catalog/delete/test_multiple_deletes_multiple_delete_files.test
@@ -30,7 +30,7 @@ statement ok
 DROP TABLE IF EXISTS my_datalake.default.test;
 
 statement ok
-CREATE TABLE my_datalake.default.test WITH ('write.parquet.compression-codec'='uncompressed') AS SELECT i id FROM range(5000000) t(i);
+CREATE TABLE my_datalake.default.test WITH ('write.parquet.compression-codec'='snappy') AS SELECT i id FROM range(5000000) t(i);
 
 query I
 select count(*) from iceberg_metadata(my_datalake.default.test);


src/storage/iceberg_insert.cpp
```diff
@@ -395,6 +385,17 @@ vector<IcebergManifestEntry> IcebergInsert::GetInsertManifestEntries(IcebergInse
 	return std::move(global_state.written_files);
 }
 
+void ExposeProperty(const unique_ptr<CopyInfo> &copy_info, case_insensitive_map_t<string> table_properties,
```
Tishj: Can we instead define an array of `IcebergParquetOptionMapping` ?

```c++
namespace {

struct IcebergParquetOptionMapping {
	const char *iceberg_option;
	const char *parquet_option;
};

static const IcebergParquetOptionMapping ICEBERG_TABLE_PROPERTY_MAPPING[] = {
    {"write.parquet.row-group-size-bytes", "row_group_size_bytes"},
    {"write.parquet.compression-codec", "codec"},
     ... // fill with other options
};

static const idx_t ICEBERG_TABLE_PROPERTY_MAPPING_SIZE = sizeof(ICEBERG_TABLE_PROPERTY_MAPPING) / sizeof(IcebergParquetOptionMapping);

} // namespace
```

So we can iterate through the mapping options, do a find in the `table_properties` map and transform it to the duckdb parquet option

(see https://iceberg.apache.org/docs/1.10.0/configuration/?h=commit.retry.num+retries#write-properties for the iceberg table property names)

src/storage/iceberg_insert.cpp
```diff
@@ -380,6 +381,35 @@ vector<IcebergManifestEntry> IcebergInsert::GetInsertManifestEntries(IcebergInse
...
+    {"write.parquet.compression-codec", "codec"},
+    {"write.parquet.compression-level", "compression_level"},
+    {"write.parquet.dict-size-bytes", "string_dictionary_page_size_limit"},
+    {"write.parquet.row-group-size", "row_group_size"},
+    {"write.parquet.page-size-bytes", "chunk_size"},
+    {"write.parquet.row-groups-per-file", "row_groups_per_file"},
+    // Parquet native options
+    {"row_group_size", "row_group_size"},
+    {"chunk_size", "chunk_size"},
+    {"row_groups_per_file", "row_groups_per_file"}};
```
Tishj: This is covered already by `write.parquet.row-groups-per-file`, no?

test/sql/local/irc_any_catalog/delete/test_deletes_from_multiple_files.test
```diff
@@ -30,29 +30,28 @@ statement ok
 DROP TABLE IF EXISTS my_datalake.default.test;
 
 statement ok
-CREATE TABLE my_datalake.default.test AS SELECT i id FROM range(5000000) t(i);
+CREATE TABLE my_datalake.default.test WITH ('write.parquet.compression-codec'='zstd') AS SELECT i id FROM range(5000000) t(i);
```
Tishj: The test is for multiple files, I don't understand why we're adding options here that break that expectation

test/sql/local/irc_any_catalog/delete/test_multiple_deletes_multiple_delete_files.test
```diff
@@ -30,30 +30,30 @@ statement ok
 DROP TABLE IF EXISTS my_datalake.default.test;
 
 statement ok
-CREATE TABLE my_datalake.default.test AS SELECT i id FROM range(5000000) t(i);
+CREATE TABLE my_datalake.default.test WITH ('write.parquet.compression-codec'='zstd') AS SELECT i id FROM range(5000000) t(i);
```
Tishj: Same as above

test/sql/local/irc_any_catalog/delete/test_deletes_from_multiple_files.test
yan-alex: Fixture exposes this, nessie and others don't. That would result in the number of parquet files being different. Maybe we should expose zstd as default in our code?

Tishj: I don't really know what you mean by "fixture exposes this"
Do you mean the table is created with that table property by default? If so we should explicitly set it to `uncompressed`, and have a dedicated table where we explicitly set it to `zstd` to check the difference

yan-alex: Yes correct, fixture exposes this by default:
```
from iceberg_table_properties(my_datalake.default.test);
key	value	
VARCHAR	VARCHAR	
[ Rows: 2]
created-at	2026-03-06T15:40:41.274292459Z
write.parquet.compression-codec	zstd
```
[I tried now to expose `zstd` by default as a codec](https://github.com/duckdb/duckdb-iceberg/pull/765/changes/113abfafb1c9ee5c2deecea10f7da4d0b645e05c#diff-b0f5d1c7aa6e5eb94e83a68af38db9355dd221bb43064b606665ec0cdd598e43R429) but I'm now backtracking because, if I'm not mistaken this is up to the catalog to decide. We should not be making choices on this ourselves right?

scripts/data_generators/tests/insert_test_big/q00.sql
```diff
@@ -4,5 +4,6 @@ CREATE or REPLACE TABLE default.insert_test_big (
 )
 TBLPROPERTIES (
     'format-version'='2',
-    'write.update.mode'='merge-on-read'
+    'write.update.mode'='merge-on-read',
+    'write.parquet.compression-codec'='zstd'
```
yan-alex: `uncompressed`

scripts/data_generators/tests/insert_test_big/q00.sql
```diff
@@ -4,5 +4,6 @@ CREATE or REPLACE TABLE default.insert_test_big (
 )
 TBLPROPERTIES (
     'format-version'='2',
-    'write.update.mode'='merge-on-read'
+    'write.update.mode'='merge-on-read',
+    'write.parquet.compression-codec'='uncompressed'
```
Tishj: I was wrong, it's not `uncompressed` by default in duckdb, it's snappy:
```c++
duckdb_parquet::CompressionCodec::type codec = duckdb_parquet::CompressionCodec::SNAPPY;
```

PR #766, opened 2026-03-05
Implement truncate transform

jvansanten:
I've been thoroughly enjoying the iceberg extension, but now find myself needing to partition my dataset on integer-valued columns. The extension handles this gracefully if inefficiently at the moment, simply matching all partitions based on unimplemented transforms, and kicking the filtering down to the parquet scan.

This PR adds support for the [truncate](https://iceberg.apache.org/spec/#truncate-transform-details) transform, allowing iceberg_scan to prune partitions created with `truncate` at the manifest level. Truncate is defined (and implemented here) for int, long, decimal, bytes, and string types. All other types raise an error.

New sqllogic tests verify that scans return the correct results, and also skip files as intended.

It should be noted that [`iceberg_scan` can't actually read tables truncated-partitioned on decimal fields](https://github.com/duckdb/duckdb-iceberg/pull/766#issuecomment-4012347455), but will be able to after #744.

### DIFF
--- a/test/sql/local/irc_any_catalog/reads/partitioned_reads/truncate/truncate_binary.test
+++ b/test/sql/local/irc_any_catalog/reads/partitioned_reads/truncate/truncate_binary.test
@@ -9,8 +9,10 @@ require iceberg
 
 require-env DUCKDB_ICEBERG_HAVE_GENERATED_DATA
 
+require-env SECRETS_CREATED_AND_CATALOG_ATTACHED
+
 statement ok
-create view scan as select * from ICEBERG_SCAN('__WORKING_DIRECTORY__/data/generated/iceberg/spark-local/default/truncate_partitioned_binary');
+create view scan as select * from my_datalake.default.truncate_partitioned_binary;
 
 query I
 select count(*) from scan;
--- a/test/sql/local/irc_any_catalog/reads/partitioned_reads/truncate/truncate_bigint.test
+++ b/test/sql/local/irc_any_catalog/reads/partitioned_reads/truncate/truncate_bigint.test
@@ -9,8 +9,10 @@ require iceberg
 
 require-env DUCKDB_ICEBERG_HAVE_GENERATED_DATA
 
+require-env SECRETS_CREATED_AND_CATALOG_ATTACHED
+
 statement ok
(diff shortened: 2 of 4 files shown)


Tishj:
I believe they check the amount of logged parquet files that are scanned, which is affected by file pruning
```sql
query I
select count(distinct request.url) from duckdb_logs_parsed('HTTP') where request.url like '%.parquet';
----
6
```

But you can also look at `in_filter.test`:
```sql
query I
select count(*) FROM duckdb_logs() where type = 'Iceberg' and message.contains('data_file')
----
4
```

Which originates from here:
```c++
			// Check whether current data file is filtered out.
			if (!table_filters.filters.empty() && !FileMatchesFilter(manifest_entry, IcebergDataFileType::DATA)) {
				DUCKDB_LOG(context, IcebergLogType, "Iceberg Filter Pushdown, skipped 'data_file': '%s'",
				           data_file.file_path);
				//! Skip this file
				continue;
			}
```

There is also a `manifest_file` equivalent:
```sql
query I
select count(*) FROM duckdb_logs() where type = 'Iceberg' and message.contains('manifest_file')
----
4
```
Comes from here:
```c++
			if (!ManifestMatchesFilter(manifest_file)) {
				DUCKDB_LOG(context, IcebergLogType, "Iceberg Filter Pushdown, skipped 'manifest_file': '%s'",
				           manifest_file.manifest_path);
				//! Skip this manifest
				continue;
			}
```

test/sql/local/irc_any_catalog/reads/partitioned_reads/truncate/truncate_binary.test
```diff
@@ -0,0 +1,44 @@
...
+# group: [truncate]
+
+require avro
+
+require parquet
+
+require iceberg
+
+require-env DUCKDB_ICEBERG_HAVE_GENERATED_DATA
+
```
Tmonster, line 11: You can move the test files to `test/sql/local/irc_any_catalog/reads/partitioned_reads`. We have a test config that will automatically attach a catalog with the generated spark tables

jvansanten: Can do. Currently these tests just use iceberg_scan; should they use an attached catalog instead?

Tmonster: yea that would be great thanks 🙏

PR #774, opened 2026-03-09
[Testing] Use `test_env` to define required environment variables for tests

Tishj:
With https://github.com/duckdb/duckdb/pull/21099 we can now see environment variables defined in a `--test-config` file in a `require-env` statement.

This allows us to replace the `SECRETS_CREATED_AND_CATALOG_ATTACHED` with an environment variable defined in the test configs, so just the presence of the test config is the key to enabling the tests.


Tmonster (approved):
ty!

PR #794, opened 2026-03-12
If vended-credentials do not return a region for S3 storage, throw an Error

yan-alex:
https://github.com/duckdblabs/duckdb-internal/issues/6320


src/storage/iceberg_table_information.cpp
```diff
@@ -99,6 +115,17 @@ static void ParseConfigOptions(const case_insensitive_map_t<string> &config, cas
 	} else {
 		// Default to S3 parsing for backward compatibility
 		ParseS3ConfigOptions(config, options);
+
+		if (options.find("region") == options.end()) {
+			const string region = RetrieveRegion(DBConfig::GetConfig(context));
+
+			if (region.empty()) {
+				throw InvalidConfigurationException(
+				    "No region provided in vended credentials, no default region passed for the "
```
Tmonster: nit:

"No region provided was provided via the vended credentials, and no region could be found via environment variables. Please provide a default_region for the Iceberg Catalog when attaching"

PR #797, opened 2026-03-14
fix: glue default catalog

### LINKED ISSUE #769: Cannot attach the default AWS Glue catalog without providing the account id
## Description
When trying to attach an AWS Glue Catalog using the accounts default catalog (`:`) it returns a 400 bad request.
It looks like the colon is included as the warehouse query parameter as seen in the error message below.

If I provide the account id it works flawless. However, I do not want to require the user to provide the account id, but rather using the aws profile in the secret to dictate the catalog to be attached.
 
```sql
-- Works without issues
ATTACH IF NOT EXISTS '<account_id>' AS iceberg_db (
    TYPE ICEBERG,
    ENDPOINT_TYPE GLUE
);
```

## Reproduction

```sql
FORCE INSTALL iceberg;
LOAD iceberg;
INSTALL httpfs;
LOAD httpfs;

CREATE OR REPLACE SECRET aws_secret (
    TYPE S3,
    PROVIDER CREDENTIAL_CHAIN,
    PROFILE 'dataengineer',
    CHAIN 'sso'
);

-- Fails with 400 Bad Request 
ATTACH IF NOT EXISTS ':' AS iceberg_db (
    TYPE ICEBERG,
    ENDPOINT_TYPE GLUE
);
```

## Expected outcome
The accounts default glue catalog is attached.

## Actual outcome
```bash
Invalid Configuration Error:
Request to 'glue.eu-north-1.amazonaws.com/iceberg/v1/config?warehouse=%3A' returned a non-200 status code (BadRequest_400), with reason: Bad Request
DuckDB v1.4.4 (Andium) 6ddac802ff
```

## Suggested solution
I have tried with curl to just exclude the warehouse parameter and it will then default to the accounts default glue catalog. So the suggested solution would be to check if the provided warehouse is just a `:` and in that case exclude the query parameter from the request

goober:
Fixes #769


src/catalog_api.cpp
```diff
@@ -474,7 +474,9 @@ rest_api_objects::LoadTableResult IRCAPI::CommitNewTable(ClientContext &context,
 rest_api_objects::CatalogConfig IRCAPI::GetCatalogConfig(ClientContext &context, IcebergCatalog &catalog) {
 	auto url_builder = catalog.GetBaseUrl();
 	url_builder.AddPathComponent("config");
-	url_builder.SetParam("warehouse", catalog.warehouse);
+	if (catalog.warehouse != ":") {
+		url_builder.SetParam("warehouse", catalog.warehouse);
```
Tishj: I think this should be handled further up, in `GetConfig`, where we still have the `IcebergEndpointType`.
`GetCatalogConfig` should have a `const string &warehouse` parameter.

We can pass an empty string if the endpoint type is Glue and the `catalog.warehouse` is `":"`.
And the check here should be `!warehouse.empty()` instead.

goober: Thanks for the feedback, I agree with you and have updated the PR accordingly

Issue #707, opened 2026-01-29
Nessie catalog writes ignore `SET s3_*` / `CREATE SECRET` and require AWS env creds (v1.4.4)


aryahassibi:
## Summary
When using DuckDB with an Iceberg REST catalog (Nessie in this case) backed by S3, reads only work after setting S3 credentials via `SET s3_*` (not `CREATE SECRET`), but writes/commits fail with HTTP 403:
```bash
⋮
D INSERT INTO nessie.demo.users VALUES (1, 'ada'), (2, 'grace');

TransactionContext Error:
Failed to commit: Failed to commit Iceberg transaction: HTTP GET error on 'https://<bucket>.s3.us-east-1.amazonaws.com/iceberg/demo/.../metadata/...avro' (HTTP 403)
```
Unless AWS credentials are available through the AWS environment variables (e.g., `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` environment variables). This makes in-session credential configuration insufficient for insert/write operations.

This is similar to some of the issues opened before but also adds some nuance to the problem.

## Desired Behavior
If `SET s3_*` or `CREATE SECRET` is used to configure S3 credentials, both reads and writes should succeed without requiring AWS credentials from the environment or shared credentials file.

## Actual Behavior
- `SELECT` works after `SET s3_*` (not with `CREATE SECRET`).
- `INSERT` works with neither. It fails with 403 unless `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are exported in the process environment.

### Main Path

Assumptions:
- Iceberg REST endpoint is reachable (running Nessie).
- S3 bucket exists and is writable with the credentials used in steps below.
- No AWS credentials are set in the process environment.
- AWS credentials are long-lived root keys (no session token involved).

```sql
INSTALL httpfs; LOAD httpfs;
INSTALL aws; LOAD aws;
INSTALL iceberg; LOAD iceberg;

-- REST catalog
ATTACH 'warehouse' AS nessie
  (TYPE 'iceberg', ENDPOINT 'http://localhost:19120/iceberg', AUTHORIZATION_TYPE 'none');

CREATE SCHEMA IF NOT EXISTS nessie.demo;
CREATE TABLE IF NOT EXISTS nessie.demo.users (id BIGINT, name VARCHAR);

-- Configure S3 creds via SQL (non-env)
SET s3_region='us-east-1';
[...]

PR #800, opened 2026-03-16
S3 credentials can be set

yan-alex:
https://github.com/duckdblabs/duckdb-internal/issues/7475
fixes https://github.com/duckdb/duckdb-iceberg/issues/707


test/sql/local/irc_specific_catalogs/nessie/s3_credentials_can_be_set.test
```diff
@@ -0,0 +1,46 @@
...
+create secret my_secret (
+    TYPE ICEBERG,
+    CLIENT_ID 'client1',
+    CLIENT_SECRET 's3cr3t',
+    OAUTH2_SCOPE 'catalog sign',
+    OAUTH2_SERVER_URI 'http://127.0.0.1:8080/realms/iceberg/protocol/openid-connect/token'
+);
+
+statement ok
+ATTACH 'warehouse' AS nessie (TYPE 'iceberg', ENDPOINT 'http://127.0.0.1:19120/iceberg/', SECRET 'my_secret');
```
Tmonster: Should we also add a `ACCESS_DELEGATION_MODE 'none'` here?

yan-alex: So tried adding that but then the test hangs forever. If I remove this 
```patch
diff --git a/src/storage/catalog/iceberg_table_entry.cpp b/src/storage/catalog/iceberg_table_entry.cpp
index bacb44a1..0d689118 100644
--- a/src/storage/catalog/iceberg_table_entry.cpp
+++ b/src/storage/catalog/iceberg_table_entry.cpp
@@ -49,10 +49,6 @@ void IcebergTableEntry::PrepareIcebergScanFromEntry(ClientContext &context) cons
 	auto &ic_catalog = catalog.Cast<IcebergCatalog>();
 	auto &secret_manager = SecretManager::Get(context);
 
-	if (ic_catalog.attach_options.access_mode != IRCAccessDelegationMode::VENDED_CREDENTIALS) {
-		// assume secret already exists
-		return;
-	}
 	// Get Credentials from IRC API
 	auto table_credentials = table_info.GetVendedCredentials(context);
 	auto metadata_path = table_info.table_metadata.GetMetadataPath();
```
then the test passes, but i'm sure this logic is there for a reason... Any ideas? 🤔

PR #815, opened 2026-03-18
Skip bounds checking for EXPRESSION_FILTER on strings

yan-alex:
https://github.com/duckdblabs/duckdb-internal/issues/8469
TableFilterType::EXPRESSION_FILTER on strings (e.g len(my_string_col)) do not maintain lexicographic ordering properties. Therefore we skip bounds checking for those, as it makes no sense. That is because len(my_string_col) would return 16, as the bounds for strings are truncated to 16 long.


test/sql/local/irc_any_catalog/reads/should_not_do_filter_pruning_on_expression_string_filters.test
```diff
@@ -0,0 +1,37 @@
...
+attach ':memory:' as my_datalake;
+
+statement ok
+create schema my_datalake.default;
+
+statement ok
+SET unsafe_enable_version_guessing = true;
+
+query I
+SELECT item_id FROM ICEBERG_SCAN('__WORKING_DIRECTORY__/data/generated/iceberg/spark-local/default/issue_791') WHERE length(item_id) = 36;
```
Tmonster: I think this test can also be in `test/sql/local/irc_any_catalog/reads` right?

yan-alex: Well this test does  `attach ':memory:' as my_datalake;`, and all tests in `test/sql/local/irc_any_catalog` assume an attach is done via --test-config. I could make a clone of this test without the attach and add it to `test/sql/local/irc_any_catalog`... But then many more tests can go there I believe, currently all test with`ICEBERG_SCAN()` are ran with fixture only. I guess I don't fully grasp the mechanics of `ICEBERG_SCAN()`.

Is a bigger move of these ICEBERG_SCAN tests to `test/sql/local/irc_any_catalog` logical? Is there even any request being made to a catalog? I added a proxy and see no requests while running this test.

Tishj: `iceberg_scan` is for local iceberg tables, not tables managed by a catalog

test/sql/local/partitioning/foldable_expression_filter.test
```diff
@@ -56,7 +56,7 @@ where
 query I
 select count(*) FROM duckdb_logs() where type = 'Iceberg' and message.contains('data_file')
 ----
-5
+0
```
Tishj, line 59: This seems to make it so that we're never skipping files anymore for equality filters on strings, that doesn't sound correct to me?

src/planning/pruning/iceberg_predicate.cpp
```diff
@@ -209,6 +209,11 @@ bool MatchBoundsTemplated(ClientContext &context, const TableFilter &filter, con
 		case ExpressionType::COMPARE_LESSTHAN:
 		case ExpressionType::COMPARE_LESSTHANOREQUALTO:
 		case ExpressionType::COMPARE_EQUAL: {
+			// TableFilterType::EXPRESSION_FILTER on strings (e.g len(my_string_col)) do not maintain lexicographic
+			// ordering properties
+			if (stats.lower_bound.type() == LogicalType::VARCHAR) {
+				return true;
```
Tishj, line 215: This seems too limiting, we should probably inspect the expression and check if it's not just a BoundReferenceExpression, and only if that's the case we skip

test/sql/local/partitioning/foldable_expression_filter.test
yan-alex: Only when using expression filters, e.g. `len()`. In this test, `lower()` is used in query. I'm not sure if we should make an exception for certain expression filters, but `len()` surely doesn't work because the `upper_bound` is the truncated string to max 16 characters

Tmonster: `lower()` is also not safe and does not preserve ordering. Consider the following example. 
A file has 'zz' as the min and 'AA' as the max value (since lowercase is lexicographically before uppercase). 

Applying lower() on the min/max to match bounds will actually flip the values and cause the file to be skipped
`lower(zz) = 'zz'` 
`lower(AA) = 'aa'`

So the transformed values on the bounds are now flipped and it looks like the max value is less than the min. If I'm not mistaken, we only do strict bounds checking for strings, so only `<, <=, =, >=, >`.

Might actually be good to add a test case for this

Tishj: I think any function expression should void the pruning behavior
But now (I believe) it's much more strict than that even, any filter on a column of type VARCHAR gets skipped

Tmonster: @yan-alex can we add a test where we ensure filter pruning is still happening on VARCHAR columns when there is only a constant comparision filter (i.e <,<=,=,>=,>). I think 
```
select
	count(*)
from
	my_datalake.default.lineitem_partitioned_l_shipmode
where
	l_shipmode = 'RAIL'
```

Should still filter out 5 other files

PR #821, opened 2026-03-19
`make catalog` usability tweaks

yan-alex:
After using the new `make <catalog>` setup a bit, I felt like it could use some improvements. Summary of changes:
- Use dashes "-" instead of underscores "_" on the make commands for more ergonomic / faster typing. e.g. `make fixture-data`
- I noticed I spin up the catalog by itself without generated data more often than with generated data. Plus, generating data takes very long. Therefore, `make fixture` now simply spins it up, and `make fixture-data` spins it up AND generates data. This feels more usable.
- The previous command to spin up fixture also used to do `rm -rf data/generated` which would prompt a password on my laptop, which was annoying. Now this file removal only happens if the generate data command is ran.
- Added the possibility of passing a test argument to the generate data command, like so:
`TEST=all_types_table make fixture-data` 
which is then passed on to the generate_data script. The script will now only generate the needed data for that single test, which is faster.

### DIFF
--- a/make/catalogs/nessie.mk
+++ b/make/catalogs/nessie.mk
@@ -18,7 +18,7 @@ nessie: nessie-clone nessie-stop
 	(cd .catalogs/nessie/docker/catalog-auth-s3 && docker compose up -d)
 	$(call set_active_catalog,nessie)
 
-nessie-data_only:
+nessie-data-only:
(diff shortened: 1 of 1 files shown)


Tishj (approved):
Thanks, looks good to me 👍

PR #833, opened 2026-03-20
[Dev] Rework manifest scan tasks

### LINKED ISSUE #832: Improve `IcebergManifestEntry` parallel scan logic
Currently in `IcebergMultiFileList` we use a bunch of buffers, and have one big `vector<IcebergManifestEntry>` that we pool all entries we want to scan into.
This was done to enable scanning manifests in parallel, while still being able to start a scan before all manifests are scanned.
If the vector that owns the `IcebergManifestEntry` gets resized, the references to that vector become invalid.

But, the `manifest_file` tells us exactly how many files are inside of the manifest, and we can use this information to reserve for the total amount up front, guaranteeing that the references never become stale.
When we do that, we can simplify the logic a lot and just let each scan task add to the `vector<IcebergManifestEntry>` of the `IcebergManifestListEntry` that it's scanning.

Tishj:
This PR fixes #832 

Previously we had a `current_manifest_entries` vector, that all scan tasks would write to.
Inside of `GetDataFile` we would consume the `current_manifest_entries` buffer and move it into `manifest_entries` or discard it, if the data file is filtered out.

Now we put the entries of each manifest into the corresponding `IcebergManifestListEntry`, and we push `ManifestReadBatch` into a shared queue instead, which contain the index of the list entry and the start+end index of the entries to read.
Inside of `GetDataFile` we consume this queue and push a *reference* to the manifest entry into the `data_manifest_entries` vector.

This greatly simplifies the scanning logic, and reduces copies of the `IcebergManifestEntry` by a lot.
This now even allows us to cache these scans easier in the future, as the `IcebergManifestListEntry` items produced are identical to those we use for transactional (uncommitted) data.

### DIFF
--- a/src/planning/iceberg_multi_file_list.cpp
+++ b/src/planning/iceberg_multi_file_list.cpp
@@ -375,7 +375,7 @@ void IcebergMultiFileList::GetStatistics(vector<PartitionStatistics> &result) co
 		return;
 	}
 
-	if (!transaction_delete_manifests.empty() || !delete_manifests.empty()) {
+	if (!delete_manifests.empty()) {
 		//! if exist delete_manifests, return;
 		return;
 	}


Tishj:
Ah, made an error
`ManifestReadTask` now takes an `idx_t`, but there's nothing guaranteeing that it receives that manifest to scan, only the `reader.file_list_idx.GetIndex();` inside `IcebergAvroMultiFileReader::FinalizeChunk` has this information.

Perhaps we can simplify this even more, and populate the vector directly in `FinalizeChunk`, putting the `IcebergManifestReader` and `IcebergManifestListReader` directly in the `IcebergAvroMultiFileReader`

src/planning/metadata_io/manifest/iceberg_manifest_reader.cpp
```diff
@@ -194,24 +180,22 @@ idx_t ManifestReader::ReadChunk(idx_t offset, idx_t count, vector<IcebergManifes
...
-
+		idx_t index = i;
 		IcebergManifestEntry entry;
 
 		entry.status = (IcebergManifestEntryStatusType)status_data[index];
-		if (this->skip_deleted && entry.status == IcebergManifestEntryStatusType::DELETED) {
-			//! Skip this entry, we don't care about deleted entries
-			continue;
-		}
+		// if (this->skip_deleted && entry.status == IcebergManifestEntryStatusType::DELETED) {
```
Tishj: TODO: remove this dead code, this is handled elsewhere now

Issue #841, opened 2026-03-23
Better caching of read manifests


Tishj:
Currently we have:
- `transaction_data_manifests` - fully cached `IcebergManifestListEntry` vectors.
- `committed_data_manifests` - entirely uncached `IcebergManifestListEntry` vectors.

`transaction_data_manifests` are uncommitted, not written to disk yet, so nothing has to be read from disk.
`committed_data_manifests` are read again for each scan.

In the committed variant we can probably introduce some caching, so subsequent scans of the same manifests are cached, this doesn't even have to be tied to a transaction, because manifests are immutable, we should just tie it to the `manifest_file.manifest_path`.

PR #859, opened 2026-03-25
[Dev] First step towards caching `IcebergManifest`

Tishj:
First step in #841 

Introduce:
```c++
struct IcebergManifest : public ObjectCacheEntry {
public:
	IcebergManifest(idx_t expected_size) : expected_size(expected_size) {
		manifest_entries.reserve(expected_size);
	}
	IcebergManifest(vector<IcebergManifestEntry> &&entries) : manifest_entries(std::move(entries)) {
		expected_size = manifest_entries.size();
	}

public:
	string GetObjectType() override {
		return "iceberg_manifest";
	}

	//! Get the rough cache memory usage in bytes for this entry.
	//! Used for eviction decisions. Return invalid index to prevent eviction.
	optional_idx GetEstimatedCacheMemory() const override {
		return manifest_entries.size() * sizeof(IcebergManifestEntry);
	}

public:
	bool CanCache() const {
		return expected_size == manifest_entries.size();
	}
	shared_ptr<IcebergManifest> Copy() const {
		auto copy = make_shared_ptr<IcebergManifest>(expected_size);
		copy->manifest_entries = manifest_entries;
		return copy;
	}

public:
	idx_t expected_size;
	vector<IcebergManifestEntry> manifest_entries;
};
```

Wraps the `vector<IcebergManifestEntry>` in `shared_ptr<IcebergManifest>` so it can be entered into the object cache (not done in this PR yet)
Also adds getter/setters to help make sure that shared state isn't altered.

### Cache Awareness

The `IcebergManifestFileScanInfo` has been made aware of caching, and it will skip scanning manifests that are already populated (`CanCache()` returns true).
This currently happens for `existing_manifest_list`, which gets populated in a transaction when a new snapshot is added.

### DIFF
--- a/src/planning/iceberg_multi_file_list.cpp
+++ b/src/planning/iceberg_multi_file_list.cpp
@@ -355,13 +355,13 @@ unique_ptr<NodeStatistics> IcebergMultiFileList::GetCardinality(ClientContext &c
 
 	idx_t cardinality = 0;
 	for (idx_t i = 0; i < data_manifests.size(); i++) {
-		auto &manifest = data_manifests[i].entry.file;
-		cardinality += manifest.added_rows_count;
-		cardinality += manifest.existing_rows_count;
+		auto &manifest_file = data_manifests[i].entry.ManifestFile();
+		cardinality += manifest_file.added_rows_count;
+		cardinality += manifest_file.existing_rows_count;
 	}
 	for (idx_t i = 0; i < delete_manifests.size(); i++) {
-		auto &manifest = delete_manifests[i].entry.file;
-		cardinality -= manifest.added_rows_count;
+		auto &manifest_file = delete_manifests[i].entry.ManifestFile();
+		cardinality -= manifest_file.added_rows_count;
 	}
 	return make_uniq<NodeStatistics>(cardinality, cardinality);
 }
@@ -373,9 +373,9 @@ const BoundIcebergManifestEntry &IcebergMultiFileList::GetManifestEntry(idx_t fi
 const IcebergManifestFile &IcebergMultiFileList::GetManifestFileForEntry(const BoundIcebergManifestEntry &entry,
                                                                          IcebergManifestContentType type) const {
 	if (type == IcebergManifestContentType::DATA) {
-		return data_manifests[entry.manifest_file_idx].entry.file;
+		return data_manifests[entry.manifest_file_idx].entry.ManifestFile();
 	} else {
-		return delete_manifests[entry.manifest_file_idx].entry.file;
+		return delete_manifests[entry.manifest_file_idx].entry.ManifestFile();
 	}
 }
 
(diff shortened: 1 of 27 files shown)


src/planning/iceberg_multi_file_list.cpp
```diff
@@ -876,6 +876,13 @@ void IcebergMultiFileList::InitializeFiles(lock_guard<mutex> &guard) const {
 		if (!committed_delete_manifests.empty()) {
 			delete_manifest_scan = AvroScan::ScanManifest(snapshot, committed_delete_manifests, options, fs,
 			                                              iceberg_path, metadata, context);
+			for (auto &manifest_list_entry : committed_delete_manifests) {
+				// reserve upfront → guarantees no reallocation
+				auto &file = manifest_list_entry.ManifestFile();
+				idx_t reserve_size = file.existing_files_count + file.added_files_count + file.deleted_files_count;
+				manifest_list_entry.SetManifest(make_shared_ptr<IcebergManifest>(reserve_size));
```
Tishj: Perhaps we should do this as part of `IcebergManifestListEntry` creation, rather than a separate step

src/planning/metadata_io/avro/iceberg_avro_multi_file_list.cpp
```diff
@@ -23,20 +23,34 @@ IcebergManifestFileScanInfo::IcebergManifestFileScanInfo(const IcebergTableMetad
...
+		if (manifest.CanCache()) {
+			if (read_state_p) {
+				//! Push a batch for this manifest directly
+				auto &read_state = *read_state_p;
+				read_state.PushBatch(ManifestReadBatch(i, 0, manifest.manifest_entries.size()));
+			}
+			continue;
+		}
+
+		files_to_scan.emplace_back(i, manifest_list_entry);
```
Tishj, line 44: This introduces indirection which is used by FinalizeChunk, we only push entries that have not been scanned (aren't cached) to `files_to_scan` now.

src/planning/metadata_io/avro/iceberg_avro_multi_file_list.cpp
```diff
@@ -23,20 +23,34 @@ IcebergManifestFileScanInfo::IcebergManifestFileScanInfo(const IcebergTableMetad
...
+      read_state(read_state_p) {
 	unordered_set<int32_t> partition_spec_ids;
-	for (auto &manifest_list_entry : manifest_files) {
-		auto &manifest = manifest_list_entry.file;
-		partition_spec_ids.insert(manifest.partition_spec_id);
+	for (idx_t i = 0; i < manifest_files.size(); i++) {
+		auto &manifest_list_entry = manifest_files[i];
+		auto &manifest = manifest_list_entry.GetManifest();
+		if (manifest.CanCache()) {
+			if (read_state_p) {
```
Tishj, line 36: We should probably log here that the manifest was cached

PR #887, opened 2026-04-03
Add Iceberg View support for REST catalogs

### LINKED ISSUE #314: [Feature Request] Support Iceberg Views
Hey All, 

It would be really nice if we can add iceberg views support for duckdb, presently the iceberg views are multi-dialect which means each engine can store the SQLQuery it in a dialect the engine understands and update and retrive it accordingly. https://iceberg.apache.org/view-spec/#iceberg-view-spec

Engine like Trino / Spark what they do is they call loadView Api which will return back all the view representation of the view and they selectively choose their own dialect and then plumb it to the view resolution framework engine already has, please ref here : 

[1] Trino - https://github.com/trinodb/trino/blob/master/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java#L688
[2] Spark - https://github.com/apache/iceberg/blob/main/spark/v4.0/spark/src/main/java/org/apache/iceberg/spark/source/SparkView.java#L62

and integrate it E2E. Would love to know if you have this planned already

LucaSoato:
## Summary

Implements CREATE VIEW, DROP VIEW, and view loading/querying for Iceberg REST catalogs, following the [Iceberg View Spec](https://iceberg.apache.org/view-spec/).

- Views are stored with dialect `"duckdb"` per the spec's multi-dialect representation model
- When loading, prefers `"duckdb"` dialect, falls back to any available SQL representation
- Views from other engines (Spark, Trino) that can't be parsed are still visible in the catalog and droppable — querying them gives a clear error instead of crashing
- Works with any Iceberg REST catalog (tested against apache/iceberg-rest-fixture)

Closes #314

## Known workaround

`CreateViewInfo::Copy()` in DuckDB core does not copy the `names` field. We patch it after copying. Fix submitted upstream: https://github.com/duckdb/duckdb/pull/21817

## Test plan

- [x] CREATE VIEW, SELECT from view, DROP VIEW
- [x] CREATE VIEW IF NOT EXISTS (idempotent)
- [x] CREATE OR REPLACE VIEW (drop + create in same transaction)
- [x] DROP VIEW IF EXISTS on non-existent view
- [x] View with column aliases
- [x] Views from other engines give clear error on SELECT, still droppable
- [x] Full `irc_any_catalog` test suite passes (169 cases, 83,661 assertions, 0 regressions)

### DIFF
--- a/src/catalog/rest/catalog_entry/schema/iceberg_schema_entry.cpp
+++ b/src/catalog/rest/catalog_entry/schema/iceberg_schema_entry.cpp
@@ -13,6 +13,9 @@
 #include "duckdb/planner/expression_binder/table_function_binder.hpp"
 #include "duckdb/execution/expression_executor.hpp"
 
+#include "duckdb/parser/parsed_data/create_view_info.hpp"
+#include "duckdb/catalog/catalog_entry/view_catalog_entry.hpp"
+
 #include "catalog/rest/catalog_entry/table/iceberg_table_information.hpp"
 #include "catalog/rest/iceberg_catalog.hpp"
 #include "catalog/rest/transaction/iceberg_transaction.hpp"
@@ -106,21 +109,55 @@ void IcebergSchemaEntry::DropEntry(ClientContext &context, DropInfo &info) {
 }
 
 void IcebergSchemaEntry::DropEntry(ClientContext &context, DropInfo &info, bool delete_entry) {
-	auto table_name = info.name;
-	// find if info has a table name, if so look for it in
-	auto table_info_it = tables.GetEntries().find(table_name);
+	auto entry_name = info.name;
+
+	// Handle VIEW_ENTRY drops
+	if (info.type == CatalogType::VIEW_ENTRY) {
+		auto &transaction = IcebergTransaction::Get(context, catalog).Cast<IcebergTransaction>();
+		auto view_key = IcebergTableInformation::GetTableKey(namespace_items, entry_name);
+
+		// Check if view was created in this transaction — just remove from created_views
+		if (transaction.created_views.erase(view_key) > 0) {
+			tables.InvalidateViewCache(entry_name);
+			return;
+		}
+
+		// Load view entries if not yet loaded, then check if view exists
+		tables.LoadViewEntries(context);
+		auto view_it = tables.GetViewEntries().find(entry_name);
+		if (view_it == tables.GetViewEntries().end()) {
+			if (info.if_not_found == OnEntryNotFound::RETURN_NULL) {
+				return;
+			}
+			throw CatalogException("View %s does not exist", entry_name);
+		}
+
+		if (delete_entry) {
+			tables.InvalidateViewCache(entry_name);
+		} else {
+			IcebergTransaction::DeletedViewInfo info;
+			info.namespace_items = namespace_items;
+			info.schema_name = name;
+			info.view_name = entry_name;
+			transaction.deleted_views.emplace(view_key, std::move(info));
+		}
+		return;
+	}
+
+	// Handle TABLE_ENTRY drops (existing logic)
+	auto table_info_it = tables.GetEntries().find(entry_name);
 	if (table_info_it == tables.GetEntries().end()) {
 		if (info.if_not_found == OnEntryNotFound::RETURN_NULL) {
 			return;
 		}
-		throw CatalogException("Table %s does not exist", table_name);
+		throw CatalogException("Table %s does not exist", entry_name);
 	}
 	if (info.cascade) {
 		throw NotImplementedException("DROP TABLE <table_name> CASCADE is not supported for Iceberg tables currently");
 	}
 	if (delete_entry) {
 		// Remove the entry from the catalog
-		tables.GetEntriesMutable().erase(table_name);
+		tables.GetEntriesMutable().erase(entry_name);
 	} else {
 		// Add the table to the transaction's deleted_tables
 		auto &transaction = IcebergTransaction::Get(context, catalog).Cast<IcebergTransaction>();
@@ -153,12 +190,75 @@ optional_ptr<CatalogEntry> IcebergSchemaEntry::CreateIndex(CatalogTransaction tr
 	throw NotImplementedException("Create Index");
 }
 
-string GetUCCreateView(CreateViewInfo &info) {
-	throw NotImplementedException("Get Create View");
-}
-
 optional_ptr<CatalogEntry> IcebergSchemaEntry::CreateView(CatalogTransaction transaction, CreateViewInfo &info) {
-	throw NotImplementedException("Create View");
+	if (info.sql.empty() && !info.query) {
+		throw BinderException("Cannot create view in Iceberg without a query");
+	}
+	auto &context = transaction.GetContext();
+
+	// Handle CREATE OR REPLACE and CREATE IF NOT EXISTS
+	if (info.on_conflict == OnCreateConflict::REPLACE_ON_CONFLICT ||
+	    info.on_conflict == OnCreateConflict::IGNORE_ON_CONFLICT) {
+		auto existing_entry = GetEntry(transaction, CatalogType::VIEW_ENTRY, info.view_name);
+		if (existing_entry) {
+			if (info.on_conflict == OnCreateConflict::IGNORE_ON_CONFLICT) {
+				return existing_entry;
+			}
+			// CREATE OR REPLACE — drop first, then create
+			DropInfo drop_info;
+			drop_info.type = CatalogType::VIEW_ENTRY;
+			drop_info.name = info.view_name;
+			drop_info.cascade = false;
+			drop_info.if_not_found = OnEntryNotFound::RETURN_NULL;
+			DropEntry(context, drop_info, false);
+		}
+	} else {
+		// ERROR_ON_CONFLICT — check existence
+		auto existing_entry = GetEntry(transaction, CatalogType::VIEW_ENTRY, info.view_name);
+		if (existing_entry) {
+			throw CatalogException("View with name \"%s\" already exists", info.view_name);
+		}
+	}
+
+	// Ensure names are populated — the binder populates types+names on the original info,
+	// but as a safety net, fill them in if missing.
+	if (info.types.empty() && info.query) {
+		auto copy = unique_ptr_cast<CreateInfo, CreateViewInfo>(info.Copy());
+		auto bound = CreateViewInfo::FromSelect(transaction.GetContext(), std::move(copy));
+		info.types = bound->types;
+		info.names = bound->names;
+	} else if (info.names.empty() && !info.types.empty()) {
+		for (idx_t i = 0; i < info.types.size(); i++) {
+			if (i < info.aliases.size() && !info.aliases[i].empty()) {
+				info.names.push_back(info.aliases[i]);
+			} else {
+				info.names.push_back("col" + to_string(i));
+			}
+		}
+	}
+
+	// Track the view in the transaction
+	auto &iceberg_transaction = GetICTransaction(transaction);
+	auto view_key = IcebergTableInformation::GetTableKey(namespace_items, info.view_name);
+
+	// Workaround: CreateViewInfo::Copy() does not copy `names`, so we patch it manually.
+	// See: https://github.com/duckdb/duckdb/pull/21817
+	auto view_info = unique_ptr_cast<CreateInfo, CreateViewInfo>(info.Copy());
+	view_info->names = info.names;
+	// Preserve the SELECT SQL — ViewCatalogEntry::Initialize() will move the query out,
+	// so we need the SQL string available at commit time for the REST API request.
+	if (view_info->query) {
+		view_info->sql = view_info->query->ToString();
+	}
+
+	iceberg_transaction.created_views.erase(view_key);
+	iceberg_transaction.created_views.emplace(view_key, std::move(view_info));
+
+	// Invalidate stale caches
+	tables.InvalidateViewCache(info.view_name);
+
+	// Return a pointer to an owned entry (avoid dangling pointer)
+	return tables.GetViewEntry(transaction.GetContext(), info.view_name);
 }
 
 optional_ptr<CatalogEntry> IcebergSchemaEntry::CreateType(CatalogTransaction transaction, CreateTypeInfo &info) {
@@ -735,6 +835,10 @@ void IcebergSchemaEntry::Scan(ClientContext &context, CatalogType type,
 	if (!CatalogTypeIsSupported(type)) {
 		return;
 	}
+	if (type == CatalogType::VIEW_ENTRY) {
+		GetCatalogSet(type).ScanViews(context, callback);
+		return;
+	}
 	GetCatalogSet(type).Scan(context, callback);
 }
 void IcebergSchemaEntry::Scan(CatalogType type, const std::function<void(CatalogEntry &)> &callback) {
@@ -749,13 +853,26 @@ optional_ptr<CatalogEntry> IcebergSchemaEntry::LookupEntry(CatalogTransaction tr
 	}
 	auto &context = transaction.GetContext();
 	auto &ic_catalog = catalog.Cast<IcebergCatalog>();
+
+	// For VIEW_ENTRY, try the view path
+	if (type == CatalogType::VIEW_ENTRY) {
+		auto view_entry = GetCatalogSet(type).GetViewEntry(context, lookup_info.GetEntryName());
+		if (view_entry) {
+			return view_entry;
+		}
+		return nullptr;
+	}
+
+	// For TABLE_ENTRY, use the existing table lookup
 	auto table_entry = GetCatalogSet(type).GetEntry(context, lookup_info);
 	if (!table_entry) {
+		// Try looking up as a view — DuckDB sometimes looks up views as TABLE_ENTRY
+		auto view_entry = GetCatalogSet(type).GetViewEntry(context, lookup_info.GetEntryName());
+		if (view_entry) {
+			return view_entry;
+		}
 		// verify the schema exists
 		if (!IRCAPI::VerifySchemaExistence(context, ic_catalog, name)) {
-			// set exists to false here
-			// we would like to throw an error, but this code is also called when listing schemas,
-			// and throwing an error will abort the listing process.
 			exists = false;
 			return nullptr;
 		}
--- a/src/catalog/rest/catalog_entry/schema/iceberg_schema_entry.cpp
+++ b/src/catalog/rest/catalog_entry/schema/iceberg_schema_entry.cpp
@@ -211,28 +211,19 @@ optional_ptr<CatalogEntry> IcebergSchemaEntry::CreateView(CatalogTransaction tra
 	}
 	auto &context = transaction.GetContext();
 
-	// Handle CREATE OR REPLACE and CREATE IF NOT EXISTS
-	if (info.on_conflict == OnCreateConflict::REPLACE_ON_CONFLICT ||
-	    info.on_conflict == OnCreateConflict::IGNORE_ON_CONFLICT) {
-		auto existing_entry = GetEntry(transaction, CatalogType::VIEW_ENTRY, info.view_name);
-		if (existing_entry) {
-			if (info.on_conflict == OnCreateConflict::IGNORE_ON_CONFLICT) {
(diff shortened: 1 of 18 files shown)


Tmonster:
Hi @LucaSoato,

Thanks for the PR! This is a pretty big feature, but we will take a look soon. Unfortunately, this is a bit too big of a feature to merge for the v1.5.2 release. We will try to keep main compatible with v1.5.2 so you can install a nightly binary to use this feature if/when we merge

LucaSoato:
Hi @Tmonster, 
It's totally understandable!
I'm using it as unsigned extension right now and testing it thoroughly for several use cases.
One thing I'm noticing is that sometimes it gets stuck in the duckdb ui, I think it's related to the fact that the UI uses threads in a different way.
Let me know what I can do to help you to release this feature in 1.5.3, in the meantime I'm happy that my other PR in the duckdb main project was accepted, so we don't need a workaround in CreateViewInfo::Copy().

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -96,9 +99,43 @@ void IcebergSchemaEntry::DropEntry(ClientContext &context, DropInfo &info) {
 }
 
 void IcebergSchemaEntry::DropEntry(ClientContext &context, DropInfo &info, bool delete_entry) {
-	auto table_name = info.name;
-	// find if info has a table name, if so look for it in
-	auto table_info_it = tables.GetEntries().find(table_name);
+	auto entry_name = info.name;
+
+	// Handle VIEW_ENTRY drops
+	if (info.type == CatalogType::VIEW_ENTRY) {
```
Tmonster: nit: can we get a switch statement here for different CatalogTypes? The default case can just throw a `NotImplementedException`

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
Tmonster: nit: Can we check cascade before the View type? I think cascade is not even an Iceberg compatible keywork

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -145,12 +182,75 @@ optional_ptr<CatalogEntry> IcebergSchemaEntry::CreateIndex(CatalogTransaction tr
...
+			}
+		}
+	}
+
+	// Track the view in the transaction
+	auto &iceberg_transaction = GetICTransaction(transaction);
+	auto view_key = IcebergTableInformation::GetTableKey(namespace_items, info.view_name);
+
+	// Workaround: CreateViewInfo::Copy() does not copy `names`, so we patch it manually.
+	// See: https://github.com/duckdb/duckdb/pull/21817
```
Tmonster: Looks like https://github.com/duckdb/duckdb/pull/21819 was merged, so the workaround can be removed?

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -473,6 +478,105 @@ void IcebergTransaction::DoSchemaDeletes(ClientContext &context) {
...
+
+		// default-catalog
+		yyjson_mut_obj_add_strcpy(doc, version_obj, "default-catalog", catalog.GetName().c_str());
+
+		// representations
+		auto repr_arr = yyjson_mut_obj_add_arr(doc, version_obj, "representations");
+		auto repr_obj = yyjson_mut_arr_add_obj(doc, repr_arr);
+		yyjson_mut_obj_add_strcpy(doc, repr_obj, "type", "sql");
+		yyjson_mut_obj_add_strcpy(doc, repr_obj, "sql", view_sql.c_str());
+		yyjson_mut_obj_add_strcpy(doc, repr_obj, "dialect", "duckdb");
```
Tmonster: `duckdb` or `DuckDB` or does it not matter?

src/catalog/rest/iceberg_catalog.cpp
```diff
@@ -325,6 +325,14 @@ void IcebergCatalog::AddDefaultSupportedEndpoints() {
 	supported_urls.insert("POST /v1/{prefix}/tables/rename");
 	// commit updates to multiple tables in an atomic transaction
 	supported_urls.insert("POST /v1/{prefix}/transactions/commit");
+	// list all view identifiers
+	supported_urls.insert("GET /v1/{prefix}/namespaces/{namespace}/views");
```
Tmonster: Views are technically not in the list of default supported endpoints, so I think this should be removed

https://github.com/apache/iceberg/blob/8ac703067a2adfae1928748712dc1d47dbc3c22b/open-api/rest-catalog-open-api.yaml#L106

src/catalog/rest/iceberg_table_set.cpp
```diff
@@ -348,4 +350,196 @@ optional_ptr<CatalogEntry> IcebergTableSet::GetEntry(ClientContext &context, con
...
+	if (created_it != iceberg_transaction.created_views.end()) {
+		auto &view_info = created_it->second;
+		auto view_entry = make_uniq<ViewCatalogEntry>(catalog, schema, *view_info);
+		auto result = view_entry.get();
+		view_catalog_entries.emplace(view_name, std::move(view_entry));
+		return result;
+	}
+
+	// Check if the view endpoint is supported
+	if (ic_catalog.supported_urls.find("GET /v1/{prefix}/namespaces/{namespace}/views/{view}") ==
```
Tmonster, line 436: Should we do this check first? Any view that we create that we later call CreateViewEntryInternal will have to pass this check right?

src/catalog/rest/iceberg_table_set.cpp
```diff
@@ -348,4 +350,196 @@ optional_ptr<CatalogEntry> IcebergTableSet::GetEntry(ClientContext &context, con
...
+	    ic_catalog.supported_urls.end()) {
+		return nullptr;
+	}
+
+	// Load the view from the REST catalog
+	rest_api_objects::LoadViewResult load_result;
+	try {
+		load_result = IRCAPI::GetView(context, ic_catalog, schema, view_name);
+	} catch (...) {
+		return nullptr;
```
Tmonster: nit: Can we not have a naked catch here? Not a fan of those when it comes to debugging

src/catalog/rest/iceberg_table_set.cpp
```diff
@@ -348,4 +350,196 @@ optional_ptr<CatalogEntry> IcebergTableSet::GetEntry(ClientContext &context, con
...
+	for (auto &version : metadata.versions) {
+		if (version.version_id != metadata.current_version_id) {
+			continue;
+		}
+		// Search representations for DuckDB dialect first
+		for (auto &repr : version.representations) {
+			if (!repr.has_sqlview_representation) {
+				continue;
+			}
+			if (repr.sqlview_representation.dialect == "duckdb") {
```
Tmonster: I think we should hardcode the `"duckdb"` string somewhere. Maybe in `include/common/IcebergConstants.hpp`?

src/catalog/rest/iceberg_table_set.cpp
```diff
@@ -348,4 +350,196 @@ optional_ptr<CatalogEntry> IcebergTableSet::GetEntry(ClientContext &context, con
...
+				view_sql = repr.sqlview_representation.sql;
+				break;
+			}
+		}
+		// If no DuckDB dialect found, use the first available SQL representation
+		if (view_sql.empty()) {
+			for (auto &repr : version.representations) {
+				if (repr.has_sqlview_representation) {
+					view_sql = repr.sqlview_representation.sql;
```
Tmonster, line 482: Can we also log a warning that no DuckDB dialect has been found?
a search for `DUCKDB_LOG_WARNING` should reveal some examples

Tmonster (commented):
@LucaSoato 

Thanks! A couple of mostly minor comments.

Could you also retarget this to v1.5-variegata? That way we can try and get it in for v1.5.3

LucaSoato:
Thanks so much for your feedback, I'll get back to this asap  🚀

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
LucaSoato: Switch statement added!

https://github.com/duckdb/duckdb-iceberg/blob/28bd7f45/src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp#L121-L178

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
LucaSoato: Yes, fixed!
https://github.com/LucaSoato/duckdb-iceberg/blob/59733d610d662ddba4f317417db25abcfd831c53/src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp#L113-L125

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
LucaSoato: Done! 
https://github.com/LucaSoato/duckdb-iceberg/blob/db61a521/src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp#L252-L257

src/catalog/rest/transaction/iceberg_transaction.cpp
LucaSoato: I've used the lowercase and saved it to the constant `IcebergConstants::ViewDuckDBDialect`
https://github.com/LucaSoato/duckdb-iceberg/blob/db61a52199e69e2f2a900693752b374007342db6/src/include/common/iceberg_constants.hpp#L17-L19

In the [Iceberg View examples](https://iceberg.apache.org/view-spec/#sql-representation), I saw them always using lowercase.

src/catalog/rest/iceberg_table_set.cpp
LucaSoato: Yes, done that as well 
https://github.com/LucaSoato/duckdb-iceberg/blob/db61a52199e69e2f2a900693752b374007342db6/src/catalog/rest/iceberg_table_set.cpp#L454

src/catalog/rest/iceberg_table_set.cpp
LucaSoato: I've improved this part by refactoring `IRCAPI::GetView` to return `APIResult<unique_ptr<const LoadViewResult>>`, similarly to what happens in `GetTable`

https://github.com/LucaSoato/duckdb-iceberg/blob/db61a52199e69e2f2a900693752b374007342db6/src/catalog/rest/api/catalog_api.cpp#L529

src/catalog/rest/iceberg_table_set.cpp
LucaSoato: Yup, logged:
https://github.com/LucaSoato/duckdb-iceberg/blob/db61a52199e69e2f2a900693752b374007342db6/src/catalog/rest/iceberg_table_set.cpp#L463-L467

src/catalog/rest/iceberg_catalog.cpp
LucaSoato: Removed the rows, I've left a comment so if anyone looks into this knows why the endpoints aren't included. 

https://github.com/LucaSoato/duckdb-iceberg/commit/27f4aebcfd3a43b3876331700680623da510c418

src/catalog/rest/iceberg_table_set.cpp
LucaSoato: So right now we check: 
1.  if it's in transaction.deleted_views
2.  if it's in view_catalog_entries cache
3.  if we created it in transaction.created_views
4. and only then if views are in supported urls

Technically it's right because the first three operations are in memory and do not require URLs. The thing is that if in the same transaction we try to create a view, referencing the same view later, it will fail anyway but only during the commit of such operation for catalogs that don't support views.

If you want we can add a `POST /v1/{prefix}/namespaces/{namespace}/views` endpoint check at the top of `CreateView` so it fails immediately on no-views catalogs?

i-m-nk:
@LucaSoato can we have a support for user defined dialect value (may be from a variable/global), some Kond of override functionality. That way user can try to read multi dialects.

LucaSoato:
> @LucaSoato can we have a support for user defined dialect value (may be from a variable/global), some Kond of override functionality. That way user can try to read multi dialects.

@Tmonster what do you think about it?

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -152,12 +204,66 @@ optional_ptr<CatalogEntry> IcebergSchemaEntry::CreateIndex(CatalogTransaction tr
...
+			if (info.on_conflict == OnCreateConflict::IGNORE_ON_CONFLICT) {
+				return existing_entry;
+			}
+			// CREATE OR REPLACE — drop first, then create
+			DropInfo drop_info;
+			drop_info.type = CatalogType::VIEW_ENTRY;
+			drop_info.name = info.view_name;
+			drop_info.cascade = false;
+			drop_info.if_not_found = OnEntryNotFound::RETURN_NULL;
+			DropEntry(context, drop_info, false);
```
Tmonster: This will drop the entry regardless of if he transaction succeeds right? 
Currently DuckDB-Iceberg doesn't support `CREATE OR REPLACE` for tables because of this scenario. I would like to stay consistent with that if possible

test/sql/local/catalog_test_config_setup/catalog_agnostic/create/test_create_view.test
```diff
@@ -0,0 +1,140 @@
...
+# ─── DROP VIEW ────────────────────────────────────────────────────────────────
+
+statement ok
+drop view test_view_v1;
+
+# View should no longer exist
+statement error
+select * from test_view_v1;
+----
+does not exist
```
Tmonster: nit:  `<REGEX>:.*does not exist.*` is what we normally use

src/catalog/rest/catalog_entry/schema/iceberg_schema_entry.cpp
```diff
@@ -152,12 +204,66 @@ optional_ptr<CatalogEntry> IcebergSchemaEntry::CreateIndex(CatalogTransaction tr
...
+		}
+	} else {
+		// ERROR_ON_CONFLICT — check existence
+		auto existing_entry = GetEntry(transaction, CatalogType::VIEW_ENTRY, info.view_name);
+		if (existing_entry) {
+			throw CatalogException("View with name \"%s\" already exists", info.view_name);
+		}
+	}
+
+	// Generate default column names if the caller gave us types but no names.
```
Tmonster, line 229: Can we do some more work here and potentially bind the sql query to make sure it is valid?

Tmonster (commented):
Thanks for the updates. 
With a more thorough review, I think we are still missing some testing/functionality that I would appreciate having in this PR

There are no tests on the unhappy path. Can we add some? 

It would be nice to see
- Creating a view that selects from table columns that do not exist. (I would expect this to error on view creation similar to how it does when operating on duckdb tables)
- Selecting a column from a view where the column does not exist in the view.

Other tests I would like to see
- A view that does `select *`
- A view that does `select * exclude`
- A view that joins tables 

Also can we test how iceberg views are handled when we call 
```
from duckdb_views();
from duckdb_tables();
show all tables;
```
I don't expect full Iceberg view functionality integrated here just yet, I just want to make sure DuckDB doesn't crash in this scenario. 

One more piece of functionality I would like to see, is a table function called `iceberg_view_metadata` which returns the result of a call to the `LoadView` API. That can make view discovery on a Rest catalog a bit easier. We have a similar function for tables now, `iceberg_load_table_response`, you can use that as a template, we put a lot of the JSON into a variant.

I also like the idea of a preferred SQL dialect. Maybe that is something we can add to the view properties? I see in the [LoadViewResult](https://github.com/apache/iceberg/blob/46ae3ad2749cdcb7877d86babd0357b42f05823b/open-api/rest-catalog-open-api.yaml#L3818) we can add properties to a view.  I think that might be out of scope for this PR though, and can be added in a follow up. A config option would also work. DuckDB v2 will use the PEG parser, so we should be able to extend our parser support, and then we can have a statement like `SELECT * FROM iceberg_view WITH ('sql'='duckdb)`, but that will obvs come much later.

Tmonster:
Hi @LucaSoato is this still something you still want merged, otherwise we can close to reduce clutter?

LucaSoato:
Hi everyone,
Thanks for your patience, I should be able to push the requested changes this weekend.

LucaSoato:
Hi @Tmonster, thanks for the patience!

Just a small recap regarding the changes:
- unhappy paths are covered, you can check [here](https://github.com/duckdb/duckdb-iceberg/pull/887/changes/686885ff93d680e03b40349ce88f41561b736554#diff-600474a049dadc628430773d64d1dec810674828517d01461d823d9c6d4d4e08) and below;
- CREATE OR REPLACE now throws NotImplementedException, since it's non-transactional;
- new iceberg_view_metadata table function, returns the raw LoadView response;

Related to the preferred SQL dialect, it's a nice follow-up, let's leave it out of this PR and we can add it later. 
I'm replying also to the other comment of yours related to the binding of the sql query, it's something that we have already.

Thanks again and, as always, let me know if there's anything else left for this PR.

LucaSoato:
p.s.
I've rebased on top of v1.5-variegata, the tests go on fine except for two that fail even on a clean v1.5-variegata build.

src/catalog/rest/catalog_entry/schema/iceberg_schema_entry.cpp
LucaSoato: Hi @Tmonster,
I think that this already happens here:
https://github.com/duckdb/duckdb/blob/c028854da4348b4d32731e460b75973364f79b74/src/planner/binder/statement/bind_create.cpp#L204

LucaSoato:
(a small format fix is coming)

Tmonster:
Thanks @LucaSoato, will get to reviewing, It hurts me to say this, but we are going to limit new features to v1.5. Would you mind rebasing and retargeting `main` for this?

LucaSoato:
It's alright, will rebase to main.

src/catalog/rest/catalog_entry/schema/iceberg_schema_entry.cpp
```diff
@@ -153,12 +205,57 @@ optional_ptr<CatalogEntry> IcebergSchemaEntry::CreateIndex(CatalogTransaction tr
...
-	throw NotImplementedException("Get Create View");
-}
-
 optional_ptr<CatalogEntry> IcebergSchemaEntry::CreateView(CatalogTransaction transaction, CreateViewInfo &info) {
-	throw NotImplementedException("Create View");
+	if (info.sql.empty() && !info.query) {
+		throw BinderException("Cannot create view in Iceberg without a query");
+	}
+	auto &context = transaction.GetContext();
+
```
Tishj, line 213: We should check that `binding_mode` is set to "bind on create":
```c++
	//! Whether or not to bind the view on create
	CreateViewBindingMode binding_mode = CreateViewBindingMode::BIND_ON_CREATE;
```
Since we don't want to submit to the Catalog without verifying correctness

corleyma:
Bumping/reviving because I'd really love it for duckdb to support iceberg view spec, and I like this implementation a lot, especially:
> When loading, prefers "duckdb" dialect, falls back to any available SQL representation
> Views from other engines (Spark, Trino) that can't be parsed are still visible in the catalog and droppable — querying them gives a clear error instead of crashing

which will both make (simple) cross-engine views workable!

nicosuave:
+1 would also love to see this one land— we have a real need too

I do agree with @Tmonster this is a big changeset, maybe just simple reads would be a reasonable initial implementation, given this has gotten stale?

maxsbs:
+1 also interest on our side. Agreed with @nicosuave that a "read implementation" would helpful.

Is there any plan to bring this to an upcoming version of DuckDB? Also, are there other contributions we need to bring or is the PR just open because of the risk of impactful change?

kevinw02:
+1 also interest on our side. This would be a huge help to the maintainability + performance of our current implementation.

Tmonster:
If someone wants to pick up this PR, we will try to review it to get it in for v2.0. It will most likely need to solve a number of merge conflicts

Tmonster:
Closing this in favor of https://github.com/duckdb/duckdb-iceberg/pull/1404

PR #894, opened 2026-04-07
support both http and https schemes for REST catalog SigV4 requests

plusplusjiajia:
Previously AWSInput::BuildURI() hardcoded Aws::Http::Scheme::HTTPS, causing SigV4-signed requests to always use HTTPS regardless of the user-provided endpoint URL. This broke local dev scenarios (MinIO, LocalStack) that use HTTP endpoints.


Tmonster (commented):
thanks! 
Are we sure the test actually tests the change code? If `ACCESS_DELEGATION_MODE 'none'` is set, I don't think we go through the aws signing code.

plusplusjiajia:
> thanks! Are we sure the test actually tests the change code? If `ACCESS_DELEGATION_MODE 'none'` is set, I don't think we go through the aws signing code.

@Tmonster   Good catch — the original nessie test used OAuth2 auth. I've replaced it with a dedicated SigV4 mock server test

plusplusjiajia:
@Tmonster  Could you take another look when you get a chance?

Tmonster:
@plusplusjiajia thanks! changed the base so we can get this in the next release.

PR #902, opened 2026-04-09
Add geometry support simple

Tmonster:
This adds geometry type support to DuckDB-Iceberg.

I still want to add tests with other engines to make sure there is compatibility. It's been pretty tough getting sedonaDB to work, but I'll get there. 

This PR doesn't add support for upper bound and lower bounds for the geometry type. That is something we will add later.

This includes a cloud test to read from snowflake created geoemtry types. 
I have manually tested that snowflake can read duckdb created geometry types, except for an empty collection https://github.com/duckdblabs/duckdb-internal/issues/8847. The issue there is interesting though, since it leads me to believe that snowflake is scanning the parquet files when registering an iceberg table (which there is no reason to do that)

### DIFF
--- a/src/execution/operator/iceberg_insert.cpp
+++ b/src/execution/operator/iceberg_insert.cpp
@@ -307,10 +307,6 @@ void IcebergInsertGlobalState::AddFiles(DataChunk &chunk, const string &table_na
 				continue;
 			}
 			auto &column_info = *column_info_p;
-			if (column_info.type.id() == LogicalTypeId::GEOMETRY) {
-				//! Geometry columns have no meaningful min/max bounds
-				continue;
-			}
 			auto stats = ParseColumnStats(column_info.type, col_stats, context);
 
 			// a map type cannot violate not null constraints.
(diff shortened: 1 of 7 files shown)


src/execution/operator/iceberg_insert.cpp
```diff
@@ -307,6 +307,10 @@ void IcebergInsertGlobalState::AddFiles(DataChunk &chunk, const string &table_na
 				continue;
 			}
 			auto &column_info = *column_info_p;
+			if (column_info.type.id() == LogicalTypeId::GEOMETRY) {
+				//! Geometry columns have no meaningful min/max bounds
```
Tishj: This isn't true, I'm fine with skipping them for now, but let's at least put a FIXME
https://iceberg.apache.org/spec/#bounds-for-variant-geometry-and-geography

Tishj (changes requested):
One small point of feedback, but yea after adding a comment there we can just merge without running through all of CI again 👍

PR #904, opened 2026-04-09
[ALTER] Add `ALTER TABLE ... DROP COLUMN` support

Tishj:
This PR fixes  https://github.com/duckdblabs/duckdb-internal/issues/8606
Also fixes https://github.com/duckdblabs/duckdb-internal/issues/8827

Summary of changes:
- Fix a problem that was causing us to only be able to create 1 schema update per transaction.
- Fix a problem that was causing snapshots to always be created with the latest schema id.
- Fix a problem that was causing the `AssertCurrentSchemaId` to only be created for the first table altered in the transaction.
- `AddSetCurrentSchema` and `AddAssertCurrentSchemaId` are removed, these are instead added on commit. Since there should only ever be one of these, I deemed this the safest choice to ensure that.
- Add `DROP COLUMN` functionality

### DIFF
--- a/src/catalog/rest/api/iceberg_add_snapshot.cpp
+++ b/src/catalog/rest/api/iceberg_add_snapshot.cpp
@@ -17,6 +17,9 @@ namespace duckdb {
 
 IcebergAddSnapshot::IcebergAddSnapshot(const IcebergTableInformation &table_info)
     : IcebergTableUpdate(IcebergTableUpdateType::ADD_SNAPSHOT, table_info) {
+	//! FIXME: Do we also need to capture the current partition spec and sort order?
+	//! This is a bit of a code smell, the `IcebergTableInformation` should instead be const
+	//! and all transactional changes should live in the IcebergTransactionData
 	schema_id = table_info.table_metadata.GetCurrentSchemaId();
 }
 
--- a/src/catalog/rest/api/iceberg_add_snapshot.cpp
+++ b/src/catalog/rest/api/iceberg_add_snapshot.cpp
@@ -17,6 +17,7 @@ namespace duckdb {
 
 IcebergAddSnapshot::IcebergAddSnapshot(const IcebergTableInformation &table_info)
     : IcebergTableUpdate(IcebergTableUpdateType::ADD_SNAPSHOT, table_info) {
+	schema_id = table_info.table_metadata.GetCurrentSchemaId();
 }
 
 static rest_api_objects::TableUpdate CreateAddSnapshotUpdate(const IcebergTableInformation &table_info,
@@ -33,7 +34,7 @@ static rest_api_objects::TableUpdate CreateAddSnapshotUpdate(const IcebergTableI
 }
 
 static IcebergManifestListEntry ScanExistingManifestFile(const IcebergManifestFile &manifest_file,
-                                                         IcebergCommitState &commit_state) {
+                                                         IcebergCommitState &commit_state, int32_t schema_id) {
 	vector<IcebergManifestListEntry> manifest_files;
 	manifest_files.push_back(manifest_file);
 
@@ -43,7 +44,7 @@ static IcebergManifestListEntry ScanExistingManifestFile(const IcebergManifestFi
 
 	IcebergSnapshotScanInfo snapshot_info;
 	snapshot_info.snapshot = commit_state.latest_snapshot;
-	snapshot_info.schema_id = table_metadata.GetCurrentSchemaId();
+	snapshot_info.schema_id = schema_id;
 
 	auto manifest_scan =
 	    AvroScan::ScanManifest(snapshot_info, manifest_files, options, fs, "", table_metadata, commit_state.context);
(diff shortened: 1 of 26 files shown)


src/catalog/rest/api/iceberg_add_snapshot.cpp
```diff
@@ -17,6 +17,7 @@ namespace duckdb {
 
 IcebergAddSnapshot::IcebergAddSnapshot(const IcebergTableInformation &table_info)
     : IcebergTableUpdate(IcebergTableUpdateType::ADD_SNAPSHOT, table_info) {
+	schema_id = table_info.table_metadata.GetCurrentSchemaId();
```
Tishj, line 23: I wonder if there's similar logic that needs to happen for partition specs / sort orders ?
Something to be aware of I think

src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
```diff
@@ -286,7 +286,8 @@ optional_ptr<CatalogEntry> IcebergTableInformation::CreateSchemaVersion(const Ic
 	if (result->name.empty()) {
 		throw InternalException("IcebergTableSet::CreateEntry called with empty name");
 	}
-	schema_versions.emplace(table_schema.schema_id, std::move(table_entry));
+
+	schema_versions[table_schema.schema_id] = std::move(table_entry);
```
Tishj: Might have to undo this one, I made this change when I found out we were building up all alters into the same schema, before realizing that that in itself is not correct behavior

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -281,19 +274,55 @@ void IcebergSchemaEntry::Alter(CatalogTransaction transaction, AlterInfo &info)
...
+		}
+
+		auto &partition_spec = updated_table.table_metadata.GetLatestPartitionSpec();
+		auto partition_field = partition_spec.TryGetFieldBySourceId(column_id.GetIndex());
+		if (partition_field) {
+			throw CatalogException(
+			    "Can't drop column '%s' as it is referenced by the current partition spec's field: '%s' (field id: %d)",
+			    to_remove_column, partition_field->name, partition_field->partition_field_id);
+		}
+
```
Tishj, line 318: Should probably verify that there is at least 1 column still remaining

src/include/catalog/rest/api/iceberg_table_update.hpp
```diff
@@ -39,6 +39,7 @@ struct IcebergCommitState {
 
 public:
 	const IcebergTableInformation &table_info;
+	int32_t current_schema_id;
```
Tishj: I think this is unused

src/catalog/rest/api/iceberg_add_snapshot.cpp
Tmonster: I think we should since we are also supporting partition specs now (along with `ALTER TABLE t1 SET PARTITIONED BY ()`)

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -281,19 +274,59 @@ void IcebergSchemaEntry::Alter(CatalogTransaction transaction, AlterInfo &info)
 
 		new_iceberg_column->required = false;
 
+		auto new_schema = current_schema.Copy();
+		new_schema->schema_id++;
```
Tmonster, line 278: I think this will cause errors. I remember for partition specs, every partition spec needs to be unique, so if you have the following 
```
create table t1 (a int) partitioned by (a);
alter table set partitioned by (bucket(10, a));
alter table set partitioned by (a);
```
I could not just create a new partition spec with the same definition. I have a feeling this will be the same for schemas

Tmonster (commented):
I would be careful with just increasing the schema id by 1. I think other engines have the ability to set the schema to a previous schema to restore columns. If we just blindly increase by 1, this could lead to catalog errors being returned. 

I think it might be better to make sure we are not re-creating any schemas already in the history of the table.

Eventually (not soon), I think we will also add support to restore columns/restore a schema, so we can better test this

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
Tishj: Yea that's fair, I think schemas have the same restriction (have to be unique)
With no real definition of what makes a schema unique, but I guess we'll figure it out

Tishj:
```py
    def _commit(self) -> UpdatesAndRequirements:
        """Apply the pending changes and commit."""
        from pyiceberg.table import TableProperties

        new_schema = self._apply()

        existing_schema_id = next(
            (schema.schema_id for schema in self._transaction.table_metadata.schemas if schema == new_schema), None
        )

        requirements: tuple[TableRequirement, ...] = ()
        updates: tuple[TableUpdate, ...] = ()

        # Check if it is different current schema ID
        if existing_schema_id != self._schema.schema_id:
            requirements += (AssertCurrentSchemaId(current_schema_id=self._schema.schema_id),)
            if existing_schema_id is None:
                updates += (
                    ap(schema=new_schema),
                    SetCurrentSchemaUpdate(schema_id=-1),
                )
            else:
                updates += (SetCurrentSchemaUpdate(schema_id=existing_schema_id),)

            if name_mapping := self._name_mapping:
                updated_name_mapping = update_mapping(name_mapping, self._updates, self._adds)
                updates += (
                    SetPropertiesUpdate(updates={TableProperties.DEFAULT_NAME_MAPPING: updated_name_mapping.model_dump_json()}),
                )

        return updates, requirements
```

Tmonster: Looks like they letting the catalog determine the new schema id here? 
```
if existing_schema_id is None:
    updates += (
        ap(schema=new_schema),
        SetCurrentSchemaUpdate(schema_id=-1),
    )
else:
    updates += (SetCurrentSchemaUpdate(schema_id=existing_schema_id),)
```

Tishj: -1 is defined in the spec,as the last added schema

PR #908, opened 2026-04-10
[DEV] Enable v4 spark tests

Tmonster:
This is going to fail. I could have sworn we were testing this already, but apparently not. 
@Tishj I imagine these failures are just due to some wonky behavior with spark and row ids?

### DIFF
--- a/.github/workflows/test-fixture-catalog.yml
+++ b/.github/workflows/test-fixture-catalog.yml
@@ -85,6 +85,11 @@ jobs:
         env:
           ICEBERG_SERVER_AVAILABLE: 1
         run: |
+          # test with spark v4
+          source .venv-spark4/bin/activate
+          python3 -m pytest test/python
+          deactivate
+          # test with spark v3
           python3 -m venv .venv-spark3
           source .venv-spark3/bin/activate
           python3 -m pip install pyspark==3.5.2 pytest pyiceberg==0.9.1 pyarrow pydantic==2.9.0
--- a/test/python/test_spark_read.py
+++ b/test/python/test_spark_read.py
@@ -306,6 +306,7 @@ class TestSparkRead:
         ]
 
     # Written by Spark, read by Spark
+    # See https://github.com/duckdb/duckdb-iceberg/pull/908 on why spark behavior is unclear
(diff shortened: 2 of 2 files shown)


Tishj:
Hmm I'd have to look into these some more to say whether or not these are expected
It looks like some of the late materialization changes I made are causing this (#835 )

Tishj:
Actually, not even that, it's failing on the sanity check I added, where we are reading data that Spark produced, not even touched by DuckDB at any point:
expected:
```py
[
    Row(_last_updated_sequence_number=5, _row_id=3, id=1, data='replaced'),
    Row(_last_updated_sequence_number=8, _row_id=0, id=2, data='replaced_again'),
    Row(_last_updated_sequence_number=2, _row_id=6, id=4, data='d_u1'),
    Row(_last_updated_sequence_number=8, _row_id=1, id=6, data='replaced_again'),
    Row(_last_updated_sequence_number=7, _row_id=2, id=7, data='g_new'),
]
```

actual:
```py
[
	Row(_last_updated_sequence_number=5, _row_id=3, id=1, data='replaced'),
	Row(_last_updated_sequence_number=8, _row_id=0, id=2, data='replaced_again'),
	Row(_last_updated_sequence_number=2, _row_id=7, id=4, data='d_u1'),
	Row(_last_updated_sequence_number=8, _row_id=1, id=6, data='replaced_again'),
	Row(_last_updated_sequence_number=7, _row_id=2, id=7, data='g_new')
]
```

Tishj:
It's really hard to say if these are wrong, `_row_id` is not a very stable value, due to the way it shifts with compaction/rewrites
In our sqllogic tests we explicitly don't hardcode these anymore, instead using:
```sql
set variable id2_row_id = (select _row_id from my_datalake.default.row_lineage_test_upgraded where id == 2);
```
and
```sql
select _row_id = getvariable('id2_row_id') from my_datalake.default.row_lineage_test_upgraded where id = 2;
```
To verify the stability of the _row_id values through various UPDATE and DELETE statements

Tmonster:
@Tishj is it alright if I just skip these tests then? I can create an internal ticket to track it

PR #909, opened 2026-04-10
fix: filter vended credentials by storage type and add cross-scheme scope fallback

plusplusjiajia:
1. Filter vended credentials by storage type: When the REST catalog returns multiple vended credentials with different prefixes (e.g. s3://... and oss://...), all of them were used to create secrets regardless of the actual storage type. Now credentials are filtered by storage type (e.g. `s3`, `gcs`, `azure`), matching
  Iceberg Java S3FileIO behavior where credentials are filtered by `ROOT_PREFIX`.
  2. Cross-scheme scope fallback for credential matching: When a table location uses a different URI scheme than the vended credential prefix (e.g. table at s3a://bucket/path but credential prefix is s3://bucket/), DuckDB's scope-based secret matching fails because the s3a:// path doesn't match the s3:// scope. Now the
  table location is added as an additional scope to ensure matching works.
  3. Extract shared utility functions (`DetectStorageType`, `DecomposeHost`, `GetAwsRegion`, `GetAwsService`) into `sigv4_utils.hpp` for reuse and testability, with C++ unit tests.


plusplusjiajia:
@Tmonster @Tishj Could you take a look at this PR when you get a chance?

Tmonster:
Hi @plusplusjiajia , can you rebase on v1.5-variegata? That might also fix the flake nessie test.

plusplusjiajia:
> Hi @plusplusjiajia , can you rebase on v1.5-variegata? That might also fix the flake nessie test.

Thanks @Tmonster. Done — rebased onto v1.5-variegata and updated the base branch.

Tmonster (commented):
Hi @plusplusjiajia,

Everything looks great. I think you can remove the C++ tests though, we are probably not going to maintain those. 
If you have tested the changes yourself, that is good enough for me. 
You could also help expand our sql testing coverage if you want by adding more catalog+storage combinations

plusplusjiajia:
> Everything looks great. I think you can remove the C++ tests though, we are probably not going to maintain those.
> If you have tested the changes yourself, that is good enough for me.
> You could also help expand our sql testing coverage if you want by adding more catalog+storage combinations

Thanks @Tmonster! C++ tests removed. I've verified the changes against a REST catalog returning mixed s3/oss credentials.

Tmonster:
Thanks @plusplusjiajia

PR #912, opened 2026-04-10
[ALTER] Add `ALTER TABLE ... ALTER ... TYPE` support

Tishj:
This PR implements https://github.com/duckdblabs/duckdb-internal/issues/8607

Schema evolution is covered by https://iceberg.apache.org/spec/#schema-evolution, supported for:
- `int` -> `long`
- `float` -> `double`
- `decimal(W, S)` -> `decimal(nW, nS)` where `nW >= W` and `nS == S`
- `date` -> `timestamp`
- `date` -> `timestamp_ns` if format-version >= 3

`date` promotions are only allowed if the column isnt currently referenced by a partition transform


test/sql/local/irc_any_catalog/alter/alter_type.test
```diff
@@ -0,0 +1,200 @@
...
+query IIIII
+select * from my_datalake.default.schema_evolution_table;
+----
+23123425	1983-05-21	1934-11-05	9007199000000000.0	23456.34
+
+# Can't downgrade from INTEGER to SMALLINT
+statement error
+ALTER TABLE my_datalake.default.schema_evolution_table ALTER evolve_to_long TYPE SMALLINT;
+----
+Catalog Error: Column 'evolve_to_long' of type 'INTEGER' can't be altered to type 'SMALLINT'
```
Tmonster, line 61: Pretty sure `SMALLINT` isn't even valid Iceberg. Maybe we could add a check somewhere in the code where we call `IcebergTypeHelper::LogicalTypeToIcebergType()`? That will also throw a proper error message that the type is not a valid Iceberg Type

Tmonster (commented):
lgtm. Maybe a check to make sure types are valid iceberg types, but that can be in a follow up PR

PR #917, opened 2026-04-13
[Dev] Refactor internals of `IcebergTransaction`

Tishj:
### Limitation

Previously the internals of `IcebergTransaction` consisted of:
```c++
case_insensitive_map_t<IcebergTableInformation> updated_tables;
case_insensitive_map_t<IcebergTableInformation> deleted_tables;
```

Renaming and deleting tables use a different endpoint than other alters.
This makes renaming and recreating tables within a transaction impossible, as it would be impossible to discern which updates should be applied before and after a rename/delete.

### Refactor

To fix this, the following change was made:
```c++
vector<unique_ptr<IcebergTransactionUpdate>> transaction_updates;
case_insensitive_map_t<IcebergTransactionTableState> current_table_data;
```
`transaction_updates` represents a block of 1 or more alters that can be applied atomically (within the same endpoint)
`current_table_data` is a struct representing the current state of a table, with a reference to the `IcebergTableInformation` that currently applies to the latest state.

### IcebergTransactionTableState

```c++
enum class IcebergTableStatus : uint8_t { ALIVE, DROPPED };

enum class IcebergTableSource : uint8_t {
	//! Loaded from external source
	EXTERNAL,
	//! Version that exists within the transaction
	TRANSACTION
};

struct IcebergTransactionTableState {
public:
	IcebergTransactionTableState(IcebergTableInformation &table, IcebergTableSource source);

public:
	reference<IcebergTableInformation> table;
	IcebergTableSource source;
	IcebergTableStatus status;
};
```

### IcebergTransactionUpdate

```c++
//! Update a table with a regular alter
struct IcebergTransactionAlterUpdate : public IcebergTransactionUpdate {
public:
	static constexpr const IcebergTransactionUpdateType TYPE = IcebergTransactionUpdateType::ALTER;

public:
	IcebergTransactionAlterUpdate(IcebergTransaction &transaction);
	virtual ~IcebergTransactionAlterUpdate() override;

public:
	IcebergTableInformation &CreateTable(const string &table_key, IcebergTableInformation &&table);
	IcebergTableInformation &GetOrInitializeTable(const IcebergTableInformation &table);

public:
	//! All the tables touched in this atomic block
	case_insensitive_map_t<IcebergTableInformation> updated_tables;
};
```

```c++
//! Drop a table
struct IcebergTransactionDeleteUpdate : public IcebergTransactionUpdate {
public:
	static constexpr const IcebergTransactionUpdateType TYPE = IcebergTransactionUpdateType::DELETE;

public:
	IcebergTransactionDeleteUpdate(IcebergTransaction &transaction, const IcebergTableInformation &table);
	virtual ~IcebergTransactionDeleteUpdate() override;

public:
	const IcebergTableInformation &table;
};
```

### DIFF
--- a/src/catalog/rest/transaction/iceberg_transaction.cpp
+++ b/src/catalog/rest/transaction/iceberg_transaction.cpp
@@ -24,8 +24,13 @@
 #include "planning/metadata_io/avro/avro_scan.hpp"
 #include "iceberg_logging.hpp"
 #include "catalog/rest/api/table_update.hpp"
+#include "catalog/rest/transaction/iceberg_transaction_update.hpp"
 
 namespace duckdb {
+
+IcebergTransactionTableState::IcebergTransactionTableState(IcebergTableInformation &table, IcebergTableSource source)
+    : table(table), source(source), status(IcebergTableStatus::ALIVE) {
+}
 
 IcebergTransaction::IcebergTransaction(IcebergCatalog &ic_catalog, TransactionManager &manager, ClientContext &context)
     : Transaction(manager, context), db(*context.db), catalog(ic_catalog), access_mode(ic_catalog.access_mode) {
@@ -296,10 +301,11 @@ static bool NeedsAssertSchemaId(const IcebergTransactionData &transaction_data,
 	return initial_schema_id != table_info.table_metadata.GetCurrentSchemaId();
 }
 
-TableTransactionInfo IcebergTransaction::GetTransactionRequest(ClientContext &context) {
+TableTransactionInfo IcebergTransaction::GetTransactionRequest(IcebergTransactionAlterUpdate &alter_update,
+                                                               ClientContext &context) {
 	TableTransactionInfo info;
 	auto &transaction = info.request;
-	for (auto &updated_table : updated_tables) {
+	for (auto &updated_table : alter_update.updated_tables) {
 		auto &table_info = updated_table.second;
 		if (!table_info.transaction_data) {
 			continue;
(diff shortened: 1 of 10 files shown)


src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -442,28 +446,27 @@ void IcebergTransaction::DoTableUpdates(ClientContext &context) {
 				                          table_change.identifier.name, transaction_json);
 			}
 		}
-		updated_tables.clear();
+		// updated_tables.clear();
```
Tishj, line 449: Commented out because I'm not sure why it's there.
This would be problematic for retries - though everything is problematic for retries already as we need to make multiple commits to the catalog 🙃

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -442,28 +446,27 @@ void IcebergTransaction::DoTableUpdates(ClientContext &context) {
...
-		schema_entry.DropEntry(context, drop_info, true);
-	}
+	auto &table = delete_update.table;
+	auto schema_key = table.schema.name;
+	auto table_key = table.GetTableKey();
+	auto table_name = table.name;
+	IRCAPI::CommitTableDelete(context, catalog, table.schema.namespace_items, table.name);
+	// remove the load table result
+	//! FIXME: this can very easily be problematic
+	ic_catalog.RemoveLoadTableResult(table_key);
```
Tishj, line 463: This is eagerly removing cache from the catalog, even though we have no way to know if another transaction has already replaced the cache entry that we made

PR #918, opened 2026-04-13
additional timestamp_ns support

debugmiller:
- support for bucket transforms
- serialize
- hashes
- more tests
- a negative test for REST catalog (seems to only support v2?)

Also a drive by fix to get the docker/spark fixtures generation working for multi-homed machines.


scripts/data_generators/tests/bucket_partitioned_timestamp_ns/q00.sql
```diff
@@ -0,0 +1,7 @@
+CREATE OR REPLACE TABLE default.bucket_partitioned_timestamp_ns (
+    id INTEGER,
+    ts_val TIMESTAMP,
```
Tmonster: I don't think this is not a timestamp_ns column. Spark currently does not support it

https://iceberg.apache.org/docs/latest/spark-getting-started/#iceberg-type-to-spark-type

Tmonster (commented):
lgtm, just the one question about spark table

scripts/data_generators/tests/bucket_partitioned_timestamp_ns/q00.sql
debugmiller: ah, I forgot to remove this

debugmiller:
Ok I've removed the test that was not actually testing timestamp_ns and added skips for lakekeeper/polaris, which don't seem to support catalog defaults for timestamp_ns

scripts/data_generators/tests/bucket_partitioned_timestamp_ns/q00.sql
debugmiller: removed

Tmonster:
Thanks! Looks good to me 👍

PR #923, opened 2026-04-14
Don't support `ALTER TABLE ... ALTER COLUMN ... SET NOT NULL`

yan-alex:
https://github.com/duckdblabs/duckdb-internal/issues/8609


src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -251,6 +252,42 @@ static void VerifySchemaEvolution(const IcebergTableMetadata &table_metadata, co
...
 }
 
+//! Ensure existing data files don't contain NULL values in this column
+static void VerifyNotNullConstraint(ClientContext &context, IcebergTableInformation &updated_table,
+                                    IcebergColumnDefinition &column, int32_t current_schema_id) {
+	auto snapshot_lookup = updated_table.GetSnapshotLookup(context);
+	auto snapshot_info = updated_table.table_metadata.GetSnapshot(snapshot_lookup);
+	if (!snapshot_info.snapshot) {
+		// Column is present but there's no snapshot, thus all rows are NULL.
+		throw ConstraintException("NOT NULL constraint failed: %s.%s", updated_table.name, column.name);
```
Tishj: This is wrong, if the table is empty, there are no rows, so there are no NULL rows

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -251,6 +252,42 @@ static void VerifySchemaEvolution(const IcebergTableMetadata &table_metadata, co
...
+			auto found_column_null_count = column_null_count_it != data_file.null_value_counts.end();
+			found_column_null_count_at_least_once = found_column_null_count_at_least_once || found_column_null_count;
+			// `null_value_counts` is an optional field per the Iceberg spec.
+			if (found_column_null_count && column_null_count_it->second > 0) {
+				throw ConstraintException("NOT NULL constraint failed: %s.%s", updated_table.name, column.name);
+			}
+		}
+	}
+
+	if (!found_column_null_count_at_least_once && !column.initial_default) {
```
Tishj: `initial_default` could also be explicitly set to NULL, this only checks if its missing

Tishj: But more broadly, the `null_value_count` being missing is not an indicator of the column being missing.
Perhaps this makes a good case for reading the avro metadata, so we can determine if the data file doesn't contain a certain column

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -251,6 +252,42 @@ static void VerifySchemaEvolution(const IcebergTableMetadata &table_metadata, co
 	throw CatalogException(error);
 }
 
+//! Ensure existing data files don't contain NULL values in this column
+static void VerifyNotNullConstraint(ClientContext &context, IcebergTableInformation &updated_table,
+                                    IcebergColumnDefinition &column, int32_t current_schema_id) {
```
Tishj: `current_schema_id` is unused here I think

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
yan-alex: I added a comment describing my reasoning for this part
```
/* The column is present in current_schema but not in manifest/snapshots, without a default value, and
* table is not empty.
*
* So now we know that either:
* for this column all rows are null
* OR
* for this column the optional field `null_value_counts` is not present in any manifest.
*
* In either case the constraint fails. We could possibly read the avro file's key-value metadata to be able to
* discern between these two cases, and in the case where the optional field is missing show a different error
* message, e.g. "null_value_counts required for `ALTER ... SET NOT NULL`" , or eventually do a scan to check
* for nulls (time intensive).
*
* We would discern between the two cases as follows:
*	if (no manifest is using the current_schema_id) then we know:
*	for this column all rows are null
*	ELSE
*	for this column the optional field `null_value_counts`  is not present in any manifest.
*/
```

yan-alex:
@Tishj can you re-review? CI passed, the new commit is just a comment

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
Tishj: We can fix this for v1.5.3, it's fine with the limited information we have
I think we'd need to know the schema of the file to make this call correctly

If the file was written before this column exists, all rows in the schema will result in a NULL for the column.
If it was written after this column was added, we know if `null_value_counts` is missing for it, there are 0 NULL rows for the column, otherwise the metadata would be corrupt.

yan-alex: Yup

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -251,6 +252,83 @@ static void VerifySchemaEvolution(const IcebergTableMetadata &table_metadata, co
...
+	auto snapshot_info = updated_table.table_metadata.GetSnapshot(snapshot_lookup);
+	if (!snapshot_info.snapshot) {
+		// Table is empty
+		return;
+	}
+	IcebergOptions options;
+	auto manifest_list = IcebergManifestList::Load(updated_table.BaseFilePath(), updated_table.table_metadata,
+	                                               snapshot_info, context, options);
+	bool found_column_null_count_at_least_once = false;
+	for (auto &list_entry : manifest_list->GetManifestFilesConst()) {
```
Tishj: I have a feeling that this should also check the transactional data, to see if we didn't add any new data that is NULL in the same transaction

In the `IcebergTransactionData` we have the `existing_manifest_list`, which you should use if it's not empty, before falling back to using `IcebergManifestList::Load`
And in the transaction data we also store the `alters` which contain the transactional data (update/insert/delete) of the table

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -489,13 +480,48 @@ void IcebergSchemaEntry::Alter(CatalogTransaction transaction, AlterInfo &info)
 
 		auto &column = ResolveColumn<SetNotNullInfo>(set_not_null_info, new_schema);
 
-		VerifyNotNullConstraint(context, updated_table, column);
+		// Use IcebergTransactionData existing_manifest_list if available
+		const vector<IcebergManifestListEntry> manifest_files = !transaction_data.existing_manifest_list.empty()
```
Tishj: This doesn't use transactional data (uncommitted updates/deletes/inserts)

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -252,20 +253,28 @@ static void VerifySchemaEvolution(const IcebergTableMetadata &table_metadata, co
...
+//! Ensure existing data files don't contain NULL values in this column
+static void VerifyNotNullConstraint(IcebergColumnDefinition &column,
+                                    const vector<IcebergManifestListEntry> &manifest_files, string &table_name) {
+	if (manifest_files.empty()) {
+		// Table is empty
+		return;
+	}
 	bool found_column_null_count_at_least_once = false;
-	for (auto &list_entry : manifest_list->GetManifestFilesConst()) {
+	for (auto &list_entry : manifest_files) {
```
Tishj: I think this is also iterating over DELETE manifests right now, you should skip those

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
yan-alex: I think it does, no? I added tests to see if transaction changes are taken into account and they are passing
See:
alter_set_not_null_for_transaction_added_column.test
alter_set_not_null_for_transaction_added_column_with_default.test
alter_set_not_null_for_transaction_table.test
alter_set_not_null_with_transaction_changes.test

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
yan-alex: I do `continue;` on deletes in line [280](https://github.com/duckdb/duckdb-iceberg/pull/923/changes/BASE..8dc9179c9cf2b09ae87d99784993d1da6c77b0c2#diff-0a4190f6590b4f4616705b599b3ecab6122ca5da14122264010104f80394feb5R280)

Tishj: Yes that's good, but those are manifest entries with `status == IcebergManifestEntryStatusType::DELETED`, I am talking about **manifest files** with `content == IcebergManifestContentType::DELETE`

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
Tishj: Hmm, I'm looking at the code and I don't see how the `transaction_data.alters` is used

yan-alex: The IcebergTransactionData.manifest_list is up-to-date with transaction changes, and now we use that if available. I believe that's why it works

yan-alex: Oh actually I do need to check them I believe. I'm investigating

Tishj:
```
    frame #0: 0x000000010aeb0d4c unittest`duckdb::VerifyNotNullConstraint(column=0x000060c000140140, manifest_files=0x00000001701cb570, table_name="my_table") at iceberg_schema_entry.cpp:278:31
   275          }
   276          bool found_column_null_count_at_least_once = false;
   277          for (auto &list_entry : manifest_files) {
-> 278                  for (auto &manifest_entry : list_entry.manifest_entries) {
   279                          if (manifest_entry.status == IcebergManifestEntryStatusType::DELETED) {
   280                                  continue;
   281                          }
(lldb) p list_entry.manifest_entries
(duckdb::vector<duckdb::IcebergManifestEntry>) {
  std::__1::vector<duckdb::IcebergManifestEntry, std::__1::allocator<duckdb::IcebergManifestEntry> > = size=0 {}
```

```
lldb -- ./build/debug/test/unittest --test-config test/configs/fixture.json test/sql/local/irc_any_catalog/alter/set_not_null/alter_set_not_null_with_transaction_changes.test
```
No, that's not how it works

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -252,6 +256,87 @@ static void VerifySchemaEvolution(const IcebergTableMetadata &table_metadata, co
...
+	}
+	return existing_manifest_list;
+}
+
+//! Ensure existing data files don't contain NULL values in this column
+static void VerifyNotNullConstraint(ClientContext &context, IcebergColumnDefinition &column, const string &catalog_name,
+                                    const string &schema_name, const string &table_name) {
+	Connection con(*context.db);
+	auto query = StringUtil::Format(
+	    "SELECT COUNT(*) FROM %s.%s.%s WHERE %s IS NULL", KeywordHelper::WriteOptionallyQuoted(catalog_name),
```
Tishj: That's not exactly what I meant 😅 
This won't cover the transactional data, we need to manually start a scan here, similar to what we do in `IcebergMultiFileList::ScanDeleteFile`, only we're starting an iceberg scan and not a parquet scan

yan-alex: This does work if we don't support transaction changes though

yan-alex: So your way of scanning would take transactional changes and deletes into account?

Tishj: It should, yea

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -296,20 +299,58 @@ vector<IcebergManifestListEntry> PopulateExistingManifestList(ClientContext &con
...
+
+	// Push IS NULL filter
+	TableFilterSet filter_set;
+	filter_set.PushFilter(ColumnIndex(column_idx), make_uniq<IsNullFilter>());
+
+	auto &multi_file_list = multi_file_bind.file_list->Cast<IcebergMultiFileList>();
+	auto filtered_list = multi_file_list.PushdownInternal(context, filter_set);
+	multi_file_bind.file_list = std::move(filtered_list);
+
+	auto &return_types = multi_file_bind.types;
```
Tishj: Probably `return_types` should consist of only the type of the column we are scanning
`column_ids` should contain only `column_idx`, and projection ids (`{column_idx}` currently) should just be `vector<idx_t>()`
Forgot to fix that in the patch I sent

yan-alex: Yeah this is the error I was fixing indeed. It was trying two columns but that was throwing an assertion

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -255,6 +262,132 @@ static void VerifySchemaEvolution(const IcebergTableMetadata &table_metadata, co
 	throw CatalogException(error);
 }
 
+vector<IcebergManifestListEntry> RetrieveManifestFiles(ClientContext &context, IcebergTableInformation &updated_table) {
```
Tishj: Dead code?

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -255,6 +262,132 @@ static void VerifySchemaEvolution(const IcebergTableMetadata &table_metadata, co
...
+	if (!snapshot_info.snapshot) {
+		return std::vector<IcebergManifestListEntry>();
+	}
+	IcebergOptions options;
+	auto manifest_list = IcebergManifestList::Load(updated_table.BaseFilePath(), updated_table.table_metadata,
+	                                               snapshot_info, context, options);
+	return manifest_list->GetManifestListEntries();
+}
+
+vector<IcebergManifestListEntry> PopulateExistingManifestList(ClientContext &context,
```
Tishj: Dead code?

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -255,6 +262,132 @@ static void VerifySchemaEvolution(const IcebergTableMetadata &table_metadata, co
...
+}
+
+//! Ensure existing data files don't contain NULL values in this column
+static void VerifyNotNullConstraint(ClientContext &context, IcebergColumnDefinition &column,
+                                    IcebergTableEntry &table_entry) {
+	// Get the scan function directly from the table entry
+	unique_ptr<FunctionData> bind_data;
+	EntryLookupInfo lookup(CatalogType::TABLE_ENTRY, table_entry.name);
+	auto scan_function = table_entry.GetScanFunction(context, bind_data, lookup);
+
```
Tishj: ~~If I'm not mistaken, this is missing a `PrepareIcebergScanFromEntry` call still~~
Actually, no, it's performed by `GetScanFunction` 👍

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -255,6 +262,132 @@ static void VerifySchemaEvolution(const IcebergTableMetadata &table_metadata, co
...
+	ExecutionContext execution_context(context, thread_context, nullptr);
+
+	// Initialize scan state
+	TableFilterSet input_filter_set;
+	input_filter_set.PushFilter(ColumnIndex(0), make_uniq<IsNullFilter>());
+
+	TableFunctionInitInput input(bind_data.get(), column_ids, {vector<idx_t>()}, input_filter_set);
+	auto global_state = scan_function.init_global(context, input);
+	auto local_state = scan_function.init_local(execution_context, input, global_state.get());
+	// auto &local_state = local_state_p.Cast<ParquetReadLocalState>();
```
Tishj: dead code

Tishj (changes requested):
Can we add a test with 2 connections, each starting a transaction
first transaction alters the column to be `required`, commits
second transaction inserts NULL into the column, commits

I wonder if there's some merit to other engines not accepting `SET NOT NULL`
We have to rely on the catalog to catch these problems in this case, and the catalog is so far proving that it won't

Tishj (approved):
Thanks 👍 
Sorry for the whiplash on the "should we, should we not" question :joy:

yan-alex:
@Tishj No thank you for your help! This was a tricky one :)
CI passed [on fork ](https://github.com/yan-alex/duckdb-iceberg/pull/14) so we can merge

PR #924, opened 2026-04-14
[ALTER] Add `ALTER TABLE ... RENAME TO ...` support

Tishj:
This PR implements https://github.com/duckdblabs/duckdb-internal/issues/8611

### Known limitations

We don't allow creation of the table or other ALTER statements to happen in the same transaction as a RENAME of the table.


src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
```diff
@@ -407,6 +407,25 @@ void IcebergSchemaEntry::Alter(CatalogTransaction transaction, AlterInfo &info)
...
+		auto other_catalog_entry = tables.GetEntry(context, lookup);
+		if (other_catalog_entry) {
+			//! The table exists at this point, check if it was deleted/renamed in the transaction
+			auto &other_table_entry = other_catalog_entry->Cast<IcebergTableEntry>();
+			auto &other_table_info = other_table_entry.table_info;
+			auto other_table_key = other_table_info.GetTableKey();
+			auto state = irc_transaction.GetLatestTableState(other_table_key);
+			if (!state || state->status == IcebergTableStatus::ALIVE) {
+				throw CatalogException("Table with name \"%s\" already exists!", new_name);
+			}
```
Tmonster, line 424: What happens if the `if` statement above doesn't evaluate to true? Seems like we go ahead with the rename.
Oh, I guess the table is dropped? Should we add a D_ASSERT on the `IcebergTableStatus`?

src/catalog/rest/iceberg_table_set.cpp
```diff
@@ -306,20 +323,21 @@ optional_ptr<CatalogEntry> IcebergTableSet::GetEntry(ClientContext &context, con
 		return table_info.GetSchemaVersion(snapshot_lookup, context);
 	}
 
-	if (entries.find(table_name) != entries.end()) {
-		entries.erase(table_name);
-	}
-	auto it = entries.emplace(table_name, make_shared_ptr<IcebergTableInformation>(ic_catalog, schema, table_name));
-	auto entry = it.first;
-	auto &table_info = *entry->second;
+	shared_ptr<IcebergTableInformation> old_version;
```
Tmonster, line 327: nit: maybe a comment here explaining that we push keep the old version in the entries in case a transaction still relies on it

test/sql/local/irc_any_catalog/alter/rename_table/create_and_rename.test
```diff
@@ -0,0 +1,140 @@
...
+query I
+select * from my_datalake.default.rename_table order by all
+----
+false
+true
+
+statement error con1
+commit
+----
+<REGEX>:.*ransactionContext Error: Failed to commit:.*
```
Tmonster: nit: missing T

Tmonster (commented):
lgtm, just some small nits.
Extension ci tools submodule supposed to be updated?

src/catalog/rest/catalog_entry/iceberg_schema_entry.cpp
Tishj: We do a look up to check if it existed at the start of the transaction, if that's the case, we check what the latest state of the table in the transaction is.
If there is no state, we haven't touched it yet in the transaction, so it's still alive.
Or if the state exists and explicitly says it's alive.

The other cases are, state exists and it's either RENAMED or DROPPED, so it's free to be replaced

src/catalog/rest/iceberg_table_set.cpp
Tishj: We didn't do this before, but seeing as we did `erase` if `FillEntry` failed, I figured it was better to preserve the old entry if it existed

Tishj:
Lakekeeper CI hit this during generation of the tables :eyes:
```
26/04/16 16:05:35 WARN AuthManagers: Inferring rest.auth.type=oauth2 since property credential was provided. Please explicitly set rest.auth.type to avoid this warning.
26/04/16 16:05:35 ERROR Utils: Aborting task
org.apache.iceberg.exceptions.ServiceFailureException: Server error: Unexpected: IO error at `s3://examples/initial-warehouse/019d9705-c5bd-7241-b3a5-792958c6d9c0/019d970a-6a29-73d3-9402-217427583444/metadata/00000-019d970a-6a41-74b0-92e9-d1872d2d83ef.gz.metadata.json`: S3 write failed: Storage backend has reached its minimum free drive threshold. Please delete a few objects to proceed.
```

src/catalog/rest/iceberg_table_set.cpp
Tishj: Knowing what I know now: this is probably not necessary?
If the table was referenced by another transaction, it will be in the `tables` of that `IcebergTransaction`, so it'll stay alive that way.

If `FillEntry` fails here, that means that the table doesn't exist anymore, so it doesn't need to be seen by another other transactions, we could just remove it. No need to put the old version back.

PR #925, opened 2026-04-14
[Dev] Fix problem that was causing files of committed tables to be deleted in multi-table transactions

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/8889

This is arguably not correct behavior, as we break atomicity, but now we don't delete files of tables that successfully committed, if the commit of another table triggered a ROLLBACK.

### DIFF
--- a/src/catalog/rest/transaction/iceberg_transaction.cpp
+++ b/src/catalog/rest/transaction/iceberg_transaction.cpp
@@ -370,8 +370,8 @@ TableTransactionInfo IcebergTransaction::GetTransactionRequest(IcebergTransactio
 			update.CreateUpdate(db, context, commit_state);
 		}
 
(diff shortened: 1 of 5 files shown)


src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -520,6 +530,10 @@ void IcebergTransaction::CleanupFiles() {
 		}
 		auto &alter_update = transaction_update->Cast<IcebergTransactionAlterUpdate>();
 		for (auto &up_table : alter_update.updated_tables) {
+			if (alter_update.committed_tables.count(up_table.first)) {
+				//! Successively committed, no need to roll back
```
Tishj, line 534: Successfully*

PR #927, opened 2026-04-15
Add alter table set table options support

Tmonster:
Add suport for `ALTER TABLE SET ()`. This will allow format version updates if desired. 
Also adds support for `ALTER TABLE RESET ()`;

Provides a lot of the same functionality as the table functions, but the syntax is a bit nicer. 
```
CALL set_iceberg_table_properties(my_datalake.default.create_table_with_properties, {'property1':'value10', 'property2':'value20'});
CALL remove_iceberg_table_properties(my_datalake.default.test_target_file_size_remove, ['write.target-file-size-bytes']);
```


Tishj (changes requested):
Can we also test transactionality of these?

Tishj:
Thanks 👍

PR #932, opened 2026-04-16
[ALTER] Add `ALTER TABLE ... RENAME ... TO ...` support

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/8611

Adds support for renaming columns.
This also stumbles into the issue where we would need the `metadata-log` to accurately get the latest state of the metadata for the transaction.

### DIFF
--- a/test/sql/local/irc_any_catalog/alter/alter_rename_column.test
+++ b/test/sql/local/irc_any_catalog/alter/alter_rename_column.test
@@ -1,0 +1,170 @@
+# name: test/sql/local/irc_any_catalog/alter/alter_rename_column.test
+# group: [alter]
+
+require-env ICEBERG_SERVER_AVAILABLE
+
+require-env SECRETS_CREATED_AND_CATALOG_ATTACHED
+
+require avro
+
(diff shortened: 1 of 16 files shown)


test/sql/local/irc_any_catalog/alter/alter_rename_column.test
```diff
@@ -0,0 +1,171 @@
...
+commit
+
+query II con2
+select alias(COLUMNS(*)) from my_datalake.default.tbl limit 1
+----
+a	b
+
+# FIXME: con2 should not see the rename made by con1 yet, but we need to look at metadata-log to fix this
+statement error con2
+ALTER TABLE my_datalake.default.tbl RENAME b TO c;
```
Tmonster, line 138: I would have thought the `last_updated_ms` on the table_metadata would force the previous schema, but I guess what's happening here is that we are copying the `current_schema_id` which does reflect the changes from con1.

This behavior is fine for me 👍

Tmonster (commented):
lgtm

PR #933, opened 2026-04-16
[Dev] Use `metadata-log` when available

Tishj:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/8774
Follow up to #932 

Should make us grab the right `schema-id` in more places


Tmonster:
I'm wondering if it might be a good idea to make this a setting since it makes another request to the catalog in some cases? Or are theses cases really only to check schema updates? I guess for schema updates it's not as big of a deal

PR #934, opened 2026-04-16
Improve error message for default values in v2 tables

NiclasHaderer:
This improves the error message when trying to set `initial_default` from `Request returned HTTP 500 for HTTP POST to` to `non-null DEFAULT values are not supported for` and blocks users from adding a default with `write_default` if the iceberg version does not support it

I also added a write-default to the table creation because
* The `initial-default` must be set when a field is added and cannot change
* The `write-default` must be set when a field is added and may change
* All columns of `unknown`, `variant`, `geometry`, and `geography` types must default to null. Non-null values for `initial-default` or `write-default` are invalid.

see https://github.com/apache/iceberg/blob/main/format/spec.md


Tmonster:
Hi @NiclasHaderer ,

Looks like you missed a couple of tests? Can you take a look at the CI failures?
Otherwise looks good from my side.

Tmonster:
@NiclasHaderer do you still want this in? Thijs has done a lot of work on column defaults I think, so maybe this is fixed?

NiclasHaderer:
Its not super urgent. I mostly care about the better error messae, but is no find time to make it ready (hence tr draft pr). If you think hat is fixed, we a close this

Tishj:
I ported all the changes in this PR, I believe everything is in
I made one small additional fix though, `write_default` was being added for V2 tables with this PR, which shouldn't be done

src/catalog/rest/api/iceberg_create_table_request.cpp
```diff
@@ -258,10 +258,16 @@ shared_ptr<IcebergTableSchema> IcebergCreateTableRequest::CreateIcebergSchema(
...
+			if (logical_type.id() == LogicalTypeId::VARIANT || logical_type.id() == LogicalTypeId::GEOMETRY) {
+				if (!val.IsNull()) {
+					throw InvalidInputException("Columns of type %s must default to null", logical_type.ToString());
+				}
+			}
 			if (table_metadata.iceberg_version < 3 && !val.IsNull()) {
 				throw InvalidInputException("non-null DEFAULT values are not supported for <V3 tables");
 			}
 			iceberg_column_def->initial_default = make_uniq<Value>(val);
+			iceberg_column_def->write_default = make_uniq<Value>(val);
```
Tishj, line 270: Notably, this change was wrong

PR #947, opened 2026-04-20
Do not compare function pointers for a Cast Function

### LINKED ISSUE #941: iceberg 1.5.2 crashes with 'Unimplemented type for cast (DATE -> INTEGER)' scanning tables with hour→day partition-spec evolution (regression from 1.5.1)
## Summary

Scanning an Iceberg table whose partition spec evolved from `hour(ts)` to `day(ts)` (or any transform that emits Avro `int` + `logicalType=date`) crashes 1.5.2 with:

```
libc++abi: terminating due to uncaught exception of type duckdb::Exception:
{"exception_type":"Conversion","exception_message":"Error while reading file
\"...-m0.avro\": : Unimplemented type for cast (DATE -> INTEGER)\n\n"}
```

i.e. `NotImplementedException` surfaces as `std::terminate`, killing the process.

Downgrading the extension to 1.5.1 (`effe9de4`) on the same DuckDB 1.5.2 binary is **not** sufficient — the issue was introduced in the extension — but running on a DuckDB 1.5.1 build returns the rows correctly.

## Repro (single file, `uv run`)

```python
# /// script
# requires-python = ">=3.10"
# dependencies = [
#     "duckdb==1.5.2",
#     "pyiceberg[sql-sqlite,pyarrow]>=0.10.0",
# ]
# ///
import datetime as dt
import shutil
import tempfile

import duckdb
import pyarrow as pa
from pyiceberg.catalog.sql import SqlCatalog
from pyiceberg.partitioning import PartitionField, PartitionSpec
from pyiceberg.transforms import DayTransform, HourTransform
from pyiceberg.types import LongType, NestedField, TimestampType


def main() -> None:
    tmp = tempfile.mkdtemp(prefix="opti6353-")
    try:
        catalog = SqlCatalog(
            "default",
            uri=f"sqlite:///{tmp}/cat.db",
            warehouse=f"file://{tmp}/wh",
        )
        catalog.create_namespace("ns")

        schema = Schema(
            NestedField(1, "id", LongType(), required=False),
            NestedField(2, "ts", TimestampType(), required=False),
        )
        spec_hour = PartitionSpec(
            PartitionField(source_id=2, field_id=1000, transform=HourTransform(), name="ts_hour"),
        )
        table = catalog.create_table("ns.t", schema=schema, partition_spec=spec_hour)

        # spec 0: hour transform -> Avro 'int' (no logicalType)
        table.append(pa.table({
            "id": pa.array([1], type=pa.int64()),
            "ts": pa.array([dt.datetime(2025, 1, 1, 0, 0)]),
        }))

        # evolve: hour -> day
        with table.update_spec() as us:
            us.remove_field("ts_hour")
            us.add_field("ts", DayTransform(), "ts_day")

        # spec 1: day transform -> Avro 'int' + logicalType=date
        table.append(pa.table({
            "id": pa.array([2], type=pa.int64()),
            "ts": pa.array([dt.datetime(2025, 2, 1, 0, 0)]),
        }))

        con = duckdb.connect(config={"extension_directory": f"{tmp}/ext"})
        print(con.execute(
            f"SELECT * FROM iceberg_scan('{table.metadata_location}') ORDER BY id"
        ).fetchall())
    finally:
        shutil.rmtree(tmp, ignore_errors=True)


if __name__ == "__main__":
    main()
```

Run with `uv run repro.py`.

## Observed

- **DuckDB 1.5.2 + iceberg extension 1.5.2**: `std::terminate` on `NotImplementedException: Unimplemented type for cast (DATE -> INTEGER)` while reading the second-spec manifest (`-m0.avro`).
- **DuckDB 1.5.1 + iceberg extension `effe9de4`**: returns `[(1, 2025-01-01 00:00:00), (2, 2025-02-01 00:00:00)]`.

## Root cause

Commit [`399682e2`](https://github.com/duckdb/duckdb-iceberg/commit/399682e2) ("remove cast, fix default expressions on reader initialization for avro files") removed the globally registered `DATE → INTEGER` reinterpret cast:

```cpp
// src/iceberg_extension.cpp (pre-399682e2)
loader.RegisterCastFunction(LogicalType::DATE, LogicalType::INTEGER,
                            BoundCastInfo(DefaultCasts::ReinterpretCast), 200);
```

and replaced it with a narrower rewriter `FixSamePhysicalTypeCastsInExpr` that walks `BoundCastExpression`s bound during Avro multi-file reader init and swaps `TryVectorNullCast → ReinterpretCast` where source and target share a physical type.

The narrower rewriter misses at least one path. When the manifest reader materialises partition values from a spec that uses a day/month/year transform, Avro yields a DATE vector (because Spark / pyiceberg writes the partition field as `int` with `logicalType=date`), but downstream code expects INTEGER for the day/hour/month/year transform result type (see `IcebergDataFile::ToValue` → `partition_result_type = BIGINT` and `PartitionStructType` → `INTEGER`). A direct `Value::DefaultCastAs(INTEGER)` on a DATE value throws `NotImplementedException`, which is not caught and triggers `std::terminate`.

Possible fixes:
1. Restore the global registered `DATE → INTEGER` `ReinterpretCast` (was `cost=200` so it doesn't interfere with standard cast priority).
2. Or extend the rewriter to cover the remaining `Value::DefaultCastAs` call sites used when merging partition info across specs.

## Impact

Any production Iceberg table where a day/month/year/hour partition transform is introduced via partition-spec evolution and old-spec files haven't been re-written becomes unreadable on 1.5.2. We hit this in production on a customer table that had `hour(ts)` evolved to `day(ts)`. Compaction with `rewrite-all=true` into a single spec sidesteps it; the downgrade to 1.5.1 is a workaround.

## Environment

- macOS 25.4.0 (Darwin), arm64
- Python 3.13 via `uv run`
- DuckDB `1.5.2` from PyPI
- Iceberg extension auto-installed from community/core (1.5.2)

Tmonster:
fixes https://github.com/duckdb/duckdb-iceberg/issues/941

Turns out when Iceberg is loaded, the TryVectorNullCast function in the expression is not the one we are attempting to compare to. Now we will always reinterpret if the source is date and the target is INT **AND** there is a Day transform in the partition specs. This should avoid incorrectly casting a Date to Int. The only other time a Date field is written to avro is with a Date Identity partition, but then DuckDB will correctly assume a Date target type, and the cast will not take place. 

I have tested this manually, but I can also write a test for CI if that is desired


MartenMIK:
Thanks for the quick fix! Can the reproducing bit from the ticket be added as a simple regression test?

Tmonster:
@MartenMIK 

We actually have a number of tests for this specific case, the error only surfaces when the extension is loaded into a DuckDB core binary. Currently all of our tests run with Iceberg Built in. 

I will try and add some test cases that Load Iceberg as an extension

Tmonster:
@carlopi requesting your review to make sure building/testing without linking is sane

.github/workflows/test-fixture-catalog.yml
```diff
@@ -55,6 +60,10 @@ jobs:
         with:
           python-version: '3.13'
 
+      - name: Set LOCAL_EXTENSION_REPO
+        if: inputs.local-extension-repo != ''
+        run: echo "LOCAL_EXTENSION_REPO=${{ inputs.local-extension-repo }}" >> $GITHUB_ENV
+
```
carlopi: This is a bit of a random try, but maybe could fix it
```suggestion
      - name: Set LOCAL_EXTENSION_REPO
        if: inputs.local-extension-repo != ''
        run: LOCAL_EXTENSION_REPO=${{ inputs.local-extension-repo }} ./build/release/duckdb "FORCE INSTALL iceberg;"
```

Maybe you could try something like this?
This is basically, only for iceberg, copy from build path to install path.
I think this should make the rest just work (maybe), let's see.

PR #948, opened 2026-04-20
Warn on partitioned writes with certain table_properties

yan-alex:
https://github.com/duckdb/duckdb-iceberg/issues/942


test/sql/local/irc_any_catalog/partitioning/partitioned_write_row-group-size-bytes.test
```diff
@@ -0,0 +1,30 @@
...
+);
+
+query II
+SELECT log_level, message FROM duckdb_logs
+----
+
+statement error
+INSERT INTO my_datalake.default.my_table SELECT * FROM range(1000);
+----
+Parser Error: Unknown unit for memory: '' (expected: KB, MB, GB, TB for 1000^i units or KiB, MiB, GiB, TiB for 1024^i units)
```
yan-alex: @Tmonster should we do something about this error? Or is warning enough?

Tmonster: If the measurement doesn't have memory, we should add a memory unit ourselves. I think if you just divide the number by 1000 then add 'kb' it should be fine

Tishj:
Isn't the problem highlighted by #942 that the option passed to the parquet writer should always be formatted a certain way? Something we're fully in control of doing o.o?

test/sql/local/irc_any_catalog/partitioning/partitioned_write_row-group-size-bytes.test
Tishj: Even simpler than that, I believe we have a method for it:
`StringUtil::BytesToHumanReadableString`

yan-alex: I now just pass it as an INT if it doesn't have any memory unit specified, then the parser accepts it. If passed as a string, only then does it need memory units

Tmonster (commented):
Thanks!

I know `test/sql/local/irc_any_catalog/test_parquet_copy_options.test` tests that we write the correct sizes, but it looks like we only test with `KB` at the end. Can we also validate the parquet file size when no memory units are in the `write.parquet.row-group-size-bytes` table property?

Tmonster:
> Isn't the problem highlighted by https://github.com/duckdb/duckdb-iceberg/issues/942 that the option passed to the parquet writer should always be formatted a certain way? Something we're fully in control of doing o.o?

Not 100% sure I understand. I can imagine that this value will be formatted differently when it comes to a number of Iceberg Tables. 
I think we should add a test though to make sure we are properly these values regardless of their format. Basically with bytes, we respect the bytes, without bytes, just bytes. 

I think this PR will do that, just want to see a test to validate parquet file size without bytes

src/execution/operator/iceberg_insert.cpp
```diff
@@ -754,6 +759,14 @@ IcebergCopyOptions IcebergInsert::GetCopyOptions(ClientContext &context, Iceberg
 
 	result.use_tmp_file = false;
 	if (copy_input.partition_spec) {
+		if (table_properties.find("write.target-file-size-bytes") != table_properties.end()) {
+			DUCKDB_LOG_WARNING(context,
+			                   "Property target-file-size-bytes is currently only supported for unpartitioned writes");
```
Tmonster: Reading these Warning messages again and I feel like the UX here is not ideal. We say something is not supported, but we print a warning which I think gives the impression we did support it. 

I think either the message should be something like 
```
Table property target-file-size-bytes is currently not supported for partitioned tables. This property is being ignored.
```

Or we should throw an error, and the user will have to call 
```
SET ignore_target_file_size_bytes_for_partitioned_tables=true
```
To allow the query to succeed.

My preference is on option 2. But I know option 1 might be easier to implement. 
@Tishj any thoughts

yan-alex: @Tmonster 
Implemented option 2. Also merged variagata for test refactoring.

src/execution/operator/iceberg_insert.cpp
```diff
@@ -725,7 +725,12 @@ IcebergCopyOptions IcebergInsert::GetCopyOptions(ClientContext &context, const I
 		auto &mapping = ICEBERG_TABLE_PROPERTY_MAPPING[i];
 		auto it = table_properties.find(mapping.iceberg_option);
 		if (it != table_properties.end()) {
-			info->options[mapping.parquet_option].emplace_back(it->second);
+			if (StringUtil::CIEquals(mapping.parquet_option, "row_group_size_bytes") &&
+			    std::isdigit(it->second.back())) {
+				info->options[mapping.parquet_option].emplace_back(Value::UBIGINT(std::stoull(it->second)));
```
Tmonster: nit: Might be good to use duckdb's `StringUtil::ToUnsigned` instead of `std::stoull`

src/execution/operator/iceberg_insert.cpp
```diff
@@ -725,7 +725,12 @@ IcebergCopyOptions IcebergInsert::GetCopyOptions(ClientContext &context, const I
 		auto &mapping = ICEBERG_TABLE_PROPERTY_MAPPING[i];
 		auto it = table_properties.find(mapping.iceberg_option);
 		if (it != table_properties.end()) {
-			info->options[mapping.parquet_option].emplace_back(it->second);
+			if (StringUtil::CIEquals(mapping.parquet_option, "row_group_size_bytes") &&
+			    std::isdigit(it->second.back())) {
```
Tmonster: nit: Here to, I think you can use `StringUtil::CharacterIsDigit`

src/execution/operator/iceberg_insert.cpp
```diff
@@ -754,6 +759,26 @@ IcebergCopyOptions IcebergInsert::GetCopyOptions(ClientContext &context, const I
 
 	result.use_tmp_file = false;
 	if (copy_input.partition_spec) {
+		if (table_properties.find("write.target-file-size-bytes") != table_properties.end()) {
+			Value ignore_target_file_size_bytes_for_partitioned_tables;
+			if (!context.TryGetCurrentSetting("ignore_target_file_size_bytes_for_partitioned_tables",
+			                                  ignore_target_file_size_bytes_for_partitioned_tables) ||
+			    !ignore_target_file_size_bytes_for_partitioned_tables.GetValue<bool>()) {
+				throw InvalidInputException("Table property target-file-size-bytes is currently not supported for "
+				                            "partitioned tables. This property is being ignored.\nTo ignore this error "
```
Tmonster: I think we can remove the sentence `This property is being ignored.` It technically is not being ignored.

src/execution/operator/iceberg_insert.cpp
```diff
@@ -754,6 +759,26 @@ IcebergCopyOptions IcebergInsert::GetCopyOptions(ClientContext &context, const I
...
+				                            "run \"SET ignore_target_file_size_bytes_for_partitioned_tables=true\"");
+			}
+		}
+		if (table_properties.find("write.parquet.row-group-size-bytes") != table_properties.end()) {
+			Value ignore_row_group_size_bytes_for_partitioned_tables;
+			if (!context.TryGetCurrentSetting("ignore_row_group_size_bytes_for_partitioned_tables",
+			                                  ignore_row_group_size_bytes_for_partitioned_tables) ||
+			    !ignore_row_group_size_bytes_for_partitioned_tables.GetValue<bool>()) {
+				throw InvalidInputException("Table property row-group-size-bytes is currently not supported for "
+				                            "partitioned tables. This property is being ignored.\nTo ignore this error "
```
Tmonster: nit: same here

Tmonster (commented):
Awesome! Just the small nits then it's good to go

yan-alex:
@Tmonster addressed your feedback

Tmonster:
Thanks!

PR #949, opened 2026-04-20
fix: handle explicit null doc field in StructField JSON parsing

Amneet13:
## Problem

When an Iceberg table writes `"doc": null` explicitly in schema fields instead of omitting the key entirely, DuckDB's StructField parser fails because it treats any non-absent `doc` value as requiring a string type, causing queries to fail with:

`    StructField property 'doc' is not of type 'string', found 'null' instead`

## Fix

Added a `!yyjson_is_null` guard so an explicit JSON null is treated identically to an absent field — consistent with the Iceberg spec, which defines `doc` as an optional string where null is a valid representation of "not set".

## Iceberg Spec Reference

> `doc`: **optional string** — a comment describing the field

Null is a valid JSON representation of an absent optional field.

## Affected Input

```json
{ "id": 0, "name": "id", "required": true, "type": "int", "doc": null }

### DIFF
--- a/src/rest_catalog/objects/struct_field.cpp
+++ b/src/rest_catalog/objects/struct_field.cpp
@@ -70,7 +70,7 @@ string StructField::TryFromJSON(yyjson_val *obj) {
 		}
 	}
 	auto doc_val = yyjson_obj_get(obj, "doc");
-	if (doc_val) {
+	if (doc_val && !yyjson_is_null(doc_val)) {
 		has_doc = true;
 		if (yyjson_is_str(doc_val)) {
 			doc = yyjson_get_str(doc_val);


Tishj:
All of this is generated from `scripts/api.yaml`
We're not accepting direct changes to the generated code

The field should have been explicitly nullable in the spec if that's the case
https://github.com/duckdb/duckdb-iceberg/blob/2cfec2f99767c4118871be8579880e7bb7391fb4/scripts/api.yaml#L2225

Amneet13:
> All of this is generated from `scripts/api.yaml` We're not accepting direct changes to the generated code
> 
> The field should have been explicitly nullable in the spec if that's the case
> 
> https://github.com/duckdb/duckdb-iceberg/blob/2cfec2f99767c4118871be8579880e7bb7391fb4/scripts/api.yaml#L2225

Let me update the `scripts/api.yaml` file and generate the code. Thanks

Tishj:
That should work, but I'd prefer raising an issue [here](https://github.com/apache/iceberg/issues) because we didn't write that `api.yaml`
We already have enough small alterations we've had to make to it to make it work, and I'd like this to be upstreamed so it eventually lands in the source and we don't have to worry about migrating this every time we update the vendored `api.yaml`

Amneet13:
@Tishj just created a PR on iceberg lets see https://github.com/apache/iceberg/pull/16070

Tishj:
I was reading <https://www.speakeasy.com/openapi/schemas/null#optional-vs-nullable-vs-optional--nullable> prompted by this issue, and I think this might be something I'll push a fix for in the generator, as advised by that article:
> If the API treats null and omitted the same way, remove the nullable modifier and keep the field as just optional. 

Since the field is already optional, it might make sense to just make it behave as nullable by default

PR #951, opened 2026-04-21
Use metadata log in scan

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/8825

Similar to #946 we now also use `metadata-log` when grabbing the `IcebergTableEntry` from the Catalog.

I've added a new `iceberg_use_metadata_log` setting, which defaults to `true`.
This can be set to false to prevent making further GET requests for the relevant `metadata.json` file.
This could result in breaking atomicity guarantees that shield transactions from seeing outside changes.

This PR also changes the `table_request_cache` expiration and validation to be based on the transaction timestamp, rather than the exact time the request was made.

### DIFF
--- a/src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
+++ b/src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
@@ -474,9 +474,16 @@ optional_ptr<CatalogEntry> IcebergTableInformation::GetSchemaVersion(ClientConte
 		}
 		schema_id = snapshot.GetSchemaId();
 	} else {
+		bool use_metadata_log = true;
+		Value val;
+		if (context.TryGetCurrentSetting("iceberg_use_metadata_log", val)) {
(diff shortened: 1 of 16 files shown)


Tmonster:
Can we also check the behavior (or document it) if MAX_TABLE_STALENESS is set?

Tishj:
> Can we also check the behavior (or document it) if MAX_TABLE_STALENESS is set?

Euhm, sure?
It shouldn't affect it, because the first time the table request is made, it gets cached.
The problem this deals with is that the stored metadata is too recent for the current transaction.

In the case of max-table-staleness, that should never happen unless your transaction is so old that the time you query the table for the first time in your transaction, the cache's expiration has been reached.

We could change the max-table-staleness to act on transaction timestamp, instead of `system_clock::now()`, then even that problem disappears.

EDIT:
Changed the max-table-staleness to be based on transaction timestamp

Tmonster:
I see now this is an attach option. What do we think about making it a config option so a user can disable metadata log without having to dettach/reattach?

PR #953, opened 2026-04-22
[Dev] Update OpenAPI code generator to allow `null` for optional fields.

Tishj:
This PR supersedes #949


Amneet13:
Awesome. Could you share a rough timeline for when this will be merged to main?

PR #954, opened 2026-04-22
New table function `iceberg_load_table_response()`

yan-alex:
https://github.com/duckdblabs/duckdb-internal/issues/8946


src/function/metadata/iceberg_load_table_response.cpp
```diff
@@ -0,0 +1,257 @@
...
+	auto ret = make_uniq<IcebergLoadTableResponseBindData>();
+
+	const unique_ptr<HTTPResponse> response = MakeRequest(context, input);
+
+	// Parse the response using yyjson
+	auto doc = ICUtils::APIResultToDoc(response->body);
+	auto *root = yyjson_doc_get_root(doc.get());
+
+	// Parse via LoadTableResult for the typed fields
+	auto load_result = rest_api_objects::LoadTableResult::FromJSON(root);
```
Tishj: nit: can't we store this directly in the bind result?, then we don't need to define `struct StorageCredentialData`

src/function/metadata/iceberg_load_table_response.cpp
```diff
@@ -0,0 +1,257 @@
...
+	return_types.push_back(LogicalType::LIST(credential_struct));
+
+	// request_url
+	names.push_back("request_url");
+	return_types.push_back(LogicalType::VARCHAR);
+
+	return std::move(ret);
+}
+
+static void OutputMap(const case_insensitive_map_t<string> &config, vector<Vector>::value_type &config_vec) {
```
Tishj: nit: `Vector &config_vec` ?

Tishj (approved):
Thanks 👍

src/function/metadata/iceberg_load_table_response.cpp
```diff
@@ -0,0 +1,257 @@
...
+		                  EnumUtil::ToString(response->status), response->body);
+	}
+	return response;
+}
+
+static unique_ptr<FunctionData> IcebergLoadTableResponseBind(ClientContext &context, TableFunctionBindInput &input,
+                                                             vector<LogicalType> &return_types, vector<string> &names) {
+	auto ret = make_uniq<IcebergLoadTableResponseBindData>();
+
+	const unique_ptr<HTTPResponse> response = MakeRequest(context, input);
```
Tmonster: Is there a reason we make the request during the Bind phase? We know the schema of the response, so we should be able to make the request in `IcebergLoadTableResponseFunction` right?

src/function/metadata/iceberg_load_table_response.cpp
```diff
@@ -0,0 +1,257 @@
...
+		FlatVector::GetData<string_t>(*prefix_vec)[struct_idx] = StringVector::AddString(*prefix_vec, cred.prefix);
+
+		// config map for this credential
+		OutputMap(cred.config, *cred_config_vec);
+
+		auto inner_config_count = cred.config.size();
+		ListVector::Reserve(*cred_config_vec, inner_config_count);
+		auto &inner_key_vec = MapVector::GetKeys(*cred_config_vec);
+		auto &inner_val_vec = MapVector::GetValues(*cred_config_vec);
+		idx_t j = 0;
```
Tmonster: nit: I feel `j` could have a more descriptive variable name

Tmonster (commented):
This is gonna be super helpful for debugging. Thanks!
I think for me, the biggest thing might be moving the request to the `IcebergLoadTableResponseFunction` function call. 

Possible in the future there will be a request to expand this function so it can take a list of tables, and we return a list of the Load Table Information responses to those tables. Then the request will have to move into the function anyway

Tmonster:
@Tishj Any thoughts here on what we should do for vended credentials? Should we print those in plain text? Or should we redact them? I feel like redacting them doesn't really do much, since a user will already have access to the table, but I also feel like leaving them as plain text can lead to unforeseen consequences

Tmonster:
Thanks!

PR #958, opened 2026-04-24
Ensure transaction safety against concurrent DROP + RENAME

yan-alex:
# The Problem
Let's look at an example where we have two tables
### accounts
| id | name  | balance |
|----|-------|---------|
| 1  | Alice | 1000.00 |
| 2  | Bob   | 500.00  |

### external_accounts
| id | name  | balance  |
|----|-------|----------|
| 1  | CorpX | 99000.00 |
| 2  | CorpY | 42000.00 |

Say Alice wants to pay 500 to Bob. We would do:

```sql
UPDATE my_datalake.default.accounts
SET balance = balance - 500.00
WHERE account_id = 1;

UPDATE my_datalake.default.accounts
SET balance = balance + 500.00
WHERE account_id = 2;
```
But if at the same time another concurrent transaction would drop `accounts` and rename`external_accounts` -> `accounts`, our transaction would go through and result in:

### accounts (borked)
| id | name  | value    |
|----|-------|----------|
| 1  | CorpX | 99000.00 |
| 1  | Alice | 500.00   |
| 2  | CorpY | 42500.00 |

# The fix
For every update to a table, we now send a "assert-table-uuid" requirement with the API request to make sure our table is the one we think it is.

Also, because `LoadTableResultCache table_request_cache` was "poisoned" with the renamed table, we now keep the old `table_uuid` and never update it when `IcebergTransactionAlterUpdate::GetOrInitializeTable()` gets called. This was a less-invasive fix.
But we might also want to consider a deeper fix. We currently use `table.GetTableKey()` (which returns a string "namespace.table_name") as an index to access, not only `LoadTableResultCache table_request_cache` , but also`IcebergTransactionAlterUpdate.updated_tables` and maybe more things. We should be using a `table_uuid` instead. But that's perhaps better left to a follow-up PR, as I tried it and it didn't work out of the box.


src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -355,6 +360,12 @@ TableTransactionInfo IcebergTransaction::GetTransactionRequest(IcebergTransactio
 			commit_state.table_change.updates.push_back(std::move(set_snapshot_ref_update));
 		}
 
+		if (!info.has_assert_create && !transaction_data.alters.empty()) {
```
Tishj: `commit_state.table_info.HasTransactionUpdates` I think works better

Tmonster (approved):
Thanks!

src/catalog/rest/transaction/iceberg_transaction_update.cpp
```diff
@@ -20,6 +20,13 @@ IcebergTableInformation &IcebergTransactionAlterUpdate::GetOrInitializeTable(con
 	auto it = updated_tables.find(table_key);
 	if (it == updated_tables.end()) {
 		it = updated_tables.emplace(table_key, table.Copy(transaction)).first;
+		// Preserve the table_uuid from the original table info (resolved at transaction start).
+		// Copy() reads from the global request cache, which can be contaminated by another
+		// transaction's in-flight RENAME overwriting the entry with a different table's metadata.
+		// Only override when the original has a known UUID (skip for newly created tables).
+		if (!table.table_metadata.table_uuid.empty()) {
```
Tmonster, line 27: Actually wait, I think this can go in the `table.Copy(transaction)` function right?

Tmonster: I don't understand the comment where another transactions in-flight rename will overwrite the entry. If the transaction is in flight, the metadata for the rename will still be in the transaction no? 

Also, is there ever a case where a table is in the catalog without a UUID?

yan-alex: @Tmonster 
table uuid is required from v2 onwards. I don't think we care that is optional in v1 right?
Also you're right, it's not in flight, it has to be committed.

Tmonster: `IcebergTableMetadata::FromTableMetadata(const rest_api_objects::TableMetadata &table_metadata)` should copy the UUID for a table, is that not happening for some reason?

yan-alex: The call stack is 
`GetOrInitializeTable(const IcebergTableInformation &table)`>`IcebergTableInformation::Copy(IcebergTransaction &iceberg_transaction)`>`IcebergTableInformation::Copy(ClientContext &context)`
and this last function reads from a `table_request_cache`. The problem with that is, if we have a second connection dropping/renaming tables, the original connection will lose track of it's original table uuid because it's reading from the shared cache

Tmonster: Ah, ok I see it now. All of this transaction/IcebergTableInformation will need a refactor soon

yan-alex: Agreed

Tmonster:
I think this probably to merge with v1.5-variegata so the tests line up?

yan-alex:
@Tmonster merged with `v1.5-variegata` ,should be good to go

Tmonster:
Thanks!

PR #960, opened 2026-04-24
Get, Set, Remove schema_properties

yan-alex:
https://github.com/duckdblabs/duckdb-internal/issues/8996

Added the following functions
`iceberg_schema_properties()`
`set_iceberg_schema_properties()`
`remove_iceberg_schema_properties()`


src/function/metadata/iceberg_schema_properties_functions.cpp
```diff
@@ -0,0 +1,311 @@
...
+                                                               vector<LogicalType> &return_types,
+                                                               vector<string> &names) {
+	auto ret = make_uniq<IcebergSchemaPropertiesBindData>();
+
+	auto input_string = input.inputs[0].ToString();
+	auto iceberg_schema = IcebergUtils::GetSchemaEntry(context, input_string);
+	if (!CheckEntryIsIcebergSchema(iceberg_schema)) {
+		throw InvalidInputException("Cannot call set_iceberg_schema_properties on non-iceberg schema");
+	}
+	ret->iceberg_schema = iceberg_schema->Cast<IcebergSchemaEntry>();
```
Tishj, line 72: The endpoint to hit is optional
Need to check the config to see if the server supports it

src/catalog/rest/catalog_entry/schema/iceberg_schema_entry.cpp
```diff
@@ -652,4 +653,23 @@ IcebergTableSet &IcebergSchemaEntry::GetCatalogSet(CatalogType type) {
 	}
 }
 
+void IcebergSchemaEntry::LoadPropertiesFromIRCAPIIfNeeded(ClientContext &context) {
```
Tmonster: nit: You can omit `IfNeeded` in the function name

src/function/metadata/iceberg_schema_properties_functions.cpp
```diff
@@ -0,0 +1,311 @@
...
+	auto schema_key = iceberg_schema->GetSchemaKey();
+	if (iceberg_transaction.schema_property_updates.find(schema_key) ==
+	    iceberg_transaction.schema_property_updates.end()) {
+		// not present, create one
+		auto properties = SchemaPropertyUpdates();
+		properties.updates = bind_data.properties;
+		iceberg_transaction.schema_property_updates[schema_key] = properties;
+	} else {
+		for (auto &property : bind_data.properties) {
+			auto &schema_property_updates = iceberg_transaction.schema_property_updates[schema_key];
```
Tmonster: nit: These three lines can be outside of the for loop

src/function/metadata/iceberg_schema_properties_functions.cpp
```diff
@@ -0,0 +1,311 @@
...
+	if (iceberg_transaction.schema_property_updates.find(schema_key) ==
+	    iceberg_transaction.schema_property_updates.end()) {
+		// not present, create one
+		auto properties = SchemaPropertyUpdates();
+		properties.removals = bind_data.remove_properties;
+		iceberg_transaction.schema_property_updates[schema_key] = properties;
+
+	} else {
+		for (auto &property_to_remove : bind_data.remove_properties) {
+			auto &schema_property_updates = iceberg_transaction.schema_property_updates[schema_key];
```
Tmonster: nit: here as well, these lines can move out of the for loop

src/include/catalog/rest/transaction/iceberg_transaction.hpp
```diff
@@ -105,6 +111,7 @@ class IcebergTransaction : public Transaction {
 	vector<unique_ptr<IcebergTransactionUpdate>> transaction_updates;
 	//! The latest state of a table (either points into 'transaction_updates' or 'tables')
 	case_insensitive_map_t<IcebergTransactionTableState> current_table_data;
+	case_insensitive_map_t<case_insensitive_map_t<string>> current_schema_properties;
```
Tmonster: nit:
could you re-use the SchemaPropertyUpdates struct here as well?
i.e `case_insentive_map<SchemaPropertyUpdates>`

src/include/execution/operator/iceberg_insert.hpp
```diff
@@ -14,7 +14,7 @@
 #include "duckdb/common/index_vector.hpp"
 
 #include "catalog/rest/catalog_entry/table/iceberg_table_entry.hpp"
-#include "catalog/rest/catalog_entry/iceberg_schema_entry.hpp"
+#include "../../catalog/rest/catalog_entry/schema/iceberg_schema_entry.hpp"
```
Tmonster: no relative paths please

src/include/catalog/rest/catalog_entry/schema/iceberg_schema_information.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+#include "duckdb/common/case_insensitive_map.hpp"
+#include "duckdb/common/types/timestamp.hpp"
+
+namespace duckdb {
+struct IcebergSchemaInformation {
+	case_insensitive_map_t<string> properties;
+	bool properties_loaded = false;
+	timestamp_t properties_expire_timestamp;
```
Tmonster: nit: Is this property used anywhere?

Tmonster (commented):
Thanks! A couple of questions

src/catalog/rest/catalog_entry/schema/iceberg_schema_entry.cpp
yan-alex: I figured I would add FromIRCAPI to the function name to make it explicit where we are making requests, but it's only fired on the first call, so that's why the IfNeeded.
I'll call it `LoadProperties()` then if that's better

src/include/catalog/rest/transaction/iceberg_transaction.hpp
yan-alex: Good thing you brought it up, I removed this field for now.
 I feel like this transaction state tracking should be done as part of a bigger change to keep track of the transaction schema state which we currently don't do. So for this PR I think we just use `IcebergSchemaEntry`

yan-alex:
@Tmonster addressed your feeedback. CI failure is from networking

src/function/metadata/iceberg_schema_properties_functions.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+	// populate removals to be sent to IRC API
+	auto &iceberg_transaction = IcebergTransaction::Get(context, iceberg_schema->catalog);
+	auto schema_key = iceberg_schema->GetSchemaKey();
+	if (iceberg_transaction.schema_property_updates.find(schema_key) ==
+	    iceberg_transaction.schema_property_updates.end()) {
+		// not present, create one
+		auto properties = SchemaPropertyUpdates();
+		properties.updates = bind_data.properties;
+		iceberg_transaction.schema_property_updates[schema_key] = properties;
```
Tmonster: this smells a little bit. Here you declare `properties` on the stack, and then then set `iceberg_transaction.schema_property_updates[schema_key]` to equal that properties on the stack. I imagine a copy constructor is used in this case, which means you can use `= std::move(properties)` instead, which should move ownership directly to `iceberg_transaction.schema_property_updates[schema_key]`

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -569,6 +571,36 @@ void IcebergTransaction::DoSchemaDeletes(ClientContext &context) {
...
+		for (auto &removal : schema_property_updates.removals) {
+			yyjson_mut_arr_add_strcpy(doc, removal_arr, removal.c_str());
+		}
+		auto updates_arr = yyjson_mut_obj_add_obj(doc, root_object, "updates");
+		for (auto &update : schema_property_updates.updates) {
+			yyjson_mut_obj_add_strcpy(doc, updates_arr, update.first.c_str(), update.second.c_str());
+		}
+		auto create_body = JsonDocToString(std::move(doc_p));
+
+		IRCAPI::CommitNamespacePropertiesUpdate(context, ic_catalog, create_body, schema_name_no_catalog);
```
Tmonster: It looks like the namespace properties update does consider nested namespaces. Can we add some tests to make sure that also works?

Tmonster (commented):
Thanks! I'm not sure this will work for nested namespaces unfortunately.

src/catalog/rest/transaction/iceberg_transaction.cpp
yan-alex: @Tmonster added support for nested namespaces

Tishj (approved):
Thanks, LGTM 👍

Tmonster:
Thanks!

Issue #282, opened 2025-06-05
Error occurs when changing the column order during an equality delete select operation


nammaste6:
I added it to the test case like this in equality_deletes.test file, but an error occurs.

equality_deletes.test (Test 1)
```
query II
SELECT name, id FROM ICEBERG_SCAN('data/persistent/equality_deletes/warehouse/mydb/mytable');
----
b	2
b	1
```

Stack Trace is here.

```
================================================================================
Query unexpectedly failed (/Project/private/duckdb-iceberg/test/sql/local/equality_deletes.test:35)
================================================================================
SELECT name, id FROM ICEBERG_SCAN('data/persistent/equality_deletes/warehouse/mydb/mytable');
================================================================================
Actual result:
INTERNAL Error: Vector::Reference used on vector of different type

Stack Trace:

0        duckdb::Exception::Exception(duckdb::ExceptionType, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 64
1        duckdb::InternalException::InternalException(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 20
2        duckdb::Vector::Reference(duckdb::Vector const&) + 100
3        duckdb::ExpressionExecutor::Execute(duckdb::Expression const&, duckdb::ExpressionState*, duckdb::SelectionVector const*, unsigned long long, duckdb::Vector&) + 724
4        duckdb::ExpressionExecutor::Select(duckdb::BoundComparisonExpression const&, duckdb::ExpressionState*, duckdb::SelectionVector const*, unsigned long long, duckdb::SelectionVector*, duckdb::SelectionVector*) + 152
5        duckdb::ExpressionExecutor::Select(duckdb::BoundConjunctionExpression const&, duckdb::ExpressionState*, duckdb::SelectionVector const*, unsigned long long, duckdb::SelectionVector*, duckdb::SelectionVector*) + 1000
6        duckdb::IcebergMultiFileReader::ApplyEqualityDeletes(duckdb::ClientContext&, duckdb::DataChunk&, duckdb::IcebergMultiFileList const&, duckdb::IcebergManifestEntry const&, duckdb::vector<duckdb::MultiFileColumnDefinition, true> const&) + 4084
[...]

PR #963, opened 2026-04-26
Fix #940: short-circuit GetEqualityDeletesForFile on tables with no equality deletes

### LINKED ISSUE #940: Equality deletes crash with Vector::Reference when filter column is pruned from projection (regression of #282 / #285)
## Summary

On `duckdb 1.5.2` with the auto-loaded `iceberg` extension (commit `11fea8ed`, the build currently served by `extensions.duckdb.org` for 1.5.2), filtering on an equality-delete column while not projecting it crashes inside `IcebergMultiFileReader::ApplyEqualityDeletes`:

```
InternalException: Vector::Reference used on vector of different type (source BIGINT referenced VARCHAR)
```

This is the same signature as #282. #285 was intended to turn this crash into a clean planner-level `Not implemented Error: Equality deletes need the relevant columns to be selected`, and that merge commit (`478e7c49`) is an ancestor of the currently distributed build — so either the guardrail regressed in the 2,279 commits since, or our query shape slips past it. Either way the crash is back on released 1.5.2.

## Versions

| | |
|---|---|
| duckdb | `1.5.2` (pypi wheel) |
| iceberg extension | `11fea8ed1d405c4cb69511d13b85d600dcdb44c4` (auto-installed from `extensions.duckdb.org`) |
| avro extension | `a73b60d` |
| OS | macOS arm64 (also reproduced in Linux CI) |
| pyiceberg (fixture only) | `0.10` |

## Reproducer

Single-file, self-contained [uv](https://docs.astral.sh/uv/) script — no S3, no Java, no moto, just local filesystem + sqlite catalog. Runs in ~1s.

```bash
# Save as /tmp/repro.py and run:
cd /tmp && uv run repro.py
```

<details>
<summary>repro.py (~180 LOC)</summary>

```python
#!/usr/bin/env -S uv run --script
# /// script
# requires-python = ">=3.12,<3.13"
# dependencies = [
#     "duckdb==1.5.2",
#     "pyarrow>=17,<22",
#     "pyiceberg[sql-sqlite]==0.10.0",
# ]
# ///
"""Standalone reproducer for a DuckDB iceberg equality-delete projection-pruning crash."""

from __future__ import annotations

import json
import tempfile
import uuid
from pathlib import Path

import duckdb
import pyarrow as pa
import pyarrow.parquet as pq
from pyiceberg.catalog.sql import SqlCatalog
from pyiceberg.manifest import (
    DataFile,
    DataFileContent,
    ManifestContent,
    ManifestEntry,
    ManifestEntryStatus,
    ManifestWriterV2,
    write_manifest_list,
)
from pyiceberg.partitioning import PartitionSpec
from pyiceberg.typedef import Record
from pyiceberg.types import LongType, NestedField, StringType


class DeleteManifestWriterV2(ManifestWriterV2):
    """pyiceberg 0.10 has no public delete-manifest writer; flip content=DELETES."""

    def content(self) -> ManifestContent:
        return ManifestContent.DELETES

    @property
    def _meta(self) -> dict[str, str]:
        return {**super()._meta, "content": "deletes"}


def build_iceberg_table_with_eq_delete(tmp: Path) -> Path:
    warehouse = tmp / "warehouse"
    warehouse.mkdir()
    catalog = SqlCatalog("r", uri=f"sqlite:///{tmp}/c.db", warehouse=f"file://{warehouse}")
    catalog.create_namespace("ns")
    schema = Schema(
        NestedField(1, "id", LongType(), required=True),
        NestedField(2, "val", StringType(), required=False),
        identifier_field_ids=[1],
    )
    tbl = catalog.create_table("ns.t", schema=schema, properties={"format-version": "2"})
    tbl.append(
        pa.Table.from_pylist(
            [{"id": i, "val": chr(ord("a") + i - 1)} for i in range(1, 5)],
            schema=pa.schema(
                [
                    pa.field("id", pa.int64(), nullable=False),
                    pa.field("val", pa.string(), nullable=True),
                ]
            ),
        )
    )
    tbl.refresh()
    base_snap = tbl.current_snapshot()
    assert base_snap is not None
    (data_manifest,) = base_snap.manifests(io=tbl.io)
    table_root = Path(tbl.location().removeprefix("file://"))

    # Equality-delete parquet. PARQUET:field_id MUST match schema field id.
    delete_path = table_root / "data" / f"delete-{uuid.uuid4()}.parquet"
    delete_path.parent.mkdir(parents=True, exist_ok=True)
    pq.write_table(
        pa.Table.from_pylist(
            [{"id": 2}],
            schema=pa.schema(
                [pa.field("id", pa.int64(), nullable=False, metadata={b"PARQUET:field_id": b"1"})]
            ),
        ),
        delete_path,
        compression="zstd",
    )
    delete_size = delete_path.stat().st_size

    delete_file = DataFile.from_args(
        content=DataFileContent.EQUALITY_DELETES,
        file_path=f"file://{delete_path}",
        file_format="PARQUET",
        partition=Record(),
        record_count=1,
        file_size_in_bytes=delete_size,
        equality_ids=[1],
        spec_id=0,
        sort_order_id=None,
    )

    new_snap_id = 9876543210123456789
    seq = base_snap.sequence_number + 1
    delete_manifest_path = table_root / "metadata" / f"{uuid.uuid4()}-m0.avro"
    with DeleteManifestWriterV2(
        spec=PartitionSpec(spec_id=0),
        schema=tbl.schema(),
        output_file=tbl.io.new_output(f"file://{delete_manifest_path}"),
        snapshot_id=new_snap_id,
        avro_compression="gzip",
    ) as w:
        w.add(
            ManifestEntry.from_args(
                status=ManifestEntryStatus.ADDED,
                snapshot_id=new_snap_id,
                sequence_number=seq,
                file_sequence_number=seq,
                data_file=delete_file,
            )
        )
    delete_manifest = w.to_manifest_file()

    mlist_path = table_root / "metadata" / f"snap-{new_snap_id}-1-{uuid.uuid4()}.avro"
    with write_manifest_list(
        format_version=2,
        output_file=tbl.io.new_output(f"file://{mlist_path}"),
        snapshot_id=new_snap_id,
        parent_snapshot_id=base_snap.snapshot_id,
        sequence_number=seq,
        avro_compression="gzip",
    ) as mlw:
        mlw.add_manifests([data_manifest, delete_manifest])

    old_meta = json.loads(Path(tbl.metadata_location.removeprefix("file://")).read_text())
    ts = base_snap.timestamp_ms + 1000
    new_meta = {
        **old_meta,
        "snapshots": list(old_meta.get("snapshots", []))
        + [
            {
                "sequence-number": seq,
                "snapshot-id": new_snap_id,
                "parent-snapshot-id": base_snap.snapshot_id,
                "timestamp-ms": ts,
                "summary": {"operation": "overwrite", "added-delete-files": "1"},
                "manifest-list": f"file://{mlist_path}",
                "schema-id": 0,
            }
        ],
        "current-snapshot-id": new_snap_id,
        "last-sequence-number": seq,
        "last-updated-ms": ts,
        "refs": {**old_meta.get("refs", {}), "main": {"snapshot-id": new_snap_id, "type": "branch"}},
    }
    new_meta_path = table_root / "metadata" / "v2.metadata.json"
    new_meta_path.write_text(json.dumps(new_meta, indent=2))
    return new_meta_path


def run_query(ext_dir: Path, metadata: Path, label: str, sql: str) -> None:
    con = duckdb.connect()
    con.execute(f"SET extension_directory = '{ext_dir}';")
    con.execute("INSTALL iceberg;")
    con.execute("LOAD iceberg;")
    try:
        rows = con.execute(sql.format(t=metadata)).fetchall()
        print(f"  OK    [{label}] -> {rows}")
    except duckdb.Error as e:
        print(f"  CRASH [{label}]")
        print(f"        {type(e).__name__}: {str(e).splitlines()[0]}")
    finally:
        con.close()


def main() -> None:
    print(f"duckdb version: {duckdb.__version__}")
    tmp = Path(tempfile.mkdtemp(prefix="iceberg-eq-delete-repro-"))
    print(f"tmp dir:        {tmp}")
    metadata = build_iceberg_table_with_eq_delete(tmp)
    ext_dir = tmp / "duckdb_ext"

    print("\n--- Baseline: scan returns 3 rows (id=2 was equality-deleted) ---")
    run_query(ext_dir, metadata, "SELECT id, val",
              "SELECT id, val FROM iceberg_scan('{t}') ORDER BY id")

    print("\n--- Safe: filter on id, include id in projection ---")
    run_query(ext_dir, metadata, "SELECT id, val WHERE id=3",
              "SELECT id, val FROM iceberg_scan('{t}') WHERE id = 3")

    print("\n--- BUG: filter on id, project OMITS id ---")
    run_query(ext_dir, metadata, "SELECT val WHERE id=3",
              "SELECT val FROM iceberg_scan('{t}') WHERE id = 3")


if __name__ == "__main__":
    main()
```

</details>

**Note on the pyiceberg subclass:** pyiceberg 0.10 has no public delete-manifest writer (its concrete `ManifestWriterV2` hard-codes `content=DATA`). The `DeleteManifestWriterV2` override in the reproducer is a write-side workaround purely to produce an on-disk table shape that mirrors what a merge-on-read engine (Spark, Flink) would produce. The reported bug is on the **read** side; a table generated by any writer that produces equality deletes reproduces it.

## Actual output

```
duckdb version: 1.5.2

--- Baseline: scan returns 3 rows (id=2 was equality-deleted) ---
  OK    [SELECT id, val] -> [(1, 'a'), (3, 'c'), (4, 'd')]

--- Safe: filter on id, include id in projection ---
  OK    [SELECT id, val WHERE id=3] -> [(3, 'c')]

--- BUG: filter on id, project OMITS id ---
  CRASH [SELECT val WHERE id=3]
        InternalException: INTERNAL Error: Vector::Reference used on vector of different type (source BIGINT referenced VARCHAR)
```

## Expected

One of:
- **Ideally:** the query returns `[('c',)]` — the planner keeps the equality-delete column(s) materialized internally for `ApplyEqualityDeletes` and drops them from the final output.
- **At minimum:** the #285 planner-level guard fires and throws `Not implemented Error: Equality deletes need the relevant columns to be selected`.

Neither is happening; the `Vector::Reference` crash of #282 is back.

## Query-shape matrix

Probed against the same fixture. "Safe" shapes project the equality-delete column (or use no per-column projection, i.e. aggregates); "Crash" shapes filter on it without projecting it. Subqueries/CTEs/LIMIT/IN/CAST-in-filter do **not** block the projection pushdown.

| Query | Result |
|---|---|
| `SELECT id, val WHERE id = X` | OK |
| `SELECT id WHERE id = X` | OK |
| `SELECT CAST(id AS BIGINT), val WHERE id = X` | OK (cast in projection keeps id) |
| `SELECT count(*) WHERE id = X` | OK |
| `SELECT count(id) WHERE id = X` | OK |
| `SELECT val FROM scan WHERE val = 'c'` | OK (filter not on eq-delete col) |
| `SELECT val WHERE id = X` | **CRASH** |
| `SELECT val WHERE id = X LIMIT 1` | **CRASH** |
| `SELECT val WHERE id IN (X, Y)` | **CRASH** |
| `SELECT count(val) WHERE id = X` | **CRASH** |
| `SELECT val WHERE CAST(id AS BIGINT) = X` | **CRASH** (cast in filter doesn't help) |
| `SELECT val FROM (SELECT id, val FROM scan) WHERE id = X` | **CRASH** (subquery pushed through) |
| `WITH t AS (SELECT id, val FROM scan) SELECT val FROM t WHERE id = X` | **CRASH** (CTE pushed through) |

Related: `SELECT id FROM scan WHERE val = 'c'` also crashes with `InternalException: Attempted to access index 1 within vector of size 1` — same stack (`ApplyEqualityDeletes`), different symptom, also matches #282's second trace.

## Workaround

At the query-builder layer, ensure any column used in `WHERE` against an equality-delete table also appears in the projection. DuckDB's projection pushdown walks through subqueries and CTEs, so wrapping does not protect the inner `iceberg_scan` — it must be done at the scan's direct projection.

Related: #282, #285.

praveentandra:
## Summary

Fixes #940 by adding a fast-path early-return + defensive bounds check at the top of `IcebergMultiFileList::GetEqualityDeletesForFile`.

## The bug

```cpp
// src/planning/iceberg_multi_file_list.cpp:830
auto &manifest_file = data_manifests[bound_manifest_entry.manifest_file_idx].entry.file;
```

This line dereferences `data_manifests[manifest_file_idx]` **unconditionally** on every read of every iceberg data chunk — even when no equality delete files exist in the snapshot. When `manifest_file_idx` is corrupted (the regression described in #940), this throws `InternalException("Attempted to access index N within vector of size M")`. DuckDB then escalates to a fatal error that invalidates the entire connection (`"database has been invalidated because of a previous fatal error. The database must be restarted prior to being used again."`), cascading every subsequent query on that connection into failure.

## Reproducer context

Reproduced on `duckdb 1.5.2` (post-1.5.2 commit `a1fc7abe20` on `v1.5-variegata`) with `duckdb-iceberg` at commit `b798078f`, scanning a Snowflake-managed Iceberg v3 dynamic table. By spec, [Snowflake-managed v3 tables don't emit equality delete files](https://docs.snowflake.com/en/release-notes/2026/other/2026-03-04-iceberg-v3-support-preview) — they use deletion vectors instead — so the function should be a no-op. But the unconditional `data_manifests[manifest_file_idx]` read fires before the existing `delete_rows.empty()` early-out in `ApplyEqualityDeletes`, so a corrupted index still crashes the connection.

The crash is non-deterministic for any individual query but reproducible across a workload: a single TPC-DS Q60 (3-fact UNION ALL) run on a fresh DuckDB connection succeeds in ~15s; the same query after ~50 prior queries on the same connection crashes inside `ApplyEqualityDeletes`. Suggests `manifest_file_idx` accumulates corruption over connection lifetime, consistent with #940's "regression of #282/#285" framing.

Stack trace (symbolicated against the released binary):
```
iceberg.duckdb_extension::vector<BoundIcebergManifestListEntry>::AssertIndexInBounds   (idx=281472627900560, size=1)
iceberg.duckdb_extension::IcebergMultiFileList::GetEqualityDeletesForFile
iceberg.duckdb_extension::IcebergMultiFileReader::ApplyEqualityDeletes
libduckdb.so::MultiFileFunction<ParquetMultiFileInfo>::MultiFileScan
```

## The fix

Two layered guards added at the top of `GetEqualityDeletesForFile`:

1. **Fast-path empty check.** If `equality_delete_data.empty()` the snapshot has no equality delete files at all → return immediately without dereferencing `data_manifests`. Covers every Iceberg v3 table (which use deletion vectors) and every v2 table that doesn't use equality deletes. This alone fixes the crash for the entire class of v3 / no-equality-delete workloads.

2. **Defensive bounds check.** If `manifest_file_idx >= data_manifests.size()` return early with no deletes applied. The invariant should always hold for correctly-bound scans, so this branch is unreachable in the happy path; it's defense-in-depth for the underlying #940 binding bug while a proper fix for that lands separately. Returning "no deletes apply" is the safer failure mode than `InternalException` — incorrect results are detectable by users; a connection-killing fatal is silent.

## What this does NOT fix

The underlying #940 root cause — why `manifest_file_idx` ever becomes invalid in the first place — is presumably elsewhere in the binding/projection-pruning path and remains. This PR turns the symptom into a no-op for the workloads where equality deletes don't apply (the majority case), without changing semantics for snapshots that do contain equality delete files.

## Testing

Verified locally that with this patch applied, a 99-query TPC-DS bench on a Snowflake-managed v3 iceberg dataset completes without the cascading FATAL errors that were reproducible on `b798078f` unpatched.

A unit test covering the empty-equality-deletes fast-path would be a useful follow-up — happy to add one if the maintainers prefer.

### DIFF
--- a/src/planning/iceberg_multi_file_list.cpp
+++ b/src/planning/iceberg_multi_file_list.cpp
@@ -825,6 +825,26 @@ vector<reference<const IcebergEqualityDeleteRow>>
 IcebergMultiFileList::GetEqualityDeletesForFile(const BoundIcebergManifestEntry &bound_manifest_entry) const {
 	vector<reference<const IcebergEqualityDeleteRow>> result;
 
+	//! Fast-path: no equality delete files exist in this snapshot. Skips the
+	//! unconditional `data_manifests[manifest_file_idx]` dereference below,
+	//! which crashes on tables where `manifest_file_idx` is invalid (see
+	//! issue #940 — regression of #282/#285 where the index can be
+	//! invalidated when a filter column is pruned from the projection). Also
+	//! covers by spec all Iceberg v3 tables, which use deletion vectors
+	//! instead of equality deletes (e.g. Snowflake-managed v3 emits no
+	//! equality delete files), so the function is a no-op for them anyway.
+	if (equality_delete_data.empty()) {
+		return result;
+	}
+
+	//! Defensive bounds check on the per-entry manifest_file_idx. When the
+	//! invariant holds, this is unreachable; when it's violated (#940), we
+	//! prefer returning "no deletes apply" over an InternalException that
+	//! invalidates the entire DuckDB connection.
+	if (bound_manifest_entry.manifest_file_idx >= data_manifests.size()) {
+		return result;
+	}
+
 	//! Look through all the equality delete files with a *higher* sequence number
 	auto &manifest_entry = bound_manifest_entry.entry;
 	auto &manifest_file = data_manifests[bound_manifest_entry.manifest_file_idx].entry.file;


praveentandra:
Closing — patch will be maintained in our downstream fork only. Sorry for the noise.

Issue #903, opened 2026-04-09
DuckDB 1.5.x Iceberg Extension: Data Corruption and SIGSEGV on LIST/MAP Columns


Harvey-13:
## Description

When querying Iceberg tables containing `LIST` or `MAP` type columns via `iceberg_scan()`, DuckDB 1.5.x exhibits two classes of bugs:

1. **Data corruption**: LIST element values are replaced with NULL, and MAP key/value contents display as `{invalid}`.
2. **Memory crash**: Queries involving `LIMIT` or broad `WHERE` clauses on sufficiently large tables (~800+ rows) crash with SIGSEGV, `free(): invalid pointer`, or `double free or corruption`.

The same Parquet files read via `read_parquet()` return correct results, confirming the bug is in the Iceberg extension layer, not the Parquet reader.

## Environment

- **DuckDB**: v1.5.x (Variegata) affected
- **Iceberg extension**: loaded (or statically linked)
- **Avro extension**: loaded (or statically linked)
- **Iceberg format**: v2+, unpartitioned

### Prerequisites

- Python 3 with packages: `pyarrow`, `pyiceberg`, `fastavro`
- DuckDB 1.5.x with `iceberg` and `avro` extensions available

### Step 1: Run the reproduction script

Save the script below as `repro.py` and run it:

```bash
python3 repro.py
```

This will:
1. Create a small Iceberg table (5 rows, 1 file) at `./test_table_small/`
2. Create a larger Iceberg table (2000 rows, 2 files) at `./test_table_large/`

### Step 2: Run the test queries

```bash
DUCKDB=/path/to/duckdb

# --- Small table: data corruption without crash ---
cd test_table_small
$DUCKDB -c "
  LOAD '/path/to/avro.duckdb_extension';
  LOAD '/path/to/iceberg.duckdb_extension';
  SELECT * FROM iceberg_scan('metadata/v1.metadata.json');
"

# --- Large table: LIMIT triggers crash ---
cd ../test_table_large
$DUCKDB -c "
  LOAD '/path/to/avro.duckdb_extension';
  LOAD '/path/to/iceberg.duckdb_extension';
  SET threads TO 1;
  SELECT * FROM iceberg_scan('metadata/v1.metadata.json') LIMIT 5;
"

# --- Large table: WHERE triggers crash ---
$DUCKDB -c "
  LOAD '/path/to/avro.duckdb_extension';
  LOAD '/path/to/iceberg.duckdb_extension';
[...]

PR #975, opened 2026-05-06
Fix Iceberg LIST field name mapping

regadas:
Fixes part of #903.

Iceberg name mapping identifies list items as `element`, but DuckDB's nested column remap expects the synthetic LIST child to remain named `list`. When the Parquet/Iceberg local field is resolved through name mapping, this now looks up `element` while preserving the local child name as `list` for nested remapping.

Adds a regression query for the existing column-mapping fixture that projects the mapped LIST column directly.

### DIFF
--- a/src/planning/iceberg_multi_file_reader.cpp
+++ b/src/planning/iceberg_multi_file_reader.cpp
@@ -100,21 +100,24 @@ IcebergMultiFileReader::InitializeGlobalState(ClientContext &context, const Mult
 static void ApplyFieldMapping(MultiFileColumnDefinition &col, const vector<IcebergFieldMapping> &mappings,
                               const case_insensitive_map_t<idx_t> &fields, ClientContext &context,
                               optional_ptr<MultiFileColumnDefinition> parent = nullptr) {
+	auto name = col.name;
+	if (parent && parent->type.id() == LogicalTypeId::LIST) {
+		if (fields.find(name) == fields.end() && fields.find("element") != fields.end()) {
+			name = "element";
+		}
+		col.name = "list";
+	}
+
 	if (!col.identifier.IsNull()) {
 		return;
 	}
 
-	auto name = col.name;
 	if (parent && parent->type.id() == LogicalTypeId::MAP && StringUtil::CIEquals(name, "key_value")) {
 		//! Deal with MAP, it has a 'key_value' child, which holds the 'key' + 'value' columns
 		for (auto &child : col.children) {
 			ApplyFieldMapping(child, mappings, fields, context, parent);
 		}
 		return;
-	}
-	if (parent && parent->type.id() == LogicalTypeId::LIST && StringUtil::CIEquals(name, "list")) {
-		//! Deal with LIST, it has a 'element' child, which has the column for the underlying list data
-		name = "element";
 	}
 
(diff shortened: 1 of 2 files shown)


test/sql/local/column_mapping.test
```diff
@@ -26,3 +26,8 @@ SELECT * FROM iceberg_scan('__WORKING_DIRECTORY__/data/persistent/column_mapping
 1	Alice	25	{height=5.5, weight=130}	[85, 90]	{'email': alice@example.com, 'verified': true}
 2	Bob	30	{height=6.0, weight=180}	[78, 82, 88]	{'email': bob@example.com, 'verified': false}
 3	Charlie	35	{height=5.8, weight=160}	[92]	{'email': charlie@example.com, 'verified': true}
+
+query II
+SELECT id, scores FROM iceberg_scan('__WORKING_DIRECTORY__/data/persistent/column_mapping/warehouse/default.db/my_table') WHERE id = 1;
```
Tishj, line 31: This test passes without these changes, please make a test that shows why these changes are necessary

src/planning/iceberg_multi_file_reader.cpp
```diff
@@ -100,22 +100,25 @@ IcebergMultiFileReader::InitializeGlobalState(ClientContext &context, const Mult
 static void ApplyFieldMapping(MultiFileColumnDefinition &col, const vector<IcebergFieldMapping> &mappings,
                               const case_insensitive_map_t<idx_t> &fields, ClientContext &context,
                               optional_ptr<MultiFileColumnDefinition> parent = nullptr) {
+	auto name = col.name;
+	if (parent && parent->type.id() == LogicalTypeId::LIST) {
+		if (fields.find(name) == fields.end() && fields.find("element") != fields.end()) {
+			name = "element";
+		}
+		col.name = "list";
```
Tishj, line 108: Altering the `name` of the `MultiFileColumnDefinition` makes me uneasy, could you explain the reasoning for this?

Tishj:
I've made https://github.com/duckdb/duckdb/pull/23354 in core to prevent this from passing through unnoticed

regadas:
Sorry for dropping the ball here. After you handled the MAP case in core (duckdb/duckdb#23354), I went looking at the LIST side in core too — it's the same shape: remap keys a LIST child as `"list"`, but the reader names it after the Parquet repeated group (`array`/`bag`), so a nested LIST remap throws `Source value array not found`. Opened duckdb/duckdb#23437 for it. That fixes it for any reader, so leaving this closed in favor of the core PR.

test/sql/local/column_mapping.test
regadas: You were right — that fixture's LIST group is already named `list`, so the old code handled it. The real issue turned out to be core remap, not the extension; it's covered by a test in duckdb/duckdb#23437.

src/planning/iceberg_multi_file_reader.cpp
regadas: Agreed it was a smell. The actual fix belongs in core — remap now tolerates a non-`list` LIST child name (duckdb/duckdb#23437), so no extension-side `name` mutation is needed.

PR #976, opened 2026-05-07
Bump to duckdb latest (and update aws/avro extensions)

Tmonster:
Bump duckdb to latest, but also shorten extension configs so they can be listed as extension entries.

Fixes https://github.com/duckdblabs/duckdb-internal/issues/9227


Tmonster:
only deploy is left, gonna merge

PR #982, opened 2026-05-08
Fix stale metadata cache after table writes with MAX_TABLE_STALENESS

NiclasHaderer:
When `MAX_TABLE_STALENESS` is set, `DoTableUpdates()` commits changes to the REST catalog but never refreshes the metadata cache. Subsequent reads in new transactions hit the stale cached entry and return outdated data — in the worst case, missing all rows just written.

After a successful commit, re-fetch each committed table's metadata via `GetTable()` and store it in the cache. If the re-fetch fails, fall back to expiring the stale entry so the next read is forced to make a fresh request.

This affects INSERT, UPDATE, DELETE, MERGE INTO, ALTER TABLE, and multi-table commits — all operations that commit through `DoTableUpdates()`.


src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -474,6 +474,22 @@ void IcebergTransaction::DoTableUpdates(IcebergTransactionAlterUpdate &alter_upd
...
+	if (ic_catalog.attach_options.max_table_staleness_micros.IsValid()) {
+		for (auto &it : alter_update.committed_tables) {
+			auto table_it = alter_update.updated_tables.find(it);
+			if (table_it == alter_update.updated_tables.end()) {
+				continue;
+			}
+			auto &table_info = table_it->second;
+			auto get_table_result = IRCAPI::GetTable(context, ic_catalog, table_info.schema, table_info.name);
+			if (!get_table_result.has_error) {
+				ic_catalog.table_request_cache.SetOrOverwrite(context, it, std::move(get_table_result.result_));
```
Tishj: I don't see why we would actively make a request to populate the cache, we can just Expire and let the next query populate the cache if necessary, no?

NiclasHaderer: That would work as well. This just fits the MotherDuck model a bit better, which is why I implemented it like this.

Tishj: I'd prefer not making an additional request, as the request could be entirely pointless if the table isn't queried anymore after the table has been inserted into

Tmonster: I like not making an extra request as well, should also make the change even smaller

test/sql/local/catalog_custom_setup/fixture/test_stale_cache_after_write.test
```diff
@@ -0,0 +1,177 @@
...
+statement ok
+INSERT INTO my_datalake.default.cache_test VALUES (3, 'target');
+
+statement ok
+CREATE TABLE my_datalake.default.merge_source (a INTEGER, b VARCHAR);
+
+statement ok
+INSERT INTO my_datalake.default.merge_source VALUES (3, 'merged'), (4, 'new');
+
+statement ok
```
Tmonster, line 87: can we also add a test somewhere in here where we verify that another request has been made to refresh the cache for the table?

Tmonster (commented):
Thanks! 
Two small comments

src/catalog/rest/transaction/iceberg_transaction.cpp
NiclasHaderer: Removed the `SetOrOverwrite` call

Tmonster (approved):
Thanks!

PR #987, opened 2026-05-12
need even smaller names for those config settings

Tmonster:


Tishj:
git blame purposes: https://github.com/duckdb/duckdb-iceberg/pull/948

PR #992, opened 2026-05-14
Option that allows writing unsorted data to a sorted table

Olical:
See #851 for more information.

This PR adds a new option `unsafe_iceberg_ignore_sort_order` which allows duckdb to write _unsorted_ data to a _sorted_ table. This data will then need to be sorted at a later time by a compactor of some kind which happens automatically (with some configuration) in systems like S3 Tables.

This allows us to write many small unsorted files into an Iceberg table but have the resulting queries later read from a sorted and compacted large file that we never have to load into memory at any point.

I've added some simple tests and run the entire suite and everything looks okay on my end, unsure if there's CI too.

### DIFF
--- a/test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_insert_into_sorted_table_fails.test
+++ b/test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_insert_into_sorted_table_fails.test
@@ -31,7 +31,10 @@ insert into my_datalake.default.table_sort_order select * from lineitem limit 5;
 ----
 <REGEX>:.*Not implemented Error.*INSERT into a sorted iceberg table is not supported yet.*
 
-# With the unsafe override enabled, the INSERT succeeds (writes are NOT sorted).
+# With the unsafe override enabled, the INSERT succeeds. The written data is
+# not sorted by the table's declared sort order, which the Iceberg spec permits
+# (writers are not required to honour the declared order, and readers do not
+# assume files are sorted).
 # Wrap in a transaction so the row-count assertions below stay stable.
 statement ok
 BEGIN;
--- a/test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_insert_into_sorted_table_fails.test
+++ b/test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_insert_into_sorted_table_fails.test
@@ -31,6 +31,29 @@ insert into my_datalake.default.table_sort_order select * from lineitem limit 5;
 ----
 <REGEX>:.*Not implemented Error.*INSERT into a sorted iceberg table is not supported yet.*
 
+# With the unsafe override enabled, the INSERT succeeds (writes are NOT sorted).
+# Wrap in a transaction so the row-count assertions below stay stable.
+statement ok
+BEGIN;
+
+statement ok
+SET unsafe_iceberg_ignore_sort_order=true;
+
+statement ok
+insert into my_datalake.default.table_sort_order select * from lineitem limit 5;
+
+statement ok
+ABORT;
+
+statement ok
+SET unsafe_iceberg_ignore_sort_order=false;
+
+# Without the override the guard still applies
+statement error
+insert into my_datalake.default.table_sort_order select * from lineitem limit 5;
+----
+<REGEX>:.*Not implemented Error.*INSERT into a sorted iceberg table is not supported yet.*
+
 statement ok
 BEGIN;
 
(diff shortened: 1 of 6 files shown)


Tishj:
I'm confused: "writers are not required to write sorted data"

Then what is `sort_order` for?
If that's the case then we don't need this flag at all, and implementing the sort merely becomes an optimization

test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_insert_into_sorted_table_fails.test
```diff
@@ -31,6 +31,32 @@ insert into my_datalake.default.table_sort_order select * from lineitem limit 5;
...
+# assume files are sorted).
+# Wrap in a transaction so the row-count assertions below stay stable.
+statement ok
+BEGIN;
+
+statement ok
+SET unsafe_iceberg_ignore_sort_order=true;
+
+statement ok
+insert into my_datalake.default.table_sort_order select * from lineitem limit 5;
```
Tmonster, line 46: Can we have a read of the staged data as well?

Along with a test case where the data is committed and read back again?

Tmonster (commented):
Thanks! Looks good to me except for some more test cases

Olical:
@Tishj I was mulling over this same thought over the weekend. There's a few references in the docs about "some engines".

> Iceberg tables can be configured with a sort order that is used to automatically sort data that is written to the table in some engines. For example, MERGE INTO in Spark will use the table ordering.

https://iceberg.apache.org/docs/latest/spark-ddl/#alter-table-drop-partition-field

I think we're allowed to just write to unsorted tables if you feel like it but there should be some documentation explaining that writes aren't sorted yet. In my specific case, the S3 Tables compactor comes through later and folds the appended data into a larger file and sorts it at that point.

Would you like me to:

 - Add the tests as @Tmonster mentioned and move to get this PR merged first then consider this a future change?
 - Drop this PR and create a new PR that just drops the guards entirely?

I'm leaning towards the latter, if it's entirely optional and ignoring the sort order is just a detail the user needs to be aware of then adding this option then removing it will just be noise.

I'm happy to proceed in either direction you all prefer.

Olical:
Also worth noting, the behaviour we're talking about with no guards and just allowing unsorted writes is exactly what PyIceberg does right now. Might be a vote of confidence towards just allowing unsorted writes.

test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_insert_into_sorted_table_fails.test
Olical: I have added round trip reading tests too, but we may still want to just throw this PR out in favour of removing the guard rather than adding an option so we'll align with how PyIceberg works, but that's up to you as maintainers really as to which direction you want to go in.

If you'd like to close this and have a new PR stripping the guard so we can always write unsorted data then I'll happily do that.

Tmonster:
> Also worth noting, the behaviour we're talking about with no guards and just allowing unsorted writes is exactly what PyIceberg does right now. Might be a vote of confidence towards just allowing unsorted writes.

This is interesting to learn, and also frustrating to know that other engines do not try to respect the setting at all.

Olical:
It's just quite hard (or impossible!) to respect it in some situations I think which is why the spec doesn't enforce it and seems to encourage "just sort it later with a compactor or something" so that your writers don't need to be able to hold everything in memory at once and can instead focus on smaller writes.

Sorting is more of a hint for optimising read time and can be mixed where some big files are sorted, some aren't and some _eventually_ get sorted. I think it's nice that it's this flexible all things considered. In my case sorting my data would gain me basically nothing other than time wasted sorting a small file that's going to be sorted upon compaction anyway.

I would propose that we write unsorted like PyIceberg and the future support of sorting will need to be configuration, syntax or heuristic based so the user can decide if they want to bother with the effort of sorting or not for whatever reason they may have in their specific setup. It needs to remain flexible I think.

Olical:
Investigating the failed action now, the tests ran locally for me interestingly.

Olical:
I've added a skip for the failure like we're doing for the other tests, I think that's the "right" move for now.

Olical:
@Tmonster Should I forward port this to the main branch so it can fit into nightlies? Wondering if this branch will be left behind eventually?

Tmonster:
@Olical I think this should already be in the nightlies. I think if you run `Force install iceberg from core_nightly` you should be able to use the feature. Unless you really want it for `main`?

Olical:
My bad! Needed to bump my duckdb version and force, thanks! This is fine by me, if you want this commit on `main` for the future if that's where you cut future release branches from just let me know and I'll get it in. Only if it helps repo admin in the future is all.

Tmonster:
We will merge v1.5-variegata into main, so this will automatically be in the next release

PR #994, opened 2026-05-17
fix: flatten output_chunk before ManifestReader / ManifestListReader

mustafahasankhan:
## Summary

Reading an Iceberg table whose `current_snapshot.manifest_list` references many manifests can abort with `INTERNAL Error: Operation requires a flat vector but a non-flat vector was encountered` thrown out of `IcebergAvroMultiFileReader::FinalizeChunk`.

`ManifestReader::ReadChunk` reaches into the `data_file` STRUCT via `StructVector::GetEntries` before wrapping the children in `UnifiedVectorFormat`, so non-`FLAT_VECTOR` input misaligns the loop; the `MANIFEST_LIST` branch has the same shape. The `output_chunk.Flatten()` guards that previously held this invariant were removed alongside the move to `UnifiedVectorFormat`, but the struct access on the `data_file` path still requires a flat parent.

This restores the two `output_chunk.Flatten()` calls in `FinalizeChunk`, one per `AvroScanInfoType` branch, so the readers always see a flat chunk.

## Test plan

- [x] New sqllogictest under `catalog_test_config_setup/catalog_agnostic/reads/` walks the manifest_list past the threshold via CTAS + 20 single-row INSERTs and asserts DESCRIBE, a pruning SELECT, and COUNT(*) all succeed
- [x] Local debug build passes the new test and an existing fixture-catalog test (`iceberg_load_table_response.test`)

### DIFF
--- a/src/planning/metadata_io/manifest/iceberg_manifest_reader.cpp
+++ b/src/planning/metadata_io/manifest/iceberg_manifest_reader.cpp
@@ -165,6 +165,11 @@ void ManifestReader::ReadChunk(DataChunk &chunk, const map<idx_t, LogicalType> &
 	file_sequence_number.ToUnifiedFormat(count, file_sequence_number_format);
 
 	auto &data_file = chunk.data[vector_index++];
+	//! StructVector::GetEntries() returns the inner STRUCT's children directly when
+	//! 'data_file' is dictionary- or constant-encoded, so the children's row count
+	//! does not match 'count' and the per-child ToUnifiedFormat() builds a stale sel.
+	//! Flatten just this struct (cheap) so the children naturally span 0..count-1.
+	data_file.Flatten(count);
 	idx_t entry_index = 0;
 	auto &data_file_entries = StructVector::GetEntries(data_file);
 
--- a/test/sql/local/catalog_test_config_setup/catalog_agnostic/reads/test_manifest_list_threshold.test
+++ b/test/sql/local/catalog_test_config_setup/catalog_agnostic/reads/test_manifest_list_threshold.test
@@ -1,0 +1,53 @@
+# name: test/sql/local/catalog_test_config_setup/catalog_agnostic/reads/test_manifest_list_threshold.test
+# description: Regression for non-flat vector in ManifestListReader when manifest_list has many entries
+# group: [reads]
+
+require-env CATALOG_TEST_CONFIG_SETUP
+
+require avro
+
+require parquet
+
+require iceberg
+
+require httpfs
+
+# Do not ignore 'HTTP' error messages!
+set ignore_error_messages
+
+statement ok
+drop table if exists my_datalake.default.manifest_list_threshold;
+
+# CTAS first so the table has an initial snapshot. Subsequent single-row
+# INSERTs each add one manifest to the manifest_list.
+statement ok
+create table my_datalake.default.manifest_list_threshold as
+    select 0::INTEGER as id, 'seed' as v;
+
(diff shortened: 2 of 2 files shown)


Tishj:
Flattening is not the right solution, we use `UnifiedVectorFormat` precisely so we don't have to flatten.
I'd much rather see a PR addressing the problems in our accessing/usage of the `UnifiedVectorFormat`, instead of negating the optimization altogether, thanks 👍 

Something tells me the version you noticed these problems on was before [this PR](https://github.com/duckdb/duckdb-iceberg/pull/883) was made ?

mustafahasankhan:
Thanks for the review, you're right that a full `output_chunk.Flatten()` defeats the point of the refactor. Tested on 1.5.2 which is post #883, so the new `UnifiedVectorFormat` code path is what's affected.

The narrower gap is at `ManifestReader::ReadChunk` (post-#883):

```cpp
auto &data_file = chunk.data[vector_index++];
idx_t entry_index = 0;
auto &data_file_entries = StructVector::GetEntries(data_file);
```

When `data_file` is dictionary- or constant-encoded, `StructVector::GetEntries` transparently recurses into `DictionaryVector::Child` and returns the inner buffer's children. Those children have inner-buffer size, not `count`. The subsequent per-child `child.ToUnifiedFormat(count, ...)` hits the default branch, calls `Flatten(count)` (a no-op on a flat child smaller than `count`), and sets `sel` to the incremental 0..count-1 vector — indexing past the inner buffer.

Updated the PR to flatten only `data_file` (the one column accessed via `StructVector::GetEntries`); every other column already goes through `ToUnifiedFormat` / `RecursiveToUnifiedFormat` and is left as-is. The `MANIFEST_LIST` branch in `FinalizeChunk` is back to its current form since `ManifestListReader::ReadChunk` doesn't have the same pattern.

Happy to take an alternative shape — e.g. plumbing the outer struct's `sel` through `ReadRequiredField` / `ReadOptionalField` so nothing is flattened — if you'd prefer that over the targeted `Flatten`.

test/sql/local/catalog_test_config_setup/catalog_agnostic/reads/test_manifest_list_threshold.test
```diff
@@ -0,0 +1,53 @@
+# name: test/sql/local/catalog_test_config_setup/catalog_agnostic/reads/test_manifest_list_threshold.test
```
Tishj, line 1: This test passes on current `v1.5-variegata` branch

src/planning/metadata_io/manifest/iceberg_manifest_reader.cpp
```diff
@@ -165,6 +165,11 @@ void ManifestReader::ReadChunk(DataChunk &chunk, const map<idx_t, LogicalType> &
 	file_sequence_number.ToUnifiedFormat(count, file_sequence_number_format);
 
 	auto &data_file = chunk.data[vector_index++];
+	//! StructVector::GetEntries() returns the inner STRUCT's children directly when
+	//! 'data_file' is dictionary- or constant-encoded, so the children's row count
+	//! does not match 'count' and the per-child ToUnifiedFormat() builds a stale sel.
+	//! Flatten just this struct (cheap) so the children naturally span 0..count-1.
```
Tishj, line 171: > cheap

`data_file` is probably 90% of the entire manifest entry, so that's just not true 👍

Tishj:
Please make an issue highlighting the actual problem, the test you're adding is already passing on `v1.5-variegata` so is adding nothing. Your proposed solution is not desired.

PR #997, opened 2026-05-20
fix: pass correct operation type to IcebergAddSnapshot

p1p1bear:
Previously, IcebergAddSnapshot always hardcoded the snapshot operation as OVERWRITE, regardless of the actual operation (INSERT → APPEND, DELETE → DELETE). This fix passes the operation type through from IcebergTransactionData::AddSnapshot to the snapshot, ensuring the REST API receives the correct operation in the snapshot summary.

Also exposes the operation field in the iceberg_snapshots() table function and parses it from snapshot metadata, so users can query snapshot operation types directly.

### DIFF
--- a/src/core/metadata/snapshot/iceberg_snapshot.cpp
+++ b/src/core/metadata/snapshot/iceberg_snapshot.cpp
@@ -185,6 +185,19 @@ IcebergSnapshot IcebergSnapshot::ParseSnapshot(const rest_api_objects::Snapshot 
 	ret.manifest_list = snapshot.manifest_list;
 	ret.metrics = MetricsFromSummary(snapshot.summary.additional_properties);
 
+	auto &op = snapshot.summary.operation;
+	if (op == "append") {
+		ret.operation = IcebergSnapshotOperationType::APPEND;
+	} else if (op == "replace") {
+		ret.operation = IcebergSnapshotOperationType::REPLACE;
+	} else if (op == "overwrite") {
+		ret.operation = IcebergSnapshotOperationType::OVERWRITE;
+	} else if (op == "delete") {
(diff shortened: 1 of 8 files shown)


Tmonster (commented):
Thanks! Can you also add some tests that use `UPDATE`?

src/core/metadata/snapshot/iceberg_snapshot.cpp
```diff
@@ -185,6 +185,17 @@ IcebergSnapshot IcebergSnapshot::ParseSnapshot(const rest_api_objects::Snapshot
...
 
+	auto &op = snapshot.summary.operation;
+	if (op == "append") {
+		ret.operation = IcebergSnapshotOperationType::APPEND;
+	} else if (op == "replace") {
+		ret.operation = IcebergSnapshotOperationType::REPLACE;
+	} else if (op == "delete") {
+		ret.operation = IcebergSnapshotOperationType::DELETE;
+	} else {
+		ret.operation = IcebergSnapshotOperationType::OVERWRITE;
```
Tishj: I think there should be an explicit block for `OVERWRITE` and the fallback should throw, not just default to `OVERWRITE`

p1p1bear:
> Thanks! Can you also add some tests that use `UPDATE`?

Yes, added. :)

src/core/metadata/snapshot/iceberg_snapshot.cpp
p1p1bear: Followed your suggestion and made the change — there is now an explicit block for `OVERWRITE`, and the fallback throws instead of defaulting to `OVERWRITE`.

p1p1bear:
Fix test/configs/nessie.json

Tmonster (approved):
Thanks!

PR #1001, opened 2026-05-22
Fix #969: resolve point-in-time reads via snapshot-log, not all snapshots

### LINKED ISSUE #969: ATTACH (TYPE ICEBERG, REST): SELECT FROM tbl picks latest-by-timestamp across all branch refs instead of current-snapshot-id when side branches exist
### What happens?

  ## Summary                                                                                                                                                                                                                               
                                                                            
  When an Iceberg table is reached via a REST-attached catalog, `SELECT … FROM <ns>.<tbl>` reads from the snapshot with the largest `timestamp-ms` across **every** branch ref on the table — not from the snapshot pointed at by          
  `current-snapshot-id` (which by spec equals `refs.main.snapshot-id`).
                                                                                                                                                                                                                                           
  Path mode (`iceberg_scan('…/metadata.json')`) on the **same** metadata correctly follows `current-snapshot-id` and returns the right data. So the divergence is in the REST-attached resolver, not in the manifest/data scan layer.      
   
  In our deployment this surfaces because a daily writer creates per-typedef side branches via `truncate` (`op=DELETE, total-records=0`) and never appends to most of them. After every such truncate commit the side branch's HEAD has a  
  newer `timestamp-ms` than `refs.main`. ATTACH-mode reads then return 0 rows; PyIceberg / Daft / `iceberg_scan` path mode all return the real row count from main.
                                                                                                                                                                                                                                           
  ## Versions                                                               

  |  |  |
  |---|---|
  | duckdb | `1.5.2` (pypi wheel) |
  | iceberg extension | auto-installed for 1.5.2 from `extensions.duckdb.org` |                                                                                                                                                            
  | pyiceberg | `0.11.1` |                                                                                                                                                                                                                 
  | catalog | Apache Polaris (any Iceberg REST catalog should reproduce) |                                                                                                                                                                 
  | storage | S3 (any object store should reproduce) |  

### To Reproduce

PyIceberg sets up the table state, DuckDB observes the bug. Polaris is needed because the bug is REST-attached-specific; substitute any Iceberg REST catalog.                                                                            
   
    
  ```
import pyarrow as pa                                                      
  import duckdb
  from pyiceberg.catalog import load_catalog
  from pyiceberg.expressions import AlwaysTrue                                                                                                                                                                                             
  from pyiceberg.types import IntegerType, NestedField, StringType                                                                                                                                                                         
                                                                            
  CATALOG_URL = "http://localhost:8181/api/catalog"                                                                                                                                                                                        
  TOKEN_URL   = "http://localhost:8181/api/catalog/v1/oauth/tokens"                                                                                                                                                                        
  WAREHOUSE   = "context_store"                                                                                                                                                                                                            
  NAMESPACE   = "ae_test"                                                                                                                                                                                                                  
  TABLE       = "duckdb_branch_repro"                                                                                                                                                                                                      
                                                                                                                                                                                                                                           
  cat = load_catalog("polaris", **{                                                                                                                                                                                                        
      "type": "rest",                                                                                                                                                                                                                      
      "uri": CATALOG_URL,
      "oauth2-server-uri": TOKEN_URL,                                                                                                                                                                                                      
      "credential": "<id>:<secret>",                                                                                                                                                                                                       
      "warehouse": WAREHOUSE,                                                                                                                                                                                                              
      "scope": "PRINCIPAL_ROLE:ALL",                                                                                                                                                                                                       
      "header.X-Iceberg-Access-Delegation": "vended-credentials",                                                                                                                                                                          
  })                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                           
  # Set up the divergent state:                                                                                                                                                                                                            
  #   main: 1000 rows                                                       
  #   typedef-foo: empty side branch with newer timestamp_ms (DELETE-all)                                                                                                                                                                  
  #   current-snapshot-id == refs.main.snapshot_id  (per Iceberg spec)                                                                                                                                                                     
  cat.create_namespace_if_not_exists(NAMESPACE)                                                                                                                                                                                            
  try: cat.drop_table((NAMESPACE, TABLE))                                                                                                                                                                                                  
  except Exception: pass                                                                                                                                                                                                                   
  schema = Schema(NestedField(1, "id", IntegerType()), NestedField(2, "name", StringType()))                                                                                                                                               
  tbl = cat.create_table((NAMESPACE, TABLE), schema=schema)                                                                                                                                                                                
  tbl.append(pa.table({"id": list(range(1000)), "name": [f"r{i}" for i in range(1000)]}))                                                                                                                                                  
  main_id = tbl.refs()["main"].snapshot_id                                                                                                                                                                                                 
  with tbl.manage_snapshots() as ms:                                                                                                                                                                                                       
      ms.create_branch(snapshot_id=main_id, branch_name="typedef-foo")                                                                                                                                                                     
  tbl.refresh()                                                                                                                                                                                                                            
  tbl.delete(AlwaysTrue(), branch="typedef-foo")                                                                                                                                                                                           
  tbl.refresh()                                                                                                                                                                                                                            
                                                                                                                                                                                                                                           
  # Sanity-check the metadata is spec-compliant:                                                                                                                                                                                           
  assert tbl.metadata.current_snapshot_id == tbl.refs()["main"].snapshot_id                                                                                                                                                                
  side_id = tbl.refs()["typedef-foo"].snapshot_id                                                                                                                                                                                          
  assert tbl.metadata.snapshot_by_id(side_id).timestamp_ms > tbl.metadata.snapshot_by_id(main_id).timestamp_ms                                                                                                                             
                                                                                                                                                                                                                                           
  # PyIceberg + DuckDB iceberg_scan (path mode) — both correct                                                                                                                                                                             
  print("PyIceberg:", tbl.scan().to_arrow().num_rows)                                                                                                                                                                                      
  con = duckdb.connect()                                                                                                                                                                                                                   
  con.execute("INSTALL iceberg; LOAD iceberg; INSTALL httpfs; LOAD httpfs;")                                                                                                                                                               
  # … set up s3 secret with vended creds from tbl.config …                                                                                                                                                                                 
  print("path mode: ", con.sql(                                                                                                                                                                                                            
      f"SELECT count(*) FROM iceberg_scan('{tbl.metadata_location}')"                                                                                                                                                                      
  ).fetchone()[0])                                                                                                                                                                                                                         
                                                                                                                                                                                                                                           
  # DuckDB ATTACH (REST) — wrong                                                                                                                                                                                                           
  con.execute(f"""                                                          
      CREATE SECRET iceberg_secret (TYPE ICEBERG,                                                                                                                                                                                          
          CLIENT_ID '<id>', CLIENT_SECRET '<secret>',                                                                                                                                                                                      
          OAUTH2_SERVER_URI '{TOKEN_URL}',                                                                                                                                                                                                 
          OAUTH2_SCOPE 'PRINCIPAL_ROLE:ALL');                                                                                                                                                                                              
      ATTACH '{WAREHOUSE}' AS "{WAREHOUSE}" (TYPE ICEBERG,                                                                                                                                                                                 
          SECRET iceberg_secret, ENDPOINT '{CATALOG_URL}',                                                                                                                                                                                 
          ACCESS_DELEGATION_MODE 'none');                                                                                                                                                                                                  
      USE "{WAREHOUSE}"."{NAMESPACE}";                                                                                                                                                                                                     
  """)                                                                                                                                                                                                                                     
  print("ATTACH REST:", con.sql(f"SELECT count(*) FROM {TABLE}").fetchone()[0])          
```                                                                                                                                              
                                                                                                                                                                                                                                           
  Expected output                                                           
                                                                                                                                                                                                                                           
  PyIceberg:   1000                                                         
  path mode:   1000
  ATTACH REST: 1000

  (current-snapshot-id points at refs.main; all three should agree.)                                                                                                                                                                       
   
  Actual output                                                                                                                                                                                                                            
                                                                            
  PyIceberg:   1000
  path mode:   1000
  ATTACH REST: 0

  Adding AT (VERSION => <main_id>) to the ATTACH-mode query returns 1000 — i.e. ATTACH mode honors an explicit snapshot id, it just doesn't pick the right default.                                                                        
   
  Observations                                                                                                                                                                                                                             
                                                                            
  - SELECT * FROM iceberg_snapshots('"<wh>".<ns>.<tbl>') from the same ATTACH connection lists every snapshot, including the side-branch one. The side branch's timestamp-ms is the largest.                                               
  - The bug only fires when a non-main ref has a newer timestamp-ms than refs.main. With a single ref or with main always being newest the symptom never appears.
  - IcebergSnapshotLookup enum in src/planning/snapshot/iceberg_snapshot_lookup.cpp has only { LATEST, FROM_TIMESTAMP, FROM_ID } — no "follow ref / branch" mode is exposed, and no branch= parameter exists on iceberg_scan or in the AT  
  clause.                                                                                                                                                                                                                                  
  - Metadata file's current-snapshot-id matches refs.main.snapshot-id throughout (verified via tbl.metadata.current_snapshot_id and by reading the on-disk metadata.json). This is a reader bug, not a writer producing non-spec metadata. 

### OS:

macOS 15 / Linux x86_64 — both reproduce

### DuckDB Version:

1.5.2

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Nishant Munjal

### Affiliation:

Atlan

Yes

- [x] Yes, I have

- [x] Yes, I have

### PR DESCRIPTION
**Pull request by sankarreddy-atlan**
## Summary

  Fixes #969. When a non-main branch ref points to a snapshot with a newer `timestamp_ms` than `refs.main`'s tip (e.g. a side branch with a `DELETE`-all commit
  issued after the last write to main), default reads of an ATTACH'd REST catalog table were resolving to that side-branch snapshot and returning the wrong data —
  `SELECT count(*)` returning 0 instead of the real row count from main.
 
  ## Fix

  Per the Iceberg [spec](https://github.com/apache/iceberg/blob/main/format/spec.md), point-in-time resolution should use `snapshot-log` — the history of `current-snapshot-id`, which by spec equals `refs.main.snapshot_id`. Side-branch commits do not append to `snapshot-log`, so walking it instead of the global
  `snapshots` map gives the correct main-branch history.

### DIFF
--- a/src/core/metadata/iceberg_table_metadata.cpp
+++ b/src/core/metadata/iceberg_table_metadata.cpp
@@ -21,40 +21,39 @@ optional_ptr<const IcebergSnapshot> IcebergTableMetadata::FindSnapshotByIdIntern
 }
 
 optional_ptr<const IcebergSnapshot> IcebergTableMetadata::GetSnapshotByTimestamp(timestamp_t timestamp) const {
-	// Per Iceberg spec, point-in-time resolution must use snapshot-log
-	// (the history of refs.main). Searching the global snapshots map would
-	// incorrectly pick side-branch tips - see duckdb-iceberg#969.
+	// Per Iceberg spec, point-in-time resolution should use snapshot-log (the
+	// history of refs.main). Searching the global snapshots map would incorrectly
+	// pick side-branch tips - see duckdb-iceberg#969.
 	//
-	// Everything below is compared in raw epoch-millis: snapshot_log stores ms,
-	// last_updated_ms is ms-in-timestamp_t.value (per the parser), and we
-	// convert the incoming lookup timestamp to ms once here.
+	// All comparisons below are in raw epoch-millis: snapshot_log stores ms, and
+	// we convert the incoming lookup timestamp to ms once here.
 	auto target_ms = Timestamp::GetEpochMs(timestamp);
-	if (snapshot_log.empty()) {
-		// Table has no commits yet (freshly created, no INSERT). Original behavior
-		// returned nullptr here; GetSnapshot interprets that as "use current schema".
-		if (snapshots.empty()) {
-			return nullptr;
-		}
-		// Table has snapshots but no snapshot-log (spec allows this; rare). For the
-		// common case where the lookup is at or after the last commit,
-		// current_snapshot_id is unambiguously the right answer.
-		if (has_current_snapshot && target_ms >= last_updated_ms.value) {
-			return FindSnapshotByIdInternal(current_snapshot_id);
-		}
-		throw InvalidConfigurationException("Table has no snapshot-log; cannot resolve snapshot as of " +
-		                                    Timestamp::ToString(timestamp));
-	}
-	// snapshot_log is sorted ascending by timestamp_ms; walk newest-first and
-	// return the first entry whose snapshot still exists in the snapshots map
-	// (spec allows expired snapshots to leave dangling log entries).
-	for (auto it = snapshot_log.rbegin(); it != snapshot_log.rend(); ++it) {
-		if (it->second <= target_ms) {
-			if (auto snap = FindSnapshotByIdInternal(it->first)) {
-				return snap;
+	if (!snapshot_log.empty()) {
+		// snapshot_log is sorted ascending by timestamp_ms; walk newest-first and
+		// return the first entry whose snapshot still exists in the snapshots map
+		// (spec allows expired snapshots to leave dangling log entries).
+		for (auto it = snapshot_log.rbegin(); it != snapshot_log.rend(); ++it) {
+			if (it->second <= target_ms) {
+				if (auto snap = FindSnapshotByIdInternal(it->first)) {
+					return snap;
+				}
 			}
 		}
+		return nullptr;
 	}
-	return nullptr;
+	// snapshot-log is optional per spec. Fall back to the pre-#969 behavior:
+	// scan all snapshots for the largest timestamp_ms <= target.
+	uint64_t max_millis = NumericLimits<uint64_t>::Minimum();
+	optional_ptr<const IcebergSnapshot> max_snapshot = nullptr;
+	for (auto &it : snapshots) {
+		auto &snapshot = it.second;
+		auto curr_millis = Timestamp::GetEpochMs(snapshot.timestamp_ms);
+		if (curr_millis <= target_ms && static_cast<uint64_t>(curr_millis) >= max_millis) {
+			max_snapshot = snapshot;
+			max_millis = curr_millis;
+		}
+	}
+	return max_snapshot;
 }
 
 shared_ptr<IcebergTableSchema> IcebergTableMetadata::GetSchemaFromId(int32_t schema_id) const {
--- a/src/core/metadata/iceberg_table_metadata.cpp
+++ b/src/core/metadata/iceberg_table_metadata.cpp
@@ -30,9 +30,14 @@ optional_ptr<const IcebergSnapshot> IcebergTableMetadata::GetSnapshotByTimestamp
 	// convert the incoming lookup timestamp to ms once here.
 	auto target_ms = Timestamp::GetEpochMs(timestamp);
 	if (snapshot_log.empty()) {
-		// snapshot-log is optional per spec. For the common case where the lookup
-		// time is at or after the last commit, current_snapshot_id is unambiguously
-		// the right answer (refs.main always points there).
+		// Table has no commits yet (freshly created, no INSERT). Original behavior
+		// returned nullptr here; GetSnapshot interprets that as "use current schema".
+		if (snapshots.empty()) {
+			return nullptr;
+		}
+		// Table has snapshots but no snapshot-log (spec allows this; rare). For the
+		// common case where the lookup is at or after the last commit,
+		// current_snapshot_id is unambiguously the right answer.
 		if (has_current_snapshot && target_ms >= last_updated_ms.value) {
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by sankarreddy-atlan** [comment]
@Tishj Could you please review the fix

File: src/core/metadata/iceberg_table_metadata.cpp [hunk 1]
```diff
@@ -22,19 +22,40 @@ optional_ptr<const IcebergSnapshot> IcebergTableMetadata::FindSnapshotByIdIntern
...
+		if (snapshots.empty()) {
+			return nullptr;
+		}
+		// Table has snapshots but no snapshot-log (spec allows this; rare). For the
+		// common case where the lookup is at or after the last commit,
+		// current_snapshot_id is unambiguously the right answer.
+		if (has_current_snapshot && target_ms >= last_updated_ms.value) {
+			return FindSnapshotByIdInternal(current_snapshot_id);
+		}
+		throw InvalidConfigurationException("Table has no snapshot-log; cannot resolve snapshot as of " +
```
Tishj: Snapshot-log is optional, I'd like to preserve original behavior if its not available

sankarreddy-atlan: Hi @Tishj Thanks for the review. Added the fallback

Tishj:
Thanks, this looks correct to me
Can you please rebase your changes to `v1.5-variegata` as that branch is far more up to date than `main`

Then I think it's ready to merge 👍

sankarreddy-atlan:
> Thanks, this looks correct to me Can you please rebase your changes to `v1.5-variegata` as that branch is far more up to date than `main`
> 
> Then I think it's ready to merge 👍

@Tishj Thank you, rebased to v1.5.

PR #1007, opened 2026-05-26
If there is no Delete Source, do not throw an error. Treat the delete as empty.

Tmonster:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/8885

### DIFF
--- a/src/execution/operator/iceberg_delete.cpp
+++ b/src/execution/operator/iceberg_delete.cpp
@@ -231,6 +231,9 @@ void IcebergDelete::FlushDeletes(IcebergTransaction &transaction, ClientContext 
                                  IcebergDeleteGlobalState &global_state) const {
 	bool write_deletion_vector = table.table_info.table_metadata.iceberg_version >= 3;
 
+	if (!multi_file_list) {
+		throw InternalException("IcebergDelete multi_file_list is NULL");
+	}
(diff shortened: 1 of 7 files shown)


src/execution/operator/iceberg_delete.cpp
```diff
@@ -28,7 +29,7 @@ class IcebergDeleteGlobalState;
 class IcebergTableEntry;
 
 IcebergDelete::IcebergDelete(PhysicalPlan &physical_plan, IcebergTableEntry &table,
-                             IcebergMultiFileList &multi_file_list, PhysicalOperator &child,
+                             optional_ptr<IcebergMultiFileList> multi_file_list, PhysicalOperator &child,
```
Tishj, line 32: Why is this changed if we never check `!multi_file_list` before accessing it?

Tmonster: we access it in FlushDeletes, but if there are no deletes to flush, we never access it. I will add a check though just to be sure

Tishj: Ah I see, we change the `make_uniq<IcebergDelete>(...)` call, it's not always present anymore
that makes sense then 👍 

I was just wondering why this was now an optional_ptr in the first place, but nvm

PR #1017, opened 2026-05-29
Add REST catalog write compatibility attach options

### LINKED ISSUE #846: Writing to Snowflake Horizon is broken when used with Azure Storage
I have some iceberg table stored in ADLS and managed by Snowflake Horizon (which is basically a managed polaris offering )

pyiceberg can read and delete just fine

DuckDB, i get 

```
---------------------------------------------------------------------------
TransactionException                      Traceback (most recent call last)
[/tmp/ipykernel_16887/1376449699.py](https://localhost:8080/#) in <cell line: 0>()
----> 1 duckdb.sql(f"delete from SUMMARY where time=1845 ")

TransactionException: TransactionContext Error: Failed to commit: 403 This request is not authorized to perform this operation using this permission.
This request is not authorized to perform this operation using this permission.
RequestId:caa04d34-801f-0056-6633-bb630f000000
Time:2026-03-24T02:12:20.6266559Z
Request ID: caa04d34-801f-0056-6633-bb630f000000
```
I shared a notebook here, but you need to have your own crediential which is not very useful I guess,  

duckdb as a writer will be an awesome use case :)

https://colab.research.google.com/drive/1G6cuoLzB3mELNSwoSTi7jhUqBliA_idh?usp=sharing

dataders:
## Summary

fixes: #846

Adds REST catalog attach options for catalogs that do not support every write operation DuckDB currently tries to use.

ideally clients should not have to configure these options -- long-term this should likely like upstream in the IRC itself see [the proposal](https://docs.google.com/document/d/1F1xh6SJhS-opgWRe1pPvWh01j8VHNHRocfCCFttKNf0/edit?usp=sharing) linked in https://github.com/apache/iceberg/issues/10462.

in the short-term, I'm open to alternatives! the benefit of this implementations is that it unlocks the true data federation promise of Iceberg. Writing with just SQL to external catalogs!

docs update PR: https://github.com/duckdb/duckdb-web/pull/6918

## New attach options

These make the write path configurable for catalogs like Snowflake Horizon that reject staged create, transaction commit, create-time metadata updates, or delete operations.

| Option | Default | Effect when changed |
|---|---|---|
| `STAGE_CREATE_TABLES` | `true` | `false`: create tables directly instead of staged create + commit |
| `DISABLE_MULTI_TABLE_COMMIT` | `false` | `true`: per-table commits instead of `POST .../transactions/commit` |
| `SKIP_CREATE_TABLE_METADATA_UPDATES` | `false` | `true`: queue no follow-up metadata updates after a non-staged create (requires `STAGE_CREATE_TABLES false`, errors otherwise) |
| `REMOVE_FILES_ON_DELETE` | `true` | `false`: leave written data files in place when a failed commit is rolled back |

## Behavior changes

| Change | Impact |
|---|---|
| **`TableIsEmpty` semantics** (all catalogs) | A transaction can now write to an empty table created/updated after the transaction started — previously this threw `Table is already outdated`. An empty table has no snapshot to be outdated from. Tested by `write_to_table_created_after_transaction_start.test`. |
| **No empty commits** | Tables with no queued transaction updates are skipped at commit time. |
| **Breaking rename** | `SUPPORT_STAGE_CREATE` → `STAGE_CREATE_TABLES`; the old name now errors loudly (deprecated alias possible if preferred). Also fixes a `supports_stage_create` key typo that silently clobbered the user's value on Glue/S3Tables. |

## Notes for reviewers

@Tmonster @Tishj I'd for you to give feedback on the parameter names as they stand

| Note | Detail | Possible action |
|---|---|---|
| Three options are affirmative but `DISABLE_MULTI_TABLE_COMMIT` is negative  | affirmative: `STAGE_CREATE_TABLES`, `SKIP_CREATE_TABLE_METADATA_UPDATES`, `REMOVE_FILES_ON_DELETE`<br></bs> negated: `DISABLE_MULTI_TABLE_COMMIT`. Current names came out of the first review round. | Align if maintainers prefer — e.g. `MULTI_TABLE_COMMIT` (default `true`) |
| `REMOVE_FILES_ON_DELETE` is a misnomer (as was `ALLOW_DELETES`) | Nothing to do with `DELETE FROM` — it only gates `CleanupFiles()`, i.e. whether DuckDB deletes the data files it already wrote when a failed write transaction is rolled back. Catalogs like S3 Tables reject file deletes and garbage-collect on their own, hence the off switch. | `CLEANUP_FILES_ON_ROLLBACK` would describe the actual behavior; flagging rather than renaming again to avoid churn |
| `allows_deletes` is now user-settable | First time this internal flag is exposed; users can override the AWS Glue/S3Tables default (previously forced off). | Intended as an escape hatch — confirm that's acceptable |

## [follow-up] expanding CI to test against proprietary catalogs

The new tests run against the CI fixture server (Lakekeeper), which validates the shape
of the requests DuckDB issues (no `transactions/commit` POSTs, per-table commits happen,
no follow-up metadata updates for empty created tables). But Lakekeeper doesn't *reject* the incompatible operations the way Horizon does.
 
Enabling true end-to-end testing against Horizon and Unity Catalog is left as a follow-up PR. We (dbt Labs) would be more than happy to help set up CI against Snowflake Horizon, AWS Glue, and Databricks Unity Catalog endpoints — and can
potentially provide test instances for them. Just let me/us know!


Tishj:
Yes I agree that the Catalog should tell us these details, not the user establishing a connection to the Catalog

But in lack of that, I think this is a good alternative

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -316,7 +316,7 @@ TableTransactionInfo IcebergTransaction::GetTransactionRequest(IcebergTransactio
 			continue;
 		}
 		auto &table_info = updated_table.second;
-		if (!table_info.transaction_data) {
+		if (!table_info.transaction_data || !table_info.HasTransactionUpdates()) {
```
Tishj: Looks like this can simply be `!table_info.HasTransactionUpdates()`

src/catalog/rest/iceberg_catalog.cpp
```diff
@@ -548,7 +550,15 @@ unique_ptr<Catalog> IcebergCatalog::Attach(optional_ptr<StorageExtensionInfo> st
 		} else if (lower_name == "support_stage_create") {
 			auto result = entry.second.DefaultCastAs(LogicalType::BOOLEAN).GetValue<bool>();
 			attach_options.supports_stage_create = result;
-			set_by_attach_options.insert("supports_stage_create");
+			set_by_attach_options.insert("support_stage_create");
+		} else if (lower_name == "use_transaction_commit") {
```
Tishj: Can we rename this to `disable_multi_table_commit` and as the name suggests turn this into an opt-out rather than an opt-in ?

test/sql/local/catalog_custom_setup/fixture/attach_options/horizon_compatible_writes.test
```diff
@@ -0,0 +1,122 @@
...
+----
+3	mixed
+
+query I
+select count(*) from my_datalake.default.test_horizon_mixed_empty_create;
+----
+0
+
+query I
+select count(*) from duckdb_logs_parsed('HTTP') where request.type = 'POST'
```
Tmonster, line 117: Why is the result of this 0? 
A request should be fired to create the table even if the `skip_create_table_metadata_updates = false`?

Tmonster (commented):
Thanks! 

It's not clear to me in the code/description what catalogs require adding `SKIP_CREATE_TABLE_METADATA_UPDATES`, could you expand a little?

Tmonster:
@Tishj what do you think about renaming some the other attach options as well?
Maybe something like this?
`SUPPORT_STAGE_CREATE`->`STAGE_CREATE_TABLES`
`ALLOW_DELETES`->`REMOVE_FILES_ON_DELETE` (or something less verbose)

Tishj:
> @Tishj what do you think about renaming some the other attach options as well? Maybe something like this? `SUPPORT_STAGE_CREATE`->`STAGE_CREATE_TABLES` `ALLOW_DELETES`->`REMOVE_FILES_ON_DELETE` (or something less verbose)

Yes I agree with both of those

src/catalog/rest/transaction/iceberg_transaction.cpp
dataders: Good catch — done. `HasTransactionUpdates()` already returns `false` when `transaction_data` is null, so the `!table_info.transaction_data ||` was redundant. Simplified to `if (!table_info.HasTransactionUpdates())`.

src/catalog/rest/iceberg_catalog.cpp
dataders: Done — renamed to `disable_multi_table_commit` and flipped it to opt-out (defaults to `false`, so the multi-table `transactions/commit` path stays enabled unless a catalog explicitly disables it).

test/sql/local/catalog_custom_setup/fixture/attach_options/horizon_compatible_writes.test
dataders: This assertion checks the per-table **commit/update** endpoint (`POST .../tables/<name>`), not table creation. Creation always goes to the collection endpoint `POST .../namespaces/<ns>/tables` (no table name in the path) and does fire here.

The count is `0` because `test_horizon_mixed_empty_create` is created inside the transaction but never receives any rows. With `skip_create_table_metadata_updates = true` and a non-staged create, `CreateNewEntry` returns early without queuing any metadata updates for it, so `HasTransactionUpdates()` is false and it's skipped at commit time — no follow-up `POST .../tables/test_horizon_mixed_empty_create` is issued. The sibling table in the same transaction (`test_horizon_mixed_existing`) *does* get its per-table commit because it received an insert.

Re your `skip_create_table_metadata_updates = false` point: with it `false`, the empty table would get the standard metadata-update commits to `.../tables/<name>`, so the count would be non-zero. The attach in this test sets it `true`, which is the path being exercised. I've added a comment in the test spelling this out.

dataders:
Pushed the agreed renames in a698a43:

- `SUPPORT_STAGE_CREATE` → `STAGE_CREATE_TABLES`
- `ALLOW_DELETES` → `REMOVE_FILES_ON_DELETE`
- `USE_TRANSACTION_COMMIT` → `DISABLE_MULTI_TABLE_COMMIT` (also flipped to opt-out per @Tishj's inline note — defaults to `false`)

Left `SKIP_CREATE_TABLE_METADATA_UPDATES` as-is. Also simplified the `GetTransactionRequest` guard to `!HasTransactionUpdates()` per @Tishj, and added a clarifying comment to the empty-table HTTP assertion @Tmonster asked about (replied inline with the details).

Tmonster:
Thanks! Windows CI is failing due to the submodule bump

PR #1018, opened 2026-05-29
Scope vended credentials to Iceberg data paths

### LINKED ISSUE #792: Iceberg REST vended credentials: metadata_path scope doesn't cover data file paths, causing read failures
## Summary

Reading managed Iceberg tables via an Iceberg REST catalog with `ACCESS_DELEGATION_MODE = 'vended_credentials'` fails with a permissions error on data files, even though the catalog loads successfully and metadata is readable.

This is the correct and expected read path for managed Iceberg tables — the Iceberg REST catalog with an Iceberg reader. The credential vending scope bug is blocking this legitimate use case.

PyIceberg reads the same tables fine using the same catalog endpoint and credentials.

## Behavior

The Iceberg REST `/v1/{prefix}/namespaces/{ns}/tables/{table}` response includes a `config` block with vended credentials. The `metadata_path` in the response scopes the credentials to the metadata directory (e.g. `s3://bucket/path/__metadata/`). Data files live at `s3://bucket/path/data/`, which is outside the credentialed scope.

DuckDB appears to use the vended credentials directly to access data files without widening or re-requesting credentials for the data path. PyIceberg handles this correctly by requesting separate storage credentials for data access.

## Steps to Reproduce

```sql
INSTALL iceberg; LOAD iceberg; INSTALL httpfs; LOAD httpfs;
CREATE SECRET uc_secret (TYPE iceberg, TOKEN '<token>');
ATTACH '<catalog>' AS uc (
    TYPE iceberg,
    SECRET uc_secret,
    ENDPOINT 'https://<workspace>/api/2.1/unity-catalog/iceberg-rest',
    ACCESS_DELEGATION_MODE 'vended_credentials'
);
SELECT * FROM uc.<schema>.<native_iceberg_table> LIMIT 5;
-- IO Error: Access Denied on s3://bucket/path/data/...
```

## Expected Behavior

Data files are accessible. Reading managed Iceberg tables via an Iceberg REST catalog with an Iceberg reader should succeed. The credential scope should cover data paths, not just the metadata path.

## Environment

- DuckDB: 1.5.0
- Extension: duckdb_iceberg (latest)
- Catalog: Databricks Unity Catalog Iceberg REST endpoint
- Storage: AWS S3 (VDM managed storage)

dataders:
## Summary

fixes: #792

Scopes vended storage credentials to the Iceberg table data path when `write.data.path` is set.

Previously the REST catalog path derived the internal storage secret scope from the metadata location. That misses tables whose data files live outside the metadata directory.

This keeps the credential scope aligned with where the table data actually lives.

### DIFF
--- a/src/catalog/rest/catalog_entry/table/iceberg_table_entry.cpp
+++ b/src/catalog/rest/catalog_entry/table/iceberg_table_entry.cpp
@@ -86,13 +86,28 @@ void IcebergTableEntry::PrepareIcebergScanFromEntry(ClientContext &context) cons
 	if (table_credentials.config) {
 		auto &info = *table_credentials.config;
 		D_ASSERT(info.scope.empty());
-		string lc_storage_location = StringUtil::Lower(metadata_path);
-		size_t metadata_pos = lc_storage_location.find("metadata");
-		if (metadata_pos != string::npos) {
-			info.scope = {metadata_path.substr(0, metadata_pos)};
+		string storage_scope;
+		auto data_path = table_info.table_metadata.table_properties.find("write.data.path");
+		if (data_path != table_info.table_metadata.table_properties.end()) {
+			storage_scope = data_path->second;
 		} else {
-			DUCKDB_LOG_INFO(context, "Creating Iceberg Table secret with no scope. Returned metadata location is %s",
-			                lc_storage_location);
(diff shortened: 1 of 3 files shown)


djouallah:
thanks, it made a big difference !!!

PR #1022, opened 2026-06-01
Compute string upper bounds on code point boundaries

Olical:
Writing a string longer than 16 bytes whose leading bytes are multi-byte UTF-8 currently fails the whole insert with `Could not write upper bounds for string column`. It shows up with any longer non-ASCII text, e.g. full-width or accented characters.

The bound is built in `TruncateAndIncrementString` by truncating to 16 bytes and incrementing the last non-continuation byte. That works on raw bytes, so it splits multi-byte characters and increments a byte rather than a code point, which usually lands on invalid UTF-8. It then shrinks and retries until it gives up and throws.

The fix does it on whole code points, the same way duckdb already does for prefix filters in [`FilterCombiner::FindNextLegalUTF8`](https://github.com/duckdb/duckdb/blob/main/src/optimizer/filter_combiner.cpp): truncate on a code point boundary, then increment the last code point to the next scalar value (skipping the surrogate range). If that code point is already the maximum, drop it and carry the increment to the previous one. Only if every code point is the maximum is there no representable bound, in which case it omits it, which is allowed since bounds are optional in the spec ([Manifests / Data File Fields](https://iceberg.apache.org/spec/#manifests), field 128 `upper_bounds` is `optional`). This matches the Java side ([`UnicodeUtil.truncateStringMax`](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/util/UnicodeUtil.java)).

So these strings now get a real bound, e.g. `ＡＢＣＤＥ…` truncates and increments to `ＡＢＣＤＦ`. Lower bounds are untouched (plain truncation, no increment). Added a regression test with full-width and accented strings that checks the insert succeeds, the data round-trips, and the produced upper bound decodes to a value `>=` the column max.

### DIFF
--- a/src/core/expression/iceberg_value.cpp
+++ b/src/core/expression/iceberg_value.cpp
@@ -252,44 +252,51 @@ string IcebergValue::TruncateString(const string &input) {
 	return std::string(bytes.begin(), bytes.end());
 }
 
-string IcebergValue::TruncateAndIncrementString(const string &input) {
-	idx_t original_input_size = input.size();
-	std::vector<unsigned char> bytes(input.begin(), input.end());
-	idx_t truncated_length = std::min<idx_t>(IcebergValue::MAX_STRING_UPPERBOUND_LENGTH, bytes.size());
-	while (truncated_length > 0) {
-		// Truncate to first 16 bytes
-		bytes.resize(truncated_length);
-		// if the original string is less than our length upperbound, return the string as is.
-		if (original_input_size <= IcebergValue::MAX_STRING_UPPERBOUND_LENGTH) {
-			break;
-		}
-		// if original string size > max_upper_bound increment the last non-continuation byte
-		// so that the upper bound will be higher than the string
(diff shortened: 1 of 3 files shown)


Tmonster (commented):
thanks! 

If you want, we can also validate the string length against the `write.metadata.metrics.default` [table property](https://iceberg.apache.org/docs/latest/configuration/#write-properties).

that can also go in another PR if needed.

Seems like something is wrong with the tests though? This might be fixed with a merge from v1.5-variegata?

Olical:
> If you want, we can also validate the string length against the write.metadata.metrics.default [table property](https://iceberg.apache.org/docs/latest/configuration/#write-properties).

Oh interesting! I think this'll be a little more involved so probably deserves another PR.

Rebased the commit but still failing in some ways, investigating now.

Tmonster:
Thanks!

Olical:
Thank you for merging! (and I'll be at duckcon 😄 )

PR #1023, opened 2026-06-02
Pb/iceberg partitioning aware cardinality estimation

Tishj:
This PR supersedes #778 

I've made several changes to the branch, to reduce code duplication and make use of the data that's already present.

I've had to leave part of the fix commented out however, as we need `global_columns` and `global_column_ids` to call `ProcessDeletes`
To reduce future merge conflicts I've left that untouched for now, but likely we have to split that logic into a part that can be done without those, to then follow it up with a method that finalizes the `ProcessDeletes` logic (which looks to be purely equality-delete focused).


Tishj:
```c++
		// if the above code didn't find an equality filter (i.e country_code = "[us]")
		// and there are other table filters (i.e cost > 50), use default selectivity.
		bool has_equality_filter = (cardinality_after_filters != base_table_cardinality);
		if (!has_equality_filter && has_non_optional_filters) {
			cardinality_after_filters = MaxValue<idx_t>(
			    LossyNumericCast<idx_t>(double(base_table_cardinality) * RelationStatisticsHelper::DEFAULT_SELECTIVITY),
			    1U);
		}
```

This is causing the estimated cardinality to be reduced from `500` to `100`, `get_partition_stats` doesn't even get hit 😬

Tishj:
Actually, I think this should be using [this](https://iceberg.apache.org/spec/#partition-statistics) instead of reading all the manifest files/data files

Issue #515, opened 2025-10-03
Column metadata in DuckDB information_schema.columns system view (Polaris catalog)


m-feeser:
I am using DuckDB v1.4.0 and have attached a Fivetran maintained Polaris catalog to query Iceberg tables. I noticed that the `information_schema.columns` system view only contains a single column for each table, though the tables themselves all have more than one column. If I query the table directly using a `SELECT * FROM polaris_catalog.schema.table_name` all of the columns return as do their data types. The DuckDB UI will load the columns when selecting the table through explorer, but on hover over the table name, still shows only 1 column.

Is this a known limitation?

## Code to Reproduce
```sql
INSTALL ICEBERG;
LOAD ICEBERG;
UPDATE EXTENSION;

ATTACH 'madeup_name' AS polaris_catalog (
  TYPE iceberg,
  ENDPOINT '<catalog_URL>',
  CLIENT_ID '<client_id>',
  CLIENT_SECRET '<client_secret>'
);

-- Returns only 1 column per table, and the column_name for every table is '__' with no data_type = 'UNKNOWN'
SELECT *
FROM information_schema.columns
WHERE
  table_catalog = 'polaris_catalog'
  AND table_schema = 'schema_name'
ORDER BY table_schema, table_name, column_name;
```

## Limited Workaround
```sql
PRAGMA table_info('polaris_catalog.schema_name.table_name');
```
This only returns the data types for a single table.

Tmonster:
This is a known limitation yes. In order to get column information for a table, we have to send a request to the LoadTableInformation endpoint. If you have 100 tables, the query to show all table information will require DuckDB to send 100 requests to the catalog (one for each table), which will severely hinder performance. We can look into amortizing this over time, but if tables in the remote catalog start changing, DuckDB information will be out of date.

```
Describe polaris_catalog.schema_name.table_name;
```

Is another workaround.

PR #1026, opened 2026-06-02
Surface Iceberg field `doc` as DuckDB column comment

### LINKED ISSUE #1025: Iceberg schema field `doc` not surfaced as a DuckDB column comment
### What happens

Apache Iceberg schema fields have an optional `doc` attribute — *"a comment describing the field"* per the [Iceberg spec](https://iceberg.apache.org/spec/#schemas). When a table is attached via the iceberg extension (`ATTACH … (TYPE iceberg …)`), that `doc` is **not** surfaced as a DuckDB column comment: `duckdb_columns().comment` is `NULL` for every column, even when the table's `metadata.json` schema fields carry non‑empty `doc` values.

### Why it looks fixable

The extension already **parses and stores** `doc` — it just never applies it:

- `IcebergColumnDefinition::ParseType()` reads `field.doc` and stores it in `IcebergColumnDefinition::doc` (`src/core/metadata/schema/iceberg_column_definition.cpp`; field declared in `src/include/core/metadata/schema/iceberg_column_definition.hpp`).
- `IcebergColumnDefinition::GetColumnDefinition()` (same file) builds the DuckDB `ColumnDefinition` from only name/type/default and **never calls `ColumnDefinition::SetComment()`**, so `doc` is discarded.
- `grep -rn "SetComment" src/` returns nothing on `main`.

DuckDB core already exposes `ColumnDefinition::SetComment(const Value &)`, so the plumbing exists. (PR #949 fixed `doc` JSON parsing for explicit `null`, but did not surface it.)

### Repro

1. Create/obtain an Iceberg table whose schema fields include `doc`, e.g.
   ```json
   { "id": 1, "name": "population", "required": false, "type": "long",
     "doc": "Resident count, 2024 census" }
   ```
2. **pyiceberg** reads it fine: `table.schema().find_field("population").doc` → `"Resident count, 2024 census"`.
3. Attach the same table in DuckDB and query:
   ```sql
   SELECT column_name, comment FROM duckdb_columns() WHERE table_name = '<table>';
   ```
   → `comment` is `NULL` for all columns. Same `metadata.json`; pyiceberg reads the `doc`, DuckDB does not.

### Expected

Map `NestedField.doc` → the DuckDB column comment (i.e. call `ColumnDefinition::SetComment(Value(doc))` in `GetColumnDefinition()` when `doc` is non‑empty), so `duckdb_columns().comment` returns the field documentation.

### Related / note

`#515` (closed, not planned) explains that per‑table column metadata is loaded lazily to avoid N `LoadTable` requests, so the comment would likely only appear once a table is resolved (e.g. via `DESCRIBE` or a scan) rather than in a bare catalog listing — that constraint is fine; this request is just to stop discarding `doc` when the columns *are* materialized.

### Environment

DuckDB v1.5.3; iceberg extension `4008894c`. Confirmed still unaddressed on `main` (2026‑06).

jatorre:
Fixes #1025.

## Problem

Apache Iceberg schema fields have an optional `doc` attribute — *"a comment describing the field"* per the [spec](https://iceberg.apache.org/spec/#schemas). The extension parsed it but then discarded it, so `duckdb_columns().comment` was `NULL` for every column even when the table's schema fields carried non-empty `doc` values. (pyiceberg surfaces the same `doc` fine.)

## Changes

While wiring this up I found the `doc` was being dropped in **three** places, not one:

1. **`IcebergColumnDefinition::ParseType`** received the `doc` parameter but never assigned it to `res->doc` — so it was discarded before it could ever be used. Now stored.
2. **`IcebergColumnDefinition::Copy`** didn't copy `doc`. Now preserved.
3. **`IcebergColumnDefinition::GetColumnDefinition`** never called `SetComment`. It now maps a non-empty `doc` to the DuckDB column comment via `ColumnDefinition::SetComment(Value(doc))`.

That makes the comment present on the **resolved** table entry. But `duckdb_columns()` lists tables through `IcebergTableSet::Scan`, which returns a lazy placeholder entry (the `__` column) to avoid an `N`-table `LoadTable` storm (see #515) — so the comment still wasn't observable. To close that gap without giving up laziness:

4. **`IcebergTableSet::Scan`** now returns the already-resolved entry (which carries the real columns + comments) when a table has been resolved in the session; untouched tables still return the lazy placeholder. This matches the constraint described in #515: the comment appears once a table is resolved (e.g. via `DESCRIBE` or a scan), not in a bare catalog listing.

## Verification

Built locally and tested end-to-end against the `apache/iceberg-rest-fixture` catalog with a table whose schema fields carry `doc`:

```
-- before resolving: lazy placeholder, as today
SELECT column_name, comment FROM duckdb_columns() WHERE table_name='column_doc_comment';
-- column_name = '__', comment = NULL

-- after a DESCRIBE / scan resolves the table:
SELECT column_name, comment FROM duckdb_columns() WHERE table_name='column_doc_comment';
-- id          | Primary identifier
-- population  | Resident count, 2024 census
-- name        | NULL          (field has no doc)
```

`SHOW ALL TABLES`, `duckdb_tables()`, and mixed resolved/unresolved listings all continue to work, and data scans are unaffected.

> Note: no automated test is included here — the existing catalog tests require the Spark/docker generation pipeline and none of the fixtures currently carry a `doc`. Happy to follow up with a generator fixture + `.test` (I have one ready) if you'd like it in this PR.

### DIFF
--- a/scripts/data_generators/tests/column_doc_comment/__init__.py
+++ b/scripts/data_generators/tests/column_doc_comment/__init__.py
@@ -1,0 +1,9 @@
+from scripts.data_generators.tests.base import IcebergTest
+import pathlib
+
+
+@IcebergTest.register()
+class Test(IcebergTest):
+    def __init__(self):
+        path = pathlib.PurePath(__file__)
+        super().__init__(path.parent.name)
--- a/scripts/data_generators/tests/column_doc_comment/q00.sql
+++ b/scripts/data_generators/tests/column_doc_comment/q00.sql
@@ -1,0 +1,11 @@
+CREATE OR REPLACE TABLE default.column_doc_comment(
+  id bigint COMMENT 'Primary identifier',
+  population bigint COMMENT 'Resident count, 2024 census',
+  name string
+)
+TBLPROPERTIES (
+    'format-version'='2'
+);
+insert into default.column_doc_comment values
+(1, 1000, 'alpha'),
+(2, 2000, 'beta');
--- a/src/catalog/rest/iceberg_table_set.cpp
+++ b/src/catalog/rest/iceberg_table_set.cpp
@@ -89,6 +89,17 @@ void IcebergTableSet::Scan(ClientContext &context, const std::function<void(Cata
 		auto table_key = table_info.GetTableKey();
 		iceberg_transaction.tables[table_key] = entry.second;
 
+		if (!table_info.schema_versions.empty()) {
+			// The table has already been resolved (e.g. via DESCRIBE or a scan), so its full schema -
+			// including column comments mapped from the Iceberg field 'doc' - is available. Surface the
+			// resolved entry instead of the placeholder so listings reflect the real columns.
+			auto resolved = table_info.GetLatestSchema(context);
+			if (resolved) {
+				callback(*resolved);
+				continue;
+			}
+		}
+
 		if (table_info.dummy_entry) {
 			// FIXME: why do we need to return the same entry again?
 			auto &optional = table_info.dummy_entry.get()->Cast<CatalogEntry>();
--- a/src/core/metadata/schema/iceberg_column_definition.cpp
+++ b/src/core/metadata/schema/iceberg_column_definition.cpp
@@ -75,6 +75,7 @@ IcebergColumnDefinition::ParseType(const string &name, int32_t field_id, bool re
 	res->id = field_id;
 	res->required = required;
 	res->name = name;
+	res->doc = doc;
 
 	if (type.has_primitive_type) {
 		res->type = ParsePrimitiveType(type.primitive_type);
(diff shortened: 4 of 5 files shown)


Tmonster (commented):
Thanks! Would it be possible to add some tests as well?

jatorre:
Added tests in b06792ad:

- A Spark data generator (`scripts/data_generators/tests/column_doc_comment`) that creates a table whose columns carry `COMMENT` clauses (which Spark/Iceberg stores as the field `doc`).
- A SQL test (`test/sql/local/catalog_test_config_setup/catalog_agnostic/test_column_doc_comment.test`) that attaches the catalog, resolves the table, and asserts `duckdb_columns().comment` returns the documentation (and `NULL` for a field with no `doc`).

Verified locally end-to-end against the `apache/iceberg-rest-fixture` catalog: generated the table via Spark, then ran the `.test` through `unittest --test-config test/configs/fixture.json` — passes (9 assertions).

jatorre:
Thanks a lot, @Tmonster! 🙏

PR #1029, opened 2026-06-02
[DEV] Add forgotten tests

Tmonster:
I had some test files in my staging that I never got around to adding. Let's see if they still pass


Tishj:
You'll need to rebase this to `main`, the pytest conftest has changed in an incompatible way

PR #1030, opened 2026-06-02
[Feature] Write geometry stats and prune based on Geometry Bbox

Tmonster:
This PR implements writing stats for geometry columns to manifest files and also pruning data files based on stats available in manifest files. 

Also adds support to deserialize bounds for geometry columns. 
I still want to test this against snowflake written avro manifest files that write different geometry bounds.


Tmonster:
@jatorre would it be possible to run your iceberg-geo-testbed on this branch and let me know what the results are? 
This should implement file pruning using bbox stats. I am wondering if we have covered enough of the cases or not. 

If not no worries! We may just merge as is and tackle the other geometry types in a future PR

PR #1033, opened 2026-06-04
Pin versions in requirements.txt

NiclasHaderer:
This pins the versions used for testing to

1. Not break the tests if a new major version is published
2. Not become victim of a supply chain attach if the newest version of one of these packages gets compromised


Tmonster:
Thanks!

PR #1034, opened 2026-06-05
Write Iceberg value_counts on INSERT

The-Alchemist:
# Goal
Improve performance of queries through Iceberg without needing to open Parquet files.

# Details
Populate the manifest `value_counts` metric for data files written by `INSERT/CTAS`, so reads have per-column total value counts (including nulls) without opening Parquet footers.

- `iceberg_insert.cpp`: store the Parquet writer's `num_values` (already parsed by ParseColumnStats) into `data_file.value_counts`.
- `iceberg_manifest.cpp`: add the `value_counts` field (field ids 109/119/ 120) to the manifest Avro schema in `WriteToFile` and serialize it in `IcebergDataFile::ToValue`, in the same field position so the struct schema and values stay aligned.

Mirrors the existing null_value_counts write path. Improves null/not-null reasoning on read (FileMatchesFilter) and is a building block for more accurate eager aggregate stats.

# Feedback requested


Tmonster:
@The-Alchemist Is this still WIP?

The-Alchemist:
@Tmonster 
> @The-Alchemist Is this still WIP?

Yes, I still have to fix the CodeQuality CI failure.  I'm looking for feedback, though, to see if there's interest in merging this PR after I fix the failure.  No sense in fixing if it won't be merged ;)

The-Alchemist:
@Tmonster : ready for review.  Build is failing for unrelated reasons (`--break-system-packages` is an unsupported option)

test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_write_upper_lower_bounds_nested_types.test
```diff
@@ -110,4 +110,11 @@ name	VARCHAR	Alice	Alice
 phone_numbers	VARCHAR[]	NULL	NULL
 street	VARCHAR	123 Main St	123 Main St
 value	VARCHAR	gold	30
-zip	VARCHAR	12345	12345
\ No newline at end of file
+zip	VARCHAR	12345	12345
+
+# value_counts are written to the manifest on INSERT (one row -> value_count of 1 for the top-level columns)
+query II
+select column_name, value_count from iceberg_column_stats(my_datalake.default.duckdb_nested_types) where column_name in ('id', 'name') order by column_name;
```
Tmonster, line 120: Thanks for this test. Can we run some more scenarios? 
- only NULLs inserted
- only NaNs inserted
- values_counts on different types of nested data. Structs/Lists/Maps would be nice. 
- Also very long strings

I think for a lot of these, there are columns in `test_all_types()` that could be useful

Can you also run a test where you read from the latest manifest_path?
```
set variable avro_file = (select manifest_path from iceberg_metadata(my_datalake.default.duckdb_nested_types) limit 1);
select value_counts from (select unnest(data_file) from read_avro(getvariable('avro_file'));
```

Should give you the exact fields you want to verify.

Tmonster (commented):
Thanks! 
Just some comments on tests, logic looks fine otherwise

The-Alchemist:
seeing some intermittent failures due to a bug in `extension-ci-tools`, I believe.  upstream PR at https://github.com/duckdb/extension-ci-tools/pull/387

test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_write_upper_lower_bounds_nested_types.test
The-Alchemist: added a bunch of tests to test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_write_value_counts.test.  i think they're right, but they could definitely use a second pair of eyes

Tmonster (approved):
thanks!

PR #1035, opened 2026-06-05
feat: add iceberg_rewrite_data_files table function

p1p1bear:
## Summary

Add `iceberg_rewrite_data_files`, a table function that rewrites selected Iceberg data files through DuckDB's scan and COPY pipeline.

```sql
SELECT * FROM iceberg_rewrite_data_files(
    'my_catalog.my_schema.my_table',
    target_file_size_bytes => '128MiB',
    min_input_files => 5,
    rewrite_all => false
);
-- Returns: (rewritten_data_files BIGINT, added_data_files BIGINT, rewritten_bytes BIGINT)
```

The rewrite reads selected files through the catalog-attached Iceberg scan, so existing merge-on-read deletes are applied before new parquet files are written. Output file sizing is delegated to DuckDB COPY via `file_size_bytes`.

## Implementation notes

- Uses `table_function_bind_operator_t` and a custom logical/physical operator instead of executing generated SQL through a nested connection.
- Removes the process-local `TableLockRegistry`; correctness relies on Iceberg optimistic commit / snapshot validation.
- Removes the `strategy` option for now. The current implementation delegates file sizing to COPY rather than implementing a separate Spark-style binpacking planner.
- Uses DuckDB core `QualifiedName` instead of the custom `MaintenanceTableKey`.
- Uses COPY `RETURN_STATS` for produced file paths, row counts, and file sizes.
- Supports formatted `target_file_size_bytes` values such as `'128MiB'`, with a 100-byte lower bound.
- Supports partitioned rewrites by planning rewrite work per partition bucket.
- Allows one rewrite task to produce multiple parquet files.

## Known limitations / follow-ups

- Manifest-level column stats for rewritten files are not populated yet.
- The rewrite scan is not explicitly pinned to the planning snapshot yet. The implementation validates the table snapshot before execution and commit, so concurrent changes should fail rather than silently commit an inconsistent rewrite.
- Dedicated concurrent rewrite/equality-delete tests are left for follow-up.
- V3 tables and partition spec evolution are not supported yet.

## Test plan

- [x] Bind-time validation
- [x] Formatted `target_file_size_bytes` parsing
- [x] Basic rewrite and no-op second rewrite
- [x] Partitioned table rewrite
- [x] Position deletes remain deleted after rewrite
- [x] Data sequence number pinning for rewritten files
- [x] COPY file-size rotation where one rewrite task produces multiple parquet files

Validated locally and against a local Gravitino Iceberg REST catalog with:

```text
rewrite_data_files_commit.test
rewrite_data_files_partitioned.test
rewrite_data_files_mor_position.test
rewrite_data_files_sequence_pin.test
rewrite_data_files_file_size_rotation.test
rewrite_data_files_validation.test
```


src/core/metadata/manifest/iceberg_manifest.cpp
```diff
@@ -616,14 +616,33 @@ idx_t WriteToFile(const IcebergTableMetadata &table_metadata, const IcebergManif
 		} else {
 			chunk.SetValue(col_idx++, i, Value(LogicalType::BIGINT));
 		}
-		// sequence_number: long
-		// file_sequence_number: long
+		// sequence_number: long      ← schema column 0 (declared at line 466)
+		// file_sequence_number: long ← schema column 1 (declared at line 479)
+		// NOTE: previous writer code swapped these two slots (wrote file_seq into
```
Tmonster: nit: Do we need this note here?

src/catalog/rest/transaction/iceberg_transaction_data.cpp
```diff
@@ -89,8 +89,16 @@ void IcebergTransactionData::AddSnapshot(IcebergSnapshotOperationType operation,
 	case IcebergSnapshotOperationType::APPEND:
 		manifest_content_type = IcebergManifestContentType::DATA;
 		break;
+	case IcebergSnapshotOperationType::REPLACE:
+		//! REPLACE is what iceberg_rewrite_data_files emits: the new files are
```
Tmonster: nit: this comment feels verbose. I think we can also add this next to `case IcebergSnapshotOperationType::APPEND:` and let it fall through to `manifest_content_type = IcebergManifestContentType::DATA;`

src/core/metadata/manifest/iceberg_manifest.cpp
```diff
@@ -616,14 +616,33 @@ idx_t WriteToFile(const IcebergTableMetadata &table_metadata, const IcebergManif
...
+		// the sequence_number column and vice versa). Fixed here for both ADDED
+		// (with optional explicit seq) and EXISTING/DELETED (inherited or explicit
+		// seq via Get*) branches.
 		if (manifest_entry.status == IcebergManifestEntryStatusType::ADDED) {
-			chunk.SetValue(col_idx++, i, Value(LogicalType::BIGINT));
-			chunk.SetValue(col_idx++, i, Value(LogicalType::BIGINT));
+			//! V2 spec allows ADDED entries to carry an explicit data_sequence_number
+			//! (overriding the default NULL+inherit-from-manifest behaviour). The
```
Tmonster: It seems like this code also runs for V3 versions of a table? 

Also searching Iceberg spec §4.2.4 does not return any immediately verifiable results, so I would appreciate not linking that in a comment.

src/function/metadata/iceberg_metadata.cpp
```diff
@@ -201,6 +206,21 @@ static void IcebergMetaDataFunction(ClientContext &context, TableFunctionInput &
 			AddString(output.data[6], out, string_t(data_file.file_format));
 			//! record_count
 			FlatVector::GetData<int64_t>(output.data[7])[out] = data_file.record_count;
+			//! data_sequence_number (effective): the explicit value when the entry
+			//! carries one — rewrite_data_files pins ADDED files to the starting
```
Tmonster: nit: again the comment is very verbose here

src/core/metadata/manifest/iceberg_manifest.cpp
Tishj: Please don't leave comments that are only relevant for this PR. Prefer mentioning these in the description, so we can be aware of the change and the context, through following git-blame to the PR

src/core/metadata/manifest/iceberg_manifest.cpp
```diff
@@ -616,14 +616,33 @@ idx_t WriteToFile(const IcebergTableMetadata &table_metadata, const IcebergManif
 		} else {
 			chunk.SetValue(col_idx++, i, Value(LogicalType::BIGINT));
 		}
-		// sequence_number: long
-		// file_sequence_number: long
+		// sequence_number: long      ← schema column 0 (declared at line 466)
```
Tishj: Also preferably dont use line numbers, because this will become outdated very quickly

src/function/metadata/iceberg_snapshots.cpp
```diff
@@ -6,11 +6,27 @@
...
 #include "core/metadata/iceberg_table_metadata.hpp"
+#include "core/metadata/snapshot/iceberg_snapshot.hpp"
 
 #include <string>
 
 namespace duckdb {
 
+static string SnapshotOperationToString(IcebergSnapshotOperationType type) {
+	switch (type) {
+	case IcebergSnapshotOperationType::APPEND:
```
Tmonster, line 17: nit: I think this is duplicated somewhere else in the PR. Would be nice it we could unify thisi.

src/function/metadata/iceberg_rewrite_data_files.cpp
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	auto exec_result = ExecuteRewrite(context, plan);
+
+	if (!exec_result.new_entries.empty()) {
+		auto &table_info = *plan.table_info;
+
+		vector<string> produced_paths;
+		produced_paths.reserve(exec_result.new_entries.size());
+		for (auto &entry : exec_result.new_entries) {
```
Tmonster: just add new_entries, or also rewritten_candidates?

src/include/maintenance/table_identifier.hpp
```diff
@@ -0,0 +1,58 @@
...
+	}
+	return true;
+}
+
+//! Split "catalog.schema.table" into a `MaintenanceTableKey`. Validates that
+//! all three components are simple identifiers (no quoting, no nested
+//! namespaces). Raises InvalidInputException on malformed input.
+//!
+//! Limitation: quoted identifiers and nested namespaces are not supported yet.
+inline MaintenanceTableKey ParseMaintenanceTableIdentifier(const string &function_name, const string &identifier) {
```
Tishj: Can't we use `QualifiedName` from core here?

src/maintenance/maintenance_table_loader.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include "catalog/rest/catalog_entry/iceberg_schema_entry.hpp"
+#include "catalog/rest/catalog_entry/table/iceberg_table_information.hpp"
+#include "catalog/rest/iceberg_catalog.hpp"
+#include "catalog/rest/iceberg_schema_set.hpp"
+#include "catalog/rest/iceberg_table_set.hpp"
+
+namespace duckdb {
+
+shared_ptr<IcebergTableInformation> LoadIcebergTableShared(ClientContext &context, const MaintenanceTableKey &key,
```
Tishj: I'm not sure I understand the need for `MaintenanceTableKey`, we concatenate the components in the scan path to be able to uniquely differentiate their keys

src/maintenance/rewrite_data_files_executor.cpp
```diff
@@ -0,0 +1,264 @@
...
+	//! wrapAppend sets fileSequenceNumber=null; Trino finishOptimize sets only
+	//! rewriteFiles.dataSequenceNumber(...).
+	entry.data_file.content = IcebergManifestEntryContentType::DATA;
+	entry.data_file.file_format = "parquet";
+	entry.data_file.file_path = produced_files[0].GetValue<string>();
+	entry.data_file.record_count = count_val.GetValue<int64_t>();
+	//! SQL COPY doesn't surface file_size_bytes in this return shape; one HEAD
+	//! request per produced file (cheap relative to the COPY itself) keeps the
+	//! manifest entry honest about on-disk size for downstream readers.
+	auto file_handle = fs.OpenFile(entry.data_file.file_path, FileFlags::FILE_FLAGS_READ);
```
Tmonster: You can add `RETURN_STATS` to the COPY to get the file_size_bytes, and the stats for all the columns
[documentation](https://duckdb.org/docs/lts/sql/statements/copy#copy--to)

src/maintenance/rewrite_data_files_executor.cpp
```diff
@@ -0,0 +1,264 @@
...
+		}
+		file_list += QuoteSqlString(file_paths[i]);
+	}
+	file_list += "]";
+
+	//! Read via the catalog-attached Iceberg table so the scan layer applies
+	//! MoR position/equality deletes during read. `WHERE filename IN (...)`
+	//! uses the iceberg `filename` virtual column to scope the scan to this
+	//! rewrite group's files.
+	string sql = "COPY (SELECT * FROM ";
```
Tishj: I think it's better to construct this directly, rather than stringify and later parse the string

src/maintenance/rewrite_data_files_executor.cpp
```diff
@@ -0,0 +1,264 @@
...
+	auto sql = rewrite_executor_internal::BuildRewriteCopySql(table_key.catalog, table_key.schema, table_key.table,
+	                                                          file_paths, data_path, field_ids);
+
+	//! Fresh Connection on the same DatabaseInstance — see
+	//! `iceberg_to_ducklake.cpp:953` for the precedent. Secret manager state
+	//! is at the DB level so OSS/S3 credentials carry over without explicit
+	//! plumbing. We avoid running on `context` directly because COPY would
+	//! re-enter the table function pipeline.
+	auto &db = DatabaseInstance::GetDatabase(context);
+	Connection conn(db);
```
Tishj: I think instead of constructing a query and executing this with a new connection, we should replace the table function with an operator instead, using the optional table function callback:
```c++
	//! (Optional) Bind operator function
	//! This function is called before the regular bind function - similar to bind_replace - but allows returning a
	//! custom LogicalOperator instead.
	table_function_bind_operator_t bind_operator;
```

test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rewrite_data_files_commit.test
```diff
@@ -0,0 +1,101 @@
...
+5	e
+6	f
+7	g
+8	h
+9	i
+10	j
+11	k
+12	l
+
+# Post-commit the scanner only sees the consolidated file.
```
Tmonster: Can we also read from `iceberg_metadata(<table name>)` and `iceberg_column_stats(<table name>)` here?

src/maintenance/rewrite_data_files_executor.cpp
Tishj: Though this is likely more difficult..

I did some more thinking and I believe we can construct a UNION of the various LogicalCopyToFile operators, each feeding into a custom operator that creates the REPLACE snapshot

I took a crack at this, it's working, but it requires messing with the produced filename to identify the different copy for a group, since they all feed into the same operator

So I'm still looking to figure out a fix that won't require messing with the filename

test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rewrite_data_files_mor_position.test
```diff
@@ -0,0 +1,185 @@
...
+
+# THE CORE CHECK: deleted rows stay deleted across the REPLACE. If the rewrite had
+# dropped the position deletes without materializing them, count would jump back to
+# 12 and ids 4/7/10 would resurrect.
+query I
+select count(*) from my_datalake.default.rewrite_mor_position;
+----
+9
+
+query I
```
Tmonster, line 88: nit: you can enforce that results are the same using test headers like 

```
query III nosort result_1
----
<expected result>
```

Then later a test with 

```
query I nosort result_1
----
```

means the two results get compared

Tmonster (commented):
Thanks! I've started taking a look, and have a couple of comments so far. 

1. Since this is a big PR, I think this would be good to target main instead of v1.5-variegata. Then you might be able to use the new `Identifier` class introduced here https://github.com/duckdb/duckdb/pull/23161 for the Table locking system and in some other places

2. If you add `RETURN_STATS` to the parquet copy, you should get the `file_size_bytes` so you don't have to open the file again

4. The comments are very verbose. They might be nice to have, but I would rather the comments stay concise and short. Extra information about design decisions can be placed in the PR description. Also section 4.2.4 of the iceberg spec doesn't say much to me, I cannot search that, could you instead provide links to the spec?

5. Does it makes sense to have a target_file_size_bytes that is at least 100 ? 
You can use `StringUtil::TryParseFormattedBytes(arg, result)` and `StringUtil::BytesToHumanReadableString(` in case users submit just a regular number or a human readable string. I think 100 is probably a good lower bound. I don't know if our parquet writer can write anything less than that (copy (select 1 a) to '/tmp/blah.parquet' (FORMAT PARQUET, RETURN_STATS);) returns file_size_bytes of 192.

6. Do you have any tests that can test concurrent maintenance? I think it might be possible with larger tables and multiple connections using 

```
query I con1
----
call rewrite;

query I con2
----
call rewrite
```
If they get fired at the same time, we should see the error.


We also have the ability to create equality delete files, see test/sql/local/catalog_test_config_setup/catalog_agnostic/delete/equality_deletes/*

It requires a specific build flag though `EXT_FLAGS=-DICEBERG_ENABLE_EQUALITY_DELETE_WRITES=1`
Then you should be able to test the concurrent equality delete write

Tmonster:
Potentially @Tishj implementation is a bit more clean as well, we will have a chat to what the best way forward is

Tishj:
> 5. Do you have any tests that can test concurrent maintenance? I think it might be possible with larger tables and multiple connections using
> 
> ```
> query I con1
> ----
> call rewrite;
> 
> query I con2
> ----
> call rewrite
> ```
> 
> If they get fired at the same time, we should see the error.

I think this should use `concurrentloop` and `statement maybe` if we are expecting `rewrite_data_files` to spuriously fail in concurrent use

p1p1bear:
@Tmonster @Tishj 

Thanks for the reviews!

I pushed a small cleanup commit for now:
- removed the overly verbose design-note/AI-assistance comments
- kept only short comments for key invariants
- merged APPEND and REPLACE handling where both use DATA manifest

For the larger implementation direction, I’ll wait for your plan first and then we can decide the   next steps.

Thanks!

src/core/metadata/manifest/iceberg_manifest.cpp
p1p1bear: Yes, fixed now:)

src/core/metadata/manifest/iceberg_manifest.cpp
p1p1bear: Yes, fixed now:)

src/catalog/rest/transaction/iceberg_transaction_data.cpp
p1p1bear: Fixed now.:)

Tishj:
Another thought I had (which is present on both the current state and the `bind_operator` version) is that the `iceberg_scan` we use to feed the COPY is unguarded against concurrent writes, we don't lock these iceberg scans to a specific snapshot or table version, so changes could be made that we didn't account for during planning.

For example a new delete could be made targeting one of the data files we are set to rewrite, and we wouldn't be aware.

p1p1bear:
> Another thought I had (which is present on both the current state and the `bind_operator` version) is that the `iceberg_scan` we use to feed the COPY is unguarded against concurrent writes, we don't lock these iceberg scans to a specific snapshot or table version, so changes could be made that we didn't account for during planning.
> 
> For example a new delete could be made targeting one of the data files we are set to rewrite, and we wouldn't be aware.

@Tishj  Good point, thanks for bringing this up.

I think in the current implementation this should not silently commit an inconsistent rewrite. If a concurrent delete advances the table snapshot while the rewrite is running, the final REPLACE commit should fail because of the snapshot assertion / CAS check.

But I agree your point is still meaningful: the scan used for the rewrite is not currently pinned to the snapshot used during planning, so the execution work could observe a state that does not exactly match the planned state before the final commit rejects it.

So I think this is worth improving, even if the commit-time check should protect correctness.

Tishj:
I think on further consideration I don't think we want the `TableLockRegistry`
This would only protect against concurrent use within a duckdb process, but there is nothing stopping an outside writer to perform compaction, so we're not actually guarding against anything.

With this we're only removing our ability to test for this scenario within duckdb.
Every connection to the db should be seen as a separate reader/writer in the context of Iceberg.

p1p1bear:
> I think on further consideration I don't think we want the `TableLockRegistry` This would only protect against concurrent use within a duckdb process, but there is nothing stopping an outside writer to perform compaction, so we're not actually guarding against anything.
> 
> With this we're only removing our ability to test for this scenario within duckdb. Every connection to the db should be seen as a separate reader/writer in the context of Iceberg.

@Tishj  Thanks, I completely agree with your point.

My original reason for adding `TableLockRegistry` was to provide mutual exclusion inside a DuckDB execution node/process. But I agree that this does not really solve the Iceberg-level   concurrency problem, since external writers or other DuckDB processes can still modify the   table concurrently.

Thinking about it more, if a third-party system embeds DuckDB/libduckdb and wants process-local or node-local scheduling guarantees, that kind of lock is probably better   designed outside of this extension rather than inside `duckdb-iceberg`. 

So I’ll remove the `TableLockRegistry` direction and rely on the Iceberg optimistic commit/snapshot assertion for correctness. I’ll first fully incorporate the direction from your implementation, then make the related follow-up changes on top of that. 
Thanks again for the guidance and for putting together the bind_operator implementation!

raghav-reglobe:
@p1p1bear why can't we add support for V3 tables now, do you have any thoughts, I can contribute.

Tishj:
Because that would bloat the PR even further, when it's a perfect candidate for a later follow up

p1p1bear:
@Tmonster @Tishj Thanks again for the reviews.

I pushed a follow-up update that addresses the main points:

- moved execution to `table_function_bind_operator_t` + custom logical/physical operators, thanks @Tishj for the prototype/direction
- removed the nested SQL execution path
- removed `TableLockRegistry`
- removed the `strategy` option for now
- replaced `MaintenanceTableKey` with DuckDB core `QualifiedName`
- use COPY `RETURN_STATS` for produced paths / row counts / file sizes
- pass `target_file_size_bytes` to COPY as `file_size_bytes`
- support formatted target sizes such as `'256MiB'`
- added partitioned rewrite and COPY file-size-rotation tests

I also updated the PR description with the current implementation and remaining follow-ups. The main follow-ups are manifest-level column stats, explicitly snapshot-pinned scans, and dedicated concurrent rewrite/equality-delete tests.

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -55,9 +55,9 @@ static void AddExplicitNullSnapshotIds(yyjson_mut_doc *doc, yyjson_mut_val *root
 		if (!requirement.assert_ref_snapshot_id || requirement.assert_ref_snapshot_id->snapshot_id) {
 			continue;
 		}
-		auto requirement_json = yyjson_mut_arr_get(requirements_array, i);
-		D_ASSERT(requirement_json);
-		yyjson_mut_obj_add_null(doc, requirement_json, "snapshot-id");
+			auto requirement_json = yyjson_mut_arr_get(requirements_array, i);
```
Tishj: This will fail the formatter, and is unrelated noise that should be removed

src/catalog/rest/transaction/iceberg_transaction_data.cpp
```diff
@@ -87,10 +87,11 @@ void IcebergTransactionData::AddSnapshot(IcebergSnapshotOperationType operation,
 		manifest_content_type = IcebergManifestContentType::DELETE;
 		break;
 	case IcebergSnapshotOperationType::APPEND:
+	case IcebergSnapshotOperationType::REPLACE:
```
Tishj, line 165: For my understanding, this isn't necessarily correct, REPLACE isn't reserved just for DATA right? We just dont use it for compacting/rewriting delete files yet?

src/common/iceberg_default.cpp
```diff
@@ -43,19 +43,22 @@ namespace {
 
 //! Used to determine if the field of a struct is mapped or not
 struct StructFieldMapping {
-	case_insensitive_map_t<StructFieldMapping> child_mapping;
+	case_insensitive_map_t<unique_ptr<StructFieldMapping>> child_mapping;
```
Tishj, line 46: Why is this necessary?

src/core/metadata/manifest/iceberg_manifest.cpp
```diff
@@ -620,11 +620,19 @@ idx_t WriteToFile(const IcebergTableMetadata &table_metadata, const IcebergManif
...
+				} else {
+					chunk.data[col_idx++].Append(Value(LogicalType::BIGINT));
+				}
+				if (manifest_entry.HasFileSequenceNumber()) {
+					chunk.data[col_idx++].Append(Value::BIGINT(manifest_entry.GetExplicitFileSequenceNumber()));
+				} else {
+					chunk.data[col_idx++].Append(Value(LogicalType::BIGINT));
+				}
+			} else {
+				chunk.data[col_idx++].Append(Value::BIGINT(manifest_entry.GetSequenceNumber(manifest_file)));
```
Tishj: For my understanding, the reason we only use `GetSequenceNumber` when its not ADDED is because we should only materialize this information when the manifest file's sequence number doesn't alone with that of the manifest_entry?

Maybe a clarifying comment would help here
Bonus if it's a direct quote from the spec (see the other `//! SPEC:` comments)

src/core/metadata/manifest/iceberg_manifest_list.cpp
```diff
@@ -87,9 +87,9 @@ IcebergManifestListEntry IcebergManifestListEntry::CreateFromEntries(FileSystem
 		}
 
 		//! NOTE: this gets overwritten on commit
-		if (!manifest_file.has_min_sequence_number ||
-		    manifest_file.sequence_number < manifest_file.min_sequence_number) {
-			manifest_file.min_sequence_number = manifest_file.sequence_number;
+		auto entry_data_seq = manifest_entry.GetSequenceNumber(manifest_file);
```
Tishj, line 95: Ah, yes this was definitely wrong before, ignoring the potentially materialized sequence_number of the entry
Thanks!

src/function/metadata/iceberg_metadata.cpp
```diff
@@ -200,7 +210,13 @@ static void IcebergMetaDataFunction(ClientContext &context, TableFunctionInput &
 			//! file_format
 			AddString(output.data[6], out, string_t(data_file.file_format));
 			//! record_count
-			FlatVector::GetDataMutable<int64_t>(output.data[7])[out] = data_file.record_count;
+				FlatVector::GetDataMutable<int64_t>(output.data[7])[out] = data_file.record_count;
+				FlatVector::GetDataMutable<int64_t>(output.data[8])[out] = manifest_entry.HasSequenceNumber()
+				                                                        ? manifest_entry.GetExplicitSequenceNumber()
```
Tishj: This I don't get, this logic looks like `GetSequenceNumber` and `GetFileSequenceNumber` no?

src/function/metadata/iceberg_snapshots.cpp
```diff
@@ -114,10 +114,10 @@ static void IcebergSnapshotsFunction(ClientContext &context, TableFunctionInput
...
 		FlatVector::GetDataMutable<timestamp_t>(output.data[2])[i] = snapshot.timestamp_ms;
 		string_t manifest_string_t = StringVector::AddString(output.data[3], string_t(snapshot.manifest_list));
-		FlatVector::GetDataMutable<string_t>(output.data[3])[i] = manifest_string_t;
-		auto operation_str = SnapshotOperationToString(snapshot.operation);
-		FlatVector::GetDataMutable<string_t>(output.data[4])[i] =
-		    StringVector::AddString(output.data[4], operation_str);
+			FlatVector::GetDataMutable<string_t>(output.data[3])[i] = manifest_string_t;
+			auto operation_str = SnapshotOperationToString(snapshot.operation);
+			FlatVector::GetDataMutable<string_t>(output.data[4])[i] =
+			    StringVector::AddString(output.data[4], operation_str);
```
Tishj: More unrelated noise

src/include/core/metadata/manifest/iceberg_manifest.hpp
```diff
@@ -132,6 +132,18 @@ struct IcebergManifestEntry {
...
+	bool HasSequenceNumber() const {
+		return has_sequence_number;
+	}
+	bool HasFileSequenceNumber() const {
+		return has_file_sequence_number;
+	}
+	sequence_number_t GetExplicitSequenceNumber() const {
+		return sequence_number;
+	}
+	sequence_number_t GetExplicitFileSequenceNumber() const {
```
Tishj: I'd prefer these to return an optional actually, and remove the `Has` methods, their names are confusing at best

src/maintenance/rewrite_data_files_executor.cpp
```diff
@@ -0,0 +1,119 @@
...
+namespace duckdb {
+
+namespace {
+
+//! Recursive walk over an Iceberg column definition emitting the FIELD_IDS
+//! value tree. For primitives the value is the field_id as an integer; for
+//! nested types it's a struct with a sentinel `__duckdb_field_id` slot plus
+//! one slot per child. Mirrors the parquet field-id metadata written by the
+//! regular Iceberg insert path so APPEND and REWRITE stay compatible.
+static Value GetFieldIdValue(const IcebergColumnDefinition &column) {
```
Tishj: I feel like we already have this logic somewhere else
Can you try to reuse that instead? (I know it's added by my patch, so I apologize)

src/maintenance/rewrite_data_files_executor.cpp
```diff
@@ -0,0 +1,119 @@
...
+	values.emplace_back("__duckdb_field_id", Value::BIGINT(column.id));
+	for (auto &child : column.GetChildren()) {
+		values.emplace_back(child->name, GetFieldIdValue(*child));
+	}
+	return Value::STRUCT(std::move(values));
+}
+
+} // namespace
+
+Value BuildRewriteFieldIds(const IcebergTableSchema &schema) {
```
Tishj: The above comment also stretches to this, we shouldn't need to define this, it exists elsewhere

src/maintenance/rewrite_data_files_operator.cpp
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	ArenaLinkedList<reference<PhysicalOperator>> physical_children(planner.ArenaRef());
+	for (idx_t group_idx = 0; group_idx < children.size(); group_idx++) {
+		auto &child = children[group_idx];
+		auto &child_plan = planner.CreatePlan(*child);
+		auto child_types = child_plan.GetTypes();
+
+		vector<unique_ptr<Expression>> projection_expressions;
+		projection_expressions.push_back(make_uniq<BoundConstantExpression>(Value::UBIGINT(group_idx)));
```
Tishj, line 96: Can we add a comment explaining that we push the group idx into every COPY so it can be identified in the Sink?

Tishj (changes requested):
Thanks for the changes, left some more comments

As for tests, I think we're missing tests with UPDATE, and tests with DELETE where compaction doesn't remove all deletes from the table.
V2 supports multiple files referenced by a single delete file. I'm not sure if we ever write those, we might always separate them

I'd like to see some tests on tables created and populated with pyspark, as well as duckdb-compacted tables read by pyspark, to verify that we are keeping tables in a readable state

p1p1bear:
@Tishj Thanks for the detailed pass — I pushed a small follow-up cleanup addressing the code-level comments from this round:

- removed the unrelated formatting noise and ran `make format-check`
- clarified that the current helper writes DATA manifest entries, while REPLACE itself is not limited to data files
- changed the explicit sequence-number accessors to return optionals and reused the effective sequence-number helpers in `iceberg_metadata`
- moved the parquet field-id construction into the schema helper so INSERT and rewrite share the same logic
- added the short comment explaining the COPY stats group index

I also validated the update with the extension/unittest build, `make format-check`, the local rewrite/metadata tests, and the key Gravitino-backed rewrite tests.

I’ll look at the remaining test coverage points next, especially UPDATE/delete coverage and the PySpark interoperability cases.

Tishj:
Thanks, you should look at the tests in `scripts/data_generators/tests` to add tests with tables created by spark.
`make fixture`
`python3 -m scripts.data_generators.generate_data spark-rest --test <name_of_folder_in_data_generator_tests>`
`build/debug/test/unittest --test-config test/configs/fixture.json <path_to_duckdb_test>`

And look to the `test/python` directory for tests using spark to read duckdb-generated / altered data

src/maintenance/rewrite_data_files_executor.cpp
```diff
@@ -0,0 +1,89 @@
...
+}
+
+void CleanupRewriteFiles(ClientContext &context, const IcebergTableInformation &table_info,
+                         const vector<string> &produced_paths) {
+	auto &fs = FileSystem::GetFileSystem(context);
+	for (auto &path : produced_paths) {
+		try {
+			fs.TryRemoveFile(path);
+		} catch (...) {
+			//! Best-effort cleanup only. Failures here should not mask the original
```
Tmonster: nit: Could we log the failure with `DUCKDB_LOG_DEBUG()`

p1p1bear:
@Tishj Thanks for the pointers.

I pushed a follow-up test commit that adds the Spark/PySpark interoperability coverage:

- added a new Spark REST fixture under `scripts/data_generators/tests/spark_rewrite_mor_residual_deletes`
- the fixture creates a V2 merge-on-read table, writes multiple data files, then runs UPDATE and DELETE from Spark
- added a DuckDB sqllogictest that rewrites the Spark-created table with `iceberg_rewrite_data_files`, checks that deleted rows do not come back, and verifies residual position deletes remain on the table
- added a PySpark read test for the DuckDB-compacted table to verify Spark can still read it afterwards

I validated it with the new fixture generation, the new sqllogictest, the selected PySpark read test, `make format-check`, and the release build/unittest target.

Tishj (approved):
Thanks, I think it's ready to merge 👍

p1p1bear:
> Thanks, I think it's ready to merge 👍

Thanks @Tishj and @Tmonster for the thorough reviews and guidance throughout this PR! Really appreciate the help getting this into shape. ：)

src/function/metadata/iceberg_rewrite_data_files.cpp
```diff
@@ -0,0 +1,196 @@
...
+	copy_statement.info->is_format_auto_detected = false;
+	copy_statement.info->options["field_ids"].push_back(schema_it->second->GetFieldIds());
+	copy_statement.info->options["filename_pattern"].push_back(Value("{uuidv7}"));
+	copy_statement.info->options["file_size_bytes"].push_back(
+	    Value::UBIGINT(static_cast<uint64_t>(plan.target_file_size_bytes)));
+	//! FILE_SIZE_BYTES forces COPY through DuckDB's rotated-file path creation so
+	//! RETURN_STATS reports concrete parquet paths instead of the directory root.
+	copy_statement.info->options["return_stats"].push_back(Value::BOOLEAN(true));
+	copy_statement.info->options["per_thread_output"].push_back(Value::BOOLEAN(false));
+	copy_statement.info->options["overwrite_or_ignore"].push_back(Value::BOOLEAN(true));
```
Tmonster: Is there a reason this was left to true? I can't imagine a case where we overwrite a file, but if we do, I imagine we want to error

src/function/metadata/iceberg_rewrite_data_files.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+namespace duckdb {
+
+namespace {
+
+constexpr int64_t MIN_TARGET_FILE_SIZE_BYTES = 100;
+
+struct RewriteDataFilesOptions {
+	QualifiedName table_name;
+	int64_t target_file_size_bytes = 134217728;
```
Tmonster: Can we check if there is a `write.parquet.target-file-size-bytes` property of a table before defaulting to this value?

src/function/iceberg_functions.cpp
```diff
@@ -25,6 +25,8 @@ vector<TableFunctionSet> IcebergFunctions::GetTableFunctions(ExtensionLoader &lo
 	functions.push_back(GetIcebergToDuckLakeFunction());
 	functions.push_back(GetIcebergLoadTableResponseFunction());
```
Tmonster: nit: extra line can be removed

src/include/maintenance/maintenance_table_loader.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include "duckdb/main/client_context.hpp"
+#include "duckdb/parser/qualified_name.hpp"
+
+namespace duckdb {
+
+struct IcebergTableInformation;
+
+//! Resolve catalog.schema.table and load the latest Iceberg table metadata.
+//! The function name is only used for actionable error messages.
+IcebergTableInformation &LoadIcebergTable(ClientContext &context, const QualifiedName &table_name,
```
Tmonster: Is this used at all? (AI could not find any uses)

src/include/maintenance/maintenance_table_loader.hpp
```diff
@@ -0,0 +1,22 @@
...
+                                          const string &function_name);
+shared_ptr<IcebergTableInformation> LoadIcebergTableShared(ClientContext &context, const QualifiedName &table_name,
```
Tmonster: Same here. Seems like this function is never used

src/include/maintenance/rewrite_data_files_planner.hpp
```diff
@@ -0,0 +1,57 @@
...
+struct IcebergTableInformation;
+
+struct RewriteCandidate {
+	string file_path;
+	int64_t file_size_in_bytes = 0;
+	int64_t record_count = 0;
+	vector<IcebergPartitionInfo> partition_info;
+	int32_t partition_spec_id = 0;
+	//! Position of the original manifest entry, used when marking rewritten files as deleted.
+	idx_t manifest_idx = 0;
```
Tmonster: AI tells me this field is never read from, only written to. And indeed I can only find instances where it is written to

src/include/maintenance/rewrite_data_files_planner.hpp
```diff
@@ -0,0 +1,57 @@
...
+	idx_t entry_idx = 0;
+};
+
+struct RewritePlan {
+	bool table_is_empty = false;
```
Tmonster: This field also never seems to be read/checked, only written to.

src/include/maintenance/rewrite_data_files_planner.hpp
```diff
@@ -0,0 +1,57 @@
...
+	int64_t starting_sequence_number = 0;
+	int64_t target_file_size_bytes = 134217728;
+	//! Keep the loaded metadata alive until commit.
+	shared_ptr<IcebergTableInformation> table_info;
+	vector<RewriteCandidate> candidates;
+	//! Partition-local rewrite groups.
+	vector<vector<RewriteCandidate>> file_groups;
+};
+
+struct RewriteDataFilesPlanInput {
```
Tmonster, line 36: Would it be cumbersome to combine `RewriteDataFilesPlanInput` and `RewriteDataFilesOptions`? They seem to share a lot of the same values and have the same defaults, so one seems redundant

src/maintenance/rewrite_data_files_operator.cpp
```diff
@@ -0,0 +1,254 @@
...
+		auto produced_file = chunk.GetValue(1, row).GetValue<string>();
+		if (produced_file.empty()) {
+			throw InternalException("iceberg_rewrite_data_files: COPY for group %lld returned an empty file path",
+			                        group_idx);
+		}
+		auto count = NumericCast<int64_t>(chunk.GetValue(2, row).GetValue<uint64_t>());
+		auto file_size_in_bytes = NumericCast<int64_t>(chunk.GetValue(3, row).GetValue<uint64_t>());
+		auto &group = gstate.plan.file_groups[group_idx];
+		{
+			lock_guard<mutex> guard(gstate.lock);
```
Tmonster: Looks like we grab the lock here and again in 180? Is there a performance reason for this? Seems like BuildRewriteManifestEntry just builds the manifest entry in memory, so it's not too expensive. Then you can combine the two locked regions to prevent lock churn

Tmonster (commented):
@p1p1bear Thank you so much for the development! Still some questions from my side though

- Do we have any tests that compact from `test_all_types()`, then we can tests compaction on multiple types. `test/sql/local/catalog_test_config_setup/catalog_agnostic/create/test_create_table_as.test` does this
- I think all of the tests that have `require-env SECRETS_CREATED_AND_CATALOG_ATTACHED` are not run, they should have `require-env CATALOG_TEST_CONFIG_SETUP`

This PR still has multiple Iceberg scans that are unioned together. These scans are now per partition value. I think this is to allow pulling the partition information through to `ICEBERG_REWRITE_DATA_FILES` operator, but I think you can get rid of the multiple scans completely if there is only 1 partition schema. The Copy operator already has a way to partition files, you just need to generate the partition expression based on the schema of the table.
See `iceberg_insert.cpp::GeneratePartitionExpressions()` as to how we do that with `IcebergInsert`, From the stats returned via `RETURN_STATS` we can get the partition stats as well, so you don't have to maintain the groups for the partition values. 

Currently the plan for a rewrite looks like this on a partitioned table with 2 partitions
```
┌───────────────────────────┐
│ ICEBERG_REWRITE_DATA_FILES│
│    ────────────────────   │
│         Groups: 2         │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│           UNION           ├───────────────────────────────────────────┐
└─────────────┬─────────────┘                                           │
┌─────────────┴─────────────┐                             ┌─────────────┴─────────────┐
│         PROJECTION        │                             │         PROJECTION        │
│    ────────────────────   │                             │    ────────────────────   │
│             0             │                             │             1             │
│             #0            │                             │             #0            │
│             #1            │                             │             #1            │
│             #2            │                             │             #2            │
│             #3            │                             │             #3            │
│             #4            │                             │             #4            │
│             #5            │                             │             #5            │
│             #6            │                             │             #6            │
│                           │                             │                           │
│           ~1 row          │                             │           ~1 row          │
└─────────────┬─────────────┘                             └─────────────┬─────────────┘
┌─────────────┴─────────────┐                             ┌─────────────┴─────────────┐
│        COPY_TO_FILE       │                             │        COPY_TO_FILE       │
│    ────────────────────   │                             │    ────────────────────   │
│      FORMAT: PARQUET      │                             │      FORMAT: PARQUET      │
└─────────────┬─────────────┘                             └─────────────┬─────────────┘
┌─────────────┴─────────────┐                             ┌─────────────┴─────────────┐
│         PROJECTION        │                             │         PROJECTION        │
│    ────────────────────   │                             │    ────────────────────   │
│             #0            │                             │             #0            │
│             #1            │                             │             #1            │
│             #2            │                             │             #2            │
│                           │                             │                           │
│          ~12 rows         │                             │          ~12 rows         │
└─────────────┬─────────────┘                             └─────────────┬─────────────┘
┌─────────────┴─────────────┐                             ┌─────────────┴─────────────┐
│           FILTER          │                             │           FILTER          │
│    ────────────────────   │                             │    ────────────────────   │
│          IN (...)         │                             │          IN (...)         │
│                           │                             │                           │
│          ~12 rows         │                             │          ~12 rows         │
└─────────────┬─────────────┘                             └─────────────┬─────────────┘
┌─────────────┴─────────────┐                             ┌─────────────┴─────────────┐
│         HASH_JOIN         │                             │         HASH_JOIN         │
│    ────────────────────   │                             │    ────────────────────   │
│      Join Type: MARK      │                             │      Join Type: MARK      │
│                           │                             │                           │
│        Conditions:        ├──────────────┐              │        Conditions:        ├──────────────┐
│       filename = #0       │              │              │       filename = #0       │              │
│                           │              │              │                           │              │
│          ~12 rows         │              │              │          ~12 rows         │              │
└─────────────┬─────────────┘              │              └─────────────┬─────────────┘              │
┌─────────────┴─────────────┐┌─────────────┴─────────────┐┌─────────────┴─────────────┐┌─────────────┴─────────────┐
│        ICEBERG_SCAN       ││      COLUMN_DATA_SCAN     ││        ICEBERG_SCAN       ││      COLUMN_DATA_SCAN     │
│    ────────────────────   ││    ────────────────────   ││    ────────────────────   ││    ────────────────────   │
│         Function:         ││                           ││         Function:         ││                           │
│        ICEBERG_SCAN       ││                           ││        ICEBERG_SCAN       ││                           │
│                           ││                           ││                           ││                           │
│        Projections:       ││                           ││        Projections:       ││                           │
│             id            ││                           ││             id            ││                           │
│          category         ││                           ││          category         ││                           │
│          payload          ││                           ││          payload          ││                           │
│                           ││                           ││                           ││                           │
│          Filters:         ││                           ││          Filters:         ││                           │
│  optional: (filename IN ( ││                           ││  optional: (filename IN ( ││                           │
│          /default         ││                           ││          /default         ││                           │
│ /rewrite_partitioned/data ││                           ││ /rewrite_partitioned/data ││                           │
│ /category=a/019ef56e-3b67 ││                           ││ /category=b/019ef56e-3b68 ││                           │
│  -7d83-9b83-8caeb2ef5d7e  ││                           ││  -738f-bb10-7bc78053af51  ││                           │
│        .parquet'))        ││                           ││        .parquet'))        ││                           │
│                           ││                           ││                           ││                           │
│          ~12 rows         ││          ~0 rows          ││          ~12 rows         ││          ~0 rows          │
└───────────────────────────┘└───────────────────────────┘└───────────────────────────┘└───────────────────────────┘
```
With a union for every partition value
The mark join and column data scans are because of our optimizer, you can ignore that.
Ideally it could look like this if you use the partitioned copy
```
┌───────────────────────────┐
│ ICEBERG_REWRITE_DATA_FILES│
│    ────────────────────   │
│         Groups: 2         │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         PROJECTION        │
│    ────────────────────   │
│             0             │
│             #0            │
│             #1            │
│             #2            │
│             #3            │
│             #4            │
│             #5            │
│             #6            │
│                           │
│           ~1 row          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│        COPY_TO_FILE       │
│    ────────────────────   │
│      FORMAT: PARQUET      │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         PROJECTION        │
│    ────────────────────   │
│             #0            │
│             #1            │
│             #2            │
│                           │
│          ~12 rows         │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│           FILTER          │
│    ────────────────────   │
│          IN (...)         │
│                           │
│          ~12 rows         │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│    ────────────────────   │
│         Function:         │
│        ICEBERG_SCAN       │
│                           │
│        Projections:       │
│             id            │
│          category         │
│          payload          │
│                           │
│          Filters:         │
│  optional: (filename IN ( │
|         ....... ))        |
│                           │
│          ~12 rows         |
└───────────────────────────┘
```

Eventually, this will allow us to compact a table with multiple partition schemas by running a union of `COPY` operators, where each copy operator is copying for one partition schema. Then we can maintain groups on a per-partition-schema basis. This will introduce more Iceberg Scans again, but at that point I'm willing to live with it 😅.

This isn't needed right now, and could potentially be a follow-up. If you want to wait with this, maybe add an extra filter in the planning stage so that every Iceberg Scan has a filter to only select the partition it is scanning? That will speed up the scans considerably since we have file pruning at the manifest level

p1p1bear:
Thanks for the detailed pass!

I pushed a follow-up cleanup commit for this round:

- added a `test_all_types()` rewrite test for broader primitive type coverage
- switched the rewrite catalog tests from `SECRETS_CREATED_AND_CATALOG_ATTACHED` to `CATALOG_TEST_CONFIG_SETUP`
- removed unused maintenance loader helpers / planner fields
- merged the rewrite option structs so the defaults live in one place
- added table-property based target file size resolution before falling back to the default
- combined the adjacent lock regions in the rewrite operator
- changed the COPY mode to `append=true`

For the partitioned COPY suggestion: I agree that the cleaner long-term shape is to use COPY's partitioned output, likely one COPY per partition schema rather than one scan/COPY per partition value. I haven't done that refactor in this commit, since it changes the execution shape quite a bit. The current implementation still keeps the existing per-partition grouping and explicitly rejects partition spec evolution, so I'd prefer to leave the partitioned-COPY refactor as a follow-up.

src/function/metadata/iceberg_rewrite_data_files.cpp
p1p1bear: Changed this to `append=true`.

`overwrite_or_ignore=false` errors when the target directory is already non-empty in this COPY path, not only when a generated filename conflicts. For rewrite we write new files into an existing Iceberg data directory and let the REPLACE snapshot decide which old files are logically replaced, so append is the behavior we want here.

src/function/metadata/iceberg_rewrite_data_files.cpp
p1p1bear: Added this in the planner. The order is now explicit argument, then `write.parquet.target-file-size-bytes`, then `write.target-file-size-bytes`, then the default.

I also added a focused test for this precedence.

src/include/maintenance/maintenance_table_loader.hpp
p1p1bear: Removed this helper now; only the reload helper that rewrite still uses is kept.

src/include/maintenance/maintenance_table_loader.hpp
p1p1bear: Removed this one as well.

src/include/maintenance/rewrite_data_files_planner.hpp
p1p1bear: Done. I removed the separate options struct and parse directly into `RewriteDataFilesPlanInput` now.

src/maintenance/rewrite_data_files_operator.cpp
p1p1bear: Agreed, combined them. `BuildRewriteManifestEntry` only builds the entry in memory, so there was no real reason to split the accounting updates across two lock acquisitions.

src/function/iceberg_functions.cpp
p1p1bear: Removed.

src/include/maintenance/rewrite_data_files_planner.hpp
p1p1bear: Good catch, removed this field.

src/include/maintenance/rewrite_data_files_planner.hpp
p1p1bear: Good catch, this was an unused no-op marker. I removed it as well; an empty `file_groups` already represents the no-op case.

Tishj:
> so I'd prefer to leave the partitioned-COPY refactor as a follow-up.

That also has my preference
Also because I'm not convinced we can get the same level of control over created files without making significant changes to PhysicalCopyToFile

src/include/maintenance/rewrite_data_files_planner.hpp
Tmonster: Seems like this field is still in the struct?

Tmonster (approved):
LGTM, just the 1 comment. 

Unsure how open you are to future improvements, but one thing that will really help is file pruning based on file names, or adding a filter to the iceberg_scans that filters on the partitioned value. Right now I think rewrite data files will be pretty slow, but with one of these two improvements I think it will be much faster

Tmonster:
@p1p1bear if you can just fix the merge conflicts, then I think this is g2g

src/include/maintenance/rewrite_data_files_planner.hpp
p1p1bear: Removed now in ebbbcc60. The leftover candidate position fields are gone from `RewriteCandidate`, and the planner now iterates candidates directly instead of storing unused manifest/entry indexes.

p1p1bear:
> @p1p1bear if you can just fix the merge conflicts, then I think this is g2g

Fixed now :)

p1p1bear:
> LGTM, just the 1 comment.
> 
> Unsure how open you are to future improvements, but one thing that will really help is file pruning based on file names, or adding a filter to the iceberg_scans that filters on the partitioned value. Right now I think rewrite data files will be pretty slow, but with one of these two improvements I think it will be much faster

Thanks, agreed. I’ll keep this as a follow-up rather than folding it into this PR. 

The current implementation focuses on correctness and commit semantics; adding scan-side pruning by selected file paths or partition values sounds like the right next performance improvement.

test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rewrite_data_files_all_types.test
```diff
@@ -0,0 +1,126 @@
+# name: test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rewrite_data_files_all_types.test
+# description: rewrite_data_files preserves primitive test_all_types columns
+# group: [maintenance]
+
+require-env ICEBERG_SERVER_AVAILABLE
```
Tishj: I have a hunch we don't use this anymore, and this might result in the test never getting run?

p1p1bear: @Tishj  Thanks for catching this. You were right — `ICEBERG_SERVER_AVAILABLE` is not set by the fixture catalog jobs, so these tests could be skipped.

I removed that stale gate from this test and audited the other `rewrite_data_files` maintenance tests as well. The related cases have all been cleaned up now. :)

Tmonster:
Thanks!

The-Alchemist:
fantastic work, guys!  looking forward to using it!

because this is against the `main` branch, I'm guessing this will be released as part of `1.6.x` series?

Tmonster:
@The-Alchemist yup, probably [2.0.0 actually](https://duckdb.org/release_calendar)!

PR #1036, opened 2026-06-08
Resolve virtual columns in GetScanFunction like Ducklake

NiclasHaderer:
`IcebergTableEntry::GetScanFunction(context, bind_data, lookup)` calls `iceberg_scan_function.bind()` to create `MultiFileBindData`, but never populates `MultiFileBindData::virtual_columns`. The `get_virtual_columns` callback (`IcebergVirtualColumns`) normally handles this during plan creation, but callers that invoke `GetScanFunction` directly and skip the planner never trigger that callback.

DuckLake equivalent for reference 
https://github.com/duckdb/ducklake/blob/d897bc5a35f887c9087403bc20efd92d99272e69/src/storage/ducklake_table_entry.cpp#L324


Tmonster:
Thanks!

PR #1040, opened 2026-06-08
WIP: Map write.parquet.format-version to DuckDB Parquet V2 writes.

The-Alchemist:
Allow Iceberg tables to opt into Parquet file format version 2 via a table property instead of relying on the DuckDB default of V1.

No other behavior should change.


Tishj:
I generally don't see a problem with this, other than it not being a standard

None of these options are defined in the spec, which bothers me. They're only defined in the docs of the iceberg runtime.

But this one isnt even defined there.
Without a proper spec for these options, I could easily see a world where 3 different paths exist with the same meaning, and I don't want to contribute to that

Tishj:
My advise to get this accepted is to take it to apache/iceberg

Tmonster:
I also don't see a problem, but would like to know if other engines have similar options/capabilities before we this option to DuckDB

The-Alchemist:
As per your suggestion @Tishj , I made an equivalent PR for apache/iceberg at https://github.com/apache/iceberg/pull/16763

The-Alchemist:
Closing until https://github.com/apache/iceberg/pull/16763 gets some traction.

PR #1048, opened 2026-06-09
Round up the partition field summary string upper bound

### LINKED ISSUE #1047: DuckDB writes an invalid partition `upper_bound` into the manifest list.
### What happens?

For a string partition column, the partition summary `upper_bound` (`manifest_file.partitions[].upper_bound`) is the value truncated to 16 bytes with no round-up. When a partition value is longer than 16 bytes, that bound ends up *smaller* than the real value — so the manifest claims it does not contain its own partition. (The spec allows truncation, but an upper bound must be rounded up so it stays `>=` every value.)

DuckDB already does this correctly for column metrics (`data_file.upper_bounds`); only the partition summary is missing the round-up. The repro prints both from the same manifest: column metric `...g` (correct), partition summary `...f` (wrong).

Why it matters: engines that prune by these summaries skip the manifest. On Trino / Athena over a DuckDB-written table, `WHERE p = '<long key>'` silently returns too few rows, and `OPTIMIZE` / `rewrite_data_files` fails with `Missing required files to delete` for files that exist. A full scan (no partition filter) is fine, so it is easy to miss. DuckDB's own reads are not affected.

Setup: REST catalog, table created and written by DuckDB. The repro partitions with `truncate(20, p)` to get >16-byte string partition values from plain SQL; the same happens with `identity` partitioning on long keys (the usual real-world case).

Likely cause (one-line slip): in `IcebergManifestList`'s field-summary computation (`src/core/metadata/manifest/iceberg_manifest_list.cpp`, on both `main` and `v1.5-variegata`) the upper bound is serialized with `SerializeBound::LOWER_BOUND`:

```cpp
lower_result = IcebergValue::SerializeValue(min_values[i]..., SerializeBound::LOWER_BOUND);
upper_result = IcebergValue::SerializeValue(max_values[i]..., SerializeBound::LOWER_BOUND);  // should be UPPER_BOUND
```

In `iceberg_value.cpp`, `SerializeBound::UPPER_BOUND` routes through `TruncateAndIncrementString` (rounds up — the function PR #1022 fixed), while `LOWER_BOUND` uses plain `TruncateString`. The column-metrics writer already passes `UPPER_BOUND`, which is why that bound is correct and why #1022 did not touch this path. Passing `SerializeBound::UPPER_BOUND` for `upper_result` should fix it.

### To Reproduce

Start a local Iceberg REST catalog:

```bash
docker run -d -p 8181:8181 -v /tmp/icewh:/tmp/icewh \
  -e CATALOG_WAREHOUSE=/tmp/icewh \
  -e CATALOG_IO__IMPL=org.apache.iceberg.hadoop.HadoopFileIO \
  apache/iceberg-rest-fixture:latest
```

Write a string-partitioned table with DuckDB and inspect the manifest it wrote (`fastavro` is only used to read the Avro metadata):

```python
import duckdb, glob, os, fastavro

con = duckdb.connect()
con.execute("ATTACH '' AS ice (TYPE ICEBERG, ENDPOINT 'http://127.0.0.1:8181', AUTHORIZATION_TYPE 'none')")
con.execute("CREATE SCHEMA ice.demo")
con.execute("CREATE TABLE ice.demo.tt (id INTEGER, p VARCHAR)")
con.execute("ALTER TABLE ice.demo.tt SET PARTITIONED BY (truncate(20, p))")
con.execute("INSERT INTO ice.demo.tt VALUES (1, '0123456789abcdef0123456789abcdef')")

snap = sorted(glob.glob('/tmp/icewh/demo/tt/metadata/snap-*.avro'), key=os.path.getmtime)[-1]
mf = next(iter(fastavro.reader(open(snap, 'rb'))))
summary = mf['partitions'][0]
value = b'0123456789abcdef0123'  # truncate(20) of the inserted key

print("partition value:      ", value)
print("summary lower_bound:  ", summary['lower_bound'])
print("summary upper_bound:  ", summary['upper_bound'])
print("upper_bound >= value: ", summary['upper_bound'] >= value)

data_file = next(iter(fastavro.reader(open(mf['manifest_path'], 'rb'))))['data_file']
col_upper = next(kv['value'] for kv in data_file['upper_bounds'] if kv['key'] == 2)
print("column upper_bound(p):", col_upper)
```

Output:

```
partition value:       b'0123456789abcdef0123'
summary lower_bound:   b'0123456789abcdef'
summary upper_bound:   b'0123456789abcdef'
upper_bound >= value:  False
column upper_bound(p): b'0123456789abcdeg'
```

The partition-summary `upper_bound` (`...f`) is less than the partition value, while the column-metrics `upper_bound` for the same column is correctly rounded up (`...g`).

### OS:

macOS 26.3.1, arm64 (Darwin 25.3.0)

### DuckDB Version:

1.5.3

### DuckDB Client:

Python (also reproducible from the CLI — the table is built with plain SQL)

### Hardware:

_No response_

### Full Name:

Dmitrii Tikhonov

### Affiliation:

Behavix Oy

- [x] Yes, I have

- [x] Yes, I have

tikhdm:
The manifest-list partition field summary `upper_bound` for a string partition is serialized with `SerializeBound::LOWER_BOUND`, so it is a plain 16-byte truncation instead of being rounded up. For any partition value longer than 16 bytes the truncated `upper_bound` sorts *less than* the value, so the summary claims the manifest cannot contain its own partition.

Engines that prune by manifest-list partition summaries (Trino/Athena, Spark) then skip the manifest: partition-equality filters silently drop rows, and `OPTIMIZE` / `rewrite_data_files` aborts with `Missing required files to delete` for files that exist.

The column-metrics writer already passes `SerializeBound::UPPER_BOUND` (which routes through `TruncateAndIncrementString`, the path fixed in #1022); only the partition-summary path was wrong. This passes `UPPER_BOUND` there too.

Before/after with `truncate(20, p)` on a 32-byte key (partition value `0123456789abcdef0123`):

| | upper_bound | `>= value` |
| --- | --- | --- |
| before | `0123456789abcdef` | false |
| after | `0123456789abcdeg` | true |

Added a regression test under `insert/partitions/` using `iceberg_partition_stats` (red before, green after).

Fixes #1047.

### DIFF
--- a/src/core/metadata/manifest/iceberg_manifest_list.cpp
+++ b/src/core/metadata/manifest/iceberg_manifest_list.cpp
@@ -216,7 +216,7 @@ void ManifestPartitions::Create(const IcebergTableMetadata &metadata, const Iceb
 			lower_result = IcebergValue::SerializeValue(min_values[i].DefaultCastAs(LogicalType::VARCHAR),
 			                                            min_values[i].type(), SerializeBound::LOWER_BOUND);
 			upper_result = IcebergValue::SerializeValue(max_values[i].DefaultCastAs(LogicalType::VARCHAR),
-			                                            max_values[i].type(), SerializeBound::LOWER_BOUND);
+			                                            max_values[i].type(), SerializeBound::UPPER_BOUND);
 		}
 
 		if (lower_result.HasValue()) {
--- a/test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/partitions/test_partition_summary_string_upper_bound.test
+++ b/test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/partitions/test_partition_summary_string_upper_bound.test
@@ -1,0 +1,45 @@
+# name: test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/partitions/test_partition_summary_string_upper_bound.test
+# description: The manifest-list partition field summary upper_bound for a string partition must be rounded up (>= the partition value), not a plain 16-byte truncation. A truncated-down upper bound wrongly prunes the partition from its own manifest (see duckdb-iceberg#1047).
+# group: [partitions]
+
+require-env CATALOG_TEST_CONFIG_SETUP
+
+require avro
+
+require parquet
+
+require iceberg
+
+require httpfs
+
+# Do not ignore 'HTTP' error messages!
+set ignore_error_messages
+
+statement ok
+drop table if exists my_datalake.default.partition_summary_upper_bound;
+
+statement ok
+create table my_datalake.default.partition_summary_upper_bound (id integer, p varchar);
+
+# truncate(20) produces a 20-byte partition value, longer than the 16-byte bound truncation.
+statement ok
+alter table my_datalake.default.partition_summary_upper_bound set partitioned by (truncate(20, p));
+
+statement ok
+insert into my_datalake.default.partition_summary_upper_bound values (1, '0123456789abcdef0123456789abcdef');
+
+# The partition value is truncate(20, key) = '0123456789abcdef0123'. The field-summary
+# upper_bound must be >= that value. Previously it was the plain 16-byte truncation
+# '0123456789abcdef', which is < the value and excludes the partition from its own manifest.
+query I
+select upper_bound >= '0123456789abcdef0123'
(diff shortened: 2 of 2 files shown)


Tmonster (approved):
LGTM thanks!
just waiting for tests to pass

PR #1051, opened 2026-06-10
Add credential auto-refresh support for web_identity/STS chains

haoxu0:
## Summary

Adds credential auto-refresh support for `web_identity` and `sts` credential chains (e.g., IRSA on EKS). Without this, STS credentials expire after ~1 hour and Iceberg queries fail with `ExpiredToken`.

## Problem

The iceberg extension was:
1. **Stripping `refresh_info`** from the catalog secret when building per-table secrets
2. **Snapshotting static credentials** for SigV4 catalog API calls without triggering refresh

Even with duckdb-aws#136 (`web_identity` chain with auto-refresh) and duckdb-httpfs#165 (refresh in more locations), the iceberg extension bypassed the refresh path entirely.

### `sigv4.cpp` / `sigv4.hpp` — Serialized credential refresh before SigV4 signing

Adds `MaybeRefreshSecret()` method to `SIGV4Authorization` that:
- Checks if enough time has passed since the last refresh (300s interval, configurable via `REFRESH_INTERVAL_SECONDS`)
- Acquires a mutex with `try_to_lock` — concurrent callers skip instead of blocking
- Double-checks timing after acquiring the lock to avoid redundant refreshes
- Reconstructs a `CreateSecretInput` from the secret's `refresh_info` struct and re-runs the credential chain via `SecretManager::CreateSecret` to get fresh STS tokens
- On failure, skips updating the timestamp so the next request retries

The `refresh_info` is a STRUCT containing the original named parameters used to create the secret (e.g., `chain`, `region`, `profile`). These are stored by the aws extension when `refresh='auto'` is set (automatically for `web_identity` and `sts` chains).

### `iceberg_table_information.cpp` — Propagate refresh_info to per-table secrets

- Adds `TryRefreshCatalogSecret()` helper that mirrors httpfs's refresh logic using public DuckDB APIs (no cross-extension dependency)
- Before snapshotting catalog credentials into per-table secrets, triggers a refresh to ensure fresh values
- Stops stripping `refresh_info` so httpfs can auto-refresh per-table secrets during long-running data reads

## Dependencies

- duckdb-aws#136 (`web_identity` chain) — merged
- duckdb-httpfs#165 (refresh in GLOBs/range reads) — open; our patch handles catalog API calls independently

## Testing

- Added `test_web_identity_credential_refresh.test` guarded by `mode skip` with environment documentation
- Tested in production on EKS with IRSA — pods running 10+ hours with zero `ExpiredToken` errors, credentials refreshing transparently


src/catalog/rest/storage/authorization/sigv4.cpp
```diff
@@ -105,8 +105,39 @@ AWSInput SIGV4Authorization::CreateAWSInput(ClientContext &context, const IRCEnd
...
+			CreateSecretInput refresh_input;
+			refresh_input.on_conflict = OnCreateConflict::REPLACE_ON_CONFLICT;
+			refresh_input.persist_type = SecretPersistType::TEMPORARY;
+			refresh_input.type = kv_secret_pre.GetType();
+			refresh_input.name = kv_secret_pre.GetName();
+			refresh_input.provider = kv_secret_pre.GetProvider();
+			refresh_input.storage_type = secret_entry->storage_mode;
+			refresh_input.scope = kv_secret_pre.GetScope();
+
+			auto child_count = StructType::GetChildCount(refresh_info.type());
```
Tmonster: Can you leave a comment as to what children are needed in a secret refresh? I can't think of any

src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
```diff
@@ -27,6 +27,43 @@
 
 namespace duckdb {
 
+//! Attempt to refresh the catalog's storage secret if it supports auto-refresh.
+//! Rebuilds the secret via its original provider (e.g., web_identity/sts credential chain)
+//! which fetches fresh STS tokens. No-op if the secret lacks refresh_info.
+static bool TryRefreshStorageSecret(ClientContext &context, const SecretEntry &secret_entry) {
```
Tmonster: I think this can be renamed to `TryRefreshCatalogSecret`?

Tmonster (commented):
I would like to see some test for this, even if they are guarded by `mode skip` + a description of what kind of environment should be set up. 

I think there is also a nicer way to handle the Vcpkg ports, I'll come back to this

src/catalog/rest/storage/authorization/sigv4.cpp
haoxu0: Done — added a comment explaining that `refresh_info` is a STRUCT containing the original named parameters used to create the secret (e.g., `chain`, `region`, `profile`). These are stored by the aws extension when `refresh='auto'` is set. Re-passing them to `SecretManager::CreateSecret` re-runs the credential chain provider, which fetches fresh STS tokens.

src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
haoxu0: Done — renamed to `TryRefreshCatalogSecret`.

Tmonster (commented):
Thanks! Looks good to me, are we worried at all that this might refresh the secret unnecessarily? I.e if are credentials are valid for an hour, we don't need to refresh every time we read the table within that hour right?

haoxu0:
@Tmonster actually the biggest problem for us it got into a race situation when multiple queries landed and all try to refresh the creds. And I'm trying to resolve this issue now

src/catalog/rest/storage/authorization/sigv4.cpp
```diff
@@ -12,6 +13,11 @@
 
 namespace duckdb {
 
+// Static member definitions — shared across all SIGV4Authorization instances
+// AND TryRefreshStorageSecret (via GetRefreshMutex/GetLastRefreshTime accessors)
+std::mutex SIGV4Authorization::refresh_mutex;
```
Tmonster: Is there a reason we need this shared across all SigV4Authorization instances? If I attach different sigv4 catalogs with different secrets, the refresh logic for both of them should not interfere

src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
```diff
@@ -23,10 +23,74 @@
...
+//! Uses SIGV4Authorization's shared mutex to prevent write-write conflicts with
+//! MaybeRefreshSecret (both call CreateSecret on the same aws_secret).
+static bool TryRefreshCatalogSecret(ClientContext &context, const SecretEntry &secret_entry) {
+	auto &refresh_mutex = SIGV4Authorization::GetRefreshMutex();
+	auto &last_refresh_time = SIGV4Authorization::GetLastRefreshTime();
+	static constexpr int REFRESH_INTERVAL_SECONDS = 300;
+
+	// Fast path: skip if refreshed recently
+	auto now = std::chrono::steady_clock::now();
+	auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(now - last_refresh_time).count();
```
Tmonster: Seems like we run the `if we refreshed recently return` check here and also in `MaybeRefreshSecret`? I think we can gfet away with only doing this once?

Tmonster (commented):
I don't think having one sigv4 refresh for all sigv4 authenticated catalogs is a good idea. Can we make the refresh a per catalog thing? 

Also, if you can figure out a way to add tests that would be awesome. I understand it will require some more setup, but if you could include the setup in some script unders `scripts`, you can guard your test with a `mode skip`

src/catalog/rest/storage/authorization/sigv4.cpp
haoxu0: Good catch on the wording — the mutex and `last_refresh_time` are actually **instance members** of `SIGV4Authorization` (declared in `sigv4.hpp`), not static/shared. Each attached catalog gets its own `SIGV4Authorization` instance, so the refresh logic is already per-catalog and won't interfere between different catalogs with different secrets.

The `#include <mutex>` at the top of the header is just the standard library include — it doesn't make anything shared.

src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
haoxu0: You're right — these two serve different purposes but the duplication is confusing:

1. **`MaybeRefreshSecret()` in sigv4.cpp** — called on every catalog API request, time-gated to 300s intervals. This is the hot path.
2. **`TryRefreshCatalogSecret()` in iceberg_table_information.cpp** — called once during `GetVendedCredentials()` to ensure fresh creds are snapshotted into per-table secrets. This isn't called per-request.

I'll simplify by removing the time-gate from `TryRefreshCatalogSecret` (it doesn't need one since it's not in the hot path) and just keeping it as a simple "refresh if refresh_info exists" helper. The time-gating only matters in `MaybeRefreshSecret` where it's called on every request.

Tmonster:
Hi @HaoXuAI ,

Can you check if this functionality is not provided by https://github.com/duckdb/duckdb-iceberg/pull/1158 and the httpfs related PR https://github.com/duckdb/duckdb-httpfs/pull/358 ? (I think it was that one)

These do not eagerly refresh, but I believe they get most of the way there. 
It was also merged for v1.5.5, so you can just test it on the latest release 👍

haoxu0:
Thanks — checked both, and they do overlap with part of this PR. I've dropped that part.

**#1158 + httpfs#358 cover the vended, scoped table storage secrets:** httpfs sees a 403, dispatches to `IcebergTableSecretProvider`, and the provider re-loads the table to get freshly vended credentials. Better than what this PR originally did (propagating the catalog secret's `refresh_info` onto per-table secrets), so that's removed.

**What's left uncovered is the catalog secret that signs the REST requests themselves.** `sigv4.cpp` has no refresh on either `main` or `v1.5-variegata`, and httpfs never uses that secret for S3, so its 403 path never fires for it. With a `web_identity`/`sts` chain the token just expires and every catalog call then fails — the `ExpiredToken` we were hitting in production.

**They're complementary:** `ReVendVendedCredentials` re-vends via `IRCAPI::GetTable(...)`, which is signed with the catalog secret — so if that credential has expired, re-vending fails too ("Could not refresh Iceberg vended credentials ... GetTableInformation returned response code"). Keeping it fresh is what lets #1158's path work.

**Tested on v1.5.5 as you suggested:** rebased onto `v1.5-variegata` (duckdb `d8cdaa33`), builds clean, and the refresh test passes with #1158's provider in the same build. There the change is only `sigv4.{hpp,cpp}` — no files shared with #1158.

Tmonster:
Thanks! this is looking much better and I am happy this is getting done.

Can you use an annotated mutex for the `refresh_mutex` in sigv4.hpp?

Something like 
```c++
// sigv4.hpp
#include "duckdb/common/mutex.hpp"
#include "duckdb/common/thread_annotation.hpp"

annotated_mutex refresh_mutex;
std::chrono::steady_clock::time_point last_refresh_time DUCKDB_GUARDED_BY(refresh_mutex) =
      std::chrono::steady_clock::now();

// sigv4.cpp -- fast path deleted, try_lock hoisted out
void SIGV4Authorization::MaybeRefreshSecret(ClientContext &context) {
      if (!refresh_mutex.try_lock()) {
              // Another thread is refreshing; proceed with the current credentials
              return;
      }
      annotated_lock_guard<annotated_mutex> guard(refresh_mutex, std::adopt_lock);
      auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(
                         std::chrono::steady_clock::now() - last_refresh_time).count();
        if (elapsed < REFRESH_INTERVAL_SECONDS) {
                return;
        }
        ...                                                                                                                                 
  }
```

This can help against some race conditions.

I'm going to try and re-add the vended credential testing from https://github.com/duckdb/duckdb-iceberg/pull/1158/changes, hopefully we can make use of that here as well

Tmonster:
Actually, if you merge with main, you will see that `test-fixture-catalog` already tests vended credentials logic, could you add a similar workflow to also test your changes?

src/catalog/rest/storage/authorization/sigv4.cpp
```diff
@@ -116,6 +118,74 @@ AWSInput SIGV4Authorization::CreateAWSInput(ClientContext &context, const IRCEnd
...
+		auto &key = StructType::GetChildName(refresh_info.type(), i);
+		refresh_input.options[key.GetIdentifierName()] = children[i];
+	}
+
+	try {
+		auto &secret_manager = context.db->GetSecretManager();
+		(void)secret_manager.CreateSecret(context, refresh_input);
+		last_refresh_time = std::chrono::steady_clock::now();
+	} catch (std::exception &) {
+		// Leave last_refresh_time alone so the next call retries
```
Tmonster, line 185: nit: Could we add a `DUCKDB_LOG_DEBUG` here with a message that the secret was not updated?

Tmonster (approved):
Thanks!
Just the one nit, but if tests pass I will just merge.

Tmonster:
Awesome thanks!

PR #1053, opened 2026-06-11
Don't prune files on filter expressions whose column side is not a direct reference (fixes #1052)

### LINKED ISSUE #1052: Stats-based file pruning evaluates non-monotone expressions against min/max bounds — silently wrong results (e.g. WHERE pk % 8 = 2 returns 0 rows)
### What happens

On a build containing #1038 (`merge_into_dv_fix`), repeated `MERGE INTO` statements whose update legs target rows in the **same data file** still do not consolidate deletion vectors — and the failure mode is now **silent data loss** instead of 1.5.3's loud `two or more deletion vectors for the same referenced_data_file` error:

- **MERGE 1** (updates 7 rows of the seed file): OK — writes DV #1 (7 positions).
- **MERGE 2** (updates 7 *different* rows of the seed file): commits, but writes an **independent second 7-row DV** against the same data file instead of consolidating into one 14-position DV. Reads remain correct because scans union the disjoint DVs — so nothing notices yet.
- **MERGE 3** (updates 6 more rows of the seed file): **silently drops its entire update leg.** The snapshot commits containing only the `WHEN NOT MATCHED` insert file — no third DV, no rewritten-rows file. The 6 matched rows are never updated → duplicate "current" rows / lost updates, with no error or warning.

The on-disk state was cross-verified with Spark/Iceberg-Java 1.11.0 (reading the same REST catalog): after the 3 merges, `<table>.files` shows **two `content=1` delete files of 7 rows each** and six data files (50-row seed, 7+7 from merge 1, 7+7 from merge 2, and only a 6-row insert file from merge 3 — merge 3's update artifacts are absent entirely).

Reproduces identically when each MERGE runs in a **fresh CLI session** (fresh ATTACH) and with `MAX_TABLE_STALENESS '0 seconds'`, so it's persistent DV state, not session caching.

### Versions

- duckdb-iceberg: `main` @ `c5221e1e` (includes #1038), built from source
- duckdb: `v1.6.0-dev7566` (`6c4baf23e76d`, the submodule pin at that extension SHA)
- Catalog: Apache Polaris (Iceberg REST), S3 storage, format-version 3 table

### Reproduction

```sql
LOAD iceberg; LOAD httpfs; LOAD avro;
CREATE OR REPLACE SECRET s3s (TYPE s3, PROVIDER credential_chain, REGION '<region>');
CREATE OR REPLACE SECRET cat (TYPE iceberg, CLIENT_ID '…', CLIENT_SECRET '…',
  OAUTH2_SERVER_URI '<endpoint>/v1/oauth/tokens', OAUTH2_SCOPE 'PRINCIPAL_ROLE:ALL');
ATTACH 'mycat' AS cat (TYPE iceberg, SECRET cat, ENDPOINT '<endpoint>');

CREATE SCHEMA IF NOT EXISTS cat.repro;
DROP TABLE IF EXISTS cat.repro.dv_gate;

-- seed: 50 rows in ONE data file
CREATE TABLE cat.repro.dv_gate AS
SELECT r.range::INT AS pk, 'v0' AS val,
       TIMESTAMP '2026-01-01 00:00:00' AS _valid_from,
       CAST(NULL AS TIMESTAMP) AS _valid_to,
       true AS _is_current, CAST(0 AS BIGINT) AS _cdc_offset
FROM range(50) r;

-- run this MERGE three times with i = 1, 2, 3
-- (substituting i into 'v{i}', the timestamp hour, the offset, and pk % 8 = {i-1}):
MERGE INTO cat.repro.dv_gate t USING (
  SELECT pk, 'v1' AS val, TIMESTAMP '2026-01-01 01:00:00' AS _valid_from,
         CAST(NULL AS TIMESTAMP) AS _valid_to, true AS _is_current, CAST(1 AS BIGINT) AS _cdc_offset
  FROM range(50) r(pk) WHERE pk % 8 = 0
  UNION ALL
  SELECT pk, val, _valid_from, TIMESTAMP '2026-01-01 01:00:00' AS _valid_to,
         false AS _is_current, _cdc_offset
  FROM cat.repro.dv_gate WHERE _is_current AND pk % 8 = 0
) s
ON t.pk = s.pk AND t._valid_from = s._valid_from AND t._cdc_offset = s._cdc_offset
WHEN MATCHED THEN UPDATE SET _valid_to = s._valid_to, _is_current = s._is_current
WHEN NOT MATCHED THEN INSERT (pk, val, _valid_from, _valid_to, _is_current, _cdc_offset)
  VALUES (s.pk, s.val, s._valid_from, s._valid_to, s._is_current, s._cdc_offset);

-- invariant check after each merge: exactly one current row per pk
SELECT count(*) AS total,
       count(*) FILTER (_is_current) AS cur,
       count(DISTINCT pk) FILTER (_is_current) AS cur_pks
FROM cat.repro.dv_gate;
```

Each merge updates a *different* slice (`pk % 8 = 0`, `= 1`, `= 2`) of the **original seed file**, mirroring a slowly-changing-dimension writer demoting current rows.

### Observed

| after merge | total | current | expected total | expected current |
|---|---|---|---|---|
| 1 | 57 | 50 | 57 | 50 ✅ |
| 2 | 64 | 50 | 64 | 50 ✅ (but 2 stacked DVs on the seed file, not 1 consolidated) |
| 3 | **70** | **56** | 70 | 50 ❌ — 6 PKs now have 2 "current" rows; merge 3's updates were silently dropped |

### Expected

Merge 2 consolidates into a single DV per data file (the V3 invariant); merge 3's updates apply. If consolidation cannot happen for some reason, an error (as in 1.5.3) would be strictly preferable to silently committing a snapshot that's missing the update leg.

### Notes

- #1038's "multiple merge intos" test may not cover the case where the second merge's deletes target the **same data file** as the first (here every merge hits the original seed file); standalone `DELETE`/`UPDATE` statements consolidate correctly in the same build.
- Happy to test a fix branch — we have this wired as an automated gate.

raghav-reglobe:
## What

Fixes #1052 — stats-based file pruning silently returns wrong results for pushed-down filters whose column side is not a direct column reference.

`MatchBoundsExpression` evaluated the column-side expression at the file's lower/upper bound and treated `[f(min), f(max)]` as the range of `f` over the file. That is only sound when `f` is monotone over the bound interval — which cannot be established in general. For `pk % 8 = 4` over a file with bounds `[0, 49]` it computes `[0 % 8, 49 % 8] = [0, 1]` and prunes the file, silently dropping all matching rows from the result (residues 0 and 1 "work" by coincidence, 2–7 lose data). Same class applies to `day(ts)`, `abs(x)`, etc.

## Fix

Conservative: only direct column-reference comparisons participate in bounds pruning (the existing `MatchBoundsConstant` paths); any other comparison shape returns "cannot prune". Removes the now-unused `MatchTransformedBounds` / `BoundExpressionReplacer`. A follow-up could reclaim pruning for provably monotone shapes (e.g. widening casts) behind an explicit allowlist — happy to take guidance on whether that's wanted.

## Validation

- New catalog-agnostic test (`reads/filter_pushdown_non_monotone.test`): `pk % 8 ∈ {2, 7, 0}` counts plus a direct-reference control.
- Against a live Polaris REST catalog + S3, on this branch: the previously failing queries return correct counts, and an 8-tick repeated-MERGE SCD2 soak that was derailed by this pruning (its source subquery used `pk % 8 = residue` and read zero files at tick 3) now passes 8/8 with the SCD2 invariant intact — both with modulo and with range predicates.
- Confirmed the bug reproduces on the v1.5.2 release and `main` (`c5221e1e`) before the fix.

### DIFF
--- a/src/planning/pruning/iceberg_predicate.cpp
+++ b/src/planning/pruning/iceberg_predicate.cpp
@@ -18,28 +18,6 @@
 
 namespace duckdb {
 
-namespace {
-
-struct BoundExpressionReplacer : public LogicalOperatorVisitor {
-public:
-	BoundExpressionReplacer(const Value &val) : val(val) {
-	}
-
-public:
-	unique_ptr<Expression> VisitReplace(BoundReferenceExpression &expr, unique_ptr<Expression> *expr_ptr) override {
-		if (expr.index != 0) {
-			return nullptr;
-		}
-		auto &return_type = expr.GetReturnType();
-		return make_uniq<BoundConstantExpression>(val.DefaultCastAs(return_type, true));
-	}
-
-public:
-	const Value &val;
-};
-
-} // namespace
-
 template <class TRANSFORM>
 bool MatchBoundsTemplated(ClientContext &context, const ExpressionFilter &filter, const IcebergPredicateStats &stats,
                           const IcebergTransform &transform);
@@ -90,37 +68,6 @@ static bool MatchBoundsIsNullFilter(const IcebergPredicateStats &stats, const Ic
 template <class TRANSFORM>
 static bool MatchBoundsIsNotNullFilter(const IcebergPredicateStats &stats, const IcebergTransform &transform) {
 	return stats.has_not_null == true;
-}
-
-template <class TRANSFORM>
-bool MatchTransformedBounds(ClientContext &context, ExpressionType comparison_type, const Expression &left,
-                            const Expression &right, const IcebergPredicateStats &stats,
-                            const IcebergTransform &transform) {
-	BoundExpressionReplacer lower_replacer(stats.lower_bound);
-	BoundExpressionReplacer upper_replacer(stats.upper_bound);
-	auto lower_copy = left.Copy();
-	auto upper_copy = left.Copy();
-	lower_replacer.VisitExpression(&lower_copy);
-	upper_replacer.VisitExpression(&upper_copy);
-
-	Value right_constant;
-	if (!ExpressionExecutor::TryEvaluateScalar(context, right, right_constant)) {
-		return true;
-	}
-
-	Value transformed_lower_bound;
-	Value transformed_upper_bound;
-	if (!ExpressionExecutor::TryEvaluateScalar(context, *lower_copy, transformed_lower_bound)) {
-		return true;
-	}
-	if (!ExpressionExecutor::TryEvaluateScalar(context, *upper_copy, transformed_upper_bound)) {
-		return true;
-	}
-	IcebergPredicateStats transformed_stats(stats);
-	transformed_stats.lower_bound = transformed_lower_bound;
-	transformed_stats.upper_bound = transformed_upper_bound;
-
-	return MatchBoundsConstant<TRANSFORM>(right_constant, comparison_type, transformed_stats, transform);
 }
 
 // template <class TRANSFORM>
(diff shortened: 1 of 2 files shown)


Tishj:
I believe you're right.
We can't make this assumption, because we have no property on an expression to guarantee this

PR #1055, opened 2026-06-11
backport #1053

Tishj:
This PR backports #1053


Tishj:
It passed CI, but I'm SUPER skeptical of it suddenly passing `test_tpch.test` when it was consistently failing before

Issue #790, opened 2026-03-11
Support MergeAppend insert strategy


Tishj:
Currently every `INSERT/UPDATE/DELETE` operation performed in a transaction creates a new `manifest_file`.
This is convenient because it means we never have to look at any existing manifest files, but it is not efficient for reads, because it causes an unbounded grow in the amount of files that have to be read for scan planning.
This is known as the "FastAppend" strategy.

To make reads more efficient, the "MergeAppend" strategy can be used instead.
In a MergeAppend, the writer should analyze the current state of the metadata and potentially add its new `manifest_entry` items to existing manifest file(s), rewriting the existing files in the process.

Starting with #728 we have added support for rewriting existing manifest files, so a portion of the work is already done.

Relevant table properties that should be respected for this are (see [docs](https://iceberg.apache.org/docs/1.10.0/configuration/?h=commit.retry.num+retries#table-behavior-properties)):
- `commit.manifest-merge.enabled`
- `commit.manifest.min-count-to-merge`
- `commit.manifest.target-size-bytes`

cjnoname:
Implemented in #1056 (together with #786, since merge runs as a step of the per-attempt `apply`, so the two are coupled).

How the issue's points are addressed:

- **MergeAppend strategy** — when committing, the writer analyzes the current manifest list and bin-packs manifests into fewer/larger files, rewriting them (building on the #728 rewrite support). DATA and DELETE manifests are merged independently, grouped by `partition_spec_id`. The bin-packing matches Java `ManifestMergeManager` / PyIceberg `ListPacker.pack_end` (first-fit, `lookback=1`, pack-from-tail) so the under-filled bin lands on the newest manifests (merged next time).
- **Table properties respected** (Java defaults):
  - `commit.manifest-merge.enabled` (default `true`)
  - `commit.manifest.min-count-to-merge` (default `100`)
  - `commit.manifest.target-size-bytes` (default `8 MB`)
- Merged-in committed entries are written `EXISTING` with their original sequence numbers; entries new in this transaction stay `ADDED` with inherited (NULL) sequence numbers, which keeps them retry-immune (#786). `min_sequence_number` and (V3) `first_row_id` are recomputed for merged manifests so scan-planning pruning stays correct.
- Bonus aligned with Java: manifests are written with `write.manifest.compression-codec` (default deflate) instead of uncompressed.

[...]

Issue #786, opened 2026-03-11
Supporting retries for commits


Tishj:
Iceberg uses an optimistic concurrency model, which means it is built with efficient retries in mind.

Currently if the commit request to the REST Catalog fails, the transaction aborts immediately.
We should instead perform a retry of the commit, by rewriting the necessary data and performing the commit request again, up to a set maximum of retries.

That will require some changes to how we structure commits and transactional data currently.
I believe all operations can be retried, but not all operations can be *easily* retried.
I think we should first start with retrying operations that limit the changes to just the `IcebergSnapshot` and the `IcebergManifestFile` level.

`IcebergSnapshot` (`snapshot` in the spec) will need to be rewritten for a retry.
`IcebergManifestFile` (`manifest_file` in the spec) will also need to be rewritten for a retry.
`IcebergManifestEntry` (`manifest_entry` in the spec) is structured in such a way that it should never require rewriting for a retry[0].

[0] This is *mostly* true, but changes to the `partition-spec-id` and `sort-order-id` can invalidate this assumption.
If those changes happen in the same transaction as the creation of the `manifest_entry` (the manifest file), we can not retry the commit without having to rewrite the created manifest, as the `sort-order-id` or `partition-spec-id` used could be in use by another commit, and those are materialized in the `manifest_entry`.

Apart from those, V3 deletion vectors might also have to be rewritten, and thereby also force a rewrite of the accompanying `manifest_entry`. If another commit has made a deletion targeting the same `referenced_data_file`, we will need to combine those with our own deletes.

## Required changes

While the `manifest_entry` is built to be immune to retries, we are not writing it as such.
We should not be materializing the following for newly added data:
- `manifest_entry.snapshot_id`
- `manifest_entry.sequence_number`
- `manifest_entry.file_sequence_number`
- `manifest_entry.data_file.first_row_id`

We should also bundle all data changes (insert/update/delete) of a table into a single snapshot.
This will be required to make UPDATE operations that target transaction-local (uncommitted) data immune to retries.
An UPDATE requires materializing the `_row_id` that's inferred from the scan in the parquet file, to preserve it.
To make this immune to retries, the `_row_id` that should be inferred by the UPDATE's scan should be `NULL` for transaction-local data, that way it doesn't get invalidated by any additional inserts that are made in between retry attempts.

cjnoname:
Hi @Tishj , 

[...]

PR #1056, opened 2026-06-12
MergeAppend (#790) and commit retries (#786)

cjnoname:
Implements **MergeAppend (#790)** and **optimistic-concurrency commit retries (#786)** as one coupled change (merge runs as a step of a pure, per-attempt `apply`, so retries regenerate merged manifests correctly). Validated end-to-end against **AWS S3 Tables** and a local **apache/iceberg-rest-fixture + MinIO** catalog.

> [!IMPORTANT]
> **Depends on duckdb/duckdb-avro#105** (implements `copy_to_get_written_statistics`). Until that merges, `extension_config.cmake` temporarily points avro at the PR branch; it should be reverted to the upstream avro repo + merge commit once #105 lands. Note this repo's pre-existing `iceberg_delete.cpp` already *calls* that callback unconditionally, so the implementation is a real dependency, not new coupling.

## MergeAppend (#790)
- First-fit bin-packing aligned with Java `ManifestMergeManager` / PyIceberg `ListPacker.pack_end` (lookback=1, pack-from-tail), grouped by `partition_spec_id`, DATA and DELETE manifests merged independently.
- Merge decision per Java: a bin with a manifest new in this transaction only merges at `commit.manifest.min-count-to-merge`; merged-in committed entries are written `EXISTING` with their original sequence numbers, new `ADDED` entries keep inherited (NULL) sequence numbers (retry-immune).
- `min_sequence_number` and (V3) `first_row_id` are recomputed correctly for merged manifests; V3 new-data manifests are left unmerged for one commit to preserve row-lineage inheritance (converges next commit).
- Empty merged bins are dropped (never write an empty manifest).
- Manifest Avro compression (`write.manifest.compression-codec`, default deflate, matching Java) via an object-store-safe temp-file recompress; forced uncompressed when the catalog forbids client deletes (e.g. S3 Tables).

## Commit retries (#786)
- Optimistic-concurrency retry loop: on conflict, refresh each table, re-`apply` against the new parent, re-submit. Per-table stable snapshot id; sequence number / parent / row-id / manifests regenerated each attempt. Data files are written once at sink and reused across attempts.
- Backoff with **decorrelated jitter**, **`total-timeout-ms` enforcement**, and **`Retry-After`** support. Multi-table commits fold every table's `commit.retry.*` into the most lenient policy.
- Commit-status classification: 409/429 -> CONFLICT, 408/5xx/transport-throw -> UNKNOWN, other -> FATAL. An **UNKNOWN** outcome is resolved by a status-check (look for the stable snapshot id) and **never cleans up files** (avoids deleting a commit that actually landed).
- **`ValidateRetrySafe`** guards delete/overwrite retries (ancestry/rollback, data-file existence, no concurrent new delete on a targeted file) and aborts non-retryable rather than double-applying. Pure appends skip validation.
- Manifest/manifest-list orphans from failed attempts are tracked and cleaned up (respecting `allows_deletes`); data files are never deleted between attempts.
- Snapshot `operation` is classified from the staged change (APPEND/DELETE/OVERWRITE) — converges with #997, merged here.
- Single-table commit refills the table cache from the commit response's `LoadTableResult` (avoids a redundant GET); manifest length comes from the avro write statistics (avoids a read-back HEAD).

## Testing
- **C++ logic suite** `test/cpp/test_merge_retry_logic.cpp` (`make test_logic`, wired into CI): bin-pack edges, merge decision, backoff/decorrelated-jitter/overflow, retry-config parsing, MostLenient, total-timeout, Retry-After (incl. overflow clamp), commit-status classification, ancestry guard, delete attribution (V2 bounds + V3 ref), no-new-delete decision, codec resolution, and a real Avro OCF deflate round-trip (cross-checked against fastavro).
- **Catalog-free SQL**: manifest deflate read path.
- **Catalog-required SQL** (auto-picked up by the catalog CI jobs): merge / merge-disabled / V3-delete-merge / compression / retry-config; a concurrent-delete test driving the `ValidateRetrySafe` FATAL paths; both concurrent tests gated to `LAKEKEEPER_SERVER_AVAILABLE` because the SQLite-backed rest-fixture cannot enforce serializable commits under parallelism (returns 200 to commits that should 409).
- **Fault injection** (mitmproxy addon, fixture CI): `UNKNOWN -> ALL_COMMITTED` (commit landed, response lost) and `UNKNOWN -> NONE_COMMITTED -> retry-to-success` (never reached the catalog).
- **Real catalogs**: full suite green on AWS S3 Tables (CRUD, operation labels, merge, V3 deletes, retry config, 15-/50-way concurrent appends with zero loss, concurrent-delete safety net, cache-refill read-your-own-writes) and on Docker rest-fixture + MinIO (incl. on-disk deflate verification that S3 Tables can't exercise).

See `docs/merge_append_and_retry_design.md` for the full design, the pre-implementation correctness analysis (with inline resolution markers), the real-catalog hardening, and the connection-caching / performance notes.


src/catalog/rest/transaction/iceberg_transaction_data.cpp
```diff
@@ -24,16 +24,61 @@ IcebergTransactionData::IcebergTransactionData(ClientContext &context, const Ice
 	initial_schema_id = table_info.table_metadata.GetCurrentSchemaId();
 }
 
-void IcebergTransactionData::CacheExistingManifestList(lock_guard<mutex> &guard, const IcebergTableMetadata &metadata) {
-	if (!alters.empty()) {
+void IcebergTransactionData::RefreshForRetry(ClientContext &context) {
```
Tishj, line 27: I would imagine this requires a much larger refactor, as all our transaction-data is heavily bound to the state of the metadata it was created against

Don't we need to drop all this data and recreate it from rerunning the undobuffer?

Tishj:
Can you explain why there's a full fledged avro reader+writer in the PR?

src/core/metadata/manifest/iceberg_manifest_list.cpp
```diff
@@ -457,16 +461,32 @@ void WriteToFile(const IcebergTableMetadata &table_metadata, const IcebergManife
...
 
-	auto global_state = copy.copy_to_initialize_global(context, *bind_data, manifest_list.GetPath());
+	auto global_state = copy.copy_to_initialize_global(context, *bind_data, copy_path);
 	auto local_state = copy.copy_to_initialize_local(execution_context, *bind_data);
 
 	copy.copy_to_sink(execution_context, *bind_data, *global_state, *local_state, data);
 	copy.copy_to_combine(execution_context, *bind_data, *global_state, *local_state);
 	copy.copy_to_finalize(context, *bind_data, *global_state);
+
+	if (compress) {
```
Tishj, line 484: You can remove this and make a follow up PR later, I see no reason to include this in this already massive PR

Tishj:
Thanks for your work, the feature(s) sound interesting, but in the current state this is not reviewable.
Please separate these into smaller dedicated PRs and we'll be happy to take a look, thanks for understanding

src/catalog/rest/transaction/iceberg_transaction_data.cpp
Tishj: Just correcting my old comment, we don't need to rerun from the undobuffer.
#1115 is landed for raw INSERT, but delete/update is harder.

PR #1059, opened 2026-06-12
Support write.manifest.compression-codec for manifest writes

cjnoname:
Honors Iceberg's `write.manifest.compression-codec` when writing manifests and manifest lists.

The Avro COPY function this extension writes manifests through only emits the `null` (uncompressed) codec. For a compressing codec, the writer now COPYs to a temporary uncompressed path and recompresses each Avro data block with raw DEFLATE (via DuckDB's vendored miniz) into the final path as a single fresh write, then deletes the temp. Writing a brand-new final object (never reopening/truncating) keeps this object-store safe (S3/Azure/GCS).

## Behavior

- Default matches Java: `gzip` ⇒ Avro `deflate`. `none`/`null` ⇒ uncompressed. Unsupported values throw rather than silently writing an unrequested codec.
- The codec is resolved once per apply and **forced to uncompressed when the catalog forbids client deletes** (e.g. S3 Tables): those catalogs cannot delete the temp and run their own compaction, so uncompressed manifests are fine.
- Uncompressed path takes the manifest length from the COPY's written statistics when available, falling back to a `GetFileSize` probe otherwise.
- The reader already handles DEFLATE-coded manifests (the Iceberg default written by Java/Spark).

## Testing

- Catalog-free manifest-compression **read** test (`iceberg_manifest_compression_read.test`).
- Catalog-backed **write** test (`test_manifest_compression.test`): default/gzip/none codecs round-trip through a real catalog.
- Standalone C++ logic suite (`make test_logic`, wired into CI): codec resolution + Avro OCF deflate round-trip.

Full release build is green; the catalog-free read test and the C++ logic suite pass locally.

### DIFF
--- a/.github/workflows/BuildIceberg.yml
+++ b/.github/workflows/BuildIceberg.yml
@@ -50,6 +50,10 @@ jobs:
         run: |
           make release
 
+      - name: Run C++ logic unit tests (manifest compression)
+        run: |
+          make test_logic
+
       - name: Upload DuckDB-Iceberg no link folder
         uses: actions/upload-artifact@v4
         with:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,7 @@ set(EXTENSION_SOURCES
 	src/core/metadata/iceberg_table_metadata.cpp
 	src/core/metadata/manifest/iceberg_manifest_list.cpp
 	src/core/metadata/manifest/iceberg_manifest.cpp
+	src/core/metadata/manifest/iceberg_avro_codec.cpp
 	src/core/metadata/partition/iceberg_partition_spec.cpp
 	src/core/metadata/schema/iceberg_column_definition.cpp
 	src/core/metadata/schema/iceberg_field_mapping.cpp
--- a/make/util.mk
+++ b/make/util.mk
@@ -15,3 +15,14 @@ define set_active_catalog
 	@mkdir -p $(dir $(ACTIVE_CATALOG_FILE))
 	@echo "$(1)" > $(ACTIVE_CATALOG_FILE)
 endef
+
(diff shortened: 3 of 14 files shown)


Tishj:
As mentioned in #1060 I think this needs to be implemented in `duckdb-avro` (and drags into `duckdb-avro-c` most likely)
All of these codecs exist in `avro-c`, we just need to communicate this through the COPY options

cjnoname:
Agreed — moved the compression out of duckdb-iceberg and into the Avro layer as you suggested. The codecs already exist in avro-c; the writer just couldn't select one on the path this extension uses, so I wired that through instead of recompressing here.

Two PRs:
- **duckdb/duckdb-avro-c#6** — adds `avro_file_writer_create_from_writers_with_metadata_and_codec`. The *from-writers* creation path (the one the COPY writer uses to stream through DuckDB's FileSystem) hard-coded the `null` codec; this lets it set one, like the path/FILE* writers already can. Purely additive — the existing function delegates with `NULL`.
- **duckdb/duckdb-avro#106** — adds a `CODEC` option to `COPY ... TO (FORMAT AVRO)` that passes the codec straight to avro-c (native compression, no post-processing). Accepted codecs follow whatever avro-c was built with. Draft until #6 lands + the vcpkg-duckdb-ports avro-c port is bumped.

Once those land, duckdb-iceberg drops the manifest recompression entirely and just writes with `CODEC 'deflate'` via COPY. I'll strip the compression bits from #1059/#1060 accordingly so those PRs stay focused.

Verified locally: deflate shrinks a 20k-row Avro file ~6x and reads back identical.

Tishj:
Sorry for the radio silence, with #898 merged, you can rewrite this PR targeting `main` and bump the duckdb-avro in `extension_config.cmake` to include https://github.com/duckdb/duckdb-avro/pull/106

Should make the PR a lot smaller 👍

PR #1060, opened 2026-06-12
Implement MergeAppend manifest merging on commit (#790)

cjnoname:
Implements **MergeAppend ([#790](https://github.com/duckdb/duckdb-iceberg/issues/790))**: physically repack the manifest list on commit to bound manifest-file growth, instead of always fast-appending a new manifest per write.

> [!NOTE]
> **Stacked on #1059** (manifest compression). This branch is based on that PR, so the diff here will include its commit until #1059 merges; review the second commit (`Implement MergeAppend manifest merging on commit`) for this PR's changes. The only overlap is that the merge writer passes the already-resolved manifest codec into the manifest write — independent of the compression logic itself.

## What it does

- First-fit bin-packing aligned with Apache Iceberg Java's `ManifestMergeManager` / PyIceberg's `ListPacker.pack_end` (lookback=1, pack from the tail), grouped by `partition_spec_id`, with DATA and DELETE manifests merged independently and never mixed.
- Merge decision matches Java: a bin containing a manifest **new in this transaction** merges only once it reaches `commit.manifest.min-count-to-merge`; merged-in committed entries are written `EXISTING` with their original sequence numbers, while new `ADDED` entries keep inherited (NULL) sequence numbers.
- `min_sequence_number` and (V3) `first_row_id` are recomputed for merged manifests; new V3 data manifests are left unmerged for one commit to preserve row-lineage inheritance and converge on the next commit.
- Empty merged bins are dropped so an empty manifest is never written.
- Controlled by `commit.manifest-merge.enabled` / `commit.manifest.min-count-to-merge` / `commit.manifest.target-size-bytes` (the Iceberg property names).

The per-manifest scan helper used by both the existing delete-rewrite path and the merge path is unified into `ScanManifestEntries`.

## Testing

- Catalog-backed `test_merge_append` / `test_merge_append_disabled` / `test_merge_append_v3_deletes` SQL tests.
- Standalone C++ logic suite (`make test_logic`, in CI): bin-packing edges + merge decision.

Full release build is green; the C++ logic suite and the catalog-free read test pass locally.


src/core/metadata/manifest/iceberg_manifest_list.cpp
```diff
@@ -307,7 +311,7 @@ static Value FieldSummaryFieldIds() {
 }
 
 void WriteToFile(const IcebergTableMetadata &table_metadata, const IcebergManifestList &manifest_list,
-                 CopyFunction &copy, DatabaseInstance &db, ClientContext &context) {
+                 CopyFunction &copy, DatabaseInstance &db, ClientContext &context, const string &avro_codec) {
```
Tishj, line 314: Either the `avro_codec` can be forwarded to duckdb-avro in a future PR (my preference), or we can implement this compression in duckdb-iceberg in a future PR.

I see no reason to include this in this PR.

Tishj:
I'd like to see some transaction tests, what happens when we are performing multiple inserts in the same transaction?

cjnoname:
Added tests covering this: `test_merge_append_same_transaction.test` (+ a merge-disabled control). Since each in-transaction write produces its own snapshot, a transaction assembles N chained snapshots committed atomically — the tests exercise merge-on-commit across them:

- multiple inserts in one transaction → read-your-own-writes, then the per-insert data manifests merge into one on COMMIT (with EXISTING-status entries asserted)
- ROLLBACK of a multi-insert transaction discards everything atomically
- insert + delete of the just-inserted rows in the same transaction (a DATA snapshot followed by a DELETE snapshot in one commit)
- many small single-row inserts in one transaction, merged to a single manifest
- control: same multi-insert transaction with `commit.manifest-merge.enabled=false` keeps the manifests separate, data still correct

I ran these locally against the apache/iceberg-rest-fixture catalog and they pass (70 + 22 assertions). So multiple inserts in the same transaction behave correctly with MergeAppend.

.github/workflows/BuildIceberg.yml
```diff
@@ -50,6 +50,10 @@ jobs:
         run: |
           make release
 
+      - name: Run C++ logic unit tests (manifest compression + merge)
```
Tishj: Please remove this, we only do e2e tests, no c++ unit tests.

make/util.mk
```diff
@@ -14,4 +14,17 @@ endef
...
\ No newline at end of file
+endef
+
+# Standalone C++ logic tests. Dependency-free (the compression suite uses zlib for the Avro OCF
+# deflate round-trip), so they build and run without a catalog or the DuckDB unittest harness. They
+# cover codec resolution / OCF recompress and the manifest bin-packing / merge-decision logic the SQL
+# integration tests cannot reach. Run with `make test_logic`.
+CXX ?= c++
+.PHONY: test_logic
+test_logic:
```
Tishj: https://github.com/duckdb/duckdb-iceberg/pull/1060/changes#r3407910543

test/cpp/test_compression_logic.cpp
```diff
@@ -0,0 +1,388 @@
+// Standalone unit tests for the pure-logic pieces of the manifest Avro compression implementation.
```
Tishj: https://github.com/duckdb/duckdb-iceberg/pull/1060/changes#r3407910543

test/cpp/test_merge_logic.cpp
```diff
@@ -0,0 +1,231 @@
+// Standalone unit tests for the pure-logic pieces of the MergeAppend manifest bin-packing
```
Tishj: https://github.com/duckdb/duckdb-iceberg/pull/1060/changes#r3407910543

cjnoname:
Thanks — addressed both:

**C++ unit tests** — removed. Deleted `test/cpp/test_compression_logic.cpp` and `test/cpp/test_merge_logic.cpp`, the `make test_logic` target, and the CI step that ran them. MergeAppend is covered by the e2e SQL tests instead: `test_merge_append.test`, `test_merge_append_disabled.test`, `test_merge_append_v3_deletes.test`, plus the same-transaction ones. (commit feb01d12)

**Manifest compression (`avro_codec`)** — agreed it doesn't belong here. It's actually part of the separate compression PR #1059; this MergeAppend branch is *stacked on top of* #1059, which is why the codec changes show up in this PR's diff (the PR base has to be `v1.5-variegata` since I can't push the intermediate branch to this repo). The MergeAppend commit itself (`b007c704`) touches no compression code — please review that commit for this PR's actual changes. Once #1059 merges, the codec diff drops out of here automatically.

Separately, per your earlier point, I've moved the compression *implementation* into the Avro layer as you suggested rather than post-processing in this extension — duckdb/duckdb-avro-c#6 + duckdb/duckdb-avro#106 add a native `CODEC` option, and #1059 will use that instead of recompressing manifests here.

cjnoname:
Closing in favor of a clean rewrite targeting main (this PR was based on the old v1.5-variegata line and stacked under the compression PR). The MergeAppend implementation is reopened as a fresh PR against main.

PR #1061, opened 2026-06-12
Support optimistic-concurrency commit retries (#786)

cjnoname:
Implements **optimistic-concurrency commit retries ([#786](https://github.com/duckdb/duckdb-iceberg/issues/786))**: on a commit conflict, refresh each table, re-apply the staged changes against the new parent, and re-submit — instead of failing the transaction.

> [!NOTE]
> **Stacked on #1059 (compression) and #1060 (MergeAppend).** This branch is based on those, so the diff includes their commits until they merge; review the third commit (`Support optimistic-concurrency commit retries`) for this PR's changes. The retry loop re-runs `ApplyTableChanges` (the pure core extracted here), which internally calls the merge step from #1060 — but retry itself does not depend on merge being enabled.

## What it does

- **Retry loop** over the whole transaction commit: on conflict, each table's metadata is reloaded, manifests/snapshots are regenerated against the refreshed parent, and the commit is re-submitted. Data files are written once at sink and reused across attempts; only metadata is regenerated.
- **Backoff** with decorrelated jitter, `commit.retry.total-timeout-ms` enforcement, and `Retry-After` support. Multi-table commits fold every table's `commit.retry.*` into the most lenient policy.
- **Commit-status classification**: 409/429 → CONFLICT, 408/5xx/transport-throw → UNKNOWN, other → FATAL. An **UNKNOWN** outcome is resolved by a status-check that looks for this commit's stable snapshot id in the refreshed table, and **never cleans up files** (so a commit that actually landed is never corrupted).
- **`ValidateRetrySafe`** guards delete/overwrite retries (ancestry/rollback check, data-file existence, no concurrent new delete on a targeted file) and aborts non-retryable conflicts rather than double-applying. Pure appends skip it.
- Manifest / manifest-list files written by a failed or discarded attempt are tracked and cleaned up (subject to the catalog allowing client deletes); data files are never deleted between attempts.
- `CreateUpdate`'s body becomes a pure, repeatable `ApplyTableChanges`; the snapshot id is generated once and kept stable across attempts so the status-check can find it. Single-table commit refills the table cache from the commit response's `LoadTableResult`.

Controlled by `commit.retry.num-retries` / `min-wait-ms` / `max-wait-ms` / `total-timeout-ms` (the Iceberg property names).

## Testing

- Catalog-backed retry-config + concurrent-append / concurrent-delete SQL tests.
- mitmproxy **fault injection** for the UNKNOWN status-check: commit landed but response lost (ALL_COMMITTED), and never-reached-catalog retry-to-success (NONE_COMMITTED).
- Standalone C++ logic suite (`make test_logic`, in CI): backoff/jitter, status classification, Retry-After parsing, config folding, total-timeout, ancestry walk, delete attribution.

Full release build is green; the three C++ logic suites and the catalog-free read test pass locally.

### DIFF
--- a/.github/workflows/BuildIceberg.yml
+++ b/.github/workflows/BuildIceberg.yml
@@ -50,7 +50,7 @@ jobs:
         run: |
           make release
 
-      - name: Run C++ logic unit tests (manifest compression)
+      - name: Run C++ logic unit tests (manifest compression + merge)
         run: |
           make test_logic
 
--- a/.github/workflows/BuildIceberg.yml
+++ b/.github/workflows/BuildIceberg.yml
@@ -50,6 +50,10 @@ jobs:
         run: |
           make release
 
+      - name: Run C++ logic unit tests (manifest compression)
+        run: |
+          make test_logic
+
       - name: Upload DuckDB-Iceberg no link folder
         uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/BuildIceberg.yml
+++ b/.github/workflows/BuildIceberg.yml
@@ -50,7 +50,7 @@ jobs:
         run: |
           make release
 
-      - name: Run C++ logic unit tests (manifest compression + merge)
+      - name: Run C++ logic unit tests (manifest compression + merge + commit retry)
         run: |
           make test_logic
 
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ __pycache__/
 
 tmp/
 .catalogs/
+.cache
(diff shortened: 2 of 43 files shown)


cjnoname:
Closing — superseded. This PR's append/INSERT commit-retry is now covered by @Tishj's #1115 (merged), and the exponential-backoff-with-jitter + total-timeout piece is split out into #1122. What remains unique here (delete/overwrite retry with ValidateRetrySafe, and Retry-After handling) is best done as a focused follow-up on top of #1115/#1122 rather than as this large branch. Closing to avoid a stale, conflicting PR.

PR #1068, opened 2026-06-14
Populate snapshot-summary delete-file metrics

raghav-reglobe:
## Problem

When a snapshot adds deletion vectors / delete files, duckdb-iceberg writes the delete
manifests correctly but does **not** populate the standard Iceberg snapshot-summary
delete-file counters. `IcebergSnapshotMetricType` has no delete-file metric types at all, and
`IcebergSnapshotMetrics::AddManifestFile` counts every manifest — data *and* delete — into the
data-file metrics.

Result: a snapshot that adds DVs reports `added-data-files` (slightly inflated by the delete
manifest) while `added-delete-files` / `total-delete-files` / `added-position-deletes` /
`total-position-deletes` are entirely absent from the summary. Any tooling that reads those
standard counters (compaction schedulers, maintenance, monitoring) is blind to the table's DV
pressure even though the DVs are physically present in the manifests.

Verified against a Polaris REST catalog: a table with 108 live DVs in its manifests had `None`
for every delete counter across all snapshot summaries, while a Spark-written snapshot on the
same table correctly reported `total-delete-files=0`.

## Fix

- Add the standard delete metric types: `ADDED_DELETE_FILES`, `ADDED_POSITION_DELETES`,
  `TOTAL_DELETE_FILES`, `TOTAL_POSITION_DELETES`.
- Branch `AddManifestFile` on `manifest_file.content`: DELETE manifests accumulate into the
  delete metrics (mirroring the existing data-file accounting) instead of being mis-counted as
  data files. This also corrects `added-data-files` / `total-data-files`, previously inflated by
  delete manifests. The delete manifest's `added_files_count` (= number of DV/delete files) maps
  to `added-delete-files`; `added_rows_count` (= deleted positions) maps to `added-position-deletes`.
- Seed + carry-forward the new totals (`EmptyMetrics`, copy constructor) and register the
  metric-key strings so they round-trip through the summary.

## Testing

`iceberg_snapshots()` doesn't surface the summary `additional_properties`, so a SQL-logic
assertion isn't directly possible. Verified manually end-to-end against a Polaris REST catalog
(write a DV → read the snapshot summary → delete counters now present). Happy to add a test if
you can point me at the right harness (extend `iceberg_snapshots()` to expose the summary, or a
python metadata test).


src/core/metadata/snapshot/iceberg_snapshot.cpp
```diff
@@ -34,9 +36,48 @@ IcebergSnapshotMetrics::IcebergSnapshotMetrics(const IcebergSnapshot &snapshot)
...
 void IcebergSnapshotMetrics::AddManifestFile(const IcebergManifestFile &manifest_file) {
+	if (manifest_file.content == IcebergManifestContentType::DELETE) {
+		//! Delete manifest (V2 position/equality delete files, V3 Puffin deletion vectors).
+		//! Account for it under the Iceberg snapshot-summary delete metrics, mirroring the
+		//! data-file accounting below — otherwise added DVs are silently invisible to any
+		//! tooling that reads `total-delete-files` / `total-position-deletes` (e.g. compaction
+		//! schedulers), and are erroneously counted as added data files.
+		metrics.emplace(IcebergSnapshotMetricType::ADDED_DELETE_FILES, 0).first->second +=
+		    manifest_file.added_files_count;
+		metrics.emplace(IcebergSnapshotMetricType::ADDED_POSITION_DELETES, 0).first->second +=
```
Tishj, line 65: Thanks, this looks good to me, I just have one small note about assuming positional deletes.
Please wrap this in a `ifndef ICEBERG_ENABLE_EQUALITY_DELETE_WRITES` block

src/core/metadata/snapshot/iceberg_snapshot.cpp
```diff
@@ -34,9 +36,48 @@ IcebergSnapshotMetrics::IcebergSnapshotMetrics(const IcebergSnapshot &snapshot)
...
 }
 
 void IcebergSnapshotMetrics::AddManifestFile(const IcebergManifestFile &manifest_file) {
+	if (manifest_file.content == IcebergManifestContentType::DELETE) {
+		//! Delete manifest (V2 position/equality delete files, V3 Puffin deletion vectors).
+		//! Account for it under the Iceberg snapshot-summary delete metrics, mirroring the
+		//! data-file accounting below — otherwise added DVs are silently invisible to any
+		//! tooling that reads `total-delete-files` / `total-position-deletes` (e.g. compaction
+		//! schedulers), and are erroneously counted as added data files.
+		metrics.emplace(IcebergSnapshotMetricType::ADDED_DELETE_FILES, 0).first->second +=
```
Tishj, line 56: Please also add `removed-delete-files` and `removed-position-delete-files`

raghav-reglobe:
Thanks for the review @Tishj! Both addressed in the latest commit:

- Added `removed-delete-files` and `removed-position-delete-files`.
- Wrapped the position-delete-specific accounting (`added-position-deletes`, `removed-position-delete-files`, `total-position-deletes`) in `#ifndef ICEBERG_ENABLE_EQUALITY_DELETE_WRITES` so equality deletes won't be mislabeled as positional once equality-delete writes land. The `*-delete-files` file counts stay unconditional.

Tmonster (commented):
Thanks! We have testing infrastructure where you can query the `load_table_information` response, which should give you access to the summary
Can you add a test that verifies this information is now added in your PR? 

I think you can access the information with  `select * from iceberg_load_table_response(my_datalake.default.tbl1)`

raghav-reglobe:
All the catalog jobs are green now. One correction to my note above — after CI surfaced a few catalog-portability issues, the test landed a bit differently than I first described:

- **Default (v2) format-version**, not v3. The delete-file metric accounting is identical for v2/v3, and v2 deletes are supported on every configured catalog (the Nessie fixture rejects v3 creates).
- Asserts the **unconditional** metrics: `added-delete-files` / `total-delete-files` (=1) on the delete snapshot, `total-delete-files`=0 on the append snapshot, and `total-data-files`=1 on the delete snapshot (the delete manifest is no longer mis-counted as a data file).
- Does **not** assert `added-position-deletes` / `total-position-deletes` — those are compiled out under `ICEBERG_ENABLE_EQUALITY_DELETE_WRITES`, so asserting them failed the equality-deletes build. The delete-file counts above are emitted in every build.
- Reads the summaries via `iceberg_load_table_response` and selects snapshots by `summary.operation` (not array position — `metadata.snapshots` order isn't consistent across catalogs).
- Skipped on Nessie, whose fixture doesn't expose snapshot history — the same limitation under which `test_snapshot_operation_type.test` is already skipped there.

Tishj:
Thanks!

PR #1076, opened 2026-06-15
[V3] Add `UNKNOWN` type

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/9696


Tmonster:
Thanks, I know this is probably really hard to set up, but do you know what happens if some other engine registers a table with an UNKNOWN type and we insert into it? I feel like I tried to do this once and something went wrong with parquet writing a SQLNULL type

PR #1079, opened 2026-06-16
Fix scan failure after schema-reducing alter

NiclasHaderer:
A schema-reducing alter (DROP COLUMN/DROP DEFAULT) can revert the current schema to an earlier, identical schema whose id is reused. Snapshots written under the higher id then exceed the current schema id, and the scan path rejected this valid state with an internal error. Allow it when scanning with the table's current schema (iceberg projects by field id);


Tishj:
Thanks, that indeed looks like a leftover piece of logic before we implemented ALTER and realized that making an alteration that matches an existing schema reuses that schema-id rather than create a new one

Tishj:
Though I am wary of accepting this change, because it's located in `IcebergTableEntry::GetScanFunction`, at this point in the process, the schema is already locked (schema is associated with an `IcebergTableEntry`)

I wonder if we can fix this in `IcebergTableSet::GetEntry`, as that is where the schema is selected.

PR #1080, opened 2026-06-16
Support other iceberg_scan implementations

JelteF:
It's possible for other iceberg_scan functions to exist. This makes sure we don't crash in those cases by additionally checking if the `get_multi_file_reader` field contains a pointer to our multi file reader.


Tmonster (approved):
Thanks!
Just waiting for CI to pass

PR #1081, opened 2026-06-16
Fix cloud tests

Tmonster:


Tmonster:
This just modifies CI, and I want to test some cloud tests before the release, so merging with Code quality checked

PR #1082, opened 2026-06-16
Fix operator visitor pattern in pre-optimizer

JelteF:
The pre-optimizer would use `return` to early exit if the first child of an operator did not match. There's no guarantee though that an iceberg table with equality deletes is the first child of e.g. a join operator. The included test would trigger a segfault without this change.


Tmonster (approved):
Thanks!
Just waiting for CI to pass

PR #1083, opened 2026-06-17
Bump duckdb 1.6 submodule again

Tishj:
This PR supersedes #856 

Merge in a patch to the MultiFileReader required for filtering on the virtual `_row_id` column


Tishj:
`INTERNAL Error: ManagedAsyncWriteQueue still tracks external pending writes`
I think this means we are not flushing some state somehow o.o

PR #1088, opened 2026-06-19
Cast CTAS output to Iceberg storage types

NiclasHaderer:
CTAS derives its columns from the SELECT output types, but the table (and its write pipeline) is created with the Iceberg storage types. These differ for types that have no Iceberg equivalent (e.g. HUGEINT).

The CTAS write appends the SELECT chunks straight into a ColumnDataCollection typed with the storage types, so the types disagree (HUGEINT vs DECIMAL(38,0)). This trips D_ASSERT(types == input_types) in assertion builds; in release it is masked only because HUGEINT and DECIMAL(38,0) share the INT128 layout. 

**Fix:**
Insert a projection above the CTAS plan that casts each column to its Iceberg storage type before the write. The storage types come from IcebergCreateTableRequest::CreateIcebergSchema (the same source the table is created from), and the cast reuses BoundCastExpression::AddCastToType so nested types are handled recursively. The projection is skipped when no column needs a cast.

### DIFF
--- a/src/execution/operator/iceberg_insert.cpp
+++ b/src/execution/operator/iceberg_insert.cpp
@@ -12,6 +12,7 @@
 #include "duckdb/function/function_binder.hpp"
 #include "duckdb/planner/expression/bound_reference_expression.hpp"
 #include "duckdb/planner/expression/bound_constant_expression.hpp"
+#include "duckdb/planner/expression/bound_cast_expression.hpp"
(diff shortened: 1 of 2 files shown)


Tishj:
Thanks!

NiclasHaderer:
As an aside, should I start targeting main with my prs?

Issue #746, opened 2026-02-26
Feature Request: Dynamic filtering of files for Top-N queries using bounds


renaudb:
When dealing with event datasets, where the data is roughly sorted by time descending, Top-N queries like:

```
SELECT
  *
FROM
  iceberg_scan(...)
WHERE
  timestamp BETWEEN {start_time} AND {end_time}
ORDER BY
  timestamp DESC
LIMIT 10
```

Grow slower and slower as the query time range expands with more files being read.

However, using dynamic filtering and bounds data contained in Iceberg manifests, it should be possible to sort the files by upper bound and read only a couple files before filtering all the other ones out.

Something similar was implemented for [ducklake](https://github.com/duckdb/ducklake/pull/668). I'm relatively new to the duckdb code, but it looks like `IcebergMultiFileList` already implements the dynamic filter pushdown logic, so it might just be a matter for updating `InitializeFiles` to sort manifest files by upper/lower bound based on the existence of a dynamic filter. Please let me know if there's a better approach.

tomlarkworthy:
There is a `set_scan_order` setting for MultiFileList, I have a prototype that works well querying the latest rows of our timeseries

PR #1089, opened 2026-06-20
feat: Order iceberg_scan and prune matching

tomlarkworthy:
Addresses duckdb/duckdb-iceberg#746: Top-N queries over time-sorted data (e.g. ORDER BY ts DESC LIMIT N) read every file before the limit filters them out.

Uses DuckDB core's set_scan_order table-function callback (used by native table_scan). When a LIMIT is present over an ORDER BY on a numeric/temporal column, reorder the data files by the order column's per-file bounds and drop files that cannot contain a top-N row, so the scan touches only the relevant files. 

When a WHERE is present the ordering also helps choose a more efficient order for expansion (done by existing core optimizers).

A fair amount of this was generated by AI. Let me know if this is helpful or not (this is quite a good win though for timeseries usecases)

### DIFF
--- a/src/planning/iceberg_multi_file_list.cpp
+++ b/src/planning/iceberg_multi_file_list.cpp
@@ -3,6 +3,7 @@
 #include "duckdb/catalog/catalog_entry/table_function_catalog_entry.hpp"
 #include "duckdb/common/exception.hpp"
 #include "duckdb/common/file_system.hpp"
+#include "duckdb/common/numeric_utils.hpp"
 #include "duckdb/execution/execution_context.hpp"
 #include "duckdb/parallel/thread_context.hpp"
 #include "duckdb/parallel/event.hpp"
@@ -18,6 +19,7 @@
 #include "duckdb/execution/executor.hpp"
 #include "duckdb/optimizer/filter_combiner.hpp"
 #include "duckdb/function/scalar/struct_utils.hpp"
+#include "duckdb/storage/table/row_group_reorderer.hpp"
 
 #include "planning/iceberg_multi_file_reader.hpp"
 #include "function/iceberg_functions.hpp"
@@ -36,6 +38,8 @@
 #include "planning/metadata_io/manifest/iceberg_manifest_reader.hpp"
 #include "planning/metadata_io/manifest_list/iceberg_manifest_list_reader.hpp"
 #include "planning/metadata_io/manifest_list/bound_iceberg_manifest_list_entry.hpp"
+
+#include <algorithm>
 
 namespace duckdb {
 
@@ -296,6 +300,11 @@ void IcebergMultiFileList::SetOptions(const IcebergOptions &options) {
 	shared_state->options = options;
 }
 
+void IcebergMultiFileList::SetScanOrder(unique_ptr<RowGroupOrderOptions> options) {
+	scan_order_options = std::move(options);
+	scan_order_applied = false;
+}
+
 unique_ptr<ExpressionFilter> IcebergMultiFileList::GetFilterForColumnIndex(const IcebergTableFilters &filter_set,
                                                                            const ColumnIndex &column_index) const {
 	auto primary_index = column_index.GetPrimaryIndex();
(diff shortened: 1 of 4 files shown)


Tmonster (commented):
There are only happy path tests. 
Can we also check what happens if we run a large insert and the values we eventually order on are not ordered?

i.e 
```
create table t1 (a int);
insert into t1 select (range%60) a from range(100); -- file 1, lower bound 0
insert into t1 select (range%60) + 5 a from range(100); -- file 2, lower bound 5
insert into t1 select (range%60) + 2 a from range(100); -- file 3, lower bound 2
insert into t1 select (range%60) + 3 a from range(100); -- file 4, lower bound 3
insert into t1 select (range%60) + 1 a from range(100); -- file 4, lower bound 1

select a from t1 order by a asc limit 10;
```

Here the lower bounds for each file are scattered a bit more, and I wonder how the logic reacts to this. I would expect every file to still be scanned because there is no guarantee there are multiple instances of the lower bound per file.

Tmonster:
Also, can you retarget main for this?

tomlarkworthy:
ok added a new test for equally overlapping files, it scanned everything but still got the right answer (pruning is not breaking correctness). 
Also added a test for mostly overlapping files, it scanned *less* than all of them, indicating the pruning has value in realistic scenarios when things are not perfectly mutually exclusive ranges.

rebased on main. Thanks for checking it!

Tishj:
I understand the changes, I'm just worried that this introduces full metadata materialization
So I'm leaning on the side of not merging this

EDIT:
I took a better look, it runs after pushdown, we would end up materializing anyways, only now we're doing it eagerly as opposed to lazily.

That's fine actually, if there's a limit anyways and an order by, that sounds like the right call

src/planning/iceberg_multi_file_list.cpp
```diff
@@ -393,6 +402,9 @@ unique_ptr<IcebergMultiFileList> IcebergMultiFileList::PushdownInternal(ClientCo
 	filtered_list->names = names;
 	filtered_list->types = types;
 	filtered_list->have_bound = true;
+	if (scan_order_options) {
```
Tishj, line 405: nit: I wonder if this should live in the `IcebergMultiFileListSharedState`, but I'm not sure, since filters are also kept outside of this.

Actually.. probably not, let's keep it as is

src/planning/iceberg_multi_file_list.cpp
```diff
@@ -965,10 +977,143 @@ optional_ptr<const BoundIcebergManifestEntry> IcebergMultiFileList::GetDataFile(
...
+		}
+		//! lower/upper bounds are stored as raw Iceberg-encoded blobs; decode to typed Values before comparing.
+		auto stats = IcebergPredicateStats::DeserializeBounds(lower_it->second, upper_it->second, order_column.name,
+		                                                      order_column.type);
+		if (!stats.has_lower_bounds || !stats.has_upper_bounds || stats.lower_bound.IsNull() ||
+		    stats.upper_bound.IsNull()) {
+			return;
+		}
+		auto null_it = data_file.null_value_counts.find(field_id);
+		if (null_it != data_file.null_value_counts.end() && null_it->second > 0) {
```
Tishj: `if (null_it == data_file.null_value_counts.end() || null_it->second > 0)` ?
I thought we couldn't make any assumptions about there being nulls or not if it was omitted

Tishj (changes requested):
I have one request regarding the null counts, but other than that I think it's ready to merge

src/planning/iceberg_multi_file_list.cpp
tomlarkworthy: yeah good catch

tomlarkworthy:
> I understand the changes, I'm just worried that this introduces full metadata materialization So I'm leaning on the side of not merging this
> 
> EDIT: I took a better look, it runs after pushdown, we would end up materializing anyways, only now we're doing it eagerly as opposed to lazily.
> 
> That's fine actually, if there's a limit anyways and an order by, that sounds like the right call

Yes exactly, we only cut down on IO work or perform the IO work in a better order.

Tishj (approved):
Thanks!

tomlarkworthy:
FWIW we have seen incredible performance improvement on top-k queries since we put this in prod.

PR #1097, opened 2026-06-22
Write delete manifests with content='deletes' file metadata

raghav-reglobe:
## Problem

`WriteToFile` in `src/core/metadata/manifest/iceberg_manifest.cpp` hardcodes the Avro
file-metadata `content` key to `"data"` for *every* manifest it writes, including delete
manifests:

```cpp
metadata_values.emplace_back("content", "data");
```

Per the Iceberg spec, a manifest's `content` metadata must reflect what it holds — `"data"`
for a data manifest, `"deletes"` for a delete manifest. So a v3 `DELETE` (which writes a
deletion-vector delete manifest) produces an Avro file whose `content` metadata says
`data`, contradicting the manifest list entry, which correctly records it as a delete
manifest.

Readers that take a manifest's content from the manifest *list* entry tolerate this (so it
does not affect query results in duckdb-iceberg itself), but spec-strict readers cross-check
the manifest file's own `content` metadata against the list entry and reject the mismatch.
For example, iceberg-go's `ReadManifest` errors with:

```
manifest file's 'content' metadata indicates "data", but entry from manifest list indicates "deletes"
```

## Fix

Derive the value from the manifest's content type (the `IcebergManifestFile` the function
already receives):

```cpp
metadata_values.emplace_back("content",
                             manifest_file.content == IcebergManifestContentType::DATA ? "data" : "deletes");
```

## Testing

`test/sql/local/catalog_test_config_setup/catalog_agnostic/delete/test_delete_manifest_content_type.test`:
after a v3 `DELETE`, resolves the delete manifest via `iceberg_metadata` and asserts its
Avro file metadata carries `content='deletes'` (and not the mislabeled `content='data'`).

Found alongside #1095 — both are v3 deletion-vector write-correctness fixes.

### DIFF
--- a/src/core/metadata/manifest/iceberg_manifest.cpp
+++ b/src/core/metadata/manifest/iceberg_manifest.cpp
@@ -648,9 +648,6 @@ idx_t WriteToFile(const IcebergTableMetadata &table_metadata, const IcebergManif
 	metadata_values.emplace_back("partition-spec", current_partition_spec.FieldsToJSONString());
 	metadata_values.emplace_back("partition-spec-id", std::to_string(current_partition_spec.spec_id));
 	metadata_values.emplace_back("format-version", std::to_string(table_metadata.iceberg_version));
-	// The 'content' metadata must reflect what the manifest holds ("data" or "deletes"), not be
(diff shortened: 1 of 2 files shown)


Tishj (approved):
Thanks 👍

PR #1102, opened 2026-06-22
Fix cloud testing workflows on main

The-Alchemist:
## Summary

- Fixes workflow validation failures on `main` caused by duplicate `cloud-tests-main` job definitions in `CloudTestingScheduled.yml` and `CloudTestingManual.yaml` (introduced in ff1bc054).
- Keeps `main` as the default DuckDB and duckdb-iceberg ref for manual and scheduled cloud tests on the `main` branch.
- Properly wires `duckdb_iceberg_ref` through the reusable workflow: it is now declared as a `workflow_call` input, used at checkout, and exposed as a manual dispatch input so cloud tests can target a specific duckdb-iceberg branch.

## Context

Since ff1bc054, scheduled and manual cloud test workflows fail immediately at parse time:

```
'cloud-tests-main' is already defined
```

The reusable workflow also referenced `inputs.duckdb_iceberg_ref` without declaring it as a `workflow_call` input, so the passthrough never worked. This PR removes the duplicate jobs and restores the input correctly:

### DIFF
--- a/.github/workflows/CloudTestingManual.yaml
+++ b/.github/workflows/CloudTestingManual.yaml
@@ -5,11 +5,7 @@ on:
       duckdb_ref:
         description: 'DuckDB branch name or commit hash'
         required: false
-        default: 'main'
-      duckdb_iceberg_ref:
-        description: 'duckdb-iceberg branch name or commit hash'
-        required: false
-        default: 'main'
+        default: 'v1.5-variegata'
       httpfs_fork:
         description: 'URL of a duckdb-httpfs fork. Leave empty to use default duckdb/duckdb-httpfs.'
         required: false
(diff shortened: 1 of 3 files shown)


.github/workflows/CloudTestingManual.yaml
```diff
@@ -5,11 +5,7 @@ on:
       duckdb_ref:
         description: 'DuckDB branch name or commit hash'
         required: false
-        default: 'main'
-      duckdb_iceberg_ref:
-        description: 'duckdb-iceberg branch name or commit hash'
-        required: false
-        default: 'main'
+        default: 'v1.5-variegata'
```
Tishj: This looks wrong, shouldn't this be `main`?

The-Alchemist: good catch.  i also brought back `duckdb_iceberg_ref` in the latest commits. WDYT?

Tmonster (approved):
Thanks!

PR #1116, opened 2026-06-26
Support write.manifest.compression-codec via native Avro CODEC

cjnoname:
Rewrite of #1059, targeting `main` as suggested by @Tishj — now much smaller.

## What

Manifest and manifest-list files honor the `write.manifest.compression-codec` table property (default `deflate`, matching Java's `TableProperties` default).

## How

Instead of post-processing the written file, the codec is handed to the Avro COPY writer's `CODEC` option (duckdb/duckdb-avro#106), so the object container is compressed **natively in a single write** — no temp file, no recompression pass. This is the approach @Tishj asked for (do it in the Avro layer, communicate the codec through COPY).

- `ResolveAvroCodec` maps the property (`gzip`/`deflate` → `deflate`, `none`/`null`/`uncompressed` → `null`) and rejects unsupported values.
- `WriteToFile` (manifest + manifest-list) resolves the codec from the table metadata and sets the COPY `codec` option for a compressing codec; `null` is the COPY default so it is left unset. No call-site signature changes.
- Bumps `duckdb-avro` in `extension_config.cmake` to include the `CODEC` option (duckdb/duckdb-avro#106, merged).

## Testing

`test_manifest_compression.test` against the fixture REST catalog:
- default / explicit `deflate`, `none`, `gzip` alias, and an unsupported-codec error
- byte-level proof the manifest header records `avro.codec` (`deflate` vs `null`)
- `read_avro` round-trip of the compressed manifest
- per-data-file statistics (lower/upper bounds, record_count) read back intact from a compressed manifest

Verified locally that self-contained insert tests (big insert, counts, supplied columns, snapshot ops, transactional v3) all still pass with compression on, i.e. no regression in the write path. (The Spark-data-dependent tests need `make fixture-data`, which can't run here due to an unrelated local Python 3.14 / pydantic-core build issue.)

Supersedes #1059.

### DIFF
--- a/test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_manifest_compression.test
+++ b/test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_manifest_compression.test
@@ -1,7 +1,8 @@
 # name: test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_manifest_compression.test
 # description: write.manifest.compression-codec - manifests are written with the configured Avro
-#              codec (default deflate, matching Java), applied natively by the Avro COPY writer.
 # group: [insert]
+
+#              codec (default deflate, matching Java), applied natively by the Avro COPY writer.
 
 require-env CATALOG_TEST_CONFIG_SETUP
 
--- a/test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_manifest_compression.test
+++ b/test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_manifest_compression.test
@@ -183,32 +183,3 @@ SELECT record_count FROM (SELECT unnest(data_file) FROM read_avro(getvariable('m
 
 statement ok
 drop table my_datalake.default.compress_bounds;
-
-# --- v3 table with compression ----------------------------------------------------------------
-statement ok
-drop table if exists my_datalake.default.compress_v3;
-
-statement ok
-create table my_datalake.default.compress_v3 (id INTEGER, data VARCHAR) WITH ('format-version' = 3, 'write.manifest.compression-codec' = 'deflate');
-
-statement ok
-insert into my_datalake.default.compress_v3 select i, 'v' from range(300) t(i);
-
-query I
-select count(*) from my_datalake.default.compress_v3;
-----
-300
-
-statement ok
-SET VARIABLE manifest_v3 = (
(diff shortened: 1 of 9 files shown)


cjnoname:
@Tishj this is the rewrite of #1059 you asked for — targeting `main` and using the native Avro `CODEC` option (duckdb/duckdb-avro#106, merged) instead of post-processing the manifest. The diff is much smaller now: manifests/manifest-lists just pass the resolved codec through to the COPY writer, and the old miniz recompression is gone.

Tested locally against the fixture REST catalog (default/explicit deflate, none, gzip alias, unsupported-codec error, byte-level check of the avro.codec header, read_avro round-trip of the compressed manifest, per-data-file stats intact, and a v3 table). Would appreciate a review when you have a chance 🙏

test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/test_manifest_compression.test
```diff
@@ -0,0 +1,214 @@
...
+
+statement ok
+drop table my_datalake.default.compress_bounds;
+
+# --- v3 table with compression ----------------------------------------------------------------
+statement ok
+drop table if exists my_datalake.default.compress_v3;
+
+statement ok
+create table my_datalake.default.compress_v3 (id INTEGER, data VARCHAR) WITH ('format-version' = 3, 'write.manifest.compression-codec' = 'deflate');
```
Tishj: I can already tell that you want to separate this test into its own file
Not every catalog supports format v3, and this will have to be skipped on those

cjnoname:
Done — split the v3 case into its own `test_manifest_compression_v3.test` and added it to the Nessie skip list (Nessie doesn't support v3 yet), matching how the other v3 tests are handled. The main `test_manifest_compression.test` is now v2-only so it runs on every catalog. (Kept your format commit.)

Tishj:
Formatting is broken again

cjnoname:
Fixed — the v3 test's `# description:` was wrapped across two lines, which doesn't pass `format.py` (the test formatter expects a single-line description). Collapsed it to one line; both `test_manifest_compression.test` and `test_manifest_compression_v3.test` now pass the format check (ran `scripts/format.py` locally). The only remaining diff format.py reports is in `test/python/test_insert_end_to_end.py`, which this PR doesn't touch — that's a black-version artifact on my machine, not part of this PR.

Tishj:
Thanks!

PR #1117, opened 2026-06-26
[Dev] Clean up relation between `IcebergTransactionData` and `IcebergTableInformation`

Tishj:
This PR is related to #854 

We create a cleaner relationship between committed-state (IcebergTableInformation) and transaction-local-state (IcebergTransactionData)

### Old Relationship

`IcebergTableInformation` owns a `unique_ptr<IcebergTransactionData>` which holds the transaction-local state, if any.
Besides that, it holds the table metadata and the schema versions materialized as `IcebergTableEntry` items.
Whenever a transaction makes changes, those changes are directly made to the `IcebergTableInformation` and the table metadata that it holds.
Because of this, every connection needs to copy the `IcebergTableInformation`, removing any possibility for shared reuse of static data.

### New Relationship

`IcebergTableInformation` no longer holds a `unique_ptr<IcebergTransactionData>`, it should now be immutable throughout a transaction.
Instead we introduce the `IcebergTransactionTableState`, this now holds a reference to an `IcebergTableInformation`, along with `unique_ptr<IcebergTransactionData>` and new `IcebergTableEntry` items that are created during a transaction (through alter table).
The `IcebergTableInformation` is no longer modified by transaction-local changes, the `IcebergTableUpdate` items are now self-sufficient, enabling retries in the future.

This now enables a future optimization of caching the `IcebergTableInformation` between transactions, when we start using the `etag` to know that items haven't changed.

### DIFF
--- a/src/catalog/rest/api/iceberg_add_snapshot.cpp
+++ b/src/catalog/rest/api/iceberg_add_snapshot.cpp
@@ -17,7 +17,7 @@ namespace duckdb {
 
 IcebergAddSnapshot::IcebergAddSnapshot(const IcebergTableInformation &table_info,
                                        IcebergSnapshotOperationType operation)
-    : IcebergTableUpdate(IcebergTableUpdateType::ADD_SNAPSHOT, table_info), operation(operation) {
+    : IcebergTableUpdate(IcebergTableUpdateType::ADD_SNAPSHOT), operation(operation) {
 	//! FIXME: Do we also need to capture the current partition spec and sort order?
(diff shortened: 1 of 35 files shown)


Tishj:
Not super happy about how big the scope of this PR is

Might only salvage these piece(s) from it:
- Remove coupling between `IcebergTableUpdate` and `IcebergTableInformation` / `IcebergTableMetadata`, if anything, metadata change should come from applying an update, not the other way around.

PR #1120, opened 2026-06-29
Don't delete written files when the commit state is unknown

raghav-reglobe:
## Problem

`IcebergTransaction::Commit()` removes the data files a transaction wrote whenever the commit throws: the `catch` calls `CleanupFiles()`, and the `Rollback()` that follows calls it again.

That is correct for a **definitive** failure — a 4xx, e.g. a 409 conflict: the commit was rejected and did not land, so its files are genuine orphans. But it is wrong when the outcome is **unknown**: a 5xx, or a connection error with no response, may mean the commit actually **landed**, with the written files now referenced by the new, live snapshot. Deleting them then leaves the snapshot pointing at a missing file and permanently corrupts the table — surfaced as a read-time `404 / NoSuchKey` on the next scan of the current state. It gets more likely the busier the catalog is.

This is Iceberg's **"commit state unknown"** condition. The reference implementation throws `CommitStateUnknownException` precisely here and requires writers to **preserve** the written files, because the commit may have applied. Iceberg core has fixed this same class of cleanup-on-unknown-state corruption more than once:

- apache/iceberg#4673 — *Core: Fix table corruption from OOM during commit cleanup in Spark*
- apache/iceberg#13443 — *should not delete manifest files after getting `CommitStateUnknownException`*

## Fix

Two commits:

**1. Don't clean up on unknown commit state.** Classify the commit failure by HTTP status and only clean up on a **definitive 4xx** rejection. A **5xx**, or an error with **no HTTP status** (a connection error / timeout), leaves `commit_state_unknown` set, which makes `CleanupFiles()` a no-op so the written files are kept — if the commit truly failed they are unreferenced orphans (reclaimed by `remove_orphan_files`); if it landed they stay correctly referenced. The default is conservative: anything we can't read as a 4xx is treated as unknown, so we never delete a possibly-committed file. The guard lives inside `CleanupFiles()` because it has two callers (the `Commit()` catch and the subsequent `Rollback()`), so one chokepoint enforces the invariant for both.

**2. Throw `HTTPException` on non-200 REST responses.** So the status is available on the **throwing side** — `CommitStateUnknown` reads it from `ErrorData::ExtraInfo()["status_code"]` rather than parsing it out of the message. Previously `CommitResult::Throw` (the commit path) threw an `InvalidConfigurationException` carrying the status only in the message, so the consumer had to string-match it. Every REST call that produced *"returned a non-200 status code"* now throws `HTTPException` (which populates `status_code`/`reason`/`response_body` in `extra_info`): `CommitResult::Throw` plus the namespace/table/config `IRCAPI` calls. `CommitResult` carries no headers, so its `Throw()` uses the low-level `HTTPException` constructor with an empty header map. `CommitNewTable` rethrows a caught `HTTPException` as-is so a real non-200 isn't downgraded back to `InvalidConfigurationException` by its generic catch. The message-parse path in `CommitStateUnknown` is kept as a fallback.

## Testing

`test_transactional_updates_error_omits_stacktrace` exercises a 409 conflict on a multi-table commit and asserts the cleanup log fires — i.e. a definitive 4xx still cleans up, so there is no regression. Verified locally against the rest fixture. Happy to add a fixture-injected 5xx case asserting the files are preserved if that would help.

### DIFF
--- a/src/catalog/rest/transaction/iceberg_transaction.cpp
+++ b/src/catalog/rest/transaction/iceberg_transaction.cpp
@@ -379,6 +379,26 @@ static case_insensitive_set_t GetRetryTableKeys(const TableTransactionInfo &tran
 	return table_keys;
 }
 
+// 4xx = definitive rejection (commit did not land) -> delete the orphan. 5xx / no HTTP status
+// (connection error/timeout) = outcome unknown (may have landed) -> keep files. Status is in
+// extra_info (HTTPException) or the message "status code (<status>)" (fallback).
+static bool CommitStateUnknown(const ErrorData &error) {
+	auto &extra = error.ExtraInfo();
+	auto it = extra.find("status_code");
+	if (it != extra.end()) {
+		return it->second.empty() || it->second[0] != '4';
+	}
+	auto &msg = error.RawMessage();
+	auto pos = msg.find("status code (");
+	if (pos != string::npos) {
+		auto digit = msg.find_first_of("0123456789", pos);
+		if (digit != string::npos) {
+			return msg[digit] != '4';
+		}
+	}
+	return true;
+}
+
 void IcebergTransaction::Commit() {
 	if (transaction_updates.empty() && created_schemas.empty() && deleted_schemas.empty() &&
 	    schema_property_updates.empty()) {
@@ -423,6 +443,7 @@ void IcebergTransaction::Commit() {
 		DoSchemaDeletes(*temp_con_context);
 	} catch (std::exception &ex) {
 		ErrorData error(ex);
+		commit_state_unknown = CommitStateUnknown(error);
 		CleanupFiles();
 		DropSecrets(*temp_con_context);
 		temp_con.Rollback();
@@ -645,6 +666,10 @@ void IcebergTransaction::CleanupFiles() {
 		// certain catalogs don't allow deletes and will have a s3.deletes attribute in the config describing this
 		// aws s3 tables rejects deletes and will handle garbage collection on its own, any attempt to delete the files
 		// on the aws side will result in an error.
+		return;
+	}
+	if (commit_state_unknown) {
+		// Commit may have landed (CommitStateUnknownException); keep the files.
 		return;
 	}
 	ScopedTransaction temp_con(db);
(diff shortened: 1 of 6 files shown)


src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -152,231 +299,112 @@ static rest_api_objects::TableUpdate CreateSetSnapshotRefUpdate(int64_t snapshot
...
+// (connection error/timeout) = outcome unknown (may have landed) -> keep files. Status is in
+// extra_info (single-table commit) or the message "status code (<status>)" (multi-table commit).
+static bool CommitStateUnknown(const ErrorData &error) {
+	auto &extra = error.ExtraInfo();
+	auto it = extra.find("status_code");
+	if (it != extra.end()) {
+		return it->second.empty() || it->second[0] != '4';
+	}
+	auto &msg = error.RawMessage();
+	auto pos = msg.find("status code (");
```
Tishj, line 392: Perhaps fine to keep as fallback, but lets also fix this on the throwing side
Everywhere that we produce exceptions that contain: `returned a non-200 status code` should be using `HTTPException` instead

raghav-reglobe:
Thanks @Tishj — makes sense, done. I rebased onto current main while I was at it (the branch was based on an older main), so it's two commits now: the original cleanup-on-unknown-state fix, plus the throw-side change you asked for.

Every `IRCAPI` call that produced `returned a non-200 status code` now throws `HTTPException`, so the status lands in `ExtraInfo()["status_code"]` on the throwing side — including `CommitResult::Throw`, which is the commit path that previously carried the status only in the message. `CommitResult` has no headers, so its `Throw()` uses the low-level `HTTPException` constructor with an empty header map. I kept the message-parse in `CommitStateUnknown` only as the fallback you mentioned, and added a `catch (const HTTPException &) { throw; }` ahead of `CommitNewTable`'s generic catch so a real non-200 doesn't get downgraded back to `InvalidConfigurationException`.

Verified against the rest fixture — `test_transactional_updates_error_omits_stacktrace` still passes. Is this the shape you had in mind?

Tishj:
Thanks!

PR #1122, opened 2026-07-01
Add backoff to the commit-retry loop

cjnoname:
Follow-up to #1115 (append commit-retry). #1115 retries a conflicting commit **immediately**, with no wait between attempts. Under real contention — many writers committing to the same table (e.g. concurrent chunk writers against S3 Tables) — that is a thundering herd: every loser of the compare-and-swap wakes and re-collides in lockstep, so retries keep conflicting.

This adds **exponential backoff with decorrelated jitter** between retries, plus a cumulative **`commit.retry.total-timeout-ms`** budget, resolved from the existing `commit.retry.*` table properties:

- `commit.retry.min-wait-ms` / `commit.retry.max-wait-ms` — backoff bounds
- `commit.retry.total-timeout-ms` — overall retry-time budget
- `num-retries` keeps using the existing reader from #1115

Decorrelated jitter (AWS "Exponential Backoff And Jitter") starts tight near `min-wait` — a conflict's refresh+recommit window is short, so retrying quickly usually wins — and widens only as failures accumulate, which is exactly when a herd needs spreading. An independent RNG per loop de-synchronizes concurrent committers. `IcebergRetryConfig` folds a multi-table transaction's per-table policies into the most lenient one (they share one retry loop). Backoff is applied in both the single- and multi-table commit loops.

### Testing
- `test_commit_retry_config.test` — the full `commit.retry.*` family is accepted and normal commits still succeed (incl. `num-retries=0`).
- `test_commit_retry_concurrent.test` — 8 concurrent inserts to one table must all land. Gated on Lakekeeper (the SQLite-backed fixture catalog doesn't enforce serializable commits, so it can't produce real conflicts); skips cleanly elsewhere.

Verified locally against the fixture catalog (config test passes; concurrent test skips as expected). Formatting checked with clang-format 11 + format.py.

### DIFF
--- a/src/catalog/rest/api/iceberg_retry.cpp
+++ b/src/catalog/rest/api/iceberg_retry.cpp
@@ -1,0 +1,95 @@
+#include "catalog/rest/api/iceberg_retry.hpp"
+
+#include "core/metadata/iceberg_table_metadata.hpp"
+
+#include <string>
+
+namespace duckdb {
+
+namespace {
+
+constexpr const char *NUM_RETRIES = "commit.retry.num-retries";
+constexpr const char *MIN_WAIT_MS = "commit.retry.min-wait-ms";
+constexpr const char *MAX_WAIT_MS = "commit.retry.max-wait-ms";
+constexpr const char *TOTAL_WAIT_MS = "commit.retry.total-timeout-ms";
+
+constexpr idx_t NUM_RETRIES_DEFAULT = 4;
+constexpr int64_t MIN_WAIT_MS_DEFAULT = 100;
+constexpr int64_t MAX_WAIT_MS_DEFAULT = 60 * 1000;
+constexpr int64_t TOTAL_WAIT_MS_DEFAULT = 30 * 60 * 1000;
+
+template <class T>
+T ParseInt(const string &value, T fallback, bool allow_zero) {
+	if (value.empty()) {
+		return fallback;
+	}
+	try {
+		auto parsed = std::stoll(value);
+		if (parsed < 0 || (parsed == 0 && !allow_zero)) {
+			return fallback;
+		}
+		return static_cast<T>(parsed);
+	} catch (...) {
+		return fallback;
+	}
+}
+
+} // namespace
+
+IcebergRetryConfig IcebergRetryConfig::FromTableMetadata(const IcebergTableMetadata &metadata) {
+	IcebergRetryConfig config;
+	//! num-retries may legitimately be 0 (no retries, single attempt).
+	config.num_retries = ParseInt<idx_t>(metadata.GetTableProperty(NUM_RETRIES), NUM_RETRIES_DEFAULT, true);
+	config.min_wait_ms = ParseInt<int64_t>(metadata.GetTableProperty(MIN_WAIT_MS), MIN_WAIT_MS_DEFAULT, false);
+	config.max_wait_ms = ParseInt<int64_t>(metadata.GetTableProperty(MAX_WAIT_MS), MAX_WAIT_MS_DEFAULT, false);
+	config.total_wait_ms = ParseInt<int64_t>(metadata.GetTableProperty(TOTAL_WAIT_MS), TOTAL_WAIT_MS_DEFAULT, false);
+	//! Guard against a misconfigured table where min > max.
+	if (config.min_wait_ms > config.max_wait_ms) {
+		config.min_wait_ms = config.max_wait_ms;
+	}
+	return config;
+}
+
+IcebergRetryConfig IcebergRetryConfig::MostLenient(const IcebergRetryConfig &other) const {
+	//! Multiple tables committed in one atomic transaction share a single retry loop. Take the most
+	//! retry-tolerant policy so no table's (more permissive) configuration is silently dropped: the
+	//! larger num-retries / max-wait / total-timeout, and the smaller min-wait.
+	IcebergRetryConfig merged;
+	merged.num_retries = MaxValue<idx_t>(num_retries, other.num_retries);
+	merged.min_wait_ms = MinValue<int64_t>(min_wait_ms, other.min_wait_ms);
+	merged.max_wait_ms = MaxValue<int64_t>(max_wait_ms, other.max_wait_ms);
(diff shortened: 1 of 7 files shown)


cjnoname:
@Tishj this builds directly on your #1115 — it keeps the append-retry mechanism as-is and only adds backoff between attempts. #1115 retries immediately, which under contention (many writers on the same table) is a thundering herd; this adds decorrelated-jitter backoff + a total-timeout budget, driven by the existing `commit.retry.*` properties. Small and focused. Would appreciate a review when you have a chance 🙏

src/catalog/rest/api/iceberg_retry.cpp
```diff
@@ -0,0 +1,138 @@
...
+		}
+		wait *= 2;
+	}
+	if (wait > max_wait_ms) {
+		wait = max_wait_ms;
+	}
+	return wait;
+}
+
+int64_t IcebergRetryConfig::BackoffMsWithJitter(idx_t attempt, double unit_random) const {
```
Tishj: unused

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -482,6 +542,7 @@ void IcebergTransaction::DoTableRename(IcebergTransactionRenameUpdate &rename_up
 void IcebergTransaction::DoMultiTableCommitUpdates(IcebergTransactionAlterUpdate &alter_update,
                                                    ClientContext &context) {
 	const auto max_retries = GetMaxRetries(alter_update);
```
Tishj: This can be replaced with the RetryConfig

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -482,6 +542,7 @@ void IcebergTransaction::DoTableRename(IcebergTransactionRenameUpdate &rename_up
...
+	RetryBackoff backoff(GetRetryConfig(alter_update));
```
Tishj: I think this should live in `TableTransactionInfo`.

`StageSingleTableCommit` can bundle the `IcebergRetryConfig` in the `SingleTableStagedCommit`
`GetTransactionRequest` can then "merge" (MostLenient?) those configs to create the multi-table config

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -190,6 +195,61 @@ static idx_t GetMaxRetries(const IcebergTransactionAlterUpdate &alter_update) {
...
+	if (!saw_table) {
+		config = IcebergRetryConfig::FromTableMetadata_Default();
+	}
+	return config;
+}
+
+//! Per-retry-loop backoff state: decorrelated jitter (de-synchronizes a thundering herd of
+//! concurrent writers) plus a cumulative total-timeout budget. Independent RNG per loop so
+//! concurrent committers do not wake in lockstep.
+struct RetryBackoff {
```
Tishj: `IcebergRetryBackoff`

cjnoname:
Thanks, addressed all four:

- **iceberg_retry.cpp unused** — removed `BackoffMs` and `BackoffMsWithJitter` (only `DecorrelatedBackoffMs` is used).
- **`GetMaxRetries` → RetryConfig** — dropped `GetMaxRetries`; the loops now read `num_retries` straight from the resolved `IcebergRetryConfig`.
- **config lives in TableTransactionInfo** — `StageSingleTableCommit` now bundles the `IcebergRetryConfig` into `SingleTableStagedCommit`, and `GetTransactionRequest` folds them (`MostLenient`) into the multi-table config on `TableTransactionInfo`. The commit loops take the policy from there instead of recomputing it, and the backoff is created once from that config.
- **rename** — `RetryBackoff` → `IcebergRetryBackoff`.

Rebuilt, retry-config test passes, and format.py is clean.

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -481,13 +501,19 @@ void IcebergTransaction::DoTableRename(IcebergTransactionRenameUpdate &rename_up
 
 void IcebergTransaction::DoMultiTableCommitUpdates(IcebergTransactionAlterUpdate &alter_update,
                                                    ClientContext &context) {
-	const auto max_retries = GetMaxRetries(alter_update);
+	//! The retry policy is the tables' folded (most-lenient) config, produced by GetTransactionRequest.
+	//! It is stable across attempts (same tables), so the backoff state is created once, lazily, from
+	//! the first staged request and reused for every retry.
+	unique_ptr<IcebergRetryBackoff> backoff;
```
Tishj, line 507: nit: `optional<IcebergRetryBackoff>`, I don't think a heap allocation is necessary here

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -517,8 +549,9 @@ void IcebergTransaction::DoSingleTableCommitUpdates(IcebergTransactionAlterUpdat
 	D_ASSERT(catalog.supported_urls.count("POST /v1/{prefix}/namespaces/{namespace}/tables/{table}"));
 	for (auto &entry : alter_update.updated_tables) {
 		auto &table_key = entry.first;
-		const auto max_retries =
-		    entry.second.transaction_data ? GetMaxRetries(*entry.second.transaction_data) : idx_t(0);
+		//! Backoff state is created once per table (lazily, from the first staged commit's config)
+		//! and reused across this table's retries.
+		unique_ptr<IcebergRetryBackoff> backoff;
```
Tishj, line 554: nit: same as above

Tishj:
Thanks!

cjnoname:
Done - switched both backoff instances from `unique_ptr<IcebergRetryBackoff>` to `std::optional<IcebergRetryBackoff>` (lazy `emplace`, no heap allocation). Rebuilt, retry-config test passes, format.py clean.

cjnoname:
Correction: this PR was already merged before I pushed the optional change, so it didn't land here. I've opened #1147 as a small follow-up to address the two `optional<IcebergRetryBackoff>` nits. Sorry for the confusion.

PR #1123, opened 2026-07-01
Implement MergeAppend manifest merging on commit (#790)

cjnoname:
Rewrite of #1060 targeting `main` (the old PR was based on v1.5-variegata and stacked under the compression PR, now merged).

## What

On commit, repack the assembled manifest list into fewer manifests via first-fit bin-packing (mirroring Java's `ManifestMergeManager`). This bounds manifest growth so read planning doesn't degrade as small per-commit manifests accumulate — the standard Iceberg MergeAppend behavior (#790).

## How

- New `iceberg_manifest_merge` module: pure bin-packing + merge-decision logic, gated by the Java-aligned table properties `commit.manifest-merge.enabled` (default true), `commit.manifest.min-count-to-merge` (100), `commit.manifest.target-size-bytes` (8 MiB).
- `IcebergAddSnapshot::MergeManifestList` runs right before the manifest list is written. DATA and DELETE manifests are merged independently, never mixed.
- **V3 row lineage**: this transaction's brand-new DATA manifests are left unmerged — their per-row `first_row_id` isn't materialized yet, and folding them would break manifest-level inheritance. They become eligible next commit as carried-over manifests, so growth still converges (delayed one commit). New DELETE manifests carry no `first_row_id` and stay eligible.

## Testing

`test_merge_append.test` (merge happens, data intact, EXISTING-status entries present), `test_merge_append_disabled.test` (control: no merge), `test_merge_append_v3_deletes.test` (v3 delete-manifest merge). The v3 test is registered in the Nessie skip list (no v3 support there yet). All pass locally against the fixture catalog; format checked with clang-format 11 + format.py.


cjnoname:
@Tishj this is the MergeAppend rewrite (#790) targeting main — supersedes #1060, which was on the old v1.5-variegata base. It repacks the manifest list on commit via first-fit bin-packing (Java-aligned), gated by the standard commit.manifest.* properties, with DATA/DELETE merged independently and V3 new-data manifests deferred one commit for row-lineage safety. Tests included (merge / disabled / v3-deletes, v3 skipped on Nessie). Would appreciate a review 🙏

src/catalog/rest/api/iceberg_add_snapshot.cpp
```diff
@@ -152,6 +154,75 @@ void IcebergAddSnapshot::ConstructManifestList(IcebergManifestList &new_manifest
...
+	vector<MergeInputManifest> data_input;
+	vector<MergeInputManifest> delete_input;
+	vector<IcebergManifestListEntry> kept;
+	for (auto &entry : entries) {
+		auto source = entry.file.added_snapshot_id == snapshot_id ? ManifestSource::NEW_THIS_TRANSACTION
+		                                                          : ManifestSource::CARRIED_OVER;
+		//! V3 row lineage only constrains DATA manifests (first_row_id). New DELETE manifests carry
+		//! no first_row_id, so they remain eligible for merging even on V3.
+		if (is_v3 && source == ManifestSource::NEW_THIS_TRANSACTION &&
+		    entry.file.content == IcebergManifestContentType::DATA) {
```
Tishj, line 198: I'm not entirely sure that this holds, so I'm just going to write out some thoughts in the hopes they'll either lead me to realize that it *does* work, or that it doesn't.

I'm mostly worried about v2->v3 upgrades being overlooked, or that this code assumes row ids are materialized in a layer where they might not be yet.

Row-id inheritance chain is:
metadata.json -> snapshot -> manifest_file -> manifest_entry.data_file -> parquet file (_row_id)

We are dealing with the third layer here, `manifest_file`
The spec states that in V3, every manifest written should receive a `first_row_id`, we do this, in `WriteToFile` as you mentioned above.
The rest of the chain (manifest_entry.data_file -> parquet_file) is not materialized, and relies on inheritance.

However in V2, no row id information existed, and the upgrade to V3 doesn't produce a new snapshot.
So when we produce a V3 snapshot, we have to be aware that the source manifest list might have come from V2, in which case the manifest_file entries will be lacking a `manifest_file.first_row_id`.
And we'll have to assign them (see `LoadExistingManifestList`)

I *believe* that has already happened when we get to this stage of the commit though, so all existing manifest_file entries should have a first_row_id

src/catalog/rest/api/iceberg_manifest_merge.cpp
```diff
@@ -0,0 +1,348 @@
...
+		const bool carried_over = input[idx].source == ManifestSource::CARRIED_OVER;
+		if (is_v3 && member.file.has_first_row_id) {
+			if (!has_first_row_id || member.file.first_row_id < min_first_row_id) {
+				min_first_row_id = member.file.first_row_id;
+				has_first_row_id = true;
+			}
+		}
+		auto loaded = member.manifest_entries.empty() ? ScanManifestEntries(member, commit_state, schema_id) : member;
+		for (auto &entry : loaded.manifest_entries) {
+			//! DELETED entries: only deletes made by THIS snapshot are carried into the new manifest;
```
Tishj: Can we not use "deletes" here? Since that's a bit of an overloaded term otherwise. I think "dropped entry" might be less confusing

src/catalog/rest/api/iceberg_manifest_merge.cpp
```diff
@@ -0,0 +1,348 @@
...
+	while (!reader->Finished()) {
+		reader->Read();
+	}
+	return std::move(manifest_files[0]);
+}
+
+namespace {
+
+//! Merge one spec-homogeneous bin into a single new manifest. Returns the new list entry.
+IcebergManifestListEntry MergeBin(const vector<MergeInputManifest> &input, const vector<idx_t> &bin,
```
Tishj, line 241: I think we might need the avro key-value metadata of the manifests, which we haven't had need for previously.

The `schema_id` and `partition_spec_id` are stored in the key-value metadata of a manifest: https://iceberg.apache.org/spec/#manifests

Tishj: You'll likely need to use #1129

Tishj: Though we do already have the `partition_spec_id` in the manifest_file, we don't have the `schema_id`.
I think we'll need that if we're going to merge manifests, because manifests written with different schema ids *probably* don't merge correctly.

The `schema_id` influences the `manifest_entry.data_file.partition` schema, at best they add bogus columns, at worst they remove stats of existing columns.

cjnoname:
Thanks for the careful review. Addressed all three:

**schema_id safety (the important one)** — you're right that merging manifests written under different schema ids is unsafe (schema id drives the data_file.partition layout + column stats). Since a manifest_file doesn't expose its schema id yet (that's #1129), I've made merging conservative for now: it only runs when the table has a **single schema**, in which case every manifest is unambiguously that schema. Tables that have gone through schema evolution skip merging until #1129 lands and we can group by per-manifest schema_id. Added a comment referencing #1129.

**v2->v3 row-id worry** — I verified your hypothesis: by the time MergeManifestList runs, every carried-over DATA manifest already has a first_row_id. A V2->V3 upgraded snapshot assigns one to each existing DATA manifest earlier in the commit (IcebergTransactionData's upgrade handling), and new V3 data manifests are excluded from merging. Added a comment spelling this out so the invariant is explicit.

**"deletes" wording** — renamed to "dropped entry" in the merge comment to avoid confusion with delete manifests.

Rebuilt, merge/disabled/v3-deletes tests pass, format.py clean.

Tishj:
#1129 is merged now, so you should be able to get the schema_id
You'll probably need to parse the `schema` instead though, because the **reference implementation** has opted to ignore the requirement of the spec and don't populate `schema-id`, only `schema`, because it contains schema_id

I might make one more PR to also populate the `metadata` of the `IcebergManifestListEntry` items we create for transactional data, but that's not in currently.
Instead I think you can just say "if metadata.empty(), assume schema_id of the manifest is current schema id"

cjnoname:
Rebased onto main (now includes #1129) and implemented real per-schema grouping, replacing the single-schema fallback.

- `MergeManifests` now groups by `(schema id, partition spec id)` and only merges within a group.
- Schema id is resolved per manifest in `ResolveManifestSchemaId`: prefer the `schema-id` metadata key, else parse the `schema` JSON's `schema-id` (since the reference implementations only populate `schema`, not `schema-id`), else fall back to the current schema id when the manifest carries no metadata.
- Carried-over manifests only expose that Avro key-value metadata once opened, so entries are now loaded before grouping and reused by `MergeBin` (no extra file open).
- Added v2 + v3 schema-evolution merge tests proving old-schema and new-schema manifests merge among themselves but never across the schema boundary (v3 registered in nessie.json).

All merge tests pass locally, format.py clean.

cjnoname:
Fixed the failing check (`test_manifest_vector_writer_chunking.test_slow` under the Equality-Deletes fixture): that regression relies on 2050 separate manifests to exercise the chunked VectorWriter, but with manifest merge on by default the 2050 single-row inserts get packed into far fewer manifests, so the 2050-entry assertion failed. Disabled merge on that specific table so it keeps exercising a large manifest list. Verified locally.

src/catalog/rest/api/iceberg_manifest_merge.cpp
```diff
@@ -0,0 +1,414 @@
...
+//! Resolve the schema id a manifest was written under, so that only manifests sharing a schema are
+//! ever merged. The schema id drives the data_file.partition layout and column stats, so mixing
+//! schemas can drop or misalign stats. Resolution order (per the Iceberg spec's manifest key-value
+//! metadata):
+//!  1. the "schema-id" key, if present;
+//!  2. otherwise parse the "schema" key (a full schema JSON) and read its "schema-id" -- the
+//!     reference implementations (Java/PyIceberg) populate only "schema", not "schema-id";
+//!  3. otherwise (no metadata at all -- e.g. manifests we create for this transaction do not yet
+//!     carry key-value metadata) assume the table's current schema id.
+int32_t ResolveManifestSchemaId(const IcebergManifestListEntry &entry, int32_t current_schema_id) {
```
Tishj: I don't like the fallback, this should never happen
I expect that it *does* happen currently so we need to make some additional changes

Inside `MergeBin` we use `ScanManifestEntries` if it's not already loaded, but that's too late, because this metadata we're after is part of each manifest file, not the manifest list

And for manifests that are added by this transaction, we don't even have to enter this function, because we have `ManifestSource`

cjnoname: Good call - reworked it exactly as you described:

- A manifest added by this transaction now takes the current schema id straight from its `ManifestSource` (NEW_THIS_TRANSACTION); we no longer read metadata it does not carry.
- A carried-over manifest reads its schema id from the Avro header key-value metadata, which is materialized by the pre-load step before grouping (I confirmed carried-over manifests always expose `schema`/`schema-id` here).
- Dropped the current-schema fallback entirely: if a carried-over manifest is missing its schema id, that is a bug, so it now throws instead of silently guessing.

cjnoname: Update: I tried throwing when a carried-over manifest has no schema metadata, but it turns out that *does* happen a lot in practice - several existing delete/merge/alter tests produce carried-over manifests with no schema key-value metadata, and throwing broke ~5 catalog CI jobs. So I kept the ManifestSource shortcut for new-this-transaction manifests (they take the current schema id directly, no metadata read), but restored the fallback to the current schema id for carried-over manifests that lack the metadata. It is conservative (may over-group manifests written under different schemas) but never unsafe, since grouping only affects packing, not the data described. If you'd prefer to eliminate the fallback entirely, that would need the write path to always emit schema metadata on manifests first (perhaps the follow-up PR you mentioned) - happy to go that route if you want.

src/catalog/rest/api/iceberg_manifest_merge.cpp
```diff
@@ -97,7 +105,9 @@ int32_t ResolveManifestSchemaId(const IcebergManifestListEntry &entry, int32_t c
 		}
 	}
 
-	return current_schema_id;
+	throw InternalException(
```
Tishj: Let's not do `InternalException`, `InvalidConfigurationException` is better, otherwise we'll end up invalidating the database/session

cjnoname: Agreed InternalException was wrong. I've actually removed the throw altogether in the latest push and fall back to the current schema id instead, because throwing here (with any exception type) breaks a lot of existing tests: many carried-over manifests produced by delete/merge/alter genuinely have no schema key-value metadata today, so the lookup fails and the commit would error out (this took down ~5 catalog CI jobs when I tried it).

So the current behavior is: new-this-transaction manifests use the current schema id via ManifestSource; carried-over manifests read schema id from metadata, falling back to the current schema id when it is absent (conservative - may over-group - but never unsafe, since grouping only affects packing).

If you'd rather make the missing-metadata case a hard error, I think the cleaner path is a follow-up that makes the write path always emit schema metadata on manifests first, then we can drop the fallback and error (as InvalidConfigurationException) safely. Which direction do you prefer?

Tishj: I think I know why, we can create multiple snapshots per transaction, as every statement becomes a `IcebergAddSnapshotUpdate`.

We should add to the `metadata` of the `IcebergManifestListEntry` that we create for transactional data.

Kinda unrelated:
I wonder if we even write the correct schema id for cases where a schema change happened in the same transaction :thinking: 
Like: begin, insert, alter table ..., insert, commit

cjnoname: Exactly - that was the cause. Implemented your suggestion: manifest_file::WriteToFile now optionally mirrors the Avro header key-value metadata (schema, schema-id, partition-spec-id, content) into an out-param, and the three write sites that create list entries (WriteManifestListEntry, RewriteManifestFile, MergeBin) pass their entry's metadata map. So a manifest we write within a transaction now carries its metadata on the in-memory list entry, and when a later statement's commit sees it as carried-over, MergeManifests reads its real schema id instead of falling back.

Verified locally: all the delete/merge/alter tests that previously hit the fallback now resolve the real schema id (zero fallbacks). I kept the fallback (no longer a throw) purely as a safety net for manifests written by other engines that omit the metadata.

On your schema-change-within-one-transaction question - I checked it and it's actually fine: WriteToFile stamps schema-id = GetCurrentSchemaId(), and in `begin; insert; alter add column; insert; commit` the current schema id does advance after the ALTER, so the pre-ALTER manifest gets schema-id 0 and the post-ALTER one gets schema-id 1 (confirmed with a debug print). I added a test for exactly this (test_merge_append_txn_schema_change.test) plus v2/v3 schema-evolution merge tests. CI is green.

Tishj: Thanks, I am going to push on the throw one last time though, this metadata isn't optional, it's required, for all versions

src/catalog/rest/api/iceberg_add_snapshot.cpp
```diff
@@ -152,13 +155,89 @@ void IcebergAddSnapshot::ConstructManifestList(IcebergManifestList &new_manifest
...
+	//! manifests unmerged; next commit they are carried-over and become eligible, so growth still
+	//! converges, merely delayed by one commit. V2 has no row lineage, so everything is eligible.
+	const bool is_v3 = commit_state.table_info.table_metadata.iceberg_version >= 3;
+
+	//! Split by content type, tagging each manifest's origin. Excluded manifests are kept aside.
+	vector<MergeInputManifest> data_input;
+	vector<MergeInputManifest> delete_input;
+	vector<IcebergManifestListEntry> kept;
+	for (auto &entry : entries) {
+		auto source = entry.file.added_snapshot_id == snapshot_id ? ManifestSource::NEW_THIS_TRANSACTION
```
Tishj, line 193: Actually, because of what we discussed about creating multiple snapshots in the same transaction, this doesn't hold.
We can have new manifests that don't have a `manifest_file.first_row_id` assigned to them.

Instead, we should probably compare against the sequence number of the committed state.
But that "committed state" is kind of muddy right now. 
I think we'll need pieces of #1117 to confidently do this

src/catalog/rest/api/iceberg_manifest_merge.cpp
cjnoname: Done - switched it to throw InvalidConfigurationException (not Internal, so it does not invalidate the session). Now that we backfill the metadata on the in-memory entry, every manifest DuckDB writes carries its schema id, so the delete/merge/alter tests that previously broke when I tried throwing now pass locally (they resolve the real schema id and never reach the throw). Pushed; letting CI confirm nothing in the fixtures trips it.

src/catalog/rest/api/iceberg_add_snapshot.cpp
cjnoname: I dug into this and I think the situation is a bit better than the comment implies - want to check my reasoning with you.

Tracing `begin; insert(A); insert(B); commit`: when B commits, A's DATA manifest is carried-over (added_snapshot_id A != B). But A's manifest already has a valid first_row_id: CreateFromEntries assigns it eagerly at creation time for v3 (iceberg_manifest_list.cpp:57-60, has_first_row_id=true, first_row_id=next_row_id), using commit_state.next_row_id which advances per statement. It is NOT deferred to manifest_list::WriteToFile - that inheritance branch (iceberg_manifest_list.cpp:377-393) only fires when has_first_row_id is false, which never happens for manifests we create. And MergeBin gates on member.file.has_first_row_id (iceberg_manifest_merge.cpp:271-276), not on snapshot_id, so a carried-over in-transaction manifest merges with the correct min first_row_id. So I don't think row lineage actually breaks today - the code comment claiming 'not yet materialized' is what's wrong, not the merge.

That said, the current `added_snapshot_id == snapshot_id` exclusion is incomplete (it misses A's manifest within the same transaction). If we want the exclusion to be correct/self-consistent, we can identify 'created in this transaction' reliably with `sequence_number > table_metadata.last_sequence_number` - last_sequence_number is the committed baseline, immutable within the transaction (iceberg_table_update.cpp:42), so I don't think we need #1117 for this specific check.

Two options: (1) leave the merge as-is (it's safe) and just fix the misleading comment, or (2) switch the exclusion to the sequence_number predicate so it also excludes same-transaction prior-snapshot DATA manifests. Which do you prefer? Happy to do either - I'd lean (2) for correctness since it's a tiny change and doesn't need #1117.

Tishj: > But A's manifest already has a valid first_row_id: CreateFromEntries assigns it eagerly at creation time for v3 (iceberg_manifest_list.cpp:57-60, has_first_row_id=true, first_row_id=next_row_id)

If that's true then we have a retry problem, no?
Because how would we refresh assigned row-ids if we have already baked them in?

I think you're right, but that's a problem.
I'll pick up a fix for this, I think we need to delay assigning these row ids so it only happens during commit. That's the only way I see how we can make this retry compatible

No.. wait, we run `CreateCommitManifestFiles` which rebuilds a manifest file during commit
So that's what makes it retry-able currently, but that's quite wasteful I feel ...

src/catalog/rest/api/iceberg_manifest_merge.cpp
```diff
@@ -0,0 +1,426 @@
...
+		return empty;
+	}
+
+	//! CreateFromEntries computes all counts, min_sequence_number and the partition field summary
+	//! from the entries; pass the bin's own spec id so the summary is computed against the correct
+	//! (possibly historical) spec, not the table default. We pass a throwaway row-id
+	//! counter so the global one is not advanced, then restore the lineage-preserving value below.
+	int64_t scratch_row_id = 0;
+	auto result =
+	    IcebergManifestListEntry::CreateFromEntries(FileSystem::GetFileSystem(commit_state.context),
```
Tishj, line 348: I think this needs additional thought.
If `manifest_entry.first_row_id` isn't materialized, by merging the manifests we could be shuffling the final row ids of the data files.

i.e:

manifest1:
- first_row_id: 20
- manifest_entries: [ entry1, entry2, entry3 ]

manifest2:
- first_row_id: 100
- manifest_entries: [entry4, entry5]

imagine all of those entries having 10 rows:
entry1.first_row_id = 20
entry2.first_row_id = 30
entry3.first_row_id = 40
entry4.first_row_id = 100
entry5.first_row_id = 110

when we merge these, we could end up setting `entry4.first_row_id` to 20 or to 50, both are wrong

Tishj: I think a pass we should add is to assign `manifest_entry.first_row_id` to all committed entries we read.
The reason `manifest_entry.first_row_id` is left null for new manifests is so manifests don't have to be rewritten for a retry of the commit.
But when we read them here they are already committed, so we can safely materialize them, and as shown above I think we *have* to materialize them for correctness.

Which also means that this isn't correct:
```
	//! All the 'manifest_file' entries we will write to the new manifest list
	vector<IcebergManifestListEntry> manifests;
```
We have to keep the uncommitted manifests separate from the committed manifests *or* we have to perform the `manifest_entry.first_row_id` materialization step as part of `LoadExistingManifestList` ?

cjnoname: Implemented this: when merging carried-over v3 DATA manifests, I now materialize each entry's first_row_id before moving it into the merged manifest, computed against ITS source manifest's first_row_id + the record_count of preceding null-first_row_id entries (iceberg_manifest_merge.cpp, in MergeBin right after loading the entries). Per the spec's 'the inherited value of first_row_id must be written into data file metadata when creating existing and deleted entries'. CreateFromEntries/WriteToFile preserve a materialized value, so the merged row ids no longer depend on the merged manifest's first_row_id.

One honest finding while validating: I could not construct a case that actually shuffles today. I traced the ordering invariants - carried-over manifests are prepended in first_row_id order (AddToManifestEntries inserts at begin()), new manifests are appended (AddNewManifestFile push_back with a larger first_row_id), and AssignManifestFirstRowIds reassigns by physical order on read - so the merge input is always contiguous and ascending, and reading back re-derives the same ids even without materializing. A negative-control run (materialize disabled) kept _row_id correct for sequential and insert+delete+insert cases. So in the current code path this is spec-compliant + defensive rather than fixing a reproducible shuffle; it becomes load-bearing the moment any of those ordering invariants change (e.g. aligning manifest ordering with Java, or non-contiguous segments landing in one bin).

Added test_merge_append_v3_row_lineage.test as a regression guard on _row_id stability across merges (registered in nessie.json).

On your point about the mixed committed/uncommitted `manifests` vector: I key the materialization off ManifestSource (only CARRIED_OVER entries, and V3 new-DATA manifests are already excluded via `kept`), so committed vs uncommitted is distinguished at that point. Happy to go further and separate them explicitly if you'd prefer - let me know.

Tishj:
Thanks for all the work you've put into this, I thought about merging it as is, but in the end I decided it still needed some polish, which I've applied on top of your branch, here: #1169 

The biggest change is probably that merging now applies only to committed manifests, in the "load existing manifest list" path, as a result the merging logic gets a bit simpler.

cjnoname:
Thanks a lot for taking this over the finish line, and for the thorough reviews throughout - I learned a lot about the row-lineage internals from them. The committed-only approach in #1169 is cleaner: doing the merge in the load-existing-manifest-list path sidesteps the uncommitted-manifest first_row_id materialization problem entirely (exactly the messy part we kept circling), and folding it together with the existing manifest-rewrite path (WriteReplacementManifest/RewriteManifest) is a nice dedup.

I'll close this one in favor of #1169. Happy to review it or help chase down the remaining CI failures if that's useful - just let me know.

PR #1126, opened 2026-07-01
Evict caches if commiting to the rest catalog does not succeed

NiclasHaderer:
It is possible that concurrent updates to the same Iceberg table make updates fail in DuckDB Iceberg (the REST catalog rejects the commit with a 409 conflict). If you have caching enabled (MAX_TABLE_STALENESS), this has the added side effect that each retry after such a failure keeps reusing the same stale cached table metadata, so the statement fails again and again instead of succeeding on retry.

This PR evicts the cached LoadTableResult for any table touched by a failed commit, so a subsequent attempt fetches fresh metadata and can actually succeed. It also fixes ScopedTransaction's destructor, which committed via the throwing Connection::Commit() API


NiclasHaderer:
I guess this is partially fixed by the rety mechanism, but I think the underlying issue is still there if you disable that...

Tishj:
Thanks, I think this makes sense 👍

PR #1135, opened 2026-07-03
[Sorting] Enable SORTED and PARTITIONED insert/update/merge

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/9247

Note that https://github.com/duckdb/duckdb/pull/22620 doesn't enable `ORDER_BY` *without* `PARTITIONED_BY`, to work around this, we push a `PhysicalOrder` with the order-by if no partition spec is set, which I *believe* should have the same behavior.


Tishj:
follow up: https://github.com/duckdb/duckdb-iceberg/pull/1150

PR #1137, opened 2026-07-03
[Struct] Support struct field in transforms (partition-spec and sort-order)

Tishj:


src/execution/operator/iceberg_insert.cpp
```diff
@@ -222,13 +234,13 @@ void IcebergInsertGlobalState::AddFiles(DataChunk &chunk, const string &table_na
...
-		if (!AllIdentityTransforms(ic_partition_info)) {
+		if (!CanWriteIdentityPartitionsDirectly(ic_partition_info, *ic_schema)) {
 			for (auto &partition_field : ic_partition_info.fields) {
 				partition_colname_to_field.emplace(partition_field.GetPartitionSpecFieldName(), partition_field);
 			}
 		} else {
 			for (auto &partition_field : ic_partition_info.fields) {
-				auto actual_col_name = GetColumnNameBySourceId(ic_schema->columns, partition_field.source_id);
+				auto actual_col_name = GetColumnNameBySourceId(*ic_schema, partition_field.source_id);
 				partition_colname_to_field.emplace(actual_col_name, partition_field);
```
Tishj, line 244: This is problematic actually, if we partition on `nested.col` and `col`, they will both get the same `partition_colname_to_field` entry.

Joining on `_` or anything that is a valid character in a column name is also not the solution.
Perhaps we can join on `/0`?

Tishj: In any case, `GetColumnNameBySourceId` is probably a smell, if it's used in more places

Tishj: Or we should wrap every column component in a `"<component>"` ?

Issue #1136, opened 2026-07-03
String column-metric bound truncation is byte-exact and fixed-length


paultmathew:
DuckDB truncates string min/max bounds (`IcebergValue::TruncateString` in
`src/core/expression/iceberg_value.cpp`) in two problematic ways:

1. **Byte-exact.** The lower bound is cut at a raw 16-byte boundary, which can
   split a multi-byte UTF-8 code point and write an **invalid-UTF-8**
   `data_file.lower_bounds`. DuckDB's own reader tolerates it, but strict readers
   reject it — e.g. Trino/Athena reading `$partitions` fail with
   `Failed to decode value as UTF-8: java.nio.HeapByteBuffer[pos=0 lim=16 cap=16]`,
   even though the data itself reads fine. (#1022 fixed this for the upper bound
   and explicitly left the lower bound on plain truncation.)

2. **Fixed-length.** `MAX_STRING_UPPERBOUND_LENGTH` is hardcoded to 16 and ignores
   the table's configured
   [`write.metadata.metrics.default`](https://iceberg.apache.org/docs/latest/configuration/#write-properties)
   / `write.metadata.metrics.column.<name>` (`truncate(N)`), so users who store
   longer strings can't widen the bound for more selective pruning.

The length should come from the configured metrics property, and truncation should
land on a code-point boundary at that length (as Iceberg-Java's
`UnicodeUtil.truncateStringMin`/`Max` do).

<details>
<summary>Repro</summary>

```sql
CREATE TABLE my_datalake.default.lb (id INT, title VARCHAR);
-- "El cliente quer" is 15 ASCII bytes; the next code point 'í' is 0xC3 0xAD.
INSERT INTO my_datalake.default.lb VALUES (1, 'El cliente querído resolver el problema');

SET VARIABLE mp = (SELECT manifest_path FROM iceberg_metadata(my_datalake.default.lb)
                   ORDER BY manifest_sequence_number DESC LIMIT 1);
SELECT decode(lb)   -- throws: lower bound blob is not valid UTF-8
FROM (SELECT unnest(map_keys(lower_bounds)) k, unnest(map_values(lower_bounds)) lb
      FROM (SELECT unnest(data_file) FROM read_avro(getvariable('mp')))) WHERE k = 2;
```
</details>

PR #1138, opened 2026-07-03
Truncate string lower bounds on code point boundaries

paultmathew:
Similar to #1022 (which made string **upper** bounds code-point-safe and explicitly left the lower bound on plain truncation) — this fixes the **lower** bound.

## Problem

`IcebergValue::TruncateString` (the `SerializeBound::LOWER_BOUND` path for `VARCHAR`) truncates to a raw 16-byte prefix with no UTF-8 validation. When the cut lands inside a multi-byte code point, the written `data_file.lower_bounds` entry is **invalid UTF-8**. DuckDB's own reader tolerates it, but strict Iceberg readers reject it — e.g. Trino/Athena reading `$partitions` / `$manifests` fail with:

```
GENERIC_INTERNAL_ERROR: Failed to decode value as UTF-8: java.nio.HeapByteBuffer[pos=0 lim=16 cap=16]
```

even though the table data itself reads fine. This shows up in practice on free-text columns with accents / em-dashes (e.g. Spanish `í` = `0xC3 0xAD`, `—` = `0xE2 0x80 0x94`).

## Fix

Back the truncation off to the previous code-point boundary — a prefix truncated on a boundary is still a valid lower bound (a prefix is always `<=` the value), and stays valid UTF-8. This is the round-down counterpart of Iceberg-Java's `UnicodeUtil.truncateStringMin` (`TruncateAndIncrementString` already implements the round-up `truncateStringMax`).

The multi-byte back-off previously inlined in `TruncateAndIncrementString` is factored into a small `TruncateToCodePointBoundary` helper now shared by both bound paths, so the lower/upper truncation can't drift.

## Test

Adds `test_write_non_ascii_string_lower_bound.test`, mirroring the upper-bound test from #1022: inserts a long non-ASCII string, asserts the data round-trips and the written lower bound is valid UTF-8 and `<=` the column min.

## Validation

Reproduced the exact Athena failure and the fix on a real Glue catalog via an A/B: two tables with identical data differing only in the `title` lower bound — the byte-truncated bound (`…c3`) fails the `$partitions` read with the error above; the code-point-truncated bound (`…c3ad` = `"El cliente querí"`, what this patch produces) is accepted. Data scans succeed in both, confirming the fault is purely the manifest bound.

Relates to #1136.

### DIFF
--- a/src/core/expression/iceberg_value.cpp
+++ b/src/core/expression/iceberg_value.cpp
@@ -240,11 +240,19 @@ DeserializeResult IcebergValue::DeserializeValue(const string_t &blob, const Log
 
 const idx_t IcebergValue::MAX_STRING_UPPERBOUND_LENGTH;
 
+// Largest length <= MAX_STRING_UPPERBOUND_LENGTH that does not split a multi-byte
+// UTF-8 character, so a truncated bound stays valid UTF-8.
+static idx_t TruncateToCodePointBoundary(const string &input) {
+	idx_t len = std::min<idx_t>(IcebergValue::MAX_STRING_UPPERBOUND_LENGTH, input.size());
+	while (len < input.size() && (static_cast<unsigned char>(input[len]) & 0xC0) == 0x80) {
+		len--;
+	}
+	return len;
+}
+
 string IcebergValue::TruncateString(const string &input) {
-	std::vector<unsigned char> bytes(input.begin(), input.end());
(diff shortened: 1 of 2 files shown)


Tishj:
You can ignore that failing test, seems to be flaky, I'll clean it up separately

Tishj:
Thanks!

PR #1139, opened 2026-07-04
Honor write.metadata.metrics.* for string bound truncation

paultmathew:
Implements #1136 — DuckDB hardcodes the string min/max bound truncation length to 16 and ignores the table's configured [`write.metadata.metrics.*`](https://iceberg.apache.org/docs/latest/configuration/#write-properties). Users who store longer strings can't widen bounds for more selective pruning, and the metrics mode isn't respected at all.

> [!NOTE]
> **Stacked on #1138** (the string lower-bound UTF-8 fix). The first commit here is that fix; review the second commit (`Honor write.metadata.metrics.* ...`) for this change, or merge #1138 first and I'll rebase.

## What

In the insert write loop, resolve each column's metrics mode from `write.metadata.metrics.column.<name>` → `write.metadata.metrics.default` → Iceberg's `truncate(16)` default, and drive bound serialization from it:

- `none` — no metrics written for the column
- `counts` — value/null counts, no bounds
- `truncate(N)` — string bounds truncated to `N` (on a code-point boundary)
- `full` — untruncated bounds

The truncation length flows only through the string-specific path (`SerializeStringBound` / `TruncateString` / `TruncateAndIncrementString`); the generic `SerializeValue` is left unchanged. Non-string types are unaffected (no length knob).

## Test

`test_metrics_mode_configuration.test` covers: the configurable default length, per-column `full` / `counts` / `none` / `truncate(N)` overrides, both bound directions, the unset-default fallback to `truncate(16)`, and rejection of invalid mode values.

## Notes / scope

- Metrics properties govern **column metrics** (data-file bounds), so partition-summary bounds in the manifest list keep the default truncation — intentional.
- Variant columns respect `none` / `counts` gating but their bounds aren't length-adjusted (variant bound truncation is a separate concern).


src/core/expression/iceberg_value.cpp
```diff
@@ -330,22 +349,8 @@ SerializeResult IcebergValue::SerializeValue(Value input_value, const LogicalTyp
...
-		} else {
-			// for lower bound truncating is enough
-			val = IcebergValue::TruncateString(input_value.GetValue<string>());
-		}
-		auto serialized_val = Value::BLOB(reinterpret_cast<const_data_ptr_t>(val.data()), val.size());
-		auto ret = SerializeResult(column_type, serialized_val);
-		return ret;
-	}
+	case LogicalTypeId::VARCHAR:
+		return SerializeStringBound(input_value.GetValue<string>(), bound_type);
```
Tishj: I think we should just thread `IcebergMetricsConfig` into `SerializeValue`, it should probably be honored everywhere, and I could see this affecting serialization of other types in the future as well

src/execution/operator/iceberg_insert.cpp
```diff
@@ -49,6 +50,64 @@ static bool WriteSequenceNumber(IcebergInsertVirtualColumns virtual_columns) {
 	       virtual_columns == IcebergInsertVirtualColumns::WRITE_ROW_ID_AND_SEQUENCE_NUMBER;
 }
 
+// Per-column metrics collection level, from the `write.metadata.metrics.*` table
+// properties (https://iceberg.apache.org/docs/latest/configuration/#write-properties).
+enum class IcebergMetricsMode : uint8_t { NONE, COUNTS, TRUNCATE, FULL };
+
+struct IcebergMetricsConfig {
```
Tishj: I don't agree with this living in `iceberg_insert.cpp`, I think either `iceberg_metrics.hpp` or `iceberg_value.hpp` should be where this lives

Tishj:
Thanks, I'm missing tests with struct, list, map, variant and geometry, both with and without truncating/omitting metrics

paultmathew:
@Tishj thanks for the review! Added tests.

src/execution/operator/iceberg_insert.cpp
```diff
@@ -312,28 +317,38 @@ void IcebergInsertGlobalState::AddFiles(DataChunk &chunk, const string &table_na
 				auto normalized_col_name = StringUtil::Join(column_names, ".");
 				throw ConstraintException("NOT NULL constraint failed: %s.%s", table_name, normalized_col_name);
 			}
+			// Resolve the metrics mode for this column: per-column override -> table default ->
+			// Iceberg's truncate(16) default.
+			auto metrics = GetColumnMetricsConfig(table_metadata, default_metrics, StringUtil::Join(column_names, "."));
```
Tishj, line 322: I believe that this also enables targeting fields in lists/maps
But the structure of these names is very much tied to the parquet schema.

I believe it's `.key`/`.value` for a map and `.element` for a list element (will need verification)
I wonder if other writers use the same notation or we'd need to transform this to correctly map

Please add tests for those and verify whether other implementations use the same nesting semantics for them

paultmathew: Thanks for flagging this. I verified the naming lines up with the Iceberg convention, so no path normalization is needed:

- On our side the stat name from RETURN_STATS is resolved by IcebergTableSchema::GetFromPath, which descends purely by child name (column.GetChild(...)). The synthetic nested fields are named element (list) and key/value (map), so the path we look up an override with is <parent>.element / <parent>.key / <parent>.value / <struct>.field. (If we were feeding the physical parquet path like l.list.element, GetChild would miss and we'd throw "can not be found in the schema" — the nested-bounds tests passing already proves we use the logical names.)
- That's exactly what Iceberg-Java/PyIceberg use for write.metadata.metrics.column.<name>: MetricsConfig keys off Schema.findColumnName(id), which produces the same dotted field-name path with element/key/value for list/map leaves ([config docs](https://iceberg.apache.org/docs/1.10.2/configuration/)).
- I added a test that sets a different mode on each nested leaf (s.sf=truncate(4), l.element=full, m.key=truncate(8)) with m.value inheriting the table default, so each override is shown to resolve independently via the standard notation.

One thing we don't implement is write.metadata.metrics.max-inferred-column-defaults (the pre-order bound on how many columns get metrics by default) — happy to track that separately if you'd like, but it's orthogonal to the per-column naming.

Tishj: Yea that makes sense
Thanks for investigating and verifying 👍

Tishj:
Thanks!

The-Alchemist:
Is it possible to backport this to v1.5?  I'm assuming there will be another v1.5 release, probably v1.5.5?

I created a backport PR at https://github.com/duckdb/duckdb-iceberg/pull/1208

PR #1144, opened 2026-07-05
[Dev] Move sql tests that depend on generated tables into `test/python/paired`

Tishj:


Tishj:
parking this because this makes debugging a lot harder (can't just run with lldb attached, have to print the pid and halt so we can attach before it dies...), which is a major problem

PR #1147, opened 2026-07-05
Use optional instead of unique_ptr for retry backoff

cjnoname:
Small follow-up to #1122.

@Tishj left two nits on #1122 (`optional<IcebergRetryBackoff>` instead of `unique_ptr`, no heap allocation needed) but the PR was merged before I addressed them. This does exactly that: both backoff instances in `iceberg_transaction.cpp` now use `std::optional<IcebergRetryBackoff>` with lazy `emplace()` instead of `make_uniq`.

No behavior change - purely avoids the heap allocation. Rebuilt, retry-config test passes, format.py clean.


Tishj:
Thanks!

PR #1149, opened 2026-07-06
Only assign first_row_id to data manifests in the manifest list

raghav-reglobe:
### What

On V3 tables, `CreateFromEntries` assigns `first_row_id` (manifest list field 520) to **every** new manifest, including deletes manifests. Per the Iceberg V3 spec the field is only assigned to data manifests — it drives row-id inheritance, which doesn't apply to delete files — and Java's `SnapshotProducer` writes null for deletes manifests.

This PR gates the assignment on the manifest content type, so deletes manifests leave the field null (the writer already handles the null branch via `WriteNull()`).

### Why it matters

Readers that consume `first_row_id` for row-lineage purposes can mis-handle a deletes manifest that carries a value, and the output diverges from what Java writes for the same table. The row-id arithmetic itself is unchanged — `next_row_id` was only ever advanced for DATA entries — so assigned `_row_id` values are unaffected; only the serialized manifest list changes.

### Testing

Extended `test_row_lineage_write.test` with a manifest-list assertion: deletes manifests have a null `first_row_id`, data manifests a non-null one. The new assertion fails without the fix (deletes manifests show up with values) and passes with it; the existing `_row_id` / `_last_updated_sequence_number` expectations in the test are untouched, confirming the allocation behavior is unchanged.

### DIFF
--- a/src/core/metadata/manifest/iceberg_manifest_list.cpp
+++ b/src/core/metadata/manifest/iceberg_manifest_list.cpp
@@ -50,7 +50,8 @@ IcebergManifestListEntry IcebergManifestListEntry::CreateFromEntries(FileSystem 
 	IcebergManifestListEntry manifest_list_entry(manifest_file_path);
 	auto &manifest_file = manifest_list_entry.file;
 	manifest_file.manifest_path = manifest_file_path;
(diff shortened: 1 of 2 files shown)


Tishj (approved):
Thanks, makes sense

PR #1151, opened 2026-07-06
Support Iceberg v3 timestamptz_ns

### LINKED ISSUE #1019: [FEATURE] Support for Iceberg type timestamptz_ns
**Is your feature request related to a problem? Please describe.**
Iceberg's v3 table format includes the type `timestamptz_ns`, which is a nanosecond-precision `TIMESTAMP WITH TIME ZONE`.

DuckDB does not yet support this natively (see: [Timestamp Types](https://duckdb.org/docs/current/sql/data_types/timestamp), however that page has a documented warning:

> Since there is not currently a `TIMESTAMP_NS WITH TIME ZONE` data type, external columns with nanosecond precision and `WITH TIME ZONE` semantics, e.g., [Parquet timestamp columns with isAdjustedToUTC=true](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#instant-semantics-timestamps-normalized-to-utc), are converted to `TIMESTAMP WITH TIME ZONE` and thus lose precision when read using DuckDB.

This suggests that the expected behavior of DuckDB when encountering this type is to truncate to microsecond precision, interpreting such fields as `TIMESTAMP WITH TIME ZONE`.

My use case is that I am building software that creates Iceberg tables and I would like to start using Iceberg's `timestamptz_ns` type, but my users use DuckDB (among other query engines) to access the tables. Currently, that fails with the error: `Invalid Configuration Error: Unrecognized primitive type: timestamptz_ns`

**Describe the solution you'd like**
I would like duckdb-iceberg to support Iceberg's `timestamptz_ns` type by truncating to `TIMESTAMP WITH TIME ZONE`. This would allow the table to be created with nanosecond-precision timestamps that can be read by engines that support them, but remain usable from DuckDB with the same precision we have today.

Then, when DuckDB itself eventually supports `TIMESTAMP_NS WITH TIME ZONE`, duckdb-iceberg can cleanly migrate to it.

**Describe alternatives you've considered**
The only viable workaround is for users to explicitly opt in to (or out of) nanosecond-precision timestamps which is a degree of low-level configuration not required today.

jingdaws:
## Summary

  Fixes #1019.

  This adds Iceberg v3 `timestamptz_ns` support by mapping it to DuckDB's native `TIMESTAMPTZ_NS`.

  ## Changes

  - Parse Iceberg `timestamptz_ns` as DuckDB `TIMESTAMPTZ_NS`
  - Write REST catalog schemas using Iceberg `timestamptz_ns`
  - Serialize and deserialize `timestamptz_ns` bounds
  - Support `year`, `month`, `day`, `hour`, and `bucket` transforms for `TIMESTAMPTZ_NS`
  - Align nanosecond timestamp bucket hashing with Iceberg spec vectors

  ## Tests

  - `test/sql/local/iceberg_scans/timestamptz_ns_type.test`
  - `test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/partitions/temporal/
  test_timestamptz_ns_transforms.test`
  - `test/sql/local/catalog_test_config_setup/catalog_agnostic/functions/test_iceberg_bucket.test`

  Local checks run:

  - `git diff --check`
  - `clang-format --sort-includes=0 -style=file --dry-run -Werror ...`
  - `./build/debug/test/unittest test/sql/local/iceberg_scans/timestamptz_ns_type.test`
  - `./build/debug/test/unittest test/sql/local/catalog_test_config_setup/catalog_agnostic/functions/
  test_iceberg_bucket.test`
  - `./build/debug/test/unittest test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/partitions/temporal/
  test_timestamptz_ns_transforms.test --test-config /tmp/duckdb-iceberg-fixture-localhost.json`


src/core/expression/iceberg_hash.cpp
```diff
@@ -209,9 +209,14 @@ int32_t IcebergHash::HashTime(dtime_t t) {
 	return HashInt64(t.value);
 }
 
-//! Hash timestamp_ns value (Iceberg spec: int64 nanoseconds from epoch)
+//! Hash timestamp_ns bucket values as microseconds from epoch.
```
Tishj, line 212: > hashLong(microsecsFromUnixEpoch(v))

> Nanosecond timestamps must be converted to microsecond precision before hashing to ensure timestamps have the same hash value.

Sounds correct 👍

test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/partitions/temporal/test_timestamptz_ns_transforms.test
```diff
@@ -0,0 +1,61 @@
...
+require parquet
+
+require iceberg
+
+require httpfs
+
+require icu
+
+statement ok
+SET TimeZone='UTC';
```
Tishj: Can you set this to anything other than UTC? So we can actually verify that the timezone offset doesn't sneak in when serialized somehow?

Timestamp types with timezone in DuckDB are stored as instants since UTC, affected only by this setting when entering/leaving the system (converting input to UTC and rendering output with the current timezone offset applied)

test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/partitions/temporal/test_timestamptz_ns_transforms.test
```diff
@@ -0,0 +1,61 @@
...
+
+statement ok
+create schema if not exists my_datalake.default;
+
+statement ok
+drop table if exists my_datalake.default.test_timestamptz_ns_transforms;
+
+statement ok
+CREATE TABLE my_datalake.default.test_timestamptz_ns_transforms (id INT, val TIMESTAMPTZ_NS)
+PARTITIONED BY (year(val), month(val), day(val), hour(val))
```
Tishj: Can we extend this test with a `bucket` transform on the `timestamptz_ns` column? (can recreate the table and do it underneath the existing one)
We've verified that the transform produces the correct results, but we don't have a partition with it

Tishj:
Thanks, I would just like to see a couple more tests:
- ALTER TYPE `date` -> `timestamptz_ns` should fail
- ALTER TABLE ADD COLUMN with a `timestamptz_ns` type with a non-null DEFAULT value, after the table already has inserts (testing `initial-default`)
- CREATE TABLE with `timestamptz_ns` column with a non-null DEFAULT value, followed by an insert of `DEFAULT VALUES` (testing `write-default`)
- ALTER TABLE SET SORTED BY <transform>(col) tests, where transform is all the transforms we can partition on as well (hour, day, year, month, bucket) - followed by an insert, and verifying that sorting by `_row_id` is the same result as sorting by the `timestamptz_ns` column

Also it would be great to verify with Spark (`test/python`) that the data produced by us can be read back by them

jingdaws:
@Tishj Thanks for the review. I pushed an update with the requested SQL coverage: non-UTC timezone, bucket partitioning, DATE -> TIMESTAMPTZ_NS ALTER TYPE failure, initial-default/write-default, and SET SORTED BY for year/month/day/hour/bucket.

  I also moved the new write coverage to fixture-specific tests. Nessie/Lakekeeper currently reject this case at the catalog layer before DuckDB’s write behavior is exercised, so fixture-only keeps the test focused on this PR.

  I checked Spark readback as well. With the project’s Spark 4 setup (PySpark 4.0.1 + Iceberg Spark runtime 1.10.0), Spark currently fails schema conversion for `timestamptz_ns`, so I did not add a Spark readback test.

jingdaws:
Thanks for adding the extra test coverage and for splitting it by catalog capability. I pushed a small CI follow-up: the PyIceberg timestamptz_ns test now checks raw nanosecond values directly, and the bounds read test skips catalogs where the setup table is already skipped. I also moved the catalog helper to conftest.py for reuse.

Tishj:
Thanks!

PR #1152, opened 2026-07-06
Guard against cache miss when renaming a table

cjnoname:
Small robustness fix for a FIXME in `RenameTable`.

`IcebergTransaction::RenameTable` migrates the metadata cache from the old table key to the new one:
```cpp
auto cache = table_request_cache.Get(client_context, table_key, cache_guard, false);
table_request_cache.SetOrOverwriteInternal(..., cache->expire_timestamp, std::move(cache->load_table_result));
```
`Get` returns `nullptr` on a cache miss (it does not throw), so the following `cache->...` is a null dereference when the renamed table's entry is not in the cache. This now skips the migration when the entry is absent - there is nothing to migrate, and the next access under the new name repopulates it.

The FIXME said "just like the other place" - that other place is `IcebergTableInformation::Copy` (on the rename call path), which had the same pattern guarded only by a debug-only `D_ASSERT`, so a release build would null-deref on a miss. Since `Copy` genuinely needs the cached result, it now throws `InvalidConfigurationException` on a miss instead of dereferencing null (and avoids `InternalException`, which would invalidate the session).

No new test: the miss is very hard to reach from SQL (the binder resolves - and thus caches - the source table before a RENAME), so I could not construct a reproducer that exercises the miss path without the cache being repopulated first. Existing rename coverage (`rename_table_test.test`, 110 assertions) confirms the hit path is unchanged.

### DIFF
--- a/src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
+++ b/src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
@@ -600,7 +600,12 @@ IcebergTableInformation IcebergTableInformation::Copy(ClientContext &context) co
 	{
 		lock_guard<std::mutex> cache_lock(catalog.table_request_cache.Lock());
 		auto cached_result = catalog.table_request_cache.Get(context, table_key, cache_lock, false);
-		D_ASSERT(cached_result);
+		if (!cached_result) {
+			throw InvalidConfigurationException(
+			    "Cannot copy table '%s': its metadata is not present in the cache. It must be loaded before it is "
+			    "copied (e.g. renamed).",
+			    table_key);
+		}
 		auto &cached_table_result = *cached_result->load_table_result;
 		ret.InitializeFromLoadTableResult(cached_table_result, false);
 	}
--- a/src/catalog/rest/transaction/iceberg_transaction.cpp
+++ b/src/catalog/rest/transaction/iceberg_transaction.cpp
@@ -902,6 +902,8 @@ IcebergTableInformation &IcebergTransaction::RenameTable(IcebergTableInformation
 	lock_guard<mutex> cache_guard(table_request_cache.Lock());
 	auto cache = table_request_cache.Get(client_context, table_key, cache_guard, false);
 	if (cache) {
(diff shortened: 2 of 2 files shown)


Tishj:
Ah, that is indeed an oversight, but that comment was more a concern of overwriting a cached state

Imagine this:

Create table tbl;

Con1:
Begin

Con2:
Begin

Con1:
Create table tbl2;

Con1:
Select * from tbl2; -- populate cache

Con1:
Commit

Con2:
Rename tbl to tbl2; -- overwrite cache

Con2;
Commit

----------------------------------------------

Con1 doesn't even have to commit yet, it has already had to submit this create to the catalog, and therefore it *is* already committed

cjnoname:
Thanks, that clarifies the FIXME - I read it as the null-deref, but you mean the cache-overwrite hazard: rename migrating into a new_table_key that a concurrent, already-committed transaction (your Con1) has populated, so SetOrOverwriteInternal blindly clobbers tbl2's cache with tbl's metadata.

Two separate things then:

1. The null-deref this PR fixes is still a real oversight (Get returns nullptr on a miss, dereferenced unconditionally), so it's worth keeping regardless.

2. The overwrite hazard is the actual FIXME. Note there's a guard for the non-concurrent case: AlterTableType::RENAME_TABLE checks new_name existence and throws "already exists" before reaching RenameTable (iceberg_schema_entry.cpp:502-511). Your scenario slips through because tbl2 doesn't exist when Con2 does the lookup - it's created+committed by Con1 afterwards. So the migration overwrites a cache entry owned by another transaction, and ideally Con2's rename should conflict at commit via the optimistic-concurrency assertions rather than silently winning in the cache.

That's a more involved concurrency-correctness fix (and hard to cover with a deterministic test locally). Do you want me to expand this PR to handle the overwrite case (e.g. don't overwrite a cache entry created by a different transaction, and/or rely on the commit-time assertion), or keep #1152 scoped to the null-deref and track the overwrite hazard separately? I'm happy to do the former if you can point me at how you'd want the commit-time conflict surfaced.

PR #1153, opened 2026-07-07
Fix cloud tests, bump aws

Tmonster:
Since we have added Retries on insert, some cloud tests have different results

Because of https://github.com/duckdblabs/duckdb-internal/issues/9540 we've had to skip certain tests


Tishj (approved):
Thanks

Unfortunate that we have to skip some tests, but I'll trust your judgment

PR #1156, opened 2026-07-07
[Dev] Separate `IcebergTableUpdate::CreateUpdate` from transaction-local version of the table metadata

Tishj:
Salvage parts of #1117 

`IcebergTableUpdate` should contain all data it needs to represent the change (not reference a piece of `IcebergTableMetadata`), this is necessary to enable retries of updates.

### Misc change

`identifier_field_ids` of `IcebergTableSchema` was apparently not being copied/converted by existing methods


Tmonster (approved):
LGTM thanks!

PR #1158, opened 2026-07-08
Vended credentials refresh

lnkuiper:
Implementation of credential refresh by @Tmonster with a round of review and polishing by @Tishj and me.

I've first implemented credential refresh in httpfs, which I've sent PRs for. Then I bumped the httpfs hash in the DuckDB core v1.5 branch. Now that this is done, this PR bumps the DuckDB core ref here (so we can pull in the httpfs changes) and adds the functionality/testing. After this is done I can bump Iceberg in the DuckDB core v1.5 branch.

The PR diff ended up quite large but that is mostly due to testing. I've added `scripts/vended_credentials_refresh_proxy.py` which allows us to properly test 403's using `mitmproxy`. This is used in `test/sql/local/catalog_custom_setup/fixture/vended_credentials_refresh.test` which gets triggered by the newly added `Run vended credential refresh tests` CI run.

### DIFF
--- a/src/catalog/rest/storage/iceberg_table_secret_provider.cpp
+++ b/src/catalog/rest/storage/iceberg_table_secret_provider.cpp
@@ -71,8 +71,13 @@ static unique_ptr<BaseSecret> BuildVendedSecret(CreateSecretInput &input) {
 	if (input.type == "r2") {
 		auto account_id_entry = input.options.find("account_id");
 		if (account_id_entry != input.options.end()) {
-			secret->secret_map["endpoint"] = account_id_entry->second.ToString() + ".r2.cloudflarestorage.com";
-			secret->secret_map["url_style"] = "path";
+			// Match httpfs' R2 defaults, but let explicitly vended options override them.
+			if (input.options.find("endpoint") == input.options.end()) {
+				secret->secret_map["endpoint"] = account_id_entry->second.ToString() + ".r2.cloudflarestorage.com";
+			}
+			if (input.options.find("url_style") == input.options.end()) {
+				secret->secret_map["url_style"] = "path";
+			}
 		}
 	}
 
--- a/src/catalog/rest/storage/iceberg_table_secret_provider.cpp
+++ b/src/catalog/rest/storage/iceberg_table_secret_provider.cpp
@@ -11,7 +11,6 @@
 
 #include "catalog/rest/api/catalog_api.hpp"
 #include "catalog/rest/catalog_entry/schema/iceberg_schema_entry.hpp"
-#include "catalog/rest/catalog_entry/table/iceberg_table_entry.hpp"
 #include "catalog/rest/catalog_entry/table/iceberg_table_information.hpp"
 #include "catalog/rest/iceberg_catalog.hpp"
 #include "catalog/rest/storage/authorization/oauth2.hpp"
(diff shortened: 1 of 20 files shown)


Tishj (approved):
Looks good, I wonder why CI is failing but it seems unrelated

src/catalog/rest/storage/iceberg_table_secret_provider.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+static unique_ptr<BaseSecret> BuildVendedSecret(CreateSecretInput &input) {
+	auto secret = make_uniq<KeyValueSecret>(input.scope, input.type, input.provider, input.name);
+	secret->redact_keys = {"secret", "session_token", "bearer_token", "http_proxy_password"};
+
+	if (input.type == "r2") {
+		auto account_id_entry = input.options.find("account_id");
+		if (account_id_entry != input.options.end()) {
+			secret->secret_map["endpoint"] = account_id_entry->second.ToString() + ".r2.cloudflarestorage.com";
+			secret->secret_map["url_style"] = "path";
```
Tmonster: should this be path? I feel like this could come from input.options or the original secret right?

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -276,10 +276,31 @@ static rest_api_objects::TableRequirement CreateAssertNoSnapshotRequirement() {
...
+		for (auto &secret_name : created_secrets) {
+			(void)secret_manager.DropSecretByName(context, secret_name, OnEntryNotFound::RETURN_NULL,
+			                                      SecretPersistType::TEMPORARY, SecretManager::TEMPORARY_STORAGE_NAME);
+		}
+
+		if (started_transaction) {
+			context.transaction.Commit();
+		}
+	} catch (...) {
+		if (started_transaction && context.transaction.HasActiveTransaction()) {
```
Tmonster: Can we have a `DUCKDB_LOG_DEBUG()` here with some exception information?

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -390,6 +411,8 @@ TableTransactionInfo IcebergTransaction::GetTransactionRequest(IcebergTransactio
 void IcebergTransaction::Commit() {
 	if (transaction_updates.empty() && created_schemas.empty() && deleted_schemas.empty() &&
 	    schema_property_updates.empty()) {
+		// Read-only scans can create temporary vended storage secrets that later statements use
```
Tmonster: Is this in reference to the fact that we don't drop secrets?

test/sql/local/catalog_custom_setup/fixture/vended_credentials_refresh.test
```diff
@@ -0,0 +1,1024 @@
...
+query I
+SELECT count(*)
+FROM duckdb_secrets(redact=false)
+WHERE name LIKE '__internal_ic%empty_table%'
+----
+1
+
+# A real Iceberg table scan should refresh automatically installed vended
+# credentials too. The proxy maps this virtual table to table_unpartitioned,
+# lets metadata/manifest requests through with OLD_TABLE_SCAN_KEY, rejects the
```
Tmonster: I don't understand this test too much. The metadata requests to vended_table_scan_refresh works, but where does `NEW_TABLE_SCAN_KEY` come from for the revending? Is the catalog sending that information? Do we know that is how things happen in production?

Tmonster: It makes it seem like duckdb-iceberg intercepts some `NEW_TABLE_KEY` and then requests new credentials from there, is that something that is expected?

Tmonster (commented):
Thanks! 

just the question about where the other table names come from during credential vending

src/catalog/rest/transaction/iceberg_transaction.cpp
lnkuiper: Good idea, I've added logging statements

src/catalog/rest/transaction/iceberg_transaction.cpp
lnkuiper: No, these are cleaned up by the session cleanup

src/catalog/rest/storage/iceberg_table_secret_provider.cpp
lnkuiper: Yes, good call. I've added `path` only as a fallback now

test/sql/local/catalog_custom_setup/fixture/vended_credentials_refresh.test
lnkuiper: I think this is the weakest point. The test is too artificial. I'll make it more representative

lnkuiper: I've made the test better by creating actual tables (not virtual ones in the proxy server). Should be better now. I think we should also do nightly tests with a real server and wait for timeouts (instead of relying on proxy server) as we discussed.

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -292,13 +292,22 @@ void IcebergTransaction::DropSecrets(ClientContext &context) {
 		if (started_transaction) {
 			context.transaction.Commit();
 		}
+	} catch (std::exception &ex) {
```
Tmonster, line 300: This extra catch seems fishy. Looks like it does the exact same as the catch below?

Tmonster (commented):
question about the extra catch statement, but otherwise g2g

src/catalog/rest/transaction/iceberg_transaction.cpp
Tishj: The difference is that this one has a message
But we can improve upon this slightly by using ErrorData to hydrate the exception:
```c++
		ErrorData error(ex);
		auto message = error.Message();
```

Tishj:
Let's merge this when CI is green, thanks!

PR #1163, opened 2026-07-09
[Dev] Fix RENAME cache-invalidation, fix `IcebergTableInformation::Copy` weirdness

Tishj:
As mentioned in #1152 the actual problem is that we are touching the catalog-wide request cache for a RENAME TABLE while the transaction is still ongoing.

Fixing that made me realize that `IcebergTableInformation::Copy` was doing **way** too much for what the simple "copy" was insinuating, so that has also been fixed

### Misc changes
- `IcebergTableInformation::Copy` doesn't perform a refresh from catalog-wide cache anymore.
- Added the `AssertRefSnapshotId` `IcebergTableUpdate`, however it's left unused.
- Make `IcebergTableMetadata::GetSnapshot` and `GetLatestSnapshot` transactional, outputting the last snapshot visible by the transaction.
- Add check for a write-write conflict in INSERT/UPDATE/DELETE (actually, missing MERGE INTO?)
- Changed all timestamps from `timestamp_t` to `timestamp_ms_t`, as iceberg only saves `_ms` variables, and constantly having to change/downcast/use `.value` is incredibly brittle and a code smell


Tishj:
Hmm, there's a freshness issue that popped up because of this, that best shows up in `test/sql/local/catalog_test_config_setup/catalog_agnostic/transactions/test_transactional_updates.test`

In there we start a transaction, then commit a change in another transaction, and then perform an update in the transaction we started.
Result should be that the change made by our transaction is isolated and the other update made outside of the transaction isn't visible.
But that last assertion doesn't hold.

We could immediately turn this into a write-write conflict, if we detect that we're making a change to a table that has already received a newer update.
Since `con1` hasn't performed a request to load the table before the other change happened, there is a snapshot that started after `con1` has started its transaction, attempting to make an update/insert/delete or otherwise create a snapshot should immediately error.

But that feels like its treating the symptom, not the root cause
It's easier to just use the latest transaction-compatible snapshot as the base always, then it'll just fail at commit, and we can potentially retry it against the fresh state once it fails.

Tishj:
^ this comes with the problem that we only create snapshots at commit time, at which point the transaction start time is no longer available.
So we end up assigning the wrong parent-snapshot-id to the snapshot we create.

The solution I want to implement for this is checking for a write-write conflict (attempting to make an insert/update/delete on a table where the transaction timestamp is earlier than the latest snapshot)

Tishj:
For now, alters made in a transaction where another alter by a third-party already happened after the transaction started will result in a write-write conflict

In the future this should be alleviated by correctly stamping an update with the transaction start time, so we can grab the correct state at commit time, instead of pretending we constructed our update against the latest available state

PR #1171, opened 2026-07-11
Prune all-null data files for constant comparisons

tomlarkworthy:
I found "WHERE foo = 'blah'" does not skip files where foo is always NULL, but `WHERE foo = 'blah' and foo IS NOT NULL` did, and I trace that to this missing predicate. We were missing the inference that a constant implies NOT NULL and ranges that only match NULL cannot bound a constant.

(first PR attempt ran into difficult because DISTINCT foo = 'blah') does need to consider if NULL values exist

### DIFF
--- a/src/planning/pruning/iceberg_predicate.cpp
+++ b/src/planning/pruning/iceberg_predicate.cpp
@@ -61,6 +61,12 @@ static bool MatchBoundsConstant(const Value &constant, ExpressionType comparison
 		return false;
 	}
 
+	if (!stats.has_not_null && comparison_type != ExpressionType::COMPARE_DISTINCT_FROM &&
+	    comparison_type != ExpressionType::COMPARE_NOT_DISTINCT_FROM) {
+		// An all-null column (Iceberg omits its bounds) only matches a null-safe comparison.
+		return false;
+	}
+
 	if (!stats.upper_bound || !stats.lower_bound) {
 		// we do not have upper or lower bounds, assume the file matches.
 		return true;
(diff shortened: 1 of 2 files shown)


src/planning/pruning/iceberg_predicate.cpp
```diff
@@ -61,6 +61,12 @@ static bool MatchBoundsConstant(const Value &constant, ExpressionType comparison
 		return false;
 	}
 
+	if (!stats.has_not_null && comparison_type != ExpressionType::COMPARE_DISTINCT_FROM &&
+	    comparison_type != ExpressionType::COMPARE_NOT_DISTINCT_FROM) {
```
tomlarkworthy, line 65: arguably it is safer to maintain by listing the ExpressionTypes that are valid for this optimization but it was quite a big list and so I went for code succinctness, but happy to change this to being an explicit allowlist

src/planning/pruning/iceberg_predicate.cpp
```diff
@@ -61,6 +61,12 @@ static bool MatchBoundsConstant(const Value &constant, ExpressionType comparison
...
+		// An all-null column (Iceberg omits its bounds) only matches a null-safe comparison.
```
Tishj: Bounds not being present doesn't mean that it's all-null

test/sql/local/iceberg_scans/is_not_null.test
```diff
@@ -67,7 +67,19 @@ select * from my_datalake.default.is_not_null WHERE value IS NOT NULL ORDER BY i
 6	baz
 8	blah
 
-# Using the logs (which contains the data_file path), join it against the iceberg metadata to sum the record_counts of the skipped data files
+statement ok
+CALL truncate_duckdb_logs();
+
+# A comparison against a non-null constant cannot match an all-null column. Iceberg writes no
```
Tishj: As mentioned above, omitted bounds don't say anything about null or not, that's what `null_value_counts` and `value_counts` are for.
You correctly use `has_not_null` in the file pruning logic, but the comments are wrong

Tishj (changes requested):
Thanks, I think the pruning logic is correct, just correct the comments and we can merge this 👍

tomlarkworthy:
yes fixed the out-of-date comments.

Tishj (approved):
Thanks!

PR #1172, opened 2026-07-11
feat: populate manifest column stats for iceberg_rewrite_data_files

The-Alchemist:
## Summary

This is a surgical follow-up to #1035 (`iceberg_rewrite_data_files`). Rewritten data files currently land in the manifest with only `record_count`, `file_size_in_bytes`, and the partition tuple, NOT the per-column statistics from the internal `COPY ... (RETURN_STATS)`.

As a result, compacted files lacked `lower_bounds` / `upper_bounds` / `value_counts` / `null_value_counts` / `column_sizes`, so file-level pruning and Iceberg metrics (`write.metadata.metrics.column.*`) stop working for anything produced by a rewrite. This was called out as a known limitation in the original PR.

This change populates those stats for rewritten files, reusing the exact code path that `INSERT` already uses.

## Changes

- Extract the `RETURN_STATS` → `IcebergDataFile` stats population out of `IcebergInsertGlobalState::AddFiles` into a shared helper `PopulateDataFileColumnStatsFromReturnStats` (`src/storage/statistics/iceberg_data_file_stats.{hpp,cpp}`).
- `INSERT` now delegates to the helper (pure refactor, no behavior change — mostly deletions).
- `BuildRewriteManifestEntry` / the rewrite `Sink` now read the `column_statistics` map from the COPY result and call the shared helper, so rewritten files carry identical bounds/metrics (including per-column metrics modes, variant bounds, and geometry bounding boxes). Rewritten files also inherit the table `sort_order_id` when configured.
- Add `rewrite_data_files_column_stats.test` asserting rewritten files retain column bounds and that pruning still skips non-matching files.

## Notes / follow-ups (not addressed here)

- `nan_value_counts` is still not populated (same limitation as INSERT; blocked on duckdb/duckdb recording NaN counts).

### DIFF
--- a/src/include/storage/statistics/iceberg_data_file_stats.hpp
+++ b/src/include/storage/statistics/iceberg_data_file_stats.hpp
@@ -1,0 +1,23 @@
+//===----------------------------------------------------------------------===//
+//! Shared helpers for populating Iceberg data-file column metrics from COPY
+//! RETURN_STATS. Used by INSERT and iceberg_rewrite_data_files.
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/common/types/value.hpp"
+#include "duckdb/main/client_context.hpp"
+
+#include "core/metadata/manifest/iceberg_manifest.hpp"
+#include "core/metadata/iceberg_table_metadata.hpp"
+
+namespace duckdb {
+
+//! Populate lower/upper bounds, value/null counts, and column sizes on
+//! `data_file` from one COPY RETURN_STATS `column_statistics` map value.
+//! Respects write.metadata.metrics.* and enforces NOT NULL constraints.
+void PopulateDataFileColumnStatsFromReturnStats(ClientContext &context, IcebergDataFile &data_file,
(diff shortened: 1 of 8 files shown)


src/include/storage/statistics/iceberg_data_file_stats.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+#include "core/metadata/manifest/iceberg_manifest.hpp"
+#include "core/metadata/iceberg_table_metadata.hpp"
+
+namespace duckdb {
+
+//! Populate lower/upper bounds, value/null counts, and column sizes on
+//! `data_file` from one COPY RETURN_STATS `column_statistics` map value.
+//! Respects write.metadata.metrics.* and enforces NOT NULL constraints.
+void PopulateDataFileColumnStatsFromReturnStats(ClientContext &context, IcebergDataFile &data_file,
```
Tishj: Please no floating methods, wrap it into a struct as a static method instead

The-Alchemist: done!

Tishj:
Thanks!

PR #1173, opened 2026-07-11
Enable metadata-only MIN/MAX from Iceberg manifest column bounds (e.g., `MAX(updated_at)`)

The-Alchemist:
## Summary
- Wire `get_partition_stats` to expose per-column min/max from manifest `data_file.lower_bounds` / `upper_bounds` via a new `IcebergPartitionRowGroup`, so DuckDB's `StatisticsPropagator::TryExecuteAggregates` can fold `SELECT MIN(col)` / `SELECT MAX(col)` into constants without scanning Parquet.
- Exact fold for numeric/temporal types only (not `FLOAT`/`DOUBLE`/`VARCHAR`); bail out when any file lacks usable bounds or matching delete manifests are present.
- Uses **manifest column bounds**, not [Iceberg partition-statistics files](https://iceberg.apache.org/spec/#partition-statistics) (those carry counts/sizes only, no per-column min/max). Independent of #1023 (copies a leaner `IcebergPartitionRowGroup` focused on min/max; reconcile if both land).

## Test plan
- [x] `test/sql/local/iceberg_scans/aggregate_min_max_metadata.test` — correctness vs oracle, `EXPLAIN` shows `COLUMN_DATA_SCAN` for bigint/timestamp folds, fallbacks for varchar/float/`WHERE`; verify nested type behavior
- [x] `test/sql/local/iceberg_scans/scan_order_limit_pruning.test` still passes

### DIFF
--- a/src/planning/iceberg_multi_file_list.cpp
+++ b/src/planning/iceberg_multi_file_list.cpp
@@ -21,6 +21,7 @@
 #include "duckdb/execution/executor.hpp"
 #include "duckdb/optimizer/filter_combiner.hpp"
 #include "duckdb/function/scalar/struct_utils.hpp"
+#include "duckdb/storage/statistics/numeric_stats.hpp"
 #include "duckdb/storage/table/row_group_reorderer.hpp"
 
 #include "planning/iceberg_multi_file_reader.hpp"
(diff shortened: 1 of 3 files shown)


src/planning/iceberg_multi_file_list.cpp
```diff
@@ -544,15 +545,125 @@ const IcebergManifestFile &IcebergMultiFileList::GetManifestFileForEntry(const B
 	}
 }
 
+IcebergPartitionRowGroup::IcebergPartitionRowGroup(const vector<unique_ptr<IcebergColumnDefinition>> &schema_p,
+                                                   vector<reference<const IcebergDataFile>> data_files_p)
+    : schema(schema_p), data_files(std::move(data_files_p)) {
+}
+
+bool IcebergPartitionRowGroup::SupportsExactMinMaxBounds(const LogicalType &type) {
```
Tishj, line 559: I wonder if this needs to be recursive, we could for example have a struct that contains a FLOAT field
This is probably a good test to add

The-Alchemist: done!  wdyt?

src/planning/iceberg_multi_file_list.cpp
```diff
@@ -567,10 +685,25 @@ void IcebergMultiFileList::GetStatistics(vector<PartitionStatistics> &result) co
 		count += manifest.added_rows_count;
 	}
 
+	//! Materialize data-file entries so column bounds are available for MIN/MAX.
+	idx_t materialized = 0;
+	while (GetDataFile(materialized, guard)) {
+		materialized++;
```
Tishj, line 712: I'm worried this happens before filter pushdown, in which case we are now aggressively materializing all metadata, which is not ideal.

The-Alchemist: good point.  my latest commit delays materialization until necessary, although that increases the complexity

Tishj:
I think we'll need more regression tests to really test whether this doesn't degrade performance.
We've taken care to not expand all metadata up front, I am worried that this would negate that effort.

Issue #826, opened 2026-03-20
Feature request, copy to iceberg add support for append and overwrite :)


djouallah:
currently it seems it works only when the path is empty, having the option to support append and overwirte will be awesome if you don't have access to a catalog

Issue #620, opened 2025-12-08
Feature Request: Support for INSERT OVERWRITE / Atomic Table Overwrite


carc-prathyush-shankar:
### Summary
Add INSERT OVERWRITE support to enable atomic single-snapshot table overwrites, matching the behavior available in Spark, Trino, and PyIceberg. There currently does not seem to be a way to atomically replace all rows in a table with new data in a single commit.

### Use Case
Batch ETL pipelines commonly need to fully replace table contents on each run. For example:

```
INSERT OVERWRITE my_catalog.my_schema.target_table
SELECT * FROM source_table WHERE ...
```

This pattern is standard in workflows where a batch job reads from source tables, applies transformations, and overwrites a target table.

### Additional Context
The Iceberg spec already supports the overwrite snapshot operation type, and I see it's defined in the codebase (`IcebergSnapshotOperationType::OVERWRITE`). The `AddUpdateSnapshot` function already creates overwrite snapshots for UPDATE operations, so the underlying infrastructure may partially exist.

djouallah:
currently i use truncate and insert but sometimes, insert does not works and i endup with an empty table.

paultmathew:
@Tishj @Tmonster when you get a chance, could you take a look at the direction in #1178? Mainly want to check the approach before I invest further.

The core idea: a DELETE that removes every live row of a data file drops the file via a metadata-only manifest edit instead of writing a positional delete / DV, reusing the `IcebergManifestDeletes` → `RewriteManifestFile` path that `iceberg_rewrite_data_files` already uses.

One thing I want to be upfront about on the design. This PR detects the whole-file delete **at flush time** — it still scans the matching rows and only then notices they cover the entire file. So it fixes downstream read amplification (no delete files left for later scans/merges to apply), but it doesn't avoid the read cost of the delete itself. The more efficient version is a **plan-time** path that identifies fully-covered files/partitions straight from manifest metadata and skips the scan entirely (Iceberg/Spark's strict-projection `deleteWhere` model). I'm treating that as a **future phase** — it needs strict-projection onto the partition spec and a no-scan plan shape, whereas the flush-time approach here is exact and simple. It seems like a decent, if less efficient, first step, and the commit-layer plumbing (invalidate-only snapshot, file-scoped delete-file cleanup) is shared with the plan-time path either way.

Three things I'd value your read on:

1. Does this fit what you had in mind for #620 / #826?
[...]

PR #1178, opened 2026-07-14
Metadata-only delete when a DELETE removes every live row of a data file

paultmathew:
## What / why

I hit this building a delete + insert incremental event-stream pipeline (repeatedly deleting a partition and re-inserting it). Every `DELETE` writes merge-on-read delete files/DVs even when the predicate removes every live row of a data file, so the dead files pile up and every later `scan` / `MERGE` / `DELETE` still has to read them plus their delete files. When a delete fully empties a file, I'd rather drop it via metadata.

This is **Phase 1** and I'd like a sanity check on direction before going further.

## Phase 1 (this PR) — drop fully-deleted files at flush time

We already scan every matching row, so at flush we know the exact deleted positions per file. When `existing + new deleted positions == record_count` (positions are 0-based ordinals, so this holds exactly when no live row survives), invalidate the data file instead of writing a delete file; otherwise fall back to the current path.

- Exact check (no stats/null/NaN edge cases), and naturally covers whole-partition deletes.
- Reuses the existing `IcebergManifestDeletes` -> `RewriteManifestFile` commit path (the one `iceberg_rewrite_data_files` uses).
- Adds an invalidate-only `AddSnapshot` path for when a delete only drops whole files (no new manifest entries).
- Also invalidates delete files provably scoped to the dropped data file (v3 deletion vectors, and DuckDB's file-scoped positional deletes via `referenced_data_file`). Multi-file positional delete files written by other engines still apply to surviving files, so they're left for that engine's maintenance to reclaim once dangling.

## Phase 2 (follow-up) — skip the scan via strict metrics

On top of Phase 1, add a strict metrics evaluator (the strict dual of the inclusive `IcebergPredicate::MatchBounds`) to check per-file stats against the predicate *before* scanning, so provably-fully-covered files skip the scan entirely; partial files keep the MOR path. Best-effort with a safe fallback.

## Tests

Fixture tests under `.../catalog_agnostic/delete/`:
- `test_metadata_only_full_file_delete.test` — partitioned v2 (positional) + v3 (DV): whole-partition delete writes no delete file; partial delete does.
- `test_metadata_only_unpartitioned.test` — multiple/single data files, whole vs partial, and a file emptied across two deletes (existing deletes + new == record_count) which also drops the dangling file-scoped delete file.
- `test_metadata_only_with_equality_deletes.test` (gated on `EQUALITY_DELETE_WRITES_ENABLED`) — safety: with equality deletes present, deleting the remaining live rows is not mis-detected as a whole-file drop; it falls back to MOR and reads stay correct.

## Known limitations (deliberate for this phase)

- Snapshot summary counts (`deleted-data-files` / `deleted-records`) are not decremented for the metadata drop (pre-existing accounting gap); reads/cardinality are correct.
- No write-isolation conflict detection (same as today's deletes).

## Questions

1. Does this direction fit what you had in mind for #620 / #826?
2. Prefer keeping it transparent inside the `DELETE` plan (Phase 1 at flush, Phase 2 at plan time), or opt-in?
3. Anything on summary correctness or write isolation you'd want handled here vs. deferred?

### DIFF
--- a/src/catalog/rest/transaction/iceberg_transaction_data.cpp
+++ b/src/catalog/rest/transaction/iceberg_transaction_data.cpp
@@ -212,20 +212,27 @@ void IcebergTransactionData::AddUpdateSnapshot(vector<IcebergManifestEntry> &&de
 	const auto sequence_number = last_sequence_number + alters.size() + 1;
 
 	auto &fs = FileSystem::GetFileSystem(context);
-	auto delete_manifest_metadata =
-	    IcebergManifestMetadata::FromTableMetadata(table_metadata, IcebergManifestContentType::DELETE);
-	auto data_manifest_metadata =
-	    IcebergManifestMetadata::FromTableMetadata(table_metadata, IcebergManifestContentType::DATA);
-
-	auto delete_manifest_file = IcebergManifestListEntry::CreateFromEntries(
-	    fs, sequence_number, table_metadata, delete_manifest_metadata, std::move(delete_files), next_row_id);
-	// Add a manifest_file for the new insert data
-	auto data_manifest_file = IcebergManifestListEntry::CreateFromEntries(
-	    fs, sequence_number, table_metadata, data_manifest_metadata, std::move(data_files), next_row_id);
 
 	auto add_snapshot = make_uniq<IcebergAddSnapshot>(table_info);
-	add_snapshot->AddManifestFile(std::move(delete_manifest_file));
-	add_snapshot->AddManifestFile(std::move(data_manifest_file));
+
+	//! The delete side of an UPDATE can be a metadata-only whole-file delete, which produces no
+	//! delete-file entries (only altered_manifests). Only add a manifest file when it has entries;
+	//! an empty manifest is invalid.
+	if (!delete_files.empty()) {
+		auto delete_manifest_metadata =
+		    IcebergManifestMetadata::FromTableMetadata(table_metadata, IcebergManifestContentType::DELETE);
+		auto delete_manifest_file = IcebergManifestListEntry::CreateFromEntries(
+		    fs, sequence_number, table_metadata, delete_manifest_metadata, std::move(delete_files), next_row_id);
+		add_snapshot->AddManifestFile(std::move(delete_manifest_file));
+	}
+	// Add a manifest_file for the new insert data
+	if (!data_files.empty()) {
+		auto data_manifest_metadata =
+		    IcebergManifestMetadata::FromTableMetadata(table_metadata, IcebergManifestContentType::DATA);
+		auto data_manifest_file = IcebergManifestListEntry::CreateFromEntries(
+		    fs, sequence_number, table_metadata, data_manifest_metadata, std::move(data_files), next_row_id);
+		add_snapshot->AddManifestFile(std::move(data_manifest_file));
+	}
 	add_snapshot->altered_manifests = std::move(altered_manifests);
 
 	alters.push_back(*add_snapshot);
--- a/src/catalog/rest/transaction/iceberg_transaction_data.cpp
+++ b/src/catalog/rest/transaction/iceberg_transaction_data.cpp
@@ -185,6 +185,9 @@ void IcebergTransactionData::AddSnapshot(IcebergSnapshotOperationType operation,
 		auto manifest_file = IcebergManifestListEntry::CreateFromEntries(
 		    fs, temp_sequence_number, table_metadata, manifest_metadata, std::move(data_files), next_row_id);
 		add_snapshot->AddManifestFile(std::move(manifest_file));
+	} else if (operation != IcebergSnapshotOperationType::DELETE) {
+		//! A snapshot with no new manifest entries is only valid for a metadata-only delete.
+		throw InternalException("AddSnapshot: empty data_files is only valid for a metadata-only DELETE");
 	}
 
 	// make sure we are still inserting into the current schema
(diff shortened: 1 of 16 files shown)


src/catalog/rest/transaction/iceberg_transaction_data.cpp
```diff
@@ -159,29 +159,34 @@ void IcebergTransactionData::AddSnapshot(IcebergSnapshotOperationType operation,
...
+		};
+
+		auto temp_sequence_number = table_metadata.last_sequence_number + alters.size() + 1;
+
+		auto &fs = FileSystem::GetFileSystem(context);
+		auto manifest_metadata = IcebergManifestMetadata::FromTableMetadata(table_metadata, manifest_content_type);
+		auto manifest_file = IcebergManifestListEntry::CreateFromEntries(
+		    fs, temp_sequence_number, table_metadata, manifest_metadata, std::move(data_files), next_row_id);
+		add_snapshot->AddManifestFile(std::move(manifest_file));
+	}
```
Tishj, line 191: We are missing an `else` here, it should make sure that this is a delete:
`if (operation != IcebergSnapshotOperationType::DELETE) { throw InternalException(...); }`
~~And also update the `DELETED_DATA_FILES` metrics for the new snapshot~~ (that should already be done by the "rewrite manifest" path we take for altered_manifests, right?)

src/catalog/rest/transaction/iceberg_transaction_data.cpp
```diff
@@ -207,20 +212,27 @@ void IcebergTransactionData::AddUpdateSnapshot(vector<IcebergManifestEntry> &&de
...
+	//! an empty manifest is invalid.
+	if (!delete_files.empty()) {
+		auto delete_manifest_metadata =
+		    IcebergManifestMetadata::FromTableMetadata(table_metadata, IcebergManifestContentType::DELETE);
+		auto delete_manifest_file = IcebergManifestListEntry::CreateFromEntries(
+		    fs, sequence_number, table_metadata, delete_manifest_metadata, std::move(delete_files), next_row_id);
+		add_snapshot->AddManifestFile(std::move(delete_manifest_file));
+	}
+	// Add a manifest_file for the new insert data
+	if (!data_files.empty()) {
```
Tishj: Can we verify that `data_files` is allowed to be empty? I would image the answer being no. If we perform an UPDATE, we'll always have to create new data.

src/planning/iceberg_multi_file_list.cpp
```diff
@@ -971,11 +980,21 @@ optional_ptr<const BoundIcebergManifestEntry> IcebergMultiFileList::GetDataFile(
...
+			shared_state->data_file_record_count[entry_path] = data_file.record_count;
+			shared_state->data_file_record_count[data_file.file_path] = data_file.record_count;
 
 			if (manifest_entry.status == IcebergManifestEntryStatusType::DELETED) {
 				continue;
 			}
 
+			//! Skip committed data files dropped by a metadata-only delete earlier in this
+			//! transaction (the manifest rewrite only lands at commit).
+			if (!shared_state->transaction_invalidated_files.empty() &&
```
Tishj:
```
if (shared_state->transaction_invalidated_files.count(entry_path)) { continue; }
if (shared_state->transaction_invalidated_files.count(data_file.file_path)) { continue; }
```
no need for an empty check, no need to make the condition that long

Tishj:
I'd like to see some tests that use MERGE INTO, both with INSERT and UPDATE as "WHEN [NOT] MATCHED" operations, and with both operations in one statement

src/catalog/rest/transaction/iceberg_transaction_data.cpp
paultmathew: `RewriteManifestFile` marks the data file `DELETED` but doesn't touch `DELETED_DATA_FILES`/`TOTAL_DATA_FILES`/`TOTAL_RECORDS` — those only come from `AddManifestFile` over `uncommitted_manifest_files`, which is empty for a pure metadata-only delete.

#1187 adds the plumbing (`snapshot_metrics` into `RewriteManifestFile`) but only wires up the size metrics (`RemoveFileSize`), not the counts. I can add the count decrement next to that call once #1187 lands, or here if you prefer.

paultmathew:
Pushed two follow-ups so count(*) and cardinality stay correct inside a transaction after a metadata-only full-file delete.

1. Discount pending metadata-only deletes. GetCardinality and GetStatistics read row counts straight from the manifest totals, which still include files dropped by an in-transaction metadata-only delete (the manifest rewrite only lands at commit). Both now subtract those pending-deleted rows via a shared helper. GetStatistics keeps its exact-count fast path instead of bailing to a scan — it only forces file enumeration when there's actually a pending delete to discount.

2. Subtract live rows, not the full record count. When the dropped file already had a delete file, the delete-manifest totals already discount the rows it removed — so subtracting the file's full record_count double-counted those rows and skewed the estimate low whenever other live files remained (a single dropped file was masked by the zero-floor clamp). The delete now carries the count of live rows it removed (its new deletes) into the invalidation, and the read path subtracts that; wholesale invalidations like compaction still fall back to the full record count. Added a transaction-local cardinality regression test that a plain metadata-only test wouldn't have caught.

src/catalog/rest/transaction/iceberg_transaction_data.cpp
paultmathew: Now that #1187 landed I pushed the file-count metrics for the rewrite path. Metadata-only whole-file deletes now record deleted-data-files/deleted-records and decrement the totals (and the delete-file equivalents when a dangling file-scoped delete file gets dropped), reusing the same accounting AddManifestListEntry does. Added a test covering the whole-file drop metrics plus a partial-delete contrast to confirm the merge-on-read path is untouched.

paultmathew:
@Tishj Thanks again for the review — summary of what's changed since your last pass:

- Empty manifests (your data_files question): AddSnapshot/AddUpdateSnapshot no longer build empty manifests. An UPDATE always writes data — now asserted, so data_files is never empty there; only the delete side can be metadata-only, and its delete manifest is skipped when it has no entries.
- MERGE coverage: added MERGE INTO tests for INSERT-only, UPDATE-only, and both matched/not-matched in a single statement, plus v3 and a partial (merge-on-read) UPDATE.
- count(*) correctness: pending metadata-only deletes are now discounted in both the cardinality estimate and the exact statistics. When the dropped file still has positional deletes, only its remaining live rows are subtracted, so there's no double counting.
- Snapshot summary metrics: metadata-only deletes now update deleted-data-files / deleted-records and the running totals, using the plumbing from #1187 (since merged).
- Self-deadlock fix: count(*)/stats on a table with an in-transaction metadata-only delete re-locked shared_state->lock (via GetTotalFileCount()) and hung the REST-catalog CI jobs. It now enumerates files under the lock already held.

Let me know if this looks okay or if you want to break this up.

PR #1187, opened 2026-07-17
Add file size tracking to snapshot metrics

fm100:
## Summary
Add Iceberg snapshot summary metrics:
- added-files-size
- removed-files-size
- total-files-size

Sizes are derived from manifest entries. For legacy snapshots without total-files-size, the value is reconstructed from live manifest entries.

## Testing
Added coverage for appends, positional deletes, and data-file rewrites. Build, formatting, and targeted tests pass.

### DIFF
--- a/src/catalog/rest/api/iceberg_add_snapshot.cpp
+++ b/src/catalog/rest/api/iceberg_add_snapshot.cpp
@@ -9,10 +9,10 @@
 #include "duckdb/main/extension/extension_loader.hpp"
 #include "duckdb/storage/external_file_cache/caching_file_system.hpp"
 #include "duckdb/common/types/uuid.hpp"
-#include "duckdb/common/operator/add.hpp"
 
 #include "core/metadata/manifest/iceberg_manifest_list.hpp"
 #include "catalog/rest/iceberg_table_set.hpp"
+#include "common/iceberg_utils.hpp"
 
 namespace duckdb {
 
@@ -113,16 +113,6 @@ void IcebergAddSnapshot::ConstructManifestList(IcebergManifestList &new_manifest
 	commit_state.manifests.clear();
 }
 
-static void AddLiveFileSize(int64_t &total_files_size, const IcebergDataFile &data_file) {
-	if (data_file.file_size_in_bytes < 0) {
-		throw InvalidConfigurationException("Iceberg content file has negative 'file_size_in_bytes': %lld",
-		                                    data_file.file_size_in_bytes);
-	}
-	if (!TryAddOperator::Operation(total_files_size, data_file.file_size_in_bytes, total_files_size)) {
-		throw InvalidConfigurationException("Iceberg live content file sizes exceed the supported BIGINT range");
-	}
-}
-
 static int64_t ReconstructTotalFilesSize(IcebergCommitState &commit_state, int32_t schema_id) {
 	int64_t total_files_size = 0;
 	for (const auto &manifest : commit_state.manifests) {
@@ -133,7 +123,7 @@ static int64_t ReconstructTotalFilesSize(IcebergCommitState &commit_state, int32
 			if (entry.status == IcebergManifestEntryStatusType::DELETED) {
 				continue;
 			}
-			AddLiveFileSize(total_files_size, entry.data_file);
+			total_files_size = IcebergUtils::AddFileSizeChecked(total_files_size, entry.data_file.file_size_in_bytes);
 		}
 	}
 	return total_files_size;
@@ -220,7 +210,7 @@ void IcebergAddSnapshot::CreateUpdate(DatabaseInstance &db, ClientContext &conte
 
 	for (auto &manifest_list_entry : uncommitted_manifest_files) {
 		auto &manifest_file = manifest_list_entry.file;
-		new_snapshot.metrics.AddManifestFile(manifest_file);
+		new_snapshot.metrics.AddManifestListEntry(manifest_list_entry);
 
 		auto new_manifest_list_entry =
 		    WriteManifestListEntry(commit_state.table_info, manifest_list_entry, avro_copy, db, context);
--- a/src/catalog/rest/api/iceberg_add_snapshot.cpp
+++ b/src/catalog/rest/api/iceberg_add_snapshot.cpp
@@ -9,6 +9,7 @@
 #include "duckdb/main/extension/extension_loader.hpp"
 #include "duckdb/storage/external_file_cache/caching_file_system.hpp"
 #include "duckdb/common/types/uuid.hpp"
+#include "duckdb/common/operator/add.hpp"
 
 #include "core/metadata/manifest/iceberg_manifest_list.hpp"
 #include "catalog/rest/iceberg_table_set.hpp"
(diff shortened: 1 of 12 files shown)


src/catalog/rest/api/iceberg_add_snapshot.cpp
```diff
@@ -163,7 +193,6 @@ void IcebergAddSnapshot::CreateUpdate(DatabaseInstance &db, ClientContext &conte
 
 	//! Create a new manifest list, populate it with the content of the old manifest list (altered if necessary)
```
Tishj: Please move the rest of this down as well, you've broken this comment by separating the steps

src/catalog/rest/api/iceberg_add_snapshot.cpp
```diff
@@ -109,6 +113,32 @@ void IcebergAddSnapshot::ConstructManifestList(IcebergManifestList &new_manifest
 	commit_state.manifests.clear();
 }
 
+static void AddLiveFileSize(int64_t &total_files_size, const IcebergDataFile &data_file) {
```
Tishj: Duplicate of `AddFileSizeChecked` ?

src/include/core/metadata/manifest/iceberg_manifest_list.hpp
```diff
@@ -93,6 +93,12 @@ struct IcebergManifestFile {
 	idx_t existing_rows_count = 0;
 	//! deleted rows in the manifest
 	idx_t deleted_rows_count = 0;
+	//! Size in bytes of content files added by this manifest. This is derived from
+	//! manifest entries and is not serialized in the manifest list.
+	int64_t added_files_size = 0;
```
Tishj: No, this is meant to mirror the `manifest_file` spec of https://iceberg.apache.org/spec/#manifest-lists, this information shouldn't live here.

src/core/metadata/manifest/iceberg_manifest_list.cpp
```diff
@@ -103,6 +114,8 @@ IcebergManifestListEntry IcebergManifestListEntry::CreateFromEntries(FileSystem
 	manifest_file.added_rows_count = 0;
 	manifest_file.existing_rows_count = 0;
 	manifest_file.deleted_rows_count = 0;
+	manifest_file.added_files_size = 0;
```
Tishj: On top of that this shouldn't use `manifest_file` to carry this information, this is also the wrong place.
`CreateFromEntries` is used in many contexts, not just new manifests

src/catalog/rest/api/iceberg_add_snapshot.cpp
fm100: Thank you for the thorough feedback. Moved it to make them cohesive.

src/catalog/rest/api/iceberg_add_snapshot.cpp
fm100: Yes, I initially kept them separate because they serve different aggregation paths. I removed duplication by consolidating it in IcebergUtils.

src/include/core/metadata/manifest/iceberg_manifest_list.hpp
fm100: Refactored this to keep `IcebergManifestFile` struct consistent with the spec.

src/core/metadata/manifest/iceberg_manifest_list.cpp
fm100: Refactored this too. Thank you for the detailed guide.

src/core/metadata/snapshot/iceberg_snapshot.cpp
```diff
@@ -168,6 +245,9 @@ static string MetricsTypeToString(IcebergSnapshotMetricType type) {
 static IcebergSnapshotMetrics MetricsFromSummary(const case_insensitive_map_t<string> &snapshot_summary) {
 	IcebergSnapshotMetrics ret;
 	auto &metrics = ret.metrics;
+	//! Unlike the long-standing count totals, total-files-size is new to this extension.
+	//! Preserve its absence so a later write can reconstruct a legacy table's baseline.
```
Tishj: Can we replace the comment?

```
//! Remove the default for `TOTAL_FILES_SIZE`, leaving it uninitialized if not present in the summary, rather than setting it to 0.
```

src/core/metadata/snapshot/iceberg_snapshot.cpp
```diff
@@ -179,6 +259,12 @@ static IcebergSnapshotMetrics MetricsFromSummary(const case_insensitive_map_t<st
 				// Skip invalid metrics
 				continue;
 			}
+			if ((item.type == IcebergSnapshotMetricType::ADDED_FILES_SIZE ||
```
Tishj: If one of the 3 is invalid, I think it makes sense to remove all 3, no?
If you agree, let's keep track of this in a bool `file_size_metrics_are_valid = true`, and outside of this loop we can erase the items if the bool ends up false.

src/core/metadata/snapshot/iceberg_snapshot.cpp
```diff
@@ -44,9 +49,62 @@ IcebergSnapshotMetrics::IcebergSnapshotMetrics(const IcebergSnapshot &snapshot)
 	if (total_position_deletes != other_metrics.end()) {
 		metrics[IcebergSnapshotMetricType::TOTAL_POSITION_DELETES] = total_position_deletes->second;
 	}
+	auto total_files_size = other_metrics.find(IcebergSnapshotMetricType::TOTAL_FILES_SIZE);
+	if (total_files_size != other_metrics.end()) {
+		metrics[IcebergSnapshotMetricType::TOTAL_FILES_SIZE] = total_files_size->second;
+	}
+}
+
+static void AddSizeMetric(map<IcebergSnapshotMetricType, int64_t> &metrics, IcebergSnapshotMetricType type,
```
Tishj: I think this makes the most sense as a private method of `IcebergSnapshotMetrics`

src/core/metadata/snapshot/iceberg_snapshot.cpp
```diff
@@ -44,9 +49,62 @@ IcebergSnapshotMetrics::IcebergSnapshotMetrics(const IcebergSnapshot &snapshot)
...
+                          int64_t value) {
+	auto &metric = metrics.emplace(type, 0).first->second;
+	metric = IcebergUtils::AddFileSizeChecked(metric, value);
+}
+
+static void UpdateTotalFilesSize(map<IcebergSnapshotMetricType, int64_t> &metrics, int64_t added, int64_t removed) {
```
Tishj: Also makes a lot more sense as a private method

Tishj:
Thanks for the changes, I think it's almost ready, just a few small comments, I'll let CI run to see if there are other issues I missed 👍

src/core/metadata/snapshot/iceberg_snapshot.cpp
fm100: I agree. These are correlated metrics and if one of them invalid, the others may not be trustworthy.

fm100:
> Thanks for the changes, I think it's almost ready, just a few small comments, I'll let CI run to see if there are other issues I missed 👍

Thank you very much for your help. I addressed all comments 😃

Issue #1162, opened 2026-07-08
Add write isolation levels + semantic conflict detection (`write.{merge,update,delete}.isolation-level`)


paultmathew:
## Summary

The Iceberg extension's write path currently guards concurrent commits with a
single coarse `assert-ref-snapshot-id` requirement (CAS on the `main` ref) and
retries on HTTP 409. It has **no semantic conflict detection** and **no
concept of isolation levels**, so it cannot honor Iceberg's
`write.{merge,update,delete}.isolation-level` table properties the way
iceberg-java/Spark/Flink do — and it false-conflicts on *any* concurrent
commit, even to disjoint partitions/files.

This is a proposed follow-up to #786 (commit retries): once a commit can be
rebased/retried, the next step is to validate *what* changed on rebase and
decide conflict vs. reconcile based on an isolation level.

## Current behavior (as of `main`)

`src/catalog/rest/transaction/iceberg_transaction.cpp`:

- The only conflict guard staged on a table commit is
  `CreateAssertRefSnapshotIdRequirement()` — `assert-ref-snapshot-id` pinning
  `ref = "main"` to the base snapshot the writer read.
- `CommitIsRetryable()` treats **only** a 409 conflict as retryable; on retry
  `RefreshRetryTables()` → `RefreshFromCatalog()` re-reads the tip and
  re-stages. `RetryStateMatches()` throws `TransactionException` if the table
  changed incompatibly.
- The retry budget is read from table metadata via
  `IcebergRetryConfig::FromTableMetadata()` (`commit.retry.num-retries`,
  `commit.retry.min-wait-ms`, `commit.retry.max-wait-ms`,
  `commit.retry.total-timeout-ms`), with decorrelated-jitter backoff.

There is **no** `validateNoConflictingDataFiles` / `validateNoConflictingDeleteFiles`
equivalent — nothing scans the manifests of intervening snapshots to evaluate
whether concurrently-added files match an operation's conflict-detection
filter. `grep -ri isolation src/ include/` returns nothing.

DeepWiki summarizes the current guarantee as **read-committed**.

## The gap

1. **Spec/parity gap.** Iceberg defines two isolation levels (serializable,
   snapshot) selectable per row-level op. DuckDB honors neither.
2. **Correctness.** Without conflict-detection filters, there's no way to get
   an explicit serializable guarantee, and past lost-update-shaped reports
   (#631, #890) are the class of anomaly this prevents.
[...]

PR #1189, opened 2026-07-17
Retry DELETE commits on conflict when the re-apply is safe

paultmathew:
Follow-up to #786 / #1115 (INSERT commit retries) and #1181: extend commit-retry to `DELETE`, kept intentionally small per the discussion on #786.

## What

Today a `DELETE` that loses a commit conflict fails outright. This makes it retry against the refreshed tip the same way `INSERT` already does:

- `IcebergAddSnapshot::IsRetryable()` now returns true for `DELETE`.
- On retry the delete re-applies against the refreshed manifests (`IcebergCommitState::LoadExistingManifests` already re-scans the current tip when the cached list is empty), which is only safe if the data the delete targets still exists.
- `DeleteReapplyUnsafe()` inspects every snapshot committed between the delete's scan snapshot and the tip — ranging over sequence numbers, since `parent_snapshot_id` is not populated when snapshots are read from the catalog. A concurrent **append** is safe (retry and succeed); a commit that **removed/rewrote data files or added delete files** is unsafe, and `StageSingleTableCommit` throws a clear conflict error rather than silently clobbering.

This addresses the `RETRY_BLOCKER` FIXME in `ConstructManifestList` for the common cases, and the `# FIXME: In the future, retries should alleviate this error` in `test_transactional_delete`.

## Tests

- `test_delete_commit_retry` (new): concurrent append -> delete retries and succeeds; concurrent delete on the same data -> aborts with a clear error.
- `test_transactional_delete` (updated): the racing commit now succeeds via retry (its FIXME is resolved); a concurrently-appended matching row is not removed under the default (that is isolation-level work, see below).
- `test_transactional_deletes_not_retried` -> renamed `test_transactional_deletes_retried` with the serially-correct outcome.

## Out of scope (follow-ups)

- Deletion-vector merge when two deletes target the same data file.
- Positional-delete re-scan under concurrent compaction (`RewriteFiles`).
- `write.delete.isolation-level` (serializable/snapshot) — #1162 — layering serializable added-data detection on top of this.

## Notes

Tested against the local `apache/iceberg-rest-fixture` catalog. Happy to add a Lakekeeper-gated concurrency test if preferred (the SQLite-backed fixture can't enforce true serializable commits).


src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -264,6 +264,47 @@ static void CreateTableRequirements(DatabaseInstance &db, ClientContext &context
...
+	auto base = metadata.GetSnapshotById(base_snapshot_id);
+	auto tip = metadata.GetSnapshotById(tip_snapshot_id);
+	if (!base || !tip || !base->sequence_number || !tip->sequence_number) {
+		return true;
+	}
+	int64_t base_sequence = *base->sequence_number;
+	int64_t tip_sequence = *tip->sequence_number;
+	for (auto &entry : metadata.snapshots) {
+		auto &snapshot = entry.second;
+		if (!snapshot.sequence_number) {
```
Tishj, line 247: I'm not really sure what we're checking here, I believe `sequence_number` should always be set for committed data, and we aren't expecting to find uncommitted snapshots here, right?

I think we should assert that it's not missing or throw an InvalidConfigurationException if it's missing

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -264,6 +264,47 @@ static void CreateTableRequirements(DatabaseInstance &db, ClientContext &context
...
+//! Pure appends are safe. Ranges over sequence numbers (parent links aren't
+//! populated on read snapshots); an unreachable base is treated as unsafe.
+static bool DeleteReapplyUnsafe(const IcebergTableMetadata &metadata, int64_t base_snapshot_id,
+                                int64_t tip_snapshot_id) {
+	if (base_snapshot_id == tip_snapshot_id) {
+		return false;
+	}
+	auto base = metadata.GetSnapshotById(base_snapshot_id);
+	auto tip = metadata.GetSnapshotById(tip_snapshot_id);
+	if (!base || !tip || !base->sequence_number || !tip->sequence_number) {
```
Tishj: Same as the earlier comment

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -264,6 +264,47 @@ static void CreateTableRequirements(DatabaseInstance &db, ClientContext &context
...
+		int64_t sequence = *snapshot.sequence_number;
+		if (sequence <= base_sequence || sequence > tip_sequence) {
+			continue;
+		}
+		if (snapshot.operation != IcebergSnapshotOperationType::APPEND) {
+			return true;
+		}
+		auto &metrics = snapshot.metrics.metrics;
+		auto removed = metrics.find(IcebergSnapshotMetricType::DELETED_DATA_FILES);
+		if (removed != metrics.end() && removed->second > 0) {
```
Tishj: If we're relying on the presence of the metrics, it being missing should be seen as "might have deleted data files", this makes it behave as "has no deleted data files" instead

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -264,6 +264,47 @@ static void CreateTableRequirements(DatabaseInstance &db, ClientContext &context
...
+			return true;
+		}
+		auto added_deletes = metrics.find(IcebergSnapshotMetricType::ADDED_DELETE_FILES);
+		if (added_deletes != metrics.end() && added_deletes->second > 0) {
```
Tishj: Same as above

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -264,6 +264,47 @@ static void CreateTableRequirements(DatabaseInstance &db, ClientContext &context
 	}
 }
 
+//! True if a concurrent commit between the DELETE's scan snapshot and the tip
+//! removed or rewrote data, making it unsafe to re-apply the delete on retry.
+//! Pure appends are safe. Ranges over sequence numbers (parent links aren't
+//! populated on read snapshots); an unreachable base is treated as unsafe.
+static bool DeleteReapplyUnsafe(const IcebergTableMetadata &metadata, int64_t base_snapshot_id,
```
Tishj: Can we invert this method? It feels weird reading `return true;` as a negative thing (meaning can't retry)

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -284,6 +325,18 @@ static SingleTableStagedCommit StageSingleTableCommit(DatabaseInstance &db, Iceb
 	}
 	commit_state.latest_snapshot = current_snapshot;
 
+	//! DELETE commit-retry guard: reject a re-apply over a concurrent commit that
+	//! removed/rewrote the targeted data (no-op on first attempt / non-deletes).
+	if (current_snapshot && current_snapshot->snapshot_id && transaction_data.delete_scan_snapshot_id &&
```
Tishj: Let's move this entire block into a `static void VerifyDeleteRetryability(...)` method, where we can break up this giant condition with separate `if (...) { return; }` blocks with the throw at the end for the cases that fall through

paultmathew:
@Tishj Thanks for the review! Summarizing the two assumptions this now makes:

1. sequence_number is always present on committed snapshots. We rely on it to bound the (base, tip] window, and throw InvalidConfigurationException if a resolved snapshot is missing it (a spec violation). A snapshot the catalog doesn't expose at all is handled separately — treated as not-reappliable.
2. Decision is based on operation alone, not the optional file-count metrics. Per the spec, operation is the only required summary field and append means "only data files added, none removed" — exactly our safety condition. Keying off deleted-data-files/added-delete-files was unsafe since those are optional and their absence would read as "safe".

Also applied your two structural suggestions: inverted the helper to DeleteCanReapply and extracted the guard into VerifyDeleteRetryability.

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -264,6 +264,66 @@ static void CreateTableRequirements(DatabaseInstance &db, ClientContext &context
...
+static void VerifyDeleteRetryability(const IcebergTableInformation &table_info,
+                                     optional_ptr<const IcebergSnapshot> current_snapshot) {
+	if (!table_info.transaction_data) {
+		return;
+	}
+	auto &transaction_data = *table_info.transaction_data;
+	if (!transaction_data.delete_scan_snapshot_id) {
+		return;
+	}
+	if (!current_snapshot || !current_snapshot->snapshot_id) {
```
Tishj: Same as `sequence_number`, `snapshot_id` should always be present on committed snapshots

src/include/catalog/rest/transaction/iceberg_transaction_data.hpp
```diff
@@ -73,6 +73,8 @@ struct IcebergTransactionData {
 
 	//! Every insert/update/delete creates an alter of the table data
 	vector<reference<IcebergAddSnapshot>> alters;
+	//! Snapshot the DELETE scanned; drives the commit-retry safety check. Empty if no delete.
+	optional<int64_t> delete_scan_snapshot_id;
```
Tishj: let's remove the "delete" part from it, and just make it `base_snapshot_id`, and I believe we can set this in   `CacheExistingManifestList`, that'll only ever be set once per table per transaction

`LoadExistingManifestList` should probably return it the 👍

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -264,6 +264,69 @@ static void CreateTableRequirements(DatabaseInstance &db, ClientContext &context
...
+
+//! Throw if a retried DELETE can't be safely re-applied. No-op on the first
+//! attempt (tip == scan) and for non-delete transactions.
+static void VerifyDeleteRetryability(const IcebergTableInformation &table_info,
+                                     optional_ptr<const IcebergSnapshot> current_snapshot) {
+	if (!table_info.transaction_data) {
+		return;
+	}
+	auto &transaction_data = *table_info.transaction_data;
+	if (!transaction_data.delete_scan_snapshot_id) {
```
Tishj: With the addressing of the other comment, I believe this can be `if (!transaction_data.base_snapshot_id) { throw InternalException("Delete is present but base snapshot id was never registered"); }`

paultmathew:
@Tishj Addressed both:

- Renamed delete_scan_snapshot_id → base_snapshot_id and moved its capture into LoadExistingManifestList (now returns it) via CacheExistingManifestList, so it's set once per table per transaction for any commit rather than only in the delete operator.
- Since the base snapshot is now always registered, the guard treats a delete without one as an invariant violation and throws InternalException instead of returning.
- Delete detection is now derived from the update's operation type via a new IcebergAddSnapshot::GetOperation() accessor (ContainsDelete()), rather than a dedicated field.

src/catalog/rest/transaction/iceberg_transaction.cpp
paultmathew: It looks like a plain throw doesn't work — a delete can legitimately have no base_snapshot_id when it targets data created in the same transaction:

- CREATE TABLE ... AS SELECT followed by DELETE
- a same-transaction insert then delete

In those cases there's no committed snapshot when the manifest list is cached, so no base to rebase onto. Handled by gating on current_snapshot: skip the check when there's no committed tip, and only throw when a committed tip exists but the base wasn't registered.

src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -223,6 +223,76 @@ static void CreateTableRequirements(DatabaseInstance &db, ClientContext &context
...
+	//! transaction.
+	if (!current_snapshot) {
+		return;
+	}
+	if (!current_snapshot->snapshot_id) {
+		throw InvalidConfigurationException("Committed snapshot is missing a snapshot_id");
+	}
+	//! A committed tip exists, so the base snapshot the delete scanned must have been
+	//! registered when the manifest list was cached. If it wasn't, our bookkeeping is broken.
+	if (!transaction_data.base_snapshot_id) {
```
Tishj: I think this is a problem in this scenario:

table is empty (no snapshot)
t1: begin;
t2: begin;
t1: insert values(...);
t2: insert values(...);
t2: commit;
t1: delete;
t1: commit;

When we're retrying t1's commit, we won't have a base-snapshot-id, but we will have a delete (over transactional data) and the current-snapshot-id will also be set because t2 produced one.

paultmathew: Thanks for catching this. I changed the `if (!transaction_data.base_snapshot_id)` condition to return and kept the rest as is. The only conditions that throw now are `if (!current_snapshot->snapshot_id)` (genuine spec violation) or `if (DeleteCanReapply(...` evaluates to false.

Tishj:
Please fix with tests, regressions are easy to make and the only things that protect us are tests :)

paultmathew:
Of course! I appreciate the reminder :)

Tishj:
Thanks!

PR #1198, opened 2026-07-19
[Dev] Add new catalog for `fixture-latest`, using the latest version of the docker image

Tishj:
The latest version has support for new functionality, such as scan planning

We can disable the github workflow once `1.11.0` gets a published release, and bump the pinned version on the `fixture` catalog.


Tishj:
Looks like we'd also need an upgrade to the `iceberg-spark-runtime` to support the newer catalog version

PR #1203, opened 2026-07-21
Remove cloud tests that fail if retried.

Tmonster:
Should solve failures like the one reported here

https://github.com/duckdb/duckdb-iceberg/issues/1199


Tishj:
Shouldn't we just remove retrying for the cloud tests then?
Or move them to python and encapsulate the setup with the test, to make it retryable

Tmonster:
> Or move them to python and encapsulate the setup with the test, to make it retryable

This I think is a follow up, for now I just want to get these cloud tests passing consistently

PR #1204, opened 2026-07-21
Add support for server-side scanning with `tables/{table}/plan:` endpoint

Tishj:
The table's `config` can contain `scan-planning-mode`, which can be `client` or `server`.
We add support for this and force (server side) scan-planning mode if the value is `server`.
And vice-versa we won't try (server side) scan planning if the value is `client`

Adds `iceberg_use_server_side_scan_planning` setting, defaults to false.
If the value is false, in the absence of `scan-planning-mode`, sets the value to `client` (disabling server side planning).

The `fixture-latest` catalog sets `iceberg_use_server_side_scan_planning` to true, so we can test behavior.

#### Summary of (other) Changes
- IcebergMultiFileList can convert the pushed down filters to filters provided to the scan-planning endpoint.
- IcebergMultiFileList adds file paths to equality deletes, required for scan-planning because no sequence numbers are given, which is what we normally use for matching equality deletes to data files.


src/planning/iceberg_optimizer.cpp
```diff
@@ -38,14 +45,25 @@ void GuaranteeEqualityDeleteColumnsOptimizer::VisitOperator(unique_ptr<LogicalOp
...
 			continue;
 		}
 		auto &mfbd = get.bind_data->Cast<MultiFileBindData>();
 		if (!mfbd.file_list) {
 			continue;
 		}
 		auto &iceberg_list = mfbd.file_list->Cast<IcebergMultiFileList>();
+		bool requires_local_planning = below_write;
+		for (auto &column_id : get.GetColumnIds()) {
+			if (column_id.IsVirtualColumn() &&
```
Tishj, line 58: This is due to a (hopefully temporary) limitation of the server-side scan planning API
Meaning we can hopefully undo this later

src/planning/iceberg_optimizer.cpp
```diff
@@ -22,10 +22,17 @@ GuaranteeEqualityDeleteColumnsOptimizer::GuaranteeEqualityDeleteColumnsOptimizer
 }
 
 void GuaranteeEqualityDeleteColumnsOptimizer::VisitOperator(unique_ptr<LogicalOperator> &op) {
+	VisitOperator(op, false);
+}
+
+void GuaranteeEqualityDeleteColumnsOptimizer::VisitOperator(unique_ptr<LogicalOperator> &op, bool below_write) {
+	below_write = below_write || op->type == LogicalOperatorType::LOGICAL_INSERT ||
```
Tishj, line 29: This logic is a check that has a wider net than we would want, but no smaller net is possible at the moment.

The goal is to identify scans that are part of a update/insert/delete/merge-into, so we can disable server-side planning for them, which is necessary because those operators rely on metadata collected by the IcebergMultiFileList

src/planning/iceberg_multi_file_list.cpp
```diff
@@ -153,12 +154,151 @@ void IcebergMultiFileList::SetScanOrder(unique_ptr<RowGroupOrderOptions> options
...
+	}
+	auto result = make_uniq<rest_api_objects::Expression>();
+	result->and_or_expression.emplace();
+	result->and_or_expression->type.value = "and";
+	result->and_or_expression->left = std::move(left);
+	result->and_or_expression->right = std::move(right);
+	return result;
+}
+
+static optional<string> ServerSideComparisonType(ExpressionType type, bool flip) {
```
Tmonster: Can we move functions that convert duckdb expressions to rest api expressions to a separate file? `iceberg_multi_file_list.cpp` is already pretty big

src/catalog/rest/api/iceberg_scan_planning.cpp
```diff
@@ -0,0 +1,488 @@
...
+						    "Iceberg REST scan plan references one Puffin deletion vector from multiple data files");
+					}
+					delete_file.referenced_data_file = task.data_file.file.file_path;
+				}
+			}
+		}
+
+		vector<PlannedContentFile> data_files;
+		data_files.reserve(accumulator.file_tasks.size());
+		for (auto &task : accumulator.file_tasks) {
```
Tmonster: looks like we loop through `file_tasks` twice? any reason not to combine them?

src/catalog/rest/api/iceberg_scan_planning.cpp
```diff
@@ -0,0 +1,488 @@
...
+			try {
+				auto cancel_endpoint = TableEndpoint(table_info);
+				cancel_endpoint.AddPathComponent(IRCPathComponent::RegularComponent("plan"));
+				cancel_endpoint.AddPathComponent(IRCPathComponent::RegularComponent(*active_plan_id));
+				auto cancel_headers = PlanningHeaders(context);
+				cancel_headers.Insert("Idempotency-Key", UUID::ToString(UUID::GenerateRandomUUID()));
+				table_info.catalog.auth_handler->Request(RequestType::DELETE_REQUEST, context, cancel_endpoint,
+				                                         cancel_headers);
+			} catch (...) {
+				// Best-effort cleanup must not mask the planning failure or interrupt.
```
Tmonster, line 516: If something fails here, can we log the error message?

src/planning/iceberg_multi_file_list.cpp
```diff
@@ -1222,11 +1365,110 @@ void IcebergMultiFileList::InitializeView(lock_guard<mutex> &guard) const {
...
+			request.filter = std::move(server_side_filter);
+			if (scan_order_options) {
+				if (scan_order_options->row_limit.IsValid()) {
+					request.min_rows_requested = NumericCast<int64_t>(scan_order_options->row_limit.GetIndex() +
+					                                                  scan_order_options->row_group_offset);
+				}
+				if (scan_order_options->column_idx.HasPrimaryIndex() &&
+				    scan_order_options->column_idx.GetPrimaryIndex() < GetSchema().columns.size()) {
+					rest_api_objects::FieldName stats_field;
+					stats_field.value = GetSchema().columns[scan_order_options->column_idx.GetPrimaryIndex()]->name;
```
Tmonster, line 1321: maybe this needs to be a loop if multiple columns are requested to be ordered?

src/catalog/rest/api/iceberg_scan_planning.cpp
```diff
@@ -0,0 +1,488 @@
...
+						                            local_index);
+					}
+					task.delete_file_references.push_back(local_delete_indexes[local_index]);
+				}
+			}
+			result.file_tasks.push_back(std::move(task));
+		}
+	}
+
+	if (tasks.plan_tasks) {
```
Tmonster, line 278: Should we think about making this some kind of `insertion_order_preserving_map_t`? Just to make sure we don't add the same task multiple times?

src/planning/iceberg_multi_file_list.cpp
```diff
@@ -1222,11 +1365,110 @@ void IcebergMultiFileList::InitializeView(lock_guard<mutex> &guard) const {
...
+	if (scan_planning_mode == ScanPlanningMode::UNSPECIFIED) {
+		Value val;
+		if (context.TryGetCurrentSetting("iceberg_use_server_side_scan_planning", val) && !val.IsNull() &&
+		    val.type().id() == LogicalTypeId::BOOLEAN && !val.GetValue<bool>()) {
+			//! Without 'iceberg_use_server_side_scan_planning', only use client-side planning
+			scan_planning_mode = ScanPlanningMode::CLIENT_SIDE_ONLY;
+		}
+	}
+	if (!table_entry || HasTransactionData() || table_entry->table_info.IsRenamed() ||
+	    scan_planning_mode == ScanPlanningMode::CLIENT_SIDE_ONLY) {
```
Tmonster, line 1290: Should we throw an error here if the catalog has `scan-plan = Server-side-only`? To not silently fall back to client side planning?

Tmonster (commented):
Should we think about adding some tests that do the following

- Check that `EXPLAIN` output shows `server side` scan or something similar (but this can be future work)

src/planning/iceberg_multi_file_list.cpp
Tishj: Actually, that already happens
Because of the check at the bottom

src/planning/iceberg_multi_file_list.cpp
Tishj: I checked, `RowGroupOrderOptions` can only be ordered by one column (`column_idx`)

PR #1209, opened 2026-07-23
Re-enable duckdb tests

Tmonster:
DuckDB tests were disabled a while ago. Now that you can set a default format-version, we can re-enable these tests and run more variant tests.


Tishj:
Looks like last one is `test/sql/function/string/regex_operators_v1_5_view.test`

PR #1216, opened 2026-07-23
[Dev] Refactor IcebergMultiFileList internals, preparing for the switch needed between client-side and server-side planning behavior

Tishj:
This PR is part of #1204


Tishj:
This has the problem causing linux tests to fail

PR #1217, opened 2026-07-23
Fix equality-delete scan when the delete file has extra columns

NiclasHaderer:
Reading an Iceberg table with equality deletes can crash in `ScanEqualityDeleteFile`, because an equality-delete file may contain more columns than the reader models for it (e.g. a partition column added from the hive path on a transform-partitioned table), so the scanned chunk is wider than the modeled columns. The code assumed a positional match and dereferenced past the end. Now each equality field is resolved to its physical position in the delete chunk by name.


Tishj:
Thanks!

PR #1220, opened 2026-07-24
[WIP] skip scanning positional deletes that are not in the selected partition

Tmonster:


src/planning/iceberg_multi_file_list.cpp
```diff
@@ -1243,6 +1243,19 @@ void IcebergMultiFileList::EnumerateDeleteManifestEntriesInternal() const {
 	GetScanPlanProvider().EnumerateDeleteManifestEntries(*this);
 }
 
+bool IcebergMultiFileList::DeleteEntryMatchesFilters(const BoundIcebergManifestEntry &bound_manifest_entry) const {
+	auto manifest_idx = bound_manifest_entry.manifest_file_idx;
+	if (!delete_manifest_matches[manifest_idx]) {
```
Tishj, line 1248: This works, but in a follow up we should really fix this at a higher level, because right now we're always scanning every delete manifest, which is wasteful at best

Tmonster:
```
benchmark/file_pruning_benchmarks/prune_positional_deletes_from_unread_partitions.benchmark
Old timing: 0.335211
New timing: 0.009812
```

Benchmark shaves off a ton of time ❤️

But can be even faster if we skip reading delete manifest files

test/sql/local/catalog_test_config_setup/catalog_agnostic/reads/positional_delete_partition_pruning.test
```diff
@@ -0,0 +1,134 @@
...
+----
+99
+
+# Every positional delete file that was opened and materialized during the single-day scan, together
+# with the data file it references. A v2 positional delete file records the full referenced data-file
+# path, which for a day-partitioned table contains .../event_date_day=YYYY-MM-DD/...
+statement ok
+CREATE TEMP TABLE single_day_reads AS
+SELECT DISTINCT
+    regexp_extract(message, 'read ''positional_delete_file'': ''([^'']*)''', 1) AS delete_file,
```
Tishj, line 109: nit: you can use `$$text that contains ' without needing escapes$$`

test/sql/local/catalog_test_config_setup/catalog_agnostic/reads/positional_delete_partition_pruning.test
```diff
@@ -0,0 +1,134 @@
...
+    regexp_extract(message, 'referencing ''data_file'': ''([^'']*)''', 1) AS referenced_data_file
+FROM duckdb_logs
+WHERE type = 'Iceberg' AND message LIKE 'Iceberg Delete Scan, read ''positional_delete_file''%';
+
+# All 7 positional delete files are read even though the query only touches one of the 7 partitions.
```
Tishj, line 114: The comment says "all 7", but the result has `1`, I'm confused

PR #1222, opened 2026-07-25
[AI] Add case_sensitive ATTACH option for catalog/schema/table identifier lookups

dataders:
## Summary

related: https://github.com/duckdb/duckdb/pull/24375

Adds a `case_sensitive` boolean ATTACH option (tri-state: unset/true/false) controlling whether catalog/schema/table identifier lookups against an Iceberg REST catalog are case-sensitive or case-insensitive.

The extension currently has an unintentional, inconsistent case-sensitivity split:
- **Schemas** fall back to a genuinely case-insensitive cache once warm (`IcebergSchemaSet::GetEntry` returns a cached case-insensitive match with no re-verification).
- **Tables** are effectively always case-sensitive, regardless of prior listing — `IcebergTableSet::CreateEntryInternal` overwrites any case-insensitive cache match with a literal-cased object and re-fetches, so a case mismatch 404s against a real (exact-string) REST catalog even after a prior `SHOW TABLES`.

This PR makes both cases explicitly controllable and internally consistent, without changing default (unset) behavior.

## What's included

- New `CaseAwareIdentifierMap<T>`/`CaseAwareIdentifierSet` wrapper types (`src/include/catalog/rest/case_aware_identifier_container.hpp`), each backed by either DuckDB's `case_insensitive_map_t`/`case_insensitive_set_t` or a plain `unordered_map`/`unordered_set` depending on a tri-state mode. `Insert`/`Emplace` raise an ambiguity error on a same-fold key collision, but *only* in explicit insensitive mode — `unset` preserves today's silent behavior exactly.
- `IcebergSchemaSet`/`IcebergTableSet::entries`, `IcebergTransaction`'s `current_table_data`/`created_schemas`/`deleted_schemas`/`listed_schemas`/`looked_up_entries`/`schema_property_updates`, and `IcebergTransactionAlterUpdate`'s `updated_tables`/`committed_tables` all migrated to the new wrapper types.
- Cold-path lookups under explicit `case_sensitive=false` resolve the canonical server-cased name via a List call before fetching, instead of trusting the user's literal casing.
- Three new SQL test files covering cold/warm lookups under all three modes for both schemas and tables, ambiguous case-fold matches (single lookup and bulk listing), and transaction-bookkeeping correctness (cross-case drop-then-query, table property updates).

Full design rationale: `docs/superpowers/specs/2026-07-24-iceberg-catalog-case-sensitivity-design.md`.


Tmonster:
Hi @dataders can you remove all the changes like the one below?

```
-	if (iceberg_transaction.created_schemas.count(name)) {
+	if (iceberg_transaction.created_schemas.Count(name)) {
```

Our formatting uses lower case names and we use older naming for these functions See [NOLINT](https://github.com/duckdb/duckdb/blob/b7ed657449badd2c1b2e8eb470114042ed582fd3/src/include/duckdb/common/vector.hpp#L21).

It clutters the PR a bit and is not relevant to the added functionality.


Can you also target `main` for this? v1.5-variegata is now only for bug fixes

Tmonster:
@dataders sorry for getting back so late. 

Is there any reason we should support case-insensitivity here? I think S3Tables only allows lowercase, but it would be good if DuckDB does not silently make the transformation.

The inconsistencies mentioned in the description are indeed a bug when it comes to the table cache. I'd be in favor of respecting case always, and passing on errors from catalogs if the catalog does not support case

PR #1229, opened 2026-07-25
[Dev] Fix possible case-sensitivity problem, add annotation for locks

Tishj:
We now use `annotated_mutex` and all the variables guarded by mutexes are annotated as such.

`positional_delete_data` was using a `case-insensitive-map`, but we access it with file paths, I don't think file paths are guaranteed to be case-insensitive on all file systems.
So I've changed it to be case-sensitive.

### DIFF
--- a/src/planning/iceberg_multi_file_list.cpp
+++ b/src/planning/iceberg_multi_file_list.cpp
@@ -126,7 +126,7 @@ IcebergScanPlanProvider &IcebergMultiFileList::GetScanPlanProvider() const {
 	return *scan_plan_provider;
 }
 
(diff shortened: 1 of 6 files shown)


src/planning/iceberg_multi_file_list.cpp
```diff
@@ -147,13 +147,14 @@ void IcebergMultiFileList::SetOptions(const IcebergOptions &options) {
 }
 
 void IcebergMultiFileList::SetScanOrder(unique_ptr<RowGroupOrderOptions> options) {
+	annotated_lock_guard<annotated_mutex> guard(shared_state->lock);
```
Tishj, line 150: Note: this lock was missing before

PR #1230, opened 2026-07-26
Honor REST catalog URI overrides

### LINKED ISSUE #1145: REST catalog: `uri` override from /v1/config is ignored; prefix appended to ATTACH ENDPOINT breaks Nessie branch-scoped endpoints
### What happens

When attaching an Iceberg REST catalog, the extension appends the `prefix` returned by `GET /v1/config` to the **ATTACH `ENDPOINT`** for all subsequent routes. It ignores the `uri` property that the config response may carry in `overrides`, which tells the client to switch its base URI before applying the prefix. pyiceberg honors this override; DuckDB does not.

For Project Nessie this breaks any branch-scoped endpoint. Nessie's config response for `ENDPOINT https://<nessie>/iceberg/main` is:

```json
{
  "defaults":  { "prefix": "main%7Cmercedes" },
  "overrides": { "uri": "https://<nessie>/iceberg/", "nessie.prefix-pattern": "{ref}|{warehouse}" }
}
```

i.e. "from now on use base `/iceberg/` with prefix `main|mercedes`". DuckDB instead builds doubled paths — captured via `CALL enable_logging()`:

```
GET https://<nessie>/iceberg/main/v1/main%7Cmercedes/namespaces/... -> 404
                          ^^^^ branch from ENDPOINT     ^^^^ branch again from prefix
```

Every namespace/table route then 404s. Namespace listing silently returns an empty catalog, and table loads fail with an empty-message error:

```
Invalid Configuration Error:
```

(native 1.5.4 surfaces the same as `_duckdb.Error` with an empty message — the swallowed 404 makes this hard to diagnose without HTTP logging).

### Reproduction

```bash
docker run -p 19120:19120 ghcr.io/projectnessie/nessie:0.107.6
```

```sql
LOAD iceberg;
CALL enable_logging();
ATTACH 'warehouse' AS cat (TYPE ICEBERG, ENDPOINT 'http://localhost:19120/iceberg/main', SUPPORT_NESTED_NAMESPACES true);
-- ATTACH succeeds (config route is fine), then:
SHOW ALL TABLES;   -- empty
SELECT request.url FROM duckdb_logs() WHERE type = 'HTTP';
-- .../iceberg/main/v1/main%7C<warehouse>/namespaces  (doubled ref, 404)
```

### Expected

After `/v1/config`, subsequent requests should use `overrides.uri` (when present) as the base URI, i.e. `{overrides.uri}/v1/{prefix}/...` → `https://<nessie>/iceberg/v1/main%7Cmercedes/namespaces/...`, which returns 200.

### Workaround

Attach with the bare base (`ENDPOINT 'http://<nessie>/iceberg'`) — but that only ever reaches the default branch, since the ref then always comes from the config default. Reaching another branch currently requires an external path-rewriting proxy.

### Versions

- duckdb 1.5.4 (osx_arm64), iceberg extension `75726455` — reproduced
- duckdb-wasm 1.33.1-dev57.0, iceberg extension `f9d5902` — reproduced
- Nessie 0.107.6

Related: #375 (prefix URL-encoding — different issue; encoding is handled correctly in the builds above, the doubled base is not).

jingdaws:
## Description

  REST catalogs can return an `overrides.uri` property from the `/v1/config`
  endpoint to replace the client-configured catalog base URI.

  DuckDB currently ignores this property. For branch-scoped Nessie endpoints
  such as `/iceberg/main`, this causes follow-up requests to be incorrectly
  constructed under `/iceberg/main/v1/...` instead of the canonical
  `/iceberg/v1/...` endpoint returned by the server.

  This change:

  - applies `overrides.uri` after loading the catalog configuration
  - removes trailing slashes before constructing subsequent REST URLs
  - keeps the server-provided prefix handling unchanged
  - extends the Nessie integration test to verify both the initial
    branch-scoped config request and the overridden follow-up request URLs

  This does not add generic branch or ref support to REST catalogs. It follows
  the standard server-provided configuration precedence.

  Fixes #1145

  ## Tests

  - `make debug`
  - Nessie integration tests: 7 test cases, 125 assertions
  - `make format-check`
  - `git diff --check`

### DIFF
--- a/src/catalog/rest/iceberg_catalog.cpp
+++ b/src/catalog/rest/iceberg_catalog.cpp
@@ -380,6 +380,11 @@ void IcebergCatalog::GetConfig(ClientContext &context, IcebergEndpointType &endp
 	auto catalog_config = IRCAPI::GetCatalogConfig(context, *this, effective_warehouse);
 	overrides = catalog_config.overrides;
 	defaults = catalog_config.defaults;
+	auto uri_override_it = overrides.find("uri");
+	if (uri_override_it != overrides.end()) {
+		uri = uri_override_it->second;
+		StringUtil::RTrim(uri, "/");
+	}
 	ParsePrefix();
 
 	if (attach_options.encode_entire_prefix) {
--- a/test/sql/local/catalog_custom_setup/nessie/test_nessie.test
+++ b/test/sql/local/catalog_custom_setup/nessie/test_nessie.test
@@ -46,20 +46,29 @@ attach '' as my_datalake (
 <REGEX>:.*Unrecognized access mode.*Supported options are 'vended_credentials' and 'none'.*
 
 statement ok
+call enable_logging(level='debug');
+
+statement ok
 attach 'warehouse' as my_datalake (
     type ICEBERG,
-    ENDPOINT 'http://127.0.0.1:19120/iceberg/',
+    ENDPOINT 'http://127.0.0.1:19120/iceberg/main',
     ACCESS_DELEGATION_MODE 'none',
     SECRET 'my_secret'
 );
 
+# make sure the branch-scoped endpoint is used to fetch the catalog config
+query I
+select request.url from duckdb_logs_parsed('HTTP') where request.url like '%/v1/config%';
+----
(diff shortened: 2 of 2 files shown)


Tishj:
Thanks!

PR #1232, opened 2026-07-26
[Dev] Parse the Catalog `prefix` once as a `vector<string>`, fix eviction of `LoadTableResultCache`

Tishj:
This removes the need for `prefix_is_one_component` and also removes the need to reparse the prefix on every endpoint builder.

We now keep track of the initialization source of a IcebergTableInformation, and use that to make sure we only evict cache if the entry is still active in the cache


src/catalog/rest/transaction/iceberg_transaction.cpp
```diff
@@ -830,17 +827,15 @@ void IcebergTransaction::EvictCachedTables() {
 	if (!catalog.attach_options.max_table_staleness_micros.IsValid()) {
 		return;
 	}
-	ScopedTransaction temp_con(db);
```
Tishj, line 833: I think the way this worked before, we would never actually evict because the transaction id didn't match

PR #1240, opened 2026-07-27
Fix equality deletes leaking across non-matching partitions

jgoldsworth:
`GetEqualityDeletesForFile` filters delete files down to those whose partition matches the data file being scanned, but the partition-value check used continue inside the inner loop over partition_info, which only advances the inner loop instead of skipping the delete file, so `result.emplace_back(delete_file)` ran unconditionally and every equality delete file sharing the data file's partition_spec_id was applied regardless of its partition values.


Tmonster:
Thanks!

PR #1250, opened 2026-07-29
Equality deletes filtering based on manifest metrics main

Tmonster:
Do not read equality delete files if a query predicate would not never match what the equality delete filters out. I.e

```sql
Create table t1 as select range%100 a from range(1000);
-- create equality delete file with the following IN values
delete from t1 where a in (5, 10, 15, 20, 30);
-- select tuples(s) that are not even in the equality delete
Select * from t1 where a = 50;
```

This required some updates to the equality delete helper file, but those changes are not shipped in releases, since we do not want to support equality deletes

### DIFF
--- a/src/execution/helpers/equality_delete_helpers.cpp
+++ b/src/execution/helpers/equality_delete_helpers.cpp
@@ -34,20 +34,6 @@ namespace duckdb {
 //! ICEBERG_ENABLE_EQUALITY_DELETE_WRITES compile flag is on - in default builds the callers
 //! (in iceberg_delete.cpp) are #ifdef'd out, so this code is dead.
 
-static bool ExpressionContainsFunction(const Expression &expr, const char *function_name) {
-	if (expr.GetExpressionClass() == ExpressionClass::BOUND_FUNCTION &&
-	    expr.Cast<BoundFunctionExpression>().Function().GetName() == function_name) {
-		return true;
-	}
-	bool found = false;
-	ExpressionIterator::EnumerateChildren(expr, [&](const Expression &child) {
-		if (!found && ExpressionContainsFunction(child, function_name)) {
-			found = true;
-		}
-	});
-	return found;
-}
-
 //! Whether a physical-filter expression is built purely from equality-delete forms, i.e. `col = const`,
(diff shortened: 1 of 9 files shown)


src/storage/statistics/iceberg_variant_statistics.cpp
```diff
@@ -356,7 +356,7 @@ bool IcebergVariantBoundsReader::RekeyBoundsVariant(const Value &bounds_variant,
 		// keys (the flat JSON paths "$['person']['address']['zip']", ...), with the leaf bounds as typed scalars.
 		Vector tmp(bounds_variant, count_t(1));
 		RecursiveUnifiedVectorFormat format;
-		Vector::RecursiveToUnifiedFormat(tmp, 1, format);
+		Vector::RecursiveToUnifiedFormat(tmp, format);
```
Tmonster, line 359: The API `RecursiveToUnifiedFormat` has been updated to no longer need count

Tmonster:
A liiiitle bit of regression improvement, but not much

```
benchmark/file_pruning_benchmarks/prune_equality_deletes.benchmark
Old timing: 0.022502
New timing: 0.019454
```

src/execution/helpers/equality_delete_helpers.cpp
```diff
@@ -27,9 +34,71 @@ namespace duckdb {
...
+
 static bool PlanContainsPhysicalFilter(PhysicalOperator &plan) {
 	if (plan.type == PhysicalOperatorType::FILTER) {
-		return true;
+		auto &filter = plan.Cast<PhysicalFilter>();
+		//! Two physical filters are compatible with writing an equality delete and must not disqualify it:
+		//!  - the 'iceberg_verify_equality_deletes' filter the read path injects to apply existing equality
+		//!    deletes (otherwise every delete after the first falls back to positional deletes), and
+		//!  - the DELETE predicate itself when it is a pure equality form (`col IN (...)` / `col = c1 OR ...`).
+		if (!ExpressionContainsFunction(*filter.expression, "iceberg_verify_equality_deletes") &&
```
Tishj: This doesn't exist anymore

PR #1264, opened 2026-07-31
Add regression test tpch sf1 with data modification rounds

Tmonster:
This adds a regression test on tpch sf=1 with 5 rounds of data modification. I tried higher scale factors, but the docker container supporting iceberg rest fixture for some reason starts to fail when the benchmark is run at higher scale factors, to avoid spurious errors, I am currently keeping it at sf=1.

This should help to benchmark improvements when it comes to data compaction & lazy loading of manifest files


Tmonster:
Waiting to see if I can get a network delay in the benchmark runner as well. This will make benchmarking local minio a bit more realistic

Tmonster:
Happy to merge this as is, we can add network delays when https://github.com/duckdb/duckdb/pull/24430 gets merged

Issue #341, opened 2025-06-27
[Predicate Pushdown] Filtering deletes based on row counts


Tishj:
### Spec

> If metrics show that a delete file has no rows that match a scan predicate, it may be ignored just as a data file would be ignored [2].

> For example, if `file_a` has rows with `id` between 1 and 10 and a delete file contains rows with `id` between 1 and 4, a scan for `id = 9` may ignore the delete file because none of the deletes can match a row that will be selected.

### Thoughts

I think this is why a positional delete can contain an optional `row` column

2147483544 row | required struct<...> [1] | Deleted row values. Omit the column when not storing deleted rows.
-- | -- | --

> When present in the delete file, `row` is `required` because all delete entries must include the row values.

The meaning of `required` here means that it can't be NULL

nicosuave:
This was written with AI so I'm not going to PR but here's an implementation for inspiration at the least.

Note that it prunes delete entries against data file bound instead of the scan predicate. In my testing on a large date partitioned table with poor compaction hygiene, it pruned all 538 delete files.

https://github.com/duckdb/duckdb-iceberg/compare/main...nicosuave:duckdb-iceberg:eqdelete-range-pruning

Tmonster:
@nicosuave , are you also sure the `row` column is present in the positional delete files? If so, can you tell us what engine you are using?

From my understanding this column is actually optional, with many engines deciding to omit it. I think many of the gains may come from partition pruning the delete files

Tmonster:
@nicosuave Nevermind, I see you are just computing the intersection of the data file bounds and the equality delete file bounds, nothing to do with the `row` column in a positional delete

PR #1269, opened 2026-08-01
[Scan] Equality delete bound-based pruning based on data file bounds

Tishj:
Implements the optimization proposed and written in https://github.com/duckdb/duckdb-iceberg/issues/341#issuecomment-5009519318

### Summary

Per field of the `equality_ids` of the equality delete file, we determine that we can skip the equality delete if the following are true:
- delete file: null counts for the field_id are set and they are 0.
- delete file: lower+upper bounds present and non-null.
- data file: lower+upper bounds present and non-null.
- if column type is FLOAT or DOUBLE:
  - delete file: nan counts for the field_id are set and they are 0.
- delete.upper_bound < data.lower_bound OR delete.lower_bound > data.upper_bound

### Future Work

Since we have the columns emitted by the data file at this point, we can probably make this pruning even more thorough, if we have bounds information for the delete file, and the data file doesn't have the field id, we can figure out what the constant value for that field-id is, and check that against the delete bounds instead.
Also we can extend this to cover NULLs as well, currently the optimization only prunes for non-null delete files.


src/planning/pruning/iceberg_file_pruner.cpp
```diff
@@ -245,6 +245,71 @@ bool IcebergFilePruner::DeleteManifestMatchesDataFile(const IcebergManifestFile
...
 }
 
+bool IcebergFilePruner::EqualityDeleteMatchesDataFile(const IcebergDataFile &delete_file,
+                                                      const IcebergDataFile &data_file) const {
+	auto &equality_ids = delete_file.equality_ids;
+
+	for (auto field_id : equality_ids) {
+		auto delete_null_count = delete_file.null_value_counts.find(field_id);
+		if (delete_null_count == delete_file.null_value_counts.end() || delete_null_count->second != 0) {
+			//! A NULL delete key can match NULL data values - require a known zero null count
```
Tmonster, line 255: If the equality delete file is just one `NULL` value the equality delete file will automatically match it seems?

src/planning/pruning/iceberg_file_pruner.cpp
```diff
@@ -245,6 +245,71 @@ bool IcebergFilePruner::DeleteManifestMatchesDataFile(const IcebergManifestFile
...
+			continue;
+		}
+
+		auto delete_lower = delete_file.lower_bounds.find(field_id);
+		auto delete_upper = delete_file.upper_bounds.find(field_id);
+		auto data_lower = data_file.lower_bounds.find(field_id);
+		auto data_upper = data_file.upper_bounds.find(field_id);
+		if (delete_lower == delete_file.lower_bounds.end() || delete_upper == delete_file.upper_bounds.end() ||
+		    data_lower == data_file.lower_bounds.end() || data_upper == data_file.upper_bounds.end()) {
+			continue;
```
Tmonster, line 265: What happens in the following case?

- We start a table with 3 columns, field_ids 1, 2, 3. & insert data
- We add a fourth column, field id 4 & insert data
- We apply an equality delete  only to field_id 4
- We scan.

I feel like this check will hit `continue` since none of the delete file field_ids are in the data file, and we will then try to apply the delete file to a data file that doesn't even have the field id?

src/planning/pruning/iceberg_file_pruner.cpp
```diff
@@ -245,6 +245,71 @@ bool IcebergFilePruner::DeleteManifestMatchesDataFile(const IcebergManifestFile
...
+				DUCKDB_LOG(context, IcebergLogType,
+				           "Iceberg Equality Delete Pruning, skipped 'equality_delete_file': '%s' for 'data_file': "
+				           "'%s', equality field '%s' (field id %d) has bounds [%s, %s] outside data bounds [%s, %s]",
+				           delete_file.file_path, data_file.file_path, column.name, field_id,
+				           delete_stats.lower_bound->ToString(), delete_stats.upper_bound->ToString(),
+				           data_stats.lower_bound->ToString(), data_stats.upper_bound->ToString());
+				return false;
+			}
+		} catch (Exception &) {
+			continue;
```
Tmonster, line 312: Can we add a `DUCKDB_LOG_DEBUG` here?

src/planning/pruning/iceberg_file_pruner.cpp
Tmonster: Or maybe this check already happens earlier?

Tmonster (commented):
Thanks! 1 or 2 questions on edge cases

Can we also get a regression test on here? I think We should defs be able to see some performance improvements in certain cases.

src/planning/pruning/iceberg_file_pruner.cpp
Tishj: That's indeed an interesting question, and one that's covered by the spec (https://iceberg.apache.org/spec/#equality-delete-files)

> If a column was added to a table and later used as a delete column in an equality delete file, the column value is read for older data files using normal projection rules (defaults to null)

So no, the delete still applies even though the file doesn't contain the field id

src/planning/pruning/iceberg_file_pruner.cpp
Tishj: Yea that's a correct observation, I'm not entirely sure why, it was just copied from the original issue.
I think we could make this optimization work even if there are NULLs.

Issue #1191, opened 2026-07-17
new parameter or new function: iceberg_rewrite_data_files(strategy: binpack|sort) or iceberg_compaction()?


The-Alchemist:
Curious what @Tishj and @Tmonster opinions on this.  I can submit a PR.

Goal is how do Spark/pyiceberg-style compaction. Not sure if it should be a part of `iceberg_rewrite_data_files`, as a parameter, or a separate function entirely (with some shared functionality).

Thanks!

Tmonster:
currently the `iceberg_rewrite_data_files` uses a binpack strategy by default. Are you asking if we can add a `sort` option?

The-Alchemist:
> currently the `iceberg_rewrite_data_files` uses a binpack strategy by default. Are you asking if we can add a `sort` option?

Exactly!  Or whether sort compaction should be a separate function

Tmonster:
We will most likely follow sort by default. We can look into adding a `sort=false` option if needed, but we would like to follow sort specs if the table has them. 

The `iceberg_rewrite_data_fiels` is not super performant in its current form, we are hoping to add performance improvements first. We might be able to get a sort option after that

The-Alchemist:
oh, ok, great! we're really hoping for the feature to follow the sort specs if the table has them (we're working on a PR), so please let us know what kind of performance improvements you'd like to see first and we'll tackle those.

Tmonster:
ok! 
Currently one of the big performance improvements I would like to see the IcebergRewriteDataFiles to re-use the functionality in IcebergInsert so that `IcebergCopyInput` and `IcebergCopyOptions` are used and the partition spec it used. Currently if you call `iceberg_rewrite_data_files` on a partitioned table, there is a separate iceberg scan per partition value, which will be very slow for large tables. 

See explain plan below for a table with 2 partitions

```
│ Groups: 5                                        │
╰─────────────────────────┬────────────────────────╯
╭─ Union ─────────────────┴────────────────────────╮
╰─────────────────────────┬────────────────────────╯
                          ├─────────────────────────────────────────────────────────────────────────────┬─
╭─ Projection ────────────┴────────────────────────╮                          ╭─ Projection ────────────┴────────────────────────╮                     
│ Projections: 0, #0, #1, #2, #3, #4, #5, #6       │                          │ Projections: 1, #0, #1, #2, #3, #4, #5, #6       │                     
│ ~1 row                                           │                          │ ~1 row                                           │                     
[...]

PR #1275, opened 2026-08-03
Optimize partitioned iceberg_rewrite_data_files via IcebergCopyOptions

The-Alchemist:
## Summary
- Reuse `IcebergInsert::GetCopyOptions` / `PlanCopyForInsert` so partitioned `iceberg_rewrite_data_files` uses one Iceberg scan and one partitioned `PhysicalCopyToFile` instead of one COPY subtree per partition value.
- Keep per-partition selection gating (`min_input_files` / target size), flatten selected candidates for a single `filename IN (...)` rewrite, and build manifests via `AddFiles` + sequence-number pinning.
- Add plan-shape / SELECT coverage and a `Regression.yml ` partitioned-rewrite test.

Conversation at https://github.com/duckdb/duckdb-iceberg/issues/1191#issuecomment-5141052632

## Notes
- Still limited to a single default partition spec (multi-spec deferred).
- Unpartitioned sorted tables can now get `PhysicalOrder` via `PlanCopyForInsert`; dedicated write-order regression tests are left for a follow-up PR.

### DIFF
--- a/benchmark/rewrite_data_files/partitioned_rewrite.benchmark
+++ b/benchmark/rewrite_data_files/partitioned_rewrite.benchmark
@@ -45,46 +45,52 @@ SELECT
     range,
     chr((65 + (range % 16))::INTEGER),
     'payload-' || range::VARCHAR
-FROM range(16000);
+FROM range(1000000);
 INSERT INTO my_datalake.default.partitioned_rewrite_bench
 SELECT
     range + 16000,
     chr((65 + (range % 16))::INTEGER),
     'payload-' || (range + 16000)::VARCHAR
-FROM range(16000);
+FROM range(1000000);
 INSERT INTO my_datalake.default.partitioned_rewrite_bench
 SELECT
     range + 32000,
     chr((65 + (range % 16))::INTEGER),
     'payload-' || (range + 32000)::VARCHAR
-FROM range(16000);
+FROM range(1000000);
 INSERT INTO my_datalake.default.partitioned_rewrite_bench
 SELECT
     range + 48000,
     chr((65 + (range % 16))::INTEGER),
     'payload-' || (range + 48000)::VARCHAR
-FROM range(16000);
+FROM range(1000000);
 INSERT INTO my_datalake.default.partitioned_rewrite_bench
 SELECT
     range + 64000,
     chr((65 + (range % 16))::INTEGER),
     'payload-' || (range + 64000)::VARCHAR
-FROM range(16000);
+FROM range(1000000);
 INSERT INTO my_datalake.default.partitioned_rewrite_bench
 SELECT
     range + 80000,
     chr((65 + (range % 16))::INTEGER),
     'payload-' || (range + 80000)::VARCHAR
-FROM range(16000);
--- Snapshot tip before rewrite; cleanup rolls back here so warmup + timed runs share state.
-SET VARIABLE pre_rewrite_snapshot = (
-    SELECT snapshot_id
+FROM range(1000000);
+drop table if exists my_datalake.default.pre_rewrite_snapshot;
+create table my_datalake.default.pre_rewrite_snapshot as
+    SELECT snapshot_id::BIGINT snapshot_id
     FROM iceberg_snapshots(my_datalake.default.partitioned_rewrite_bench)
     ORDER BY sequence_number DESC
-    LIMIT 1
-);
+    LIMIT 1;
 
 run
+set enable_external_file_cache=false;
+set variable pre_rewrite_snapshot_var = (select snapshot_id from my_datalake.default.pre_rewrite_snapshot limit 1);
+SELECT *
+FROM iceberg_rollback_to_snapshot(
+    'my_datalake.default.partitioned_rewrite_bench',
+    getvariable('pre_rewrite_snapshot_var')
+);
 SELECT *
 FROM iceberg_rewrite_data_files(
     'my_datalake.default.partitioned_rewrite_bench',
@@ -92,9 +98,3 @@ FROM iceberg_rewrite_data_files(
     rewrite_all => true
 );
 
-cleanup
-SELECT *
-FROM iceberg_rollback_to_snapshot(
-    'my_datalake.default.partitioned_rewrite_bench',
-    getvariable('pre_rewrite_snapshot')
-);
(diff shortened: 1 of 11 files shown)


benchmark/rewrite_data_files/partitioned_rewrite.benchmark
```diff
@@ -0,0 +1,86 @@
...
+INSERT INTO my_datalake.default.partitioned_rewrite_bench
+SELECT
+    range + 80000,
+    chr((65 + (range % 16))::INTEGER),
+    'payload-' || (range + 80000)::VARCHAR
+FROM range(16000);
+
+run
+SELECT *
+FROM iceberg_rewrite_data_files(
```
Tmonster, line 95: Oof, I forgot about this case. The DuckDB benchmark runner first does a warmup run, then runs the `run` statement 5 times. If we compact on the warmup run to start, the benchmark runs won't make sense. 

I see 2 ways to fix this. 

1. Add (maybe in a separate PR) a function to rollback an iceberg table. I.e something like 
```
from iceberg_rollback_to_snapshot(<table_name>);
```
Similar to the [Spark equivalent](https://iceberg.apache.org/docs/latest/spark-procedures/#rollback_to_snapshot)

You can then run the compaction, and afterwards the rollback to the snapshot right before the compaction, so every compaction is looking at the same table state.

2. There might be options in the benchmark runner that allow you to say "no warm up run" and "run only once"

I would prefer (1) to be honest, I think the table function `rollback_to_snapshot` will be very useful in the future for other iceberg write/delete benchmarking, and with 5 runs it will make it easier for us to ignore regression noise that is not repeatable.

Tmonster (commented):
Thanks!
Looks pretty good so far.

Can we also get some tests where the partition is a transform? Bucket transforms and 1-2 temporal transforms (day, hour, year etc.) would be great to see

Tmonster:
Hmm, you may also need to call `PrepareIcebergScanFromEntry(context)` so that DuckDB-Iceberg has the secrets to call the compaction functions

The-Alchemist:
> Hmm, you may also need to call `PrepareIcebergScanFromEntry(context)` so that DuckDB-Iceberg has the secrets to call the compaction functions

Do you think the new `iceberg_rollback_to_snapshot()` needs `PrepareIcebergScanFromEntry(context)` as well, you think?

benchmark/rewrite_data_files/partitioned_rewrite.benchmark
Tmonster: @The-Alchemist `rollback_to_snapshot` has been merged, so if you merge main again, you should be able to use that function to reset the snapshot and properly benchmark compaction

Tmonster:
> Do you think the new iceberg_rollback_to_snapshot() needs PrepareIcebergScanFromEntry(context) as well, you think?

I don't think so, `iceberg_rollback_to_snapshot()` is only communicating with the catalog, so only needs the catalog secret while `PrepareIcebergScanFromEntry(context)` actually then does a scan and therefore needs storage secrets

benchmark/rewrite_data_files/partitioned_rewrite.benchmark
The-Alchemist: done!  updated tests pushed

The-Alchemist:
I think the regression failures are unrelated??

Tmonster:
The regression failures were sort of related, the variable was not saved between load and run. This revealed a bug in `rollback_to_snaposhot`, an Internal exception is called when the snapshot id is null

```
SELECT *
         FROM iceberg_rollback_to_snapshot('my_datalake.default.partitioned_rewrite_bench', NULL
         );
INTERNAL Error:
Calling GetValue on a value that is NULL

LINE 2: FROM iceberg_rollback_to_snapshot('my_datalake.default.partition...
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
````

**Comment by The-Alchemist** [comment]
> The regression failures were sort of related, the variable was not saved between load and run. This revealed a bug in `rollback_to_snaposhot`, an Internal exception is called when the snapshot id is null
> 
> ```
> SELECT *
>          FROM iceberg_rollback_to_snapshot('my_datalake.default.partitioned_rewrite_bench', NULL
>          );
> INTERNAL Error:
> Calling GetValue on a value that is NULL
> 
> LINE 2: FROM iceberg_rollback_to_snapshot('my_datalake.default.partition...
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```

I added a simple guard fix against `main` at https://github.com/duckdb/duckdb-iceberg/pull/1326

File: src/maintenance/rewrite_data_files_planner.cpp [hunk 2]
```diff
@@ -144,6 +146,9 @@ RewritePlan PlanRewrite(ClientContext &context, const RewriteDataFilesPlanInput
 	snapshot_info.snapshot = latest_snapshot;
 	snapshot_info.schema_id = table_metadata.GetCurrentSchemaId();
 
+	//! Install vended storage secrets before reading manifest lists from object storage.
+	table_info.LoadCredentials(context);
```
Tmonster, line 150: This is also called during CreatePlan, I think one of them can be removed?

Tmonster (approved):
Thanks! I think this is basically ready, the regression test runner is failing because the benchmark doesn't exist on main right now.

src/maintenance/rewrite_data_files_planner.cpp
The-Alchemist: I think you're right.  looks like we can remove the call at `CreatePlan`.  we should watch the vended credentials paths (Polaris/Lakekeeper).

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -74,7 +74,9 @@ void GroupCandidates(RewritePlan &plan, int64_t target_file_size_bytes, int64_t
 		if (!rewrite_all && static_cast<int64_t>(kv.second.size()) < min_input_files) {
 			continue;
 		}
-		plan.file_groups.push_back(std::move(kv.second));
+		for (auto &cand : kv.second) {
+			plan.selected_candidates.push_back(std::move(cand));
+		}
```
Tmonster: nit: maybe as a TODO: if there are less than `min_input_files` in `plan.selected_candidates` then we can also remove all candidates.

Tmonster:
Thanks!

PR #1276, opened 2026-08-03
Write delete manifests under the partition spec of the data files they target

### LINKED ISSUE #1274: DELETE and UPDATE write delete manifests under the wrong partition spec after a spec change
On a table whose partition spec has changed, a `DELETE` or `UPDATE` against data written under the
previous spec is acknowledged and committed, but has no effect on subsequent reads.

```sql
CREATE TABLE t (id INT, ts TIMESTAMP, val VARCHAR) PARTITIONED BY (day(ts));
INSERT INTO t VALUES (1, '2021-05-01 09:00:00', 'a'), (2, '2021-05-01 10:30:00', 'b');
ALTER TABLE t SET PARTITIONED BY (hour(ts));

DELETE FROM t WHERE val = 'b';           -- returns 1, row still visible
UPDATE t SET val = 'x' WHERE val = 'b';  -- returns 1, row now visible twice
```

After the `UPDATE`, `SELECT` returns both the old and the new version of row 2. The predicate is on a
non-partition column, so no partition pruning of the data file is involved.

## Root cause: the delete manifest declares a spec that its entries are not written in

`IcebergTransactionData::AddSnapshot` and `AddUpdateSnapshot` create the delete manifest with

```cpp
IcebergManifestMetadata::FromTableMetadata(table_metadata, IcebergManifestContentType::DELETE)
```

which falls back to `table_metadata.default_spec_id` — the table's *current* spec. Meanwhile
`IcebergDelete::GenerateDeleteManifestEntries` copies the partition tuple straight off the data file
being deleted, which is in the *old* spec. The two disagree:

| | manifest `partition_spec_id` | entry partition |
|---|---|---|
| data manifest | 0 (`day_ts_2`, field 1000) | day value `18748` |
| delete manifest | **1** (`hour_ts_2`, field 1001) | day value `18748` |

Under spec 1 that partition should be an hour ordinal (`449962` for this row), and field 1001 does not
appear in the entry at all. Reading the table back through Spark's `files` metadata table shows the
inconsistency directly — the delete file claims spec 1 while carrying the `ts_day` value:

```
+-------+-------+------------------+----------------------+
|content|spec_id|partition         |referenced_data_file  |
+-------+-------+------------------+----------------------+
|0      |0      |{2021-05-01, NULL}|NULL                  |   <- data file, spec 0
|1      |1      |{2021-05-01, NULL}|s3://.../019fc5f3.parquet|<- delete file, spec 1
+-------+-------+------------------+----------------------+
```

The reader is not at fault here. The Iceberg spec, under [Scan Planning][spec], requires a position
delete file to be applied only when

> - The data file's `file_path` is equal to the delete file's `referenced_data_file` if it is non-null
> - The data file's data sequence number is _less than or equal to_ the delete file's data sequence number
> - **The data file's partition (both spec and partition values) is equal to the delete file's partition**

so `IcebergFilePruner` correctly declines to apply a delete whose spec differs. What is wrong is the
metadata being produced.

[spec]: https://iceberg.apache.org/spec/#scan-planning

## What a conformant writer does

Spark 4.0 / Iceberg 1.10 performing the same merge-on-read `DELETE` on the same catalog writes the
delete manifest under the **data file's** spec, not the table's current one:

```
=== manifests (declared partition_spec_id) ===
+-------+-----------------+
|content|partition_spec_id|
+-------+-----------------+
|0      |0                |   <- data manifest
|1      |0                |   <- delete manifest, spec 0 (not the current spec 1)
+-------+-----------------+
```

Delete partition and data partition are then equal under the same spec, and the delete applies.

Reading the same snapshot, the engines disagree about the table's contents:

| Reader | Rows after the `DELETE` |
|---|---|
| Spark 4.0 / Iceberg 1.10 | 1 (correct) |
| DuckDB | 2 |

Spark applies the delete because it indexes file-scoped deletes by `referenced_data_file` and
bypasses partition matching — leniency beyond what the spec text requires. DuckDB does not, so the
malformed metadata is visible as a silent wrong result only in DuckDB.

Conversely, DuckDB reads a **Spark-written** spec-evolved table correctly (1 row), which confirms the
read path is sound when the metadata is conformant.

## Second, independent defect: partition values compared by position

`IcebergFilePruner::DeleteFileMatchesDataFile` pairs the delete file's and data file's partition
structs up by index and throws when the lengths differ
(`src/planning/pruning/iceberg_file_pruner.cpp`).

Data manifests and delete manifests are read by separate readers, and each reader projects the union
of the partition specs it covers (`IcebergDataFile::GetFieldIdToTypeMapping`, called from
`IcebergManifestFileScanInfo`). On a table holding data files under two specs with a delete against
the older one, the data manifests span both specs and yield two partition fields per entry, while the
delete manifest spans one and yields one — so the arity check fires even though both files are in the
same spec and the same partition:

```
Invalid Configuration Error: Delete file s3://.../ts_day=2021-05-01/...-deletes.parquet
has 1 partition values, but data file s3://.../ts_day=2021-05-01/...parquet has 2
for partition spec 0
```

This reproduces on **Spark-written, spec-conformant metadata with no DuckDB writes involved**, and it
is a hard failure — the table cannot be read at all:

```sql
-- all statements run in Spark
CREATE TABLE t (id INT, ts TIMESTAMP, val STRING) USING iceberg
  PARTITIONED BY (days(ts))
  TBLPROPERTIES ('format-version'='2', 'write.delete.mode'='merge-on-read');
INSERT INTO t VALUES (1, TIMESTAMP '2021-05-01 09:00:00', 'a'),
                     (2, TIMESTAMP '2021-05-01 10:30:00', 'b');
ALTER TABLE t REPLACE PARTITION FIELD days(ts) WITH hours(ts);
INSERT INTO t VALUES (3, TIMESTAMP '2021-06-02 11:00:00', 'c');  -- data file under the new spec
DELETE FROM t WHERE val = 'b';                                   -- delete against the old-spec file
```

Spark returns rows 1 and 3; DuckDB raises the error above. Comparing partition values by field id
rather than by position, and treating a field missing from one side's projection as NULL, resolves it.

## Environment

- **OS**: macOS 26.5.2 (build 25F84), arm64
- **DuckDB version**: v1.6.0-dev11929 (Development Version) 2253530960 — the `duckdb` submodule of
  duckdb-iceberg `main`, pinned at `v1.5.5-10037-g2253530960`
- **DuckDB client**: CLI, built from source alongside the extension (`make release`, then
  `build/release/duckdb`). Also reproduced through the extension's sqllogictest runner
  (`build/*/test/unittest`).
- duckdb-iceberg `main` @ `9c71f826b`, built against the pinned submodule above
- Catalog: `apache/iceberg-rest-fixture` + MinIO (the repo's `make fixture` stack)
- Cross-engine comparison: Spark 4.0.1 with `iceberg-spark-runtime-4.0_2.13-1.10.0`

paultmathew:
Fixes #1274.

Two independent defects, one per commit. #1274 has the full analysis and the cross-engine comparison.

**Delete manifests declared the wrong partition spec.** `AddSnapshot` / `AddUpdateSnapshot` built the
delete manifest from the table's `default_spec_id`, while its entries carry the partition values of
the data file being deleted. Once the spec changes the manifest declares one spec and its entries are
written in another, so a reader comparing the delete's partition to the data file's — as the
[spec](https://iceberg.apache.org/spec/#scan-planning) requires — cannot match them, and `DELETE` /
`UPDATE` report success while having no effect on reads. Delete entries are now grouped by the spec of
the data file they reference and one manifest is emitted per spec, as Iceberg's own writer does.
Equality deletes target no single data file and stay on the current spec.

**Partition values were compared by position.** Data and delete manifests are read by separate
readers, each projecting the union of the specs it covers, so one spec can surface a different set of
fields on either side. Pairing the two structs up by index threw on any table holding data files under
two specs with a delete against the older one, which is reachable by reading a table another engine
wrote, with no DuckDB write involved. Values are now matched on field id, with a field missing from
one side's projection treated as NULL.

### Tests

Four cases, each confirmed failing before the change:

| Test | Covers |
|---|---|
| `delete/partitions/delete_after_partition_spec_evolution` | delete against an older-spec data file |
| `update/partitions/update_after_partition_spec_evolution` | update, which writes the replacement row under the current spec |
| `delete/partitions/delete_with_mixed_partition_specs` | data files under two specs, delete against the older |
| `reads/read_mor_delete_partition_evolution` | the same, on Spark-written metadata, read-only |

The last one uses a new Spark fixture (`spark_mor_delete_partition_evolution`) so a regression in
reading foreign metadata is caught independently of DuckDB's write path. Polaris skips it for the same
reason it already skips `complicated_partitioned_table` — it rejects `REPLACE PARTITION FIELD` from
Spark.

On freshly generated fixture data the full catalog-backed suite passes (401 test cases, 93,123
assertions), as do the `reads`, `delete`, `update`, `merge` and `maintenance` suites individually.

### DIFF
--- a/src/execution/helpers/equality_delete_helpers.cpp
+++ b/src/execution/helpers/equality_delete_helpers.cpp
@@ -347,6 +347,28 @@ bool IcebergDelete::TryGetEqualityDeletePredicates(ClientContext &context, Icebe
 	return true;
 }
 
+//! This writer gives an equality delete no partition values, so only an unpartitioned spec makes it apply at
+//! all; a partitioned one would scope it to a partition it never names. Lowest id, to keep the choice stable.
+static int32_t GlobalEqualityDeleteSpecId(const IcebergTableMetadata &metadata) {
+	int32_t result = 0;
+	bool found = false;
+	for (auto &entry : metadata.GetPartitionSpecs()) {
+		if (!entry.second.IsUnpartitioned()) {
+			continue;
+		}
+		if (!found || entry.first < result) {
+			result = entry.first;
+			found = true;
+		}
+	}
+	if (!found) {
+		throw NotImplementedException(
+		    "Cannot write an equality delete for this table: none of its partition specs are unpartitioned, and "
+		    "writing partition-scoped equality deletes is not supported");
+	}
+	return result;
+}
+
 void IcebergDelete::WriteEqualityDeleteFile(ClientContext &context, IcebergDeleteGlobalState &global_state) const {
 	D_ASSERT(!equality_predicates.empty());
 
@@ -433,9 +455,8 @@ void IcebergDelete::WriteEqualityDeleteFile(ClientContext &context, IcebergDelet
 	delete_file.delete_count = rows.size();
 	delete_file.file_size_bytes = stats.file_size_bytes;
 	delete_file.equality_ids = std::move(equality_ids);
-	//! An equality delete targets no single data file, so it goes into a manifest of the current spec.
-	delete_file.data_file_partitioning.partition_spec_id =
-	    NumericCast<int32_t>(table.table_info.table_metadata.default_spec_id);
+	//! An equality delete targets no single data file, so it is not tied to any one partition.
+	delete_file.partition_info.partition_spec_id = GlobalEqualityDeleteSpecId(table.table_info.table_metadata);
 
 	// Record per-field metrics for the equality-delete values so scans can prune this delete file when its
 	// equality-field range is disjoint from the scan predicate / a data file's bounds. Each field's bound
--- a/src/execution/helpers/equality_delete_helpers.cpp
+++ b/src/execution/helpers/equality_delete_helpers.cpp
@@ -433,6 +433,9 @@ void IcebergDelete::WriteEqualityDeleteFile(ClientContext &context, IcebergDelet
 	delete_file.delete_count = rows.size();
 	delete_file.file_size_bytes = stats.file_size_bytes;
 	delete_file.equality_ids = std::move(equality_ids);
+	//! An equality delete targets no single data file, so it goes into a manifest of the current spec.
+	delete_file.data_file_partitioning.partition_spec_id =
+	    NumericCast<int32_t>(table.table_info.table_metadata.default_spec_id);
 
 	// Record per-field metrics for the equality-delete values so scans can prune this delete file when its
 	// equality-field range is disjoint from the scan predicate / a data file's bounds. Each field's bound
--- a/src/include/core/metadata/manifest/iceberg_manifest.hpp
+++ b/src/include/core/metadata/manifest/iceberg_manifest.hpp
@@ -65,11 +65,11 @@ struct IcebergPartitionInfo {
 	}
 };
 
-//! How a data file is partitioned: the values, plus the spec they were produced with.
+//! The partition a file belongs to: the per-field values, plus the spec they were produced with.
 //! Partition values are only meaningful together with their spec, so the two travel as a pair.
-struct IcebergDataFilePartitioning {
+struct IcebergPartition {
 	int32_t partition_spec_id = 0;
-	vector<IcebergPartitionInfo> partition_info;
+	vector<IcebergPartitionInfo> fields;
 };
 
 struct IcebergDataFile {
@@ -145,9 +145,12 @@ private:
 	optional<sequence_number_t> file_sequence_number;
 };
 
-//! Delete-manifest entries grouped by the partition spec of the data files they target, so each group can be
-//! written into a manifest declaring that spec.
-using IcebergDeleteManifestEntries = map<int32_t, vector<IcebergManifestEntry>>;
+//! Manifest entries grouped by partition spec id, so each group can be written into a manifest declaring
+//! that spec.
+using partitioned_manifest_entry_map_t = map<int32_t, vector<IcebergManifestEntry>>;
+
+//! A file's partition values looked up by partition field id.
+using partition_value_map_t = unordered_map<uint64_t, reference<const Value>>;
 
 struct IcebergManifestListEntry;
 
--- a/src/include/core/metadata/manifest/iceberg_manifest.hpp
+++ b/src/include/core/metadata/manifest/iceberg_manifest.hpp
@@ -65,6 +65,13 @@ struct IcebergPartitionInfo {
 	}
 };
 
+//! How a data file is partitioned: the values, plus the spec they were produced with.
+//! Partition values are only meaningful together with their spec, so the two travel as a pair.
+struct IcebergDataFilePartitioning {
+	int32_t partition_spec_id = 0;
+	vector<IcebergPartitionInfo> partition_info;
+};
+
 struct IcebergDataFile {
 public:
 	static map<idx_t, LogicalType> GetFieldIdToTypeMapping(const IcebergSnapshotScanInfo &snapshot_info,
@@ -137,6 +144,10 @@ private:
 	optional<sequence_number_t> sequence_number;
 	optional<sequence_number_t> file_sequence_number;
 };
+
+//! Delete-manifest entries grouped by the partition spec of the data files they target, so each group can be
+//! written into a manifest declaring that spec.
+using IcebergDeleteManifestEntries = map<int32_t, vector<IcebergManifestEntry>>;
 
 struct IcebergManifestListEntry;
 
--- a/src/include/execution/operator/iceberg_delete.hpp
+++ b/src/include/execution/operator/iceberg_delete.hpp
@@ -62,7 +62,9 @@ struct IcebergDeleteFileInfo {
 	idx_t pos_min_value;
 	optional_idx content_size_in_bytes;
 	optional_idx content_offset;
-	vector<IcebergPartitionInfo> partition_info;
+	//! The partitioning of the data file this delete targets. A delete file has to be written into a manifest
+	//! of that same spec, not the table's current one, or readers can't tell that the two partitions are equal.
+	IcebergDataFilePartitioning data_file_partitioning;
 	//! When non-empty, this is an equality-delete file; holds the field-ids it applies to
 	vector<int32_t> equality_ids;
 	//! Per-field serialized lower/upper bounds of the equality-delete values (field-id -> bound blob).
@@ -167,7 +169,7 @@ public:
 	                          OperatorSinkFinalizeInput &input) const override;
 	unique_ptr<GlobalSinkState> GetGlobalSinkState(ClientContext &context) const override;
 	unique_ptr<LocalSinkState> GetLocalSinkState(ExecutionContext &context) const override;
-	static vector<IcebergManifestEntry> GenerateDeleteManifestEntries(IcebergDeleteGlobalState &global_state);
+	static IcebergDeleteManifestEntries GenerateDeleteManifestEntries(IcebergDeleteGlobalState &global_state);
 
 	bool IsSink() const override {
 		return true;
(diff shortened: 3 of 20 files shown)


Tishj:
> Partition values were compared by position

The reason for this is that partition specs don't compare equal unless their field order is the same, so I took that to mean that we could rely on order in the `partition` field, since it should directly map to the partition spec, which is ordered.
This isn't documented in the spec though, they don't even define how partition specs should be compared for equality, this I believe was taken from pyiceberg or painstakingly learned from errors thrown by the iceberg-rest-fixture implementation.

EDIT:
I read your clarifications and they make sense, I think we can't get around mapping them.
I just request that the mapping is cached, and I have some questions regarding the comparisons around NULL and missing values.

src/execution/helpers/equality_delete_helpers.cpp
```diff
@@ -433,6 +433,9 @@ void IcebergDelete::WriteEqualityDeleteFile(ClientContext &context, IcebergDelet
 	delete_file.delete_count = rows.size();
 	delete_file.file_size_bytes = stats.file_size_bytes;
 	delete_file.equality_ids = std::move(equality_ids);
+	//! An equality delete targets no single data file, so it goes into a manifest of the current spec.
+	delete_file.data_file_partitioning.partition_spec_id =
+	    NumericCast<int32_t>(table.table_info.table_metadata.default_spec_id);
```
Tishj: Perhaps this should use the unpartitioned spec (0), so it targets all data, regardless of partition

src/include/core/metadata/manifest/iceberg_manifest.hpp
```diff
@@ -138,6 +145,10 @@ struct IcebergManifestEntry {
 	optional<sequence_number_t> file_sequence_number;
 };
 
+//! Delete-manifest entries grouped by the partition spec of the data files they target, so each group can be
+//! written into a manifest declaring that spec.
+using IcebergDeleteManifestEntries = map<int32_t, vector<IcebergManifestEntry>>;
```
Tishj: I don't like this naming convention for aliases, can we rename it to `partitioned_manifest_entry_map_t` ?

src/include/core/metadata/manifest/iceberg_manifest.hpp
```diff
@@ -65,6 +65,13 @@ struct IcebergPartitionInfo {
 	}
 };
 
+//! How a data file is partitioned: the values, plus the spec they were produced with.
+//! Partition values are only meaningful together with their spec, so the two travel as a pair.
+struct IcebergDataFilePartitioning {
```
Tishj: Can we rename this to `IcebergPartition` ?
I think DataFile is too coupled, and not only data files have a partition.

Now that I think about it, `IcebergPartitionField` would have been a better name for `IcebergPartitionInfo`, but let's leave that untouched for now

src/include/execution/operator/iceberg_delete.hpp
```diff
@@ -62,7 +62,9 @@ struct IcebergDeleteFileInfo {
 	idx_t pos_min_value;
 	optional_idx content_size_in_bytes;
 	optional_idx content_offset;
-	vector<IcebergPartitionInfo> partition_info;
+	//! The partitioning of the data file this delete targets. A delete file has to be written into a manifest
+	//! of that same spec, not the table's current one, or readers can't tell that the two partitions are equal.
+	IcebergDataFilePartitioning data_file_partitioning;
```
Tishj: Let's keep this as `partition_info`, I know that in the spec this is `data_file.partition`, but it's confusing since we're dealing with deletes here.

src/planning/pruning/iceberg_file_pruner.cpp
```diff
@@ -288,23 +288,30 @@ bool IcebergFilePruner::DeleteFileMatchesDataFile(const IcebergManifestFile &del
...
-	if (delete_file.partition_info.size() != data_file.partition_info.size()) {
-		throw InvalidConfigurationException(
-		    "Delete file %s has %llu partition values, but data file %s has %llu for partition spec %d",
-		    delete_file.file_path, delete_file.partition_info.size(), data_file.file_path,
-		    data_file.partition_info.size(), delete_manifest.partition_spec_id);
+	//! A partition struct is read as the union of the specs covered by its reader, and data and delete manifests
+	//! are read separately, so the same spec can yield a different set of fields on each side: a field belonging
+	//! to another spec is materialized as NULL for one file and missing for the other. Match on field id and
+	//! treat a missing field as NULL, rather than pairing the two structs up by position.
+	unordered_map<uint64_t, reference<const Value>> data_partition_values;
```
Tishj: Hmm I see, that makes sense, it's a side effect of how we batch reads together that these won't directly align.
Then can we cache this map and pass it to this method? Otherwise we'll be paying the cost of creating this map for each delete file.

src/planning/pruning/iceberg_file_pruner.cpp
```diff
@@ -288,23 +288,30 @@ bool IcebergFilePruner::DeleteFileMatchesDataFile(const IcebergManifestFile &del
...
+			if (!delete_partition.value.IsNull()) {
+				return false;
+			}
+			continue;
+		}
+		if (!Value::NotDistinctFrom(delete_partition.value, it->second.get())) {
+			return false;
 		}
-		if (!Value::NotDistinctFrom(delete_partition.value, data_partition.value)) {
+		data_partition_values.erase(it);
```
Tishj: Can you explain to me why we are altering the map and then iterating through the remainder?

It seems to me that all field ids we need should be present, since we union the specs, the result will be a superset, our spec should always be a subset of that result.

src/execution/helpers/equality_delete_helpers.cpp
paultmathew: Please correct me if I'm wrong but i think spec (0) could also be a partitioned spec if the table was created as partitioned. I updated it to search the table's specs for one that is unpartitioned and uses the lowest matching id. If the table has none, it throws.

The throw is because we write no partition values. Under a partitioned spec, `DeleteFileIndex` files the delete at `(specId, null)` while data files look up `(specId, <their partition>)`, so it is never found: the delete silently removes nothing, and would only ever match files whose partition value is genuinely null.

Alternatives, happy to switch:

- append an empty spec to `partition-specs` and point at it — the spec says a new spec "may
  be set as the table's default", so it needn't become one. Works on any table, but a DELETE
  then mutates table metadata.
- write per-partition equality deletes like Flink, which requires the partition source
  columns to be part of the equality fields.

src/planning/pruning/iceberg_file_pruner.cpp
paultmathew: Good call. Updated it to cache it once.

src/planning/pruning/iceberg_file_pruner.cpp
paultmathew: Thanks for catching this. The reason it was there: the old code walked the two projected partition tuples element by element, and those tuples aren't the spec's field list — each one is the union of the specs covered by the reader that produced it. Data manifests and delete manifests are read by separate readers, and IcebergManifestFileScanInfo builds partition_field_id_to_type from only the manifests in its own selection. So on a table with data files under two specs and a delete against the older one, the data side comes back with two fields and the delete side with one. That asymmetry is what used to throw:

```
Invalid Configuration Error: Delete file ...-deletes.parquet has 1 partition values,
but data file ....parquet has 2 for partition spec 0
```

Your premise holds as soon as we compare the right set of fields. Both manifests declare the same spec by that point, so both projections must contain every field of that spec — it's only fields belonging to other specs that show up on one side and not the other. The comparison now walks partition_spec.fields and looks each field id up on both sides, so there's nothing left over to sweep up.

src/execution/helpers/equality_delete_helpers.cpp
Tishj: Again something that's badly specd but from looking at other implementations, I believe spec id 0 should be reserved for unpartitioned.
They do put this in the spec for the sort order, that 0 has to be reserved for unsorted.

Tishj: I think your solution is fine then, searching for the unpartitioned spec :+1:

Tishj:
Thanks!

PR #1277, opened 2026-08-03
failing test: iceberg_rewrite_data_files must preserve table sort order

The-Alchemist:
## Summary
- `iceberg_rewrite_data_files` should Parquet in physical table sort order (`file_row_number` + `lag`) and stamp the correct `sort_order_id` on the ADDED manifest entry.
- writes today does not re-sort data; it only stamps `sort_order_id`.
- Tests-only PR


The-Alchemist:
test passed on latest `main`

PR #1278, opened 2026-08-04
Preserve explicit `ENDPOINT` ATTACH parameters when using GLUE/S3_TABLES `ENDPOINT_TYPE`

### LINKED ISSUE #906: Allow custom endpoint override with ENDPOINT_TYPE
When both `ENDPOINT_TYPE` and `ENDPOINT` are provided, `ENDPOINT_TYPE` silently overwrites the user-provided `ENDPOINT`. This means there's no way to get the catalog-specific defaults from `ENDPOINT_TYPE` (auth type, path prefix, supported operations) while pointing at a different host.

### Current behavior

```
sql
ATTACH 'my-warehouse' AS catalog (
   TYPE iceberg,
   ENDPOINT_TYPE glue,
   ENDPOINT 'my-custom-host.example.com/iceberg'
);
```

-- User-provided ENDPOINT is silently ignored
-- Connects to the hardcoded default host instead

### Proposed behavior

If the user explicitly provides `ENDPOINT`, respect it. `ENDPOINT_TYPE` should still apply its defaults (auth, path prefix, catalog-specific settings) but not overwrite the host.

```
sql
ATTACH 'my-warehouse' AS catalog (
   TYPE iceberg,
   ENDPOINT_TYPE glue,
   ENDPOINT 'my-custom-host.example.com/iceberg'
);
```
-- Connects to my-custom-host.example.com/iceberg
-- Auth, path prefix, and catalog defaults from ENDPOINT_TYPE still apply

### Use cases

- VPC endpoints / PrivateLink
- FIPS endpoints
- Region-specific hostname variations (related: #708)
- Local development with endpoint emulators

u70b3:
## Summary

Fixes the interaction between `ENDPOINT_TYPE` and an explicitly supplied
`ENDPOINT` for AWS Iceberg catalogs.

Closes #906.

## Root cause

`S3OrGlueAttachInternal` unconditionally generated an AWS endpoint from the
catalog type and region. As a result, a user-provided endpoint was silently
overwritten:

```sql
ATTACH ... (
    TYPE ICEBERG,
    ENDPOINT_TYPE 'GLUE',
    ENDPOINT 'http://localhost:8181'
);
```

Preserving the endpoint alone is not sufficient for custom hosts because SigV4
cannot infer the AWS service and region from addresses such as localhost,
emulator endpoints, PrivateLink endpoints, or FIPS endpoints.

## Changes

- Preserve `ENDPOINT` when it is explicitly provided.
- Generate the standard AWS endpoint only when `ENDPOINT` is absent.
- For custom endpoints, use `ENDPOINT_TYPE` to supply the default
  `sigv4_service` and `sigv4_region`.
- Add regression tests for Glue and S3 Tables custom endpoints.

## Validation

- Added a SigV4 mock-server regression test covering Glue and S3 Tables.
- Verified that both catalogs send requests to the custom endpoint.
- Verified that the requests contain an AWS SigV4 `Authorization` header.
- Ran all SigV4 mock tests: 20 assertions passed.
- Built the `iceberg_loadable_extension` target successfully.
- Passed formatting and `git diff --check`.


Tishj:
Thanks, changes look good to me

Tishj:
I just realized that there's nothing running these tests in CI, so it's not really tested right now
This should be hooked up to a github action workflow

u70b3:
Good catch @Tishj — you're right. Nothing in CI starts `scripts/sigv4_mock_server.py` or sets `SIGV4_MOCK_SERVER_AVAILABLE`, so these tests are silently skipped by `require-env` (this also affects the earlier `test_sigv4_http_scheme.test`).

Since it's pre-existing test infrastructure rather than specific to this fix, I've filed #1283 to track wiring it up in a follow-up PR — modeled on `test-nessie-catalog.yml`, but much simpler since the mock server is a stdlib-only Python script. I'll pick it up next.

Happy to fold it into this PR instead if you'd prefer.

Tishj:
I think the Polaris failure is unrelated. I've never seen it fail like that before though. I assume Maven or whatever it uses to grab dependent jars from is having issues.

u70b3:
All checks are green now — the Polaris failure was indeed unrelated infra flakiness, as you suspected.

Also, thanks a lot for jumping into #1286 and fixing up the local tests, the manifest reader, and the CI dependencies — the sigv4_mock tests are actually running in CI now, so the new `test_endpoint_type_custom_endpoint.test` in this PR is properly exercised.

This should be ready to merge whenever you have a moment.

Tmonster (approved):
Thanks!

PR #1280, opened 2026-08-04
[ManifestReader] Use the `format_version` and `partition_spec_id` from the manifests metadata

Tishj:
The manifest file's Avro key-value metadata contains `format_version` and `partition_spec_id`, by using this we are parsing the manifest entries from the file using the correct format version, which fixes problems where fields that are required by later versions surface as errors when reading older manifests.

This also fixes a problem where we were deserializing `IcebergDataFile` items with many more fields in the `partition` variable than the file actually has. (As can be seen by the fixes to `iceberg_column_stats`)

### DIFF
--- a/src/include/planning/metadata_io/manifest/iceberg_manifest_reader.hpp
+++ b/src/include/planning/metadata_io/manifest/iceberg_manifest_reader.hpp
@@ -4,9 +4,19 @@
 
 namespace duckdb {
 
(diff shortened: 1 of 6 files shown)


Tishj:
This may warrant a follow up even, because we currently read the `partition` field with a union of all partition specs of the manifests we're reading. That could lose type information for a field that underwent schema evolution.

PR #1281, opened 2026-08-04
Respect explicit Iceberg version guessing

### LINKED ISSUE #731: `version = '?'` does not work without `SET unsafe_enable_version_guessing = true`
The [documentation](https://duckdb.org/docs/stable/core_extensions/iceberg/overview#guessing-metadata-versions) says that passing `version = '?'` to `iceberg_scan` should be enough to enable version guessing when no `version-hint.text` is present:

```sql
SELECT * FROM iceberg_scan('path/to/table', version = '?');
```

However, on DuckDB **1.4.4** this still throws:

```
Invalid Configuration Error: Failed to read iceberg table. No version was provided and no version-hint could be found, globbing the filesystem to locate the latest version is disabled by default as this is considered unsafe and could result in reading uncommitted data. To enable this use 'SET unsafe_enable_version_guessing = true;'

LINE 1: SELECT count(*) FROM iceberg_scan('path/to/table', version = '?');
                             ^
```

The only workaround is the global setting:

```sql
SET unsafe_enable_version_guessing = true;
SELECT * FROM iceberg_scan('path/to/table');
```

### To reproduce

```sh
duckdb -c "INSTALL iceberg; LOAD iceberg; SELECT count(*) FROM iceberg_scan('path/to/table', version = '?');"
```

Where `path/to/table` is any iceberg table directory with no `version-hint.text` in its `metadata/` folder.

u70b3:
## Summary

- Treat an explicit `version = '?'` as a per-call opt-in to Iceberg metadata version guessing.
- Keep implicit version guessing protected by `unsafe_enable_version_guessing`.
- Cover `iceberg_scan`, `iceberg_metadata`, and `iceberg_snapshots`.

Closes #731.

## Root cause

`IcebergOptions::table_version` defaults to `"?"`. As a result, the metadata resolver could not distinguish between:

- a caller omitting the `version` parameter; and
- a caller explicitly supplying `version = '?'`.

Both cases reached the global `unsafe_enable_version_guessing` check, so the documented per-call form still required the global setting.

## Changes

- Added an explicitness marker to `IcebergOptions`.
- Set and propagated it through:
  - generic named-parameter parsing;
  - the multi-file Iceberg scan path;
  - `iceberg_snapshots` parameter parsing;
  - `IcebergOptions` copy assignment.
- Bypassed the global setting only when `version = '?'` was explicitly supplied.
- Preserved existing behavior for numeric versions, hint files, direct metadata paths, and omitted versions.

## Behavior

| Input | Result |
|---|---|
| No version, no hint, setting disabled | Error |
| Explicit `version = '?'`, setting disabled | Guess latest metadata version |
| No version, setting enabled | Guess latest metadata version |
| Explicit numeric version | Read the requested version |
| Existing `version-hint.text` | Use the version hint |

## Tests

- `iceberg_metadata.test`: 183 assertions passed.
  - Omitted-version safety controls.
  - Explicit `?` through `iceberg_metadata`.
  - Explicit `?` through `iceberg_scan`.
  - Explicit `?` through `iceberg_snapshots`.
- `version_name_format_error.test`: passed.
- Changed-file format check: passed.
- `git diff --check`: passed.

### DIFF
--- a/src/core/metadata/iceberg_table_metadata.cpp
+++ b/src/core/metadata/iceberg_table_metadata.cpp
@@ -268,7 +268,7 @@ string IcebergTableMetadata::GetMetaDataPath(ClientContext &context, const strin
 		version_hint = GetTableVersionFromHint(meta_path, fs, DEFAULT_VERSION_HINT_FILE);
 		return GenerateMetaDataUrl(fs, meta_path, version_hint, options);
 	}
-	if (!UnsafeVersionGuessingEnabled(context)) {
+	if (!options.version_explicitly_set && !UnsafeVersionGuessingEnabled(context)) {
 		// Make sure we're allowed to guess versions
 		throw InvalidConfigurationException(
 		    "Failed to read iceberg table. No version was provided and no version-hint could be found, globbing the "
--- a/src/function/metadata/iceberg_snapshots.cpp
+++ b/src/function/metadata/iceberg_snapshots.cpp
@@ -60,6 +60,7 @@ static unique_ptr<FunctionData> IcebergSnapshotsBind(ClientContext &context, Tab
 			options.metadata_compression_codec = StringValue::Get(kv.second);
 		} else if (loption == "version") {
 			options.table_version = StringValue::Get(kv.second);
+			options.version_explicitly_set = true;
 		} else if (loption == "version_name_format") {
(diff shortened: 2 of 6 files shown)


Tmonster (approved):
thanks!

PR #1286, opened 2026-08-04
Run sigv4_mock tests in CI, fix test-local run step

### LINKED ISSUE #1283: sigv4_mock tests are never run in CI
## Problem

The tests under `test/sql/local/catalog_custom_setup/sigv4_mock/` are never executed in CI:

- `test_sigv4_http_scheme.test`
- `test_endpoint_type_custom_endpoint.test` (added in #1278)

Both tests guard on `require-env SIGV4_MOCK_SERVER_AVAILABLE` and require a running mock server:

```bash
python3 scripts/sigv4_mock_server.py 19130
```

No workflow sets `SIGV4_MOCK_SERVER_AVAILABLE` or starts the mock server, so these tests are silently skipped everywhere — including the `test-local-catalog.yml` run that recursively covers `test/*`.

Raised by @Tishj in https://github.com/duckdb/duckdb-iceberg/pull/1278#issuecomment-5179683452

## Proposed solution

Add a dedicated reusable workflow `.github/workflows/test-sigv4-mock-catalog.yml`, modeled on `test-nessie-catalog.yml` but much simpler — the mock server is a stdlib-only Python script, so no Docker, fixtures, or health-check polling beyond a port wait:

1. Download/extract the relassert build artifact (same as the other catalog workflows)
2. Start `python3 scripts/sigv4_mock_server.py 19130` and wait for the port to accept connections
3. Run with `SIGV4_MOCK_SERVER_AVAILABLE: 1`:

   ```bash
   ./build/relassert/test/unittest --order lex "$PWD/test/sql/local/catalog_custom_setup/sigv4_mock/*"
   ```

4. Stop the server

Then wire it into `BuildIceberg.yml` alongside the other catalog test jobs.

I plan to pick this up as a follow-up to #1278.

### LINKED ISSUE #1285: test-local-catalog.yml never actually runs the tests
## Problem

The "Run local tests" step in `.github/workflows/test-local-catalog.yml` never actually runs the tests — it only lists their names.

Line 79 glues both unittest invocations onto a single line:

```bash
./build/relassert/test/unittest --order lex "$PWD/test/*" --list-test-names-only || true ./build/relassert/test/unittest --order lex "$PWD/test/*"
```

Bash parses this as:

```
unittest ... --list-test-names-only  ||  true  ./build/relassert/test/unittest --order lex "$PWD/test/*"
```

`true` ignores its arguments, so the second invocation — the actual test run — never executes. The step passes as long as listing the tests succeeds.

The intended form (as used in `test-nessie-catalog.yml`) is two separate lines:

```bash
./build/relassert/test/unittest --order lex "$PWD/test/*" --list-test-names-only || true
./build/relassert/test/unittest --order lex "$PWD/test/*"
```

This is the same class of silent-skip problem as #1283: CI is green while the tests don't run.

Note: fixing this means the step will start executing `test/*` for real. The `catalog_custom_setup/*` subdirectories all guard on their own `require-env` variables (`FIXTURE_SERVER_AVAILABLE`, `NESSIE_SERVER_AVAILABLE`, `SIGV4_MOCK_SERVER_AVAILABLE`), so they will skip cleanly in this job; the rest of the suite runs against the generated fixture data the job already prepares.

Found while reviewing the CI change for #1283.

u70b3:
## Summary

- Closes #1283 — hook the `sigv4_mock` tests up to CI (raised by @Tishj in #1278)
- Closes #1285 — fix `test-local-catalog.yml` never actually running its tests

## Changes

**New workflow `test-sigv4-mock-catalog.yml`** (reusable, wired into `BuildIceberg.yml` alongside the other catalog jobs):

The tests under `test/sql/local/catalog_custom_setup/sigv4_mock/` guard on `require-env SIGV4_MOCK_SERVER_AVAILABLE` and need `scripts/sigv4_mock_server.py` running on port 19130. No workflow provided either, so they were silently skipped everywhere.

The workflow downloads the relassert artifact, starts the mock server (stdlib-only Python — no Docker/Java/fixtures, much simpler than the other catalog workflows), waits for the port, runs the tests with the env var set, and cleans up via trap.

**Fix `test-local-catalog.yml`**: both unittest invocations were on one line, so bash parsed the second one as arguments to `true` — the "Run local tests" step only ever listed test names. Split onto two lines, matching the nessie pattern. Note this means the job will now execute `test/*` for real; the `catalog_custom_setup/*` suites skip cleanly via their own `require-env` guards, but if any other test was silently broken, CI will now surface it.

## Validation

- Reproduced both directions locally with the workflow's exact bash: without `SIGV4_MOCK_SERVER_AVAILABLE` the tests skip; with it (and the server running) all 9 assertions pass.
- All touched YAML files parse; the run block passes `bash -n`.

@Tishj could you take a look?


Tishj:
I'll push some changes to this, so hopefully we can get this green soon. Thanks for identifying the problems 👍

Tishj:
Turns out `CALL iceberg_catalog.system.add_files` in Spark doesn't write the right format version into the key-value metadata of the produced manifest file.
~~And with #1280 we now use that version, and that causes an error 😅~~

EDIT:
We shouldn't error here, the field is present, it's just NULL, which is fine.
At that point we have lost the ability to determine if the field is actually present and NULL, or not present at all.

So for nullable fields we shouldn't be using `ReadRequiredField`

Issue #1260, opened 2026-07-30
Plan-time metadata-only DELETE: drop fully-covered data files from manifest bounds without scanning


paultmathew:
### Summary

Add a plan-time optimization for `DELETE` where, if a `DELETE`'s predicate provably covers **all** rows of a data file (from the file's partition and/or manifest column bounds), the file is dropped as a **metadata-only** delete — marked deleted in a new manifest — **without scanning it** and **without writing a positional/equality delete file**. Files only *partially* covered fall back to the existing scan + row-level delete path.

### Motivation / why a new approach

I took a first cut at metadata-only deletes in #1178, but that PR resolves whole-file deletes at **flush time** — i.e., after the rows have already been scanned, once it observes that every live row of a file was deleted. In practice that path ended up entangled with manifest-rewrite / snapshot-summary / compaction-adjacent concerns, and the PR grew unwieldy without delivering the win that actually matters for large partitioned tables: **not reading the data at all**.

I'd like to step back and pursue the optimization at **plan time** instead, which is both more self-contained and where the real I/O savings are (e.g. `DELETE ... WHERE dt = '2026-01-01'` on a table partitioned by `dt` should touch zero data files).

### Proposed approach

At DELETE planning, for each candidate data file:

1. Evaluate the delete predicate against the file's **partition values** and **manifest column bounds** (lower/upper bounds, null counts).
2. If the predicate provably selects **every** row of the file → drop the file metadata-only (new manifest entry marked `DELETED`), skip the scan.
3. If the predicate provably selects **no** rows → prune the file entirely (no-op).
4. Otherwise (partial / undecidable) → fall back to the current scan + row-level delete.

### Relationship to #341

This is essentially the **write-path counterpart** to #341 ([Predicate Pushdown] Filtering deletes based on row counts). #341 asks "can this scan predicate match *no* rows of a delete file, so we can ignore it?"; this asks "does this delete predicate match *all* rows of a data file, so we can drop it whole?" Both need the same bounds-vs-predicate evaluator over manifest metrics — it would be worth sharing that machinery.

### Scope / non-goals

- **In scope:** the plan-time whole-file drop + fallback, using existing manifest partition/bounds metadata.
- **Non-goals:** compaction / manifest rewriting, snapshot-summary accounting beyond what a normal delete already does, and equality-delete-file pruning on reads (that is #341).

### Open questions

- Reuse vs. share the evaluator with #341 — is there appetite to factor out a common bounds/predicate util?
[...]

PR #1287, opened 2026-08-04
Plan-time metadata-only DELETE for covered partitions

paultmathew:
## Summary

When a DELETE's predicate provably covers every row of a data file, drop that file at the
manifest level instead of scanning it and rewriting its rows as row-level deletes. Coverage is
decided per file against the partition spec that file was written under.

### A partition value, projected

Coverage from a partition value is a *strict projection*: rewrite the predicate on data rows into a
test on the file's partition value, such that if the partition value passes, every row of that file
satisfies the original predicate. This follows the Iceberg reference implementations -
`ProjectionUtil.truncateIntegerStrict` / `truncateLongStrict` / `truncateDecimalStrict` /
`truncateArrayStrict` in Java, and `_truncate_number_strict` / `_truncate_array_strict` in pyiceberg.

The comparison against the partition value is **always strict**, and the boundary case is recovered
by stepping the constant one unit in the *source* domain before transforming:

| source predicate | test on the partition value `v` |
| --- | --- |
| `c < T` | `v < t(T)` |
| `c <= T` | `v < t(T + 1)` |
| `c > T` | `v > t(T)` |
| `c >= T` | `v > t(T - 1)` |

A non-strict comparison is unsound here. If the transform is lossy and the constant lands in the
same partition as the file, the transformed constant equals the partition value, so `v >= t(T)`
holds for a file whose rows only partly match: `day(ts)` on TIMESTAMP would drop a whole day for
`ts >= '2021-05-01 10:00' AND ts <= '2021-05-01 11:00'`, and `truncate(10, id)` would drop the whole
`10..19` file for `id >= 15`.

Stepping is `IcebergTryIncrement` / `IcebergTryDecrement` in `common/iceberg_math`: 1 for the
integer widths, a day for DATE, a microsecond for TIMESTAMP/TIMESTAMPTZ, a nanosecond for the `_NS`
variants. A type with no exact successor keeps the constant unchanged, which tests `t(c) < t(T)` -
stricter than the rule asks for, so still sound. That is how `truncate` on a string is handled,
matching `_truncate_array_strict`, and it generalises to any type without a successor.

**Identity needs no projection.** The partition value is the source value, so the predicate applies
to the bounds unchanged and `=` keeps working.

**Equality is the conjunction of the two strict inequalities.** Proving both `v > t(T - 1)` and
`v < t(T + 1)` is exactly the case where the transform separates the constant from its neighbours,
which is what makes it one-to-one there: `day` on a DATE qualifies, `day` on a TIMESTAMP does not.

**Bucket and void are never order-preserving**, so a partition value says nothing about how its
source values compare and coverage is never claimed.

### The file's own column bounds

A partition normally holds many data files, and its value cannot tell them apart. A predicate cutting
*inside* a partition therefore proves nothing about any of its files from the partition value alone -
on a `day(ts)` table, `DELETE WHERE ts <= '2021-05-01 11:00'` would rewrite every file in the boundary
day, including files whose rows all sit before 11:00, producing a delete file that shadows 100% of a
data file where a manifest edit would do.

A data file's manifest entry also records per-column bounds - for the partition source column as much
as any other - and they are exact for that file. The inclusive pruner has always read them, which is
how a file lying entirely outside the predicate is already skipped. So when no partition field settles
a filter, the proof falls back to those bounds with the identity transform: they are already in the
column's domain, so there is nothing to project and the evaluator itself needs no extra case.

## When it applies

All of the following must hold, or the delete falls back to the existing row-level path:

- a standalone DELETE - not UPDATE, not MERGE (see Caveats)
- the table is partitioned, and every pushed-down filter is on a column with a partition field
- every filter was pushed down in full - a residual filter above the scan can still reject rows
- the predicate is a conjunction of `column <cmp> constant`
- the transform is identity or order-preserving (`truncate`, `year`, `month`, `day`, `hour`), or the
  file's own bounds for that column settle it
- for equality against a partition value, the transform must separate the constant from its neighbours
- the file's partition value is not NULL

So `IS NULL`, `IN`, `OR`, prefix `LIKE`, `<>`, and any function on the column keep the row-level path.

## Deferred to follow-ups

Deliberately out of scope, in rough order of value:

**Coverage on columns with no partition field.** `DELETE ... WHERE id < 100` where `id` is not a
partition column takes the row-level path even when the file's bounds for `id` would prove it covered.
The proof already handles this - it is the same bounds-based path - but the spec-level gate rejects the
filter before it runs, and that gate is cached per spec because it is file-independent, which a
per-file bounds proof is not. That caching question is the work.

**Unfiltered DELETE.** `DELETE FROM t` with no predicate covers every row by definition, and both
references handle it by letting `alwaysTrue` flow through the same evaluator. It needs a plan-time
signal rather than an empty filter set: a predicate that nothing could be pushed down from leaves the
same empty set behind with the real filter still enforced inside the scan, so reading the empty set as
"no predicate" would silently empty the table. The check has to establish both that only projections
sit between the scan and the delete, and that the scan's own filter set is empty. The unpartitioned
case belongs with it - `DELETE FROM t` on an unpartitioned table can drop every file too, which is
what Java does, but the `IsUnpartitioned()` guard here rejects it.

**`OR` and `IN`.** Both references support these in strict projection: Java preserves the boolean
structure and evaluates the projected expression against the partition tuple, and its
`StrictMetricsEvaluator.in` reduces to "the partition value is one of the candidates". They are
unreachable here for a pushdown reason rather than a design one. DuckDB pushes a single-column
disjunction, and the `(id = 1) OR (id = 3)` a non-contiguous `IN` stands for, into the scan only as an
*optional* filter with the real predicate kept above it, so `filters_fully_pushed` is false and no file
may be dropped. If that changes, the reference rules say what the implementation should be.

**`IS NULL` / `IS NOT NULL`.** The rule is trivial - `IS NULL` holds for every row iff the column has
no non-null values - but a null check does not arrive in the filter set this evaluator sees on the
DELETE path, so it cannot be proven. Note that file *pruning* does act on these filters, so the
plumbing differs between the two paths and should be established before building on it.

**`<>` strict projection.** The references project not-equal strictly for identity, truncate, bucket
and the time transforms, and it is the one point predicate a *lossy* transform can prove, since
`v != t(T)` implies no row equals `T`. DuckDB pushes `<>` down fully, so it is reachable. `NOT IN`
does not decompose into `<>` conjunctions - it stays as `NOT (col IN (...))` - so it needs separate
work.

**Residual-style combination.** Java's `ResidualEvaluator` discharges part of an expression against
the partition value and evaluates the rest against metrics. Only meaningful for a single column whose
filter needs both sources to be settled.

## Relationship to #1260

This closes both halves of #1260 for partition source columns: coverage from partition values and
coverage from manifest column bounds. Extending the second half to columns with no partition field is
the remaining piece, above.

Two of the open questions in the issue are answered by the implementation:

- **Sharing the evaluator with #341.** `IcebergPredicate::AllRowsMatchBounds` sits beside the
  existing `MatchBounds` and reuses the same `IcebergPredicateStats` and transform machinery, so
  "does this match no rows" and "does this match all rows" are one code path with two entry
  points - and the bounds-based half is literally the same call the inclusive pruner makes, with a
  different entry point.
- **Bound tightness.** Partition values are exact rather than truncated bounds, so on that half the
  concern does not arise. On the bounds-based half it is answered by containment: a widened range can
  only make a strict test harder to satisfy, so a bound that is loose or truncated costs an
  optimization rather than correctness. Note this is the same assumption the existing inclusive prune
  has always relied on - if bounds did not contain the true range, file pruning would already return
  incomplete `SELECT` results.

## Caveats

**The returned row count can over-report.** When a covered file is also touched by an equality
delete, or by a positional delete file that does not name it in `referenced_data_file`, the rows
those deletes already removed cannot be attributed to this file. A positional delete binds to data
files by the `file_path` column inside its payload, so the split is only knowable by opening the
delete file - which is the read this optimization exists to avoid. The manifest gives only the
delete file's total `record_count`. The drop itself is correct regardless; only the count is
affected. Reachable in builds with `-DICEBERG_ENABLE_EQUALITY_DELETE_WRITES=1`.

**UPDATE and MERGE are excluded deliberately.** Both rewrite the rows they delete, so a scan they
share has to keep emitting them. UPDATE calls `IcebergDelete::PlanDelete` directly and never opts
in; MERGE plans its delete action over the scan feeding its other actions, so it goes through
`PlanDeleteInternal`, which also does not. Without the MERGE exclusion,
`MERGE INTO t USING (...) s ON t.id = s.id AND t.p = 5 WHEN MATCHED THEN DELETE` silently
destroys an uninvolved row in the `p = 5` file, because the single-table conjunct is pushed into
the target scan and covers the whole file. Both paths have regression tests.

**Same-transaction visibility.** Files dropped by an uncommitted statement are recorded on the
transaction, so reads later in that transaction see them as gone rather than waiting for the
commit to rewrite the manifests. Without it the DELETE reports N rows removed while a following
SELECT still returns them, and repeating the DELETE counts them again.

**Orphaned delete files.** A positional delete file covering several data files survives when one
of them is dropped - the others still need it - and retains rows pointing at a file that no longer
exists. Harmless on read; cleared by compaction.

## Observability

Dropping a file rather than scanning it is otherwise invisible, so:

- one Iceberg log entry per dropped file, alongside the existing filter-pushdown pruning logs
- a per-statement summary: data files dropped, records they held, delete files superseded
- EXPLAIN reports `Metadata-Only Delete: enabled`, meaning the delete is permitted to drop files,
  not that it did - that is only known after the scan

## Tests

- coverage from partition values: identity equality and ranges, multi-partition ranges, two partition
  columns under one conjunction, `day(DATE)` equality, whole-partition ranges on
  `day`/`hour`/`month`/`year`, `cast(ts AS DATE)` equality
- coverage from a file's own bounds: three files written into one day partition, one entirely inside
  the predicate, one straddling it and one entirely outside - asserting one manifest-level drop, one
  delete file, and one file untouched. The partition value is identical for all three, so this case
  fails without the bounds-based half
- data-loss guards on lossy transforms, in `delete_lossy_transform_coverage.test`: equality on
  `day(TIMESTAMP)`, `hour`, `month`, `year`, `truncate` on INT and VARCHAR and `bucket`; and a range
  landing inside a single partition on each of `day`, `hour`, `month`, `year`, `truncate(10, int)` and
  `truncate(3, varchar)`. Each asserts both the surviving rows and that a delete file was written -
  the latter is what distinguishes "took the row-level path" from "dropped the file and happened to be
  right"
- guards where the optimization must not fire: non-partition column, partial AND, OR,
  non-contiguous IN, partially-pushed prefix LIKE, unpartitioned table, NULL partition value, and
  a MERGE delete action
- interaction with existing delete files, v2 positional and v3 deletion vector: paired
  invalidation, and a count of live rather than physical rows
- same-transaction visibility, repeated drop, and rollback
- partition spec evolution: a file under the old spec is judged against its own spec
- snapshot summary metrics, and the observability output above

Nessie skips the v3 delete-file test and the snapshot-metrics test, matching its existing entries
for `test_delete_from_v3_table` and `test_snapshot_summary_delete_metrics`.

## Regression benchmarks

The file-pruning regression step globs `benchmark/file_pruning_benchmarks/*.benchmark` at run time
and writes its own CSV, so these five are picked up with no change to `Regression.yml` and no new
file under `.github/regression/`. Unlike the TPCH and metadata-late-materialization suites, this
group deliberately has no committed CSV.

The DELETE benchmarks use `require_reinit`, since a DELETE is not idempotent and the runner times
`run` five times; re-init happens outside the timer.

Local comparison against the merge base, both binaries built with `BUILD_BENCHMARKS=1` and
`-DICEBERG_ENABLE_EQUALITY_DELETE_WRITES=1`, median of 12 alternating timed runs (threshold +10% and
+0.050s):

| Benchmark | main | PR | Change |
| --- | --- | --- | --- |
| metadata_only_delete_whole_partition | 0.120s | 0.019s | -84.4% |
| metadata_only_delete_many_partitions | 0.486s | 0.030s | -93.8% |
| metadata_only_delete_partial_partition | 0.297s | 0.033s | -88.8% |
| scan_after_metadata_only_delete | 0.059s | 0.003s | -94.3% |
| row_level_delete_uncovered_predicate | 0.500s | 0.487s | -2.5% |

`metadata_only_delete_partial_partition` is the one that needs a file's own bounds: all 87 of its data
files carry the same day, so the partition value cannot separate them. With only the partition-value
half of the proof it measures 0.291s, indistinguishable from main.

`row_level_delete_uncovered_predicate` is the control: it filters on a non-partition column, so the
coverage proof can only ever fail and every row goes through the row-level path. It measures flat
within noise, which matches the mechanism - the per-spec gate rejects that predicate before any
per-file coverage proof runs, so the added work is never reached on the fallback path.


src/common/iceberg_math.cpp
```diff
@@ -0,0 +1,110 @@
...
+		return TryStepTimestamp<timestamp_tz_ns_t>(value, increment, result);
+	default:
+		return false;
+	}
+}
+
+} // namespace
+
+bool IcebergTryIncrement(const Value &value, Value &result) {
+	return TryStep(value, true, result);
```
Tishj: Since we're already templating, might as well make `increment` a template parameter as well, so the branching costs nothing at runtime

src/include/catalog/rest/transaction/iceberg_transaction_data.hpp
```diff
@@ -84,6 +84,9 @@ struct IcebergTransactionData {
 	optional<int64_t> base_snapshot_id;
 	//! The 'referenced_data_file' -> 'data_file.file_path' of the currently active transaction-local deletes
 	case_insensitive_map_t<string> transactional_delete_files;
+	//! Union of the files invalidated by the alters staged so far. The commit materializes these as DELETED
+	//! manifest entries, but reads have to treat them as gone from the moment they are staged.
+	IcebergManifestDeletes invalidated_files;
```
Tishj: I'm actually surprised that this wasn't already there, but I guess we kept these inside the `IcebergAddSnapshot` update instead.
I believe at some point there was an idea to support time traveling through transactional data, but that was dropped.
We should probably remove the `IcebergManifestDeletes` from `IcebergAddSnapshot` and use this instead.

At some point we have to revise this btw, since V4 will add manifest-level deletion vectors, in which case we should probably turn the shape of these deletes into a pair of (manifest_list_entry_index, manifest_entry_index)

src/planning/pruning/iceberg_file_pruner.cpp
```diff
@@ -9,6 +9,72 @@
...
+		if (mapping.field_id == field_id) {
+			return true;
+		}
+	}
+	return false;
+}
+
+//! The bounds and null counts a data file records for one column. False when the manifest carries nothing
+//! usable, in which case no bound-based reasoning is possible for that column.
+bool TryGetColumnBoundStats(const IcebergTableMetadata &metadata, const IcebergDataFile &data_file,
```
Tishj: Can we return `optional<IcebergPredicateStats>` instead of using the `result` as an "out" parameter and returning bool ?

src/planning/pruning/iceberg_file_pruner.cpp
```diff
@@ -171,6 +220,120 @@ bool IcebergFilePruner::FileMatchesFilter(const IcebergManifestFile &manifest_fi
...
+	}
+
+	//! Every pushed-down filter must be provably satisfied by every row of the file, either by the partition
+	//! value or by the file's own bounds for that column; otherwise fall back to the row-level path.
+	for (auto &entry : table_filters) {
+		auto &filter_column_index = entry.first;
+		auto &filter = *entry.second;
+
+		bool covered = false;
+		for (auto &field : partition_spec.fields) {
```
Tishj: Something tells me we can invert this, iterating over the fields of the partition spec in the outer loop, then looking up the `ColumnIndex` in the `source_to_column_id`, and finally using `table_filters.GetFilterForColumnIndex` to find the assigned filter (if any)

paultmathew: Inverted to iterate the partition fields, but split into two passes, because the invariant is per-filter (every filter must be covered), not per-field:

- Pass 1 (fields): for each field, resolve its column via source_to_column_id and look up the filter, run the partition-value proof with that field's transform, and record the covered filter. Iterating fields means each is visited once, so a column carrying several fields (bucket + truncate, etc.) has all of them tried — any one proving coverage settles it.
- Pass 2 (filters): the completeness check. Any filter Pass 1 didn't settle falls back to the file's own bounds; a filter neither pass proves drops us to the row-level path. This has to be filter-driven — the bounds fallback and the "every filter covered?" question are both keyed by filter, not field.

Complexity goes from O(filters × fields) (the old nested scan) to O(fields + filters), since both passes are linear and the lookups are O(1). Used TryGetFilterByColumnIndex rather than GetFilterForColumnIndex to keep it allocation-free (the latter Copy()s / extracts a sub-filter) and to preserve the existing coverage semantics.

Bonus: this extends cleanly to the deferred coverage on non-partition columns follow-up. A filter on a column with no partition field is simply never marked in Pass 1, so Pass 2 already runs the bounds fallback on it.

src/include/catalog/rest/transaction/iceberg_transaction_data.hpp
paultmathew: Thanks for reviewing. Agreed this overlaps with the per-snapshot `altered_manifests`, and consolidating
matches the existing FIXME about moving transactional state into `IcebergTransactionData`.

The catch is direction: we can't drop `altered_manifests` and just use the union.
Each statement commits its own snapshot, chained (base → S1 → S2 …), and each has to
rewrite manifests for exactly the files *it* dropped — otherwise the intermediate
snapshots over-report and time-travel breaks. The flattened union has forgotten which
statement dropped what, so it can't do that job.

So I'd flip it: make `IcebergTransactionData` the single home for pending deletes,
keyed by statement, and derive both views from it — the per-statement set for commit,
and a flattened index for the O(1) read checks. That kills the manual `Merge` + assign
dual-write, gives one source of truth, and lets `altered_manifests` come off
`IcebergAddSnapshot` — right where that FIXME points.

It's a pretty substantial refactor though — it touches the commit path and the
`Add{,Delete,Update}Snapshot` methods that INSERT/UPDATE/DELETE all share, so it's
orthogonal to the metadata-only DELETE here. Happy to do it as a dedicated follow-up
rather than blow up this PR's scope, but since it reshapes shared commit machinery I'd
want your alignment on the direction first. Same for the V4 deletion-vector reshape —
I'd fold that into the same follow-up.

Tishj: Hmm right, yea I've wanted to "fix" the situation that we produce multiple snapshots per table in a single transaction.
But I'm afraid that there are situations where we can't, for example when the schema changes in the same transaction.
Not super relevant to the conversation though ^.

I like the idea of flattening the altered-manifests.
Perhaps that's even worth doing in a separate PR before we merge this.

Since your altered-manifests at the Transaction level have the same over-reporting problem you mentioned, no?
> otherwise the intermediate
snapshots over-report and time-travel breaks

Tishj:
I still need to really grasp what we're doing here, so this is not a final review, more just things I noticed along the way.

My initial thoughts on this were too simple, that's for sure. Because of the state of pushdown in DuckDB and the sometimes limited usefulness of bounds in iceberg this has gotten a lot more complex than I was expecting, but I believe it's a necessary complexity.

src/include/catalog/rest/transaction/iceberg_transaction_data.hpp
paultmathew: Sounds good — I'll centralize pending deletes in `IcebergTransactionData` as a
separate PR and rebase this one on top.

On the over-reporting: it only bites if it's a pure flat union. The plan is to key the
transaction-level structure by statement, so each snapshot still pulls exactly its own
altered-manifests at commit — no over-reporting there. The flat union stays, but only
as a derived read-side index: mid-transaction reads want "gone if *any* statement
dropped it," so flattening is exactly right for that consumer. Same data, two
projections — per-statement for commit, flattened for reads.

And agreed we can't always collapse to a single snapshot — good
thing is this refactor doesn't need to. It keeps one snapshot per statement and just
moves the per-statement deletes into the transaction, so where the snapshots come from
doesn't change.

Let me know if you're aligned with this and I can put up the PR.

paultmathew: Opened #1301 for this.

`pending_deletes` lives on `IcebergTransactionData` now, keyed by the statement's
`IcebergAddSnapshot` (`reference_map_t`), so `IcebergAddSnapshot` holds no delete state
at all — no copy, no index. The attribution comes from the insert and is resolved by
identity at commit.

So no over-reporting: commit looks up its own entry, and each snapshot in the chain
still rewrites only the files its own statement invalidated. Fetching a snapshot's
pending deletes is O(1); the file-level question — has anything in this transaction
dropped this file — is O(snapshots), one hash lookup per entry.

Once that lands I'll rebase this PR onto it and drop `invalidated_files` and its `Merge`
in favour of deriving the read view from `pending_deletes`.

paultmathew:
@Tishj Thanks for implementing the versioned `IcebergManifestDeletes`. I've incorporated the
changes in this PR, so it no longer carries its own copy of the transaction-wide ledger.

src/execution/operator/merge_into/iceberg_merge_into.cpp
```diff
@@ -213,7 +213,7 @@ static unique_ptr<MergeIntoOperator> IcebergPlanMergeIntoAction(IcebergCatalog &
 			delete_op.expressions.push_back(std::move(ref));
 		}
 		delete_op.bound_constraints = std::move(bound_constraints);
-		result->op = catalog.PlanDelete(context, planner, delete_op, child_plan);
+		result->op = catalog.PlanDeleteInternal(context, planner, delete_op, child_plan);
```
Tishj: Could use a comment explaining why we don't use `PlanDelete` here, I assume it's because of the `EnableMetadataOnlyDelete` path that we don't want to hit here.

src/planning/iceberg_multi_file_list.cpp
```diff
@@ -718,6 +794,56 @@ IcebergDeletePlan IcebergMultiFileList::ProcessDeletes(const BoundIcebergManifes
...
+	}
+
+	vector<DroppedDataFile> result;
+	result.reserve(covered_files.size());
+	for (auto &covered : covered_files) {
+		DroppedDataFile dropped;
+		dropped.file_path = covered.file_path;
+		dropped.live_record_count = covered.record_count;
+
+		unique_ptr<BoundIcebergManifestEntry> data_manifest_entry;
```
Tishj: `optional<BoundIcebergManifestEntry>`, or preferably a method that directly constructs, so we don't need to use tricks to delay the construction.

src/execution/operator/merge_into/iceberg_merge_into.cpp
paultmathew: That's right. Added a comment.

src/planning/iceberg_multi_file_list.cpp
paultmathew: Added a method. `BindCoveredDataFile` takes the lock and returns the entry, so it's
constructed where it's declared so there's no allocation per dropped file, and no empty state to
reason about. The lock stays inside it because `ResolveApplicableDeleteFiles` takes the
lock itself, so the entry has to outlive it.

paultmathew:
The failing Regression test doesn't seem to be a timing regression. The base binary runs out of `base/` at the merge base, so the five benchmarks this branch adds report "Benchmark to run could not be found" there, while the two already on main compare fine. It only started failing now because the rewritten `test_runner.py` from the duckdb bump returns 1 on execution failures where the old one only failed on regressions.

Tishj:
Is it at all possible to split this further?
Because I find it quite hard to review when there's also moving of existing logic mixed in with the actual changes

paultmathew:
@Tishj that makes sense. I can split this into independently reviewable pieces:

**Refactors (no behaviour change)**
1. Extract `ResolveApplicableDeleteFiles` out of `ProcessDeletes` — the delete-file resolution loop lifted out verbatim.
2. Extract `ApplyNullCounts` out of `FileMatchesFilter`.
3. Extract `PlanDeleteInternal` from `PlanDelete` (`PlanDelete` just forwards — a no-op until the feature enables it).

**Purely additive**
4. `IcebergTryIncrement`/`IcebergTryDecrement` in `common/iceberg_math` (+ unit tests).
5. The coverage proof — `AllRowsMatchBounds`, `FileFullyCoveredByFilter`, `PartitionSpecHasFieldForEveryFilterColumn` (+ unit tests).

**Behaviour change**
6. Wire the metadata-only DELETE: record covered files in the scan, drop them in Finalize`, skip the empty manifest, same-transaction visibility, the MERGE exclusion, EXPLAIN line + integration tests.
7. Regression benchmarks.

1–4 are independent; 5, 6 and 7 need to be stacked on top.

Do you prefer these as separate stacked PRs, or one PR with the same breakdown as a clean, rebased commit series Either works for me — whichever is easier to review.

Tishj:
I'd say put 1-3 into one PR, if they're just moving existing logic around then it'll be easy to review, that will shrink this PR's changes and we can evaluate from there whether it makes sense for more parts to be split or its already reviewable 👍

paultmathew:
@Tishj thanks for merging #1316 in. Let me know if you want me to split it further. Also the failing regression test seems to be because the base doesn't contain the benchmark. Happy to scope out a fix.

src/include/planning/iceberg_multi_file_list.hpp
```diff
@@ -126,6 +174,16 @@ struct IcebergMultiFileList : public MultiFileList {
 	vector<string> names;
 	vector<LogicalType> types;
 	IcebergTableFilters table_filters;
+	//! True when the current filter set was pushed down in its entirety (every filter PUSHED_DOWN_FULLY),
+	//! so there is no residual filter above the scan. Required before any file may be dropped by a delete.
+	bool filters_fully_pushed = false;
+	//! True only for the scan that drives a metadata-only DELETE. Gates the skip+register branch in GetDataFile
+	//! so ordinary SELECTs never drop covered files.
+	bool metadata_only_delete_enabled = false;
```
Tmonster, line 182: nit: could we rename this to `scan_is_delete_source`? The setter function name can then also be updated

src/include/planning/iceberg_multi_file_list.hpp
```diff
@@ -126,6 +174,16 @@ struct IcebergMultiFileList : public MultiFileList {
...
+	//! Data files whose every row is covered by the fully pushed-down delete predicate, recorded during the scan.
+	mutable vector<CoveredDataFile> fully_covered_data_files DUCKDB_GUARDED_BY(shared_state->lock);
```
Tmonster: `unselected_data_file_for_delete` might be a better name?

src/include/planning/iceberg_multi_file_list.hpp
```diff
@@ -37,6 +37,33 @@ struct IcebergMultiFileReader;
 struct IcebergDeleteFileReference;
 
 struct IcebergMultiFileList : public MultiFileList {
+public:
+	//! A data file whose every row is provably covered by the (fully pushed-down) scan filter, recorded
+	//! during a metadata-only DELETE so the file can be dropped without a row-level scan.
+	struct CoveredDataFile {
```
Tmonster, line 43: do these two structs need to be here? Maye we can move them out of `IcebergMultiFileList`?
Also, is there a strong reason we should not merge these? A `CoveredDataFile` means every row is provable covered by the pushed down scan filter, it seems like that will eventually become a `DroppedDataFile` as well

src/planning/iceberg_multi_file_list.cpp
```diff
@@ -166,6 +173,26 @@ void IcebergMultiFileList::DisableServerSidePlanning() {
...
+bool IcebergMultiFileList::PartitionSpecHasFieldForEveryFilterColumn(
+    int32_t partition_spec_id, annotated_lock_guard<annotated_mutex> &guard) const {
+	auto it = spec_has_fields_for_filters.find(partition_spec_id);
+	if (it != spec_has_fields_for_filters.end()) {
+		return it->second;
+	}
+	auto can_cover = IcebergFilePruner(context, GetMetadata(), GetSchema(), table_filters)
+	                     .PartitionSpecHasFieldForEveryFilterColumn(partition_spec_id);
+	spec_has_fields_for_filters[partition_spec_id] = can_cover;
+	return can_cover;
```
Tmonster, line 193: what if one spec has filters but the other doesn't? This will only return the result for the last partition spec

Tmonster (commented):
Thanks! A couple of questions comments and then I will do another review

We have `FileMatchesFilter` functionality right now that returns `true/false`. Potentially `FileFullyCoveredByFilter` could merge with `FileMatchesFilter` and instead of returning a bool you can return an enum value like

```
enum class METADATA_STATS_PUSHDOWN { NO_ROWS_MATCH, SOME_ROWS_MATCH, ALL_ROWS_MATCH };
```

Then on a delete scan if `FileMatchesFilter` returns `METADATA_STATS_PUSHDOWN::ALL_ROWS_MATCH`, you can add it to the list of metadata delete files. 

Maybe that could make more sense?

src/planning/iceberg_multi_file_list.cpp
paultmathew: Should be fixed in #1357 since we're deciding coverage at the file level.

paultmathew:
@Tmonster thanks for reviewing! Merging `FileMatchesFilter` and `FileFullyCoveredByFilter` makes sense to me. I split it out into #1357 to (hopefully) keep things reviewable. I was planning to address the remaining comments on this PR after rebasing on top of #1357, but happy to revise my approach if you'd prefer something different.

PR #1291, opened 2026-08-05
[ManifestListReader] Add support for reading V1-produced manifest lists

### LINKED ISSUE #1284: [ManifestListReader] Manifest list reader isn't properly prepared for older versions
### What happens?

Recently we fixed a problem related to versioning in #1280 , so we properly use the version that was stamped on the manifest when it was written.

The problem is different for the manifest *list* though, because that file **doesn't** get stamped with a version.
That, combined with the fact that version upgrades don't force a new snapshot, means that we could be reading a V1 manifest list, when the table says V3.

That means we have to read every manifest list as if it was written by V1, so even if the spec says `required`, that doesn't really mean anything and we have to treat it as optional instead.

Tishj:
This PR fixes https://github.com/duckdb/duckdb-iceberg/issues/1284

Most notably, the `existing|added|deleted_(file|row)_count` used to be `optional`, and as explained by the mentioned issue, manifest lists can't be versioned, so we always have to expect to find a V1 manifest list.
So these attributes are now read as optional.

There is also another problem with these attributes in our reader, we were expecting them to non-null, which isn't guaranteed by the spec. Those are properly nullable now.
For example:
> Number of entries in the manifest that have status EXISTING (0), when null this is assumed to be non-zero

### Parallel Manifest Scans

We use these counts to be able to reserve enough size upfront to ensure that our references don't get invalidated: see https://github.com/duckdb/duckdb-iceberg/pull/844

If those counts aren't available, we can't parallelize the scan for the manifest anymore, because that might result in earlier references getting invalidated when newer entries are added and the vector reallocates.
So we have to scan those eagerly (or at least emit only 1 `ManifestReadBatch` for the manifest)


src/catalog/rest/api/iceberg_table_update.cpp
```diff
@@ -67,6 +73,17 @@ void IcebergCommitState::LoadExistingManifests(DatabaseInstance &db,
 		}
 	}
 
+	//! To deal with V1 manifest lists, which don't have added/deleted/existing file counts
+	//! We have to read all the entries of the manifest to materialize these counts on-demand
```
Tishj: Hmm these comments aren't necessarily true, this can also happen when the counts in the manifest file are `NULL`

PR #1292, opened 2026-08-05
Support `Create table SORTED BY` and `ALTER TABLE SET SORTED BY`

Tmonster:
DuckDB currently does not support altering a table and setting sorted by. This PR adds that functionality to main with some tests. 

Currently, depending on certain options and how data is inserted, DuckDB-Iceberg will not be guaranteed to write sorted data. Only if the batch writer is called, then data is written in the same order it is received, but it is not guaranteed that the batch writer is called if we start setting row group sizes.

Since writing sorted data is not required, we will let this go, but @lnkuiper is working on fixing the parquet copy to to allow writing sorted data regardless of the settings.

### DIFF
--- a/src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
+++ b/src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
@@ -394,7 +394,6 @@ IcebergSortOrder IcebergTableInformation::BuildSortOrder(const vector<OrderByNod
 		field.transform = transform;
 		field.direction = ascending ? "asc" : "desc";
 		if (order.null_order == OrderByNullType::ORDER_DEFAULT) {
-			//! Iceberg ties the default null order to the direction
 			field.null_order = ascending ? "nulls-first" : "nulls-last";
 		} else {
 			field.null_order = order.null_order == OrderByNullType::NULLS_FIRST ? "nulls-first" : "nulls-last";
--- a/src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
+++ b/src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
@@ -386,14 +386,32 @@ IcebergSortOrder IcebergTableInformation::BuildSortOrder(const vector<OrderByNod
 		}
 		auto source_id = source_columns[0].get().id;
 
+		//! An unqualified sort key (no ASC/DESC) sorts ascending
+		auto ascending = order.type != OrderType::DESCENDING;
+
 		IcebergSortOrderField field;
 		field.source_id = source_id;
 		field.transform = transform;
-		field.direction = order.type == OrderType::ASCENDING ? "asc" : "desc";
-		field.null_order = order.null_order == OrderByNullType::NULLS_FIRST ? "nulls-first" : "nulls-last";
+		field.direction = ascending ? "asc" : "desc";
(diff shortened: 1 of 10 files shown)


test/sql/local/catalog_test_config_setup/catalog_agnostic/alter/sorted_by/test_create_table_sorted_by.test
```diff
@@ -0,0 +1,241 @@
...
+unnest(metadata."sort-orders"::VARIANT[]) AS t(sort_order)
+WHERE try_cast(sort_order."order-id" AS INTEGER) = metadata."default-sort-order-id";
+----
+1	identity	asc	nulls-first
+
+statement ok
+insert into my_datalake.default.test_create_table_sorted_by values (4, 40), (2, 20), (3, 30), (1, 10);
+
+statement ok
+set variable file_path = (
```
Tishj, line 46: This will need a transaction to keep the credentials in scope

Tishj:
> Since writing sorted data is not required, we will let this go

This is true, but if we write `manifest_entry.data_file.sort_order_id` to a non-null value, then we have to make sure we're actually writing it in that order.

We didn't set that before, but starting with https://github.com/duckdb/duckdb-iceberg/pull/1135 we do

src/catalog/rest/catalog_entry/table/iceberg_table_information.cpp
```diff
@@ -399,16 +386,33 @@ IcebergSortOrder IcebergTableInformation::BuildSortOrder(const vector<OrderByNod
 		}
 		auto source_id = source_columns[0].get().id;
 
+		//! An unqualified sort key (no ASC/DESC) sorts ascending
```
Tishj, line 389: If we're going to flip the default anyways, then let's make it responsive to `default_order` and `default_null_order`, which are the settings to control this behavior in core.

See `DBConfig::ResolveNullOrder`

src/include/catalog/rest/catalog_entry/table/iceberg_table_information.hpp
```diff
@@ -47,6 +47,10 @@ struct IcebergTableInformation {
 	                                               idx_t base_partition_field_id);
 	static IcebergSortOrder BuildSortOrder(const vector<OrderByNode> &orders, const IcebergTableSchema &schema,
 	                                       int32_t sort_order_id);
+	//! Build an IcebergSortOrder from bare CREATE TABLE ... SORTED BY expressions, which carry no
+	//! direction/null-order of their own and so use the Iceberg defaults (ascending, nulls first).
```
Tishj, line 51: Where is this (ascending, nulls first) coming from?
If I look through the spec I think I found what you're referring to:
> These rows must be sorted (in ascending manner with NULL FIRST) by partition field to optimize filtering rows while scanning

But that's not talking about the sorting of the table

Tishj: Also, what is this about?:
> which carry no direction/null-order of their own

Unless the `CREATE TABLE` `SORTED BY` is parsed different than `ALTER TABLE`, this should be possible for create as well, no?
```sql
ALTER TABLE my_datalake.default.test_set_sorted_by_orderings SET SORTED BY (
	c DESC NULLS FIRST
);
```

PR #1293, opened 2026-08-05
Add iceberg_rollback_to_snapshot table function

The-Alchemist:
## Summary
- Add `iceberg_rollback_to_snapshot(catalog.schema.table, snapshot_id)` matching Spark [`rollback_to_snapshot`](https://iceberg.apache.org/docs/latest/spark-procedures/#rollback_to_snapshot): metadata-only `set-snapshot-ref` on `main`, ancestor validation, returns `previous_snapshot_id` / `current_snapshot_id`.
- Reuse a shared `SetSnapshotRef` table update for DML commits and rollback; also populate `parent_snapshot_id` when parsing snapshots so ancestry checks work.
- Add bind-time validation and catalog-agnostic maintenance coverage for restore, same-snapshot no-op, unknown ids, and non-ancestor rejection.

### DIFF
--- a/src/catalog/rest/transaction/iceberg_transaction_data.cpp
+++ b/src/catalog/rest/transaction/iceberg_transaction_data.cpp
@@ -383,4 +383,48 @@ void IcebergTransactionData::TableSetLocation() {
 	updates.push_back(make_uniq<SetLocation>(table_info.table_metadata.location));
 }
 
+static bool IsAncestorOfCurrentSnapshot(const IcebergTableMetadata &metadata, int64_t target_snapshot_id,
+                                        int64_t current_snapshot_id) {
+	if (target_snapshot_id == current_snapshot_id) {
+		return true;
+	}
+	optional_ptr<const IcebergSnapshot> cursor = metadata.FindSnapshotByIdInternal(current_snapshot_id);
+	while (cursor) {
+		if (!cursor->parent_snapshot_id) {
+			return false;
+		}
+		auto parent_id = *cursor->parent_snapshot_id;
+		if (parent_id == target_snapshot_id) {
+			return true;
+		}
+		cursor = metadata.FindSnapshotByIdInternal(parent_id);
+	}
+	return false;
+}
+
+void IcebergTransactionData::TableRollbackToSnapshot(int64_t snapshot_id) {
+	auto &metadata = table_info.table_metadata;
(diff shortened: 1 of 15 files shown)


Tishj:
Looks like you'll need to skip the test on Nessie

src/catalog/rest/transaction/iceberg_transaction_data.cpp
```diff
@@ -357,4 +357,51 @@ void IcebergTransactionData::TableSetLocation() {
...
+		throw InvalidInputException("Cannot roll back to snapshot, not an ancestor of the current state: %lld",
+		                            snapshot_id);
+	}
+
+	// Optimistic concurrency: require main still points at the snapshot we observed.
+	requirements.push_back(make_uniq<AssertRefSnapshotId>(current_snapshot_id));
+	updates.push_back(make_uniq<SetSnapshotRef>(snapshot_id));
+
+	// Align current schema with the target snapshot when it differs.
+	auto target_schema_id = target->GetSchemaId();
```
Tmonster: shouldn't this be `updates.push_back(make_uniq<SetCurrentSchemaId>(target_schema_id))` or something similar?

Tmonster (commented):
Awesome thanks!

I just tested what spark does, and they do not reset the schema id to the schema_id of the snapshot. I feel like this is bad UX, I would think a rollback would include setting the schema to the schema_id in the snapshot? Would it be crazy to add that here?

@The-Alchemist what do you think here?
CC @Tishj

Tmonster:
Seems like setting the schema to the original value has already been discussed here, with no clear resolution it seems
https://github.com/apache/iceberg/issues/5591

The-Alchemist:
@Tmonster 

Great point, I didn't know about Spark's behavior.  I feel like it makes sense to mimic Spark, even if it might be confusing.

latest commit:
* mimic https://github.com/apache/iceberg/issues/5591 snapshot behavior and `test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rollback_to_snapshot_schema_retained.test` to lock that behavior in (if we want to)

Tmonster (approved):
Yea, I would say let's stick with Spark functionality for now, although it is confusing.

Thanks! I've restarted the failing/cancelled CI

Tmonster:
Seems like you have to skip some more nessie tests

Tmonster:
The regression tests seem unrelated. I will keep my eye out for those.
Thanks!

PR #1298, opened 2026-08-06
Reduce lock holding time for GetEntry

NiclasHaderer:
Running a lot of queries simultaneously slows all of them down, because they are all waiting for the same lock in GetEntry. This PR reduces this waiting time by taking the lock for a shorter time


Tishj:
I'm sorry but I wouldn't categorize this as a bug fix, therefore I'd like this PR to be rebased to `main`, where it will land in duckdb 2.0

NiclasHaderer:
Makes sense, I retargetted to main

Tishj:
Thanks!

PR #1301, opened 2026-08-06
Move a statement's staged manifest deletes onto the transaction

paultmathew:
## Summary

Move the manifest deletes a statement stages from `IcebergAddSnapshot` onto
`IcebergTransactionData`, attributed to the snapshot of the statement that produced
them. No behaviour change — the same deletes reach the same manifest rewrite; only
where they are stored changes.

## Why

`IcebergAddSnapshot`'s constructor carries a note:

> This is a bit of a code smell, the `IcebergTableInformation` should instead be const
> and all transactional changes should live in the `IcebergTransactionData`

The deletes a statement stages are transactional change, and they were living on the
snapshot object rather than on the transaction. After this they have one home, and
`IcebergAddSnapshot` carries no transactional state of its own beyond the manifests it
is going to write.

## Why the store is keyed rather than flat

A transaction commits one snapshot per statement, chained (`base → S1 → S2 → …`). Each
snapshot's manifest rewrite has to cover exactly the files *that* statement invalidated.
A single flattened set would mark a file deleted in every snapshot of the chain,
including the ones that precede the statement that deleted it — so an intermediate
snapshot would over-report and time travel would show a file gone before the statement
that removed it. The per-statement attribution is therefore load-bearing, and
`pending_deletes` keeps it.

The key is the statement's own `IcebergAddSnapshot` (`reference_map_t`), so the
attribution is established by the insert and resolved by identity at commit. There is no
index or id for the snapshot to hold and keep aligned with a container.

## Relation to #1287

#1287 (plan-time metadata-only DELETE) adds a second consumer for these deletes: a read
inside an uncommitted transaction has to treat manifest-level dropped files as gone
before the commit rewrites anything. On that branch this is a separate flattened
`invalidated_files` on the transaction, kept in step with the per-snapshot copies by an
explicit `Merge` on every mutation — the overlap raised in review there, along with the
question of whether a transaction-level structure would inherit the same over-reporting
problem.

This PR is the answer to both. One keyed store means commit still reads only its own
entry, so there is no over-reporting; and #1287 rebases onto this to derive its read-side
view from `pending_deletes` rather than maintaining a parallel union, which removes the
dual write.

### DIFF
--- a/src/catalog/rest/api/iceberg_add_snapshot.cpp
+++ b/src/catalog/rest/api/iceberg_add_snapshot.cpp
@@ -91,6 +91,7 @@ void IcebergAddSnapshot::ConstructManifestList(IcebergManifestList &new_manifest
                                                IcebergSnapshotMetrics &snapshot_metrics) const {
 	//! Construct the manifest list
 	//! FIXME: RETRY_BLOCKER: no guarantee that no new deletes are introduced
+	auto &altered_manifests = commit_state.table_info.transaction_data->GetPendingDeletes(*this);
 	if (altered_manifests.IsEmpty()) {
 		for (auto &manifest_list_entry : commit_state.manifests) {
 			AddManifestListEntry(new_manifest_list, std::move(manifest_list_entry));
--- a/src/catalog/rest/transaction/iceberg_transaction_data.cpp
+++ b/src/catalog/rest/transaction/iceberg_transaction_data.cpp
@@ -138,6 +138,14 @@ bool IcebergTransactionData::ContainsDelete() const {
 	return false;
 }
 
+const IcebergManifestDeletes &IcebergTransactionData::GetPendingDeletes(const IcebergAddSnapshot &snapshot) const {
+	auto entry = pending_deletes.find(snapshot);
+	if (entry == pending_deletes.end()) {
+		throw InternalException("No pending deletes were staged for this snapshot");
+	}
+	return entry->second;
+}
+
 bool IcebergTransactionData::SupportsAppendRetry() const {
 	if (!requirements.empty() || pending_current_schema_id.has_value()) {
 		return false;
@@ -220,7 +228,7 @@ void IcebergTransactionData::AddSnapshot(IcebergSnapshotOperationType operation,
 	if (table_metadata.current_snapshot_id) {
 		TableAddAssertCurrentSchemaId();
 	}
-	add_snapshot->altered_manifests = std::move(altered_manifests);
+	pending_deletes.emplace(*add_snapshot, std::move(altered_manifests));
 
 	alters.push_back(*add_snapshot);
 	updates.push_back(std::move(add_snapshot));
@@ -257,7 +265,7 @@ void IcebergTransactionData::AddDeleteSnapshot(partitioned_manifest_entry_map_t 
 	if (table_metadata.current_snapshot_id) {
 		TableAddAssertCurrentSchemaId();
 	}
-	add_snapshot->altered_manifests = std::move(altered_manifests);
+	pending_deletes.emplace(*add_snapshot, std::move(altered_manifests));
 
 	alters.push_back(*add_snapshot);
 	updates.push_back(std::move(add_snapshot));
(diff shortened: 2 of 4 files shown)


Tishj:
I'm not sure that I like this, or see the benefit of this, now that I see the changes.
To me it looks like this is just a rework, but it doesn't lose or gain anything.

I like the old shape better, it (the data_file delete) is part of the change, it makes sense that it lives with the change

Tishj:
I understand now. The reason I was confused by the `IcebergManifestDeletes` outside of the `IcebergAddSnapshot` and your mention of "a read inside uncommitted transaction", is that this was already something we had to deal with.

The original reason for the existence of `IcebergManifestDeletes` is so that V3 deletes can properly invalidate existing deletes. That also has to be applied by a read.

~~I can't find this code anymore though, the only reference to `IsInvalidated` is from `RewriteManifestFile`. So I'm a little lost on how we manage to honor this actually 🤔~~

Found it, it's in `ClientSideScanPlanProvider::GetDeleteFiles`

Tishj:
I did some agent assisted planning for a way to replace the `transactional_delete_files` with `IcebergManifestDeletes`, by making it central to `IcebergTransactionData` and have every "delete" be tagged with a version (the alter index).
Through that, we get our fast "is deleted" check, and at COMMIT we can determine if the item is deleted by the current ALTER.

Here's a straight copy of the plan, for inspiration and possibly more context:

# Replace Transactional Delete Tracking with Versioned Manifest Invalidations

## Summary

Use one transaction-owned `IcebergManifestDeletes` ledger for both transaction reads and snapshot commits. Store each invalidated file path with the alter index that first invalidated it, then remove `transactional_delete_files`.

## Key Changes

- Change `IcebergManifestDeletes` from an `unordered_set<string>` to an `unordered_map<string, idx_t>`:
  - `InvalidateFile(path)` continues supporting statement-local collection.
  - Add a merge operation that imports statement-local paths under an alter version and preserves the earliest invalidating version.
  - Keep membership-based `IsInvalidated(path)` for transaction reads.
  - Add `IsInvalidatedAt(path, alter_version)` for committing one snapshot’s exact invalidations.

- Make `IcebergTransactionData` own the single versioned ledger:
  - Assign each new `IcebergAddSnapshot` the stable index it will occupy in `alters`.
  - Centralize snapshot registration so append, delete, update, and replace snapshots always merge their invalidations consistently.
  - Remove `case_insensitive_map_t<string> transactional_delete_files`.

- Give each `IcebergAddSnapshot` a non-owning reference to the transaction ledger and its alter version only when that version introduced invalidations:
  - `ConstructManifestList` uses version-specific lookup, preserving intermediate snapshot history and metrics.
  - Snapshots without new invalidations retain the existing fast path and do not rewrite manifests.
  - Commit retries reuse the same stable alter versions.

- Simplify client-side delete-file selection:
  - Skip committed `DELETED` entries as before.
  - For both committed and transaction-local entries, skip any delete file whose `data_file.file_path` is present in the transaction ledger.
  - Remove both DELETE and UPDATE code that populates `transactional_delete_files`.
  - Leave v2 positional deletes additive and equality deletes unchanged.

## Test Plan

- Run the existing committed and transaction-local deletion-vector consolidation tests.
- Run update consolidation and dynamic-filter deletion-vector regression tests.
- Add a time-travel regression with two pre-existing deletion vectors replaced by separate statements in one transaction:
  - The first committed transaction snapshot must retain the vector replaced only by the second statement.
  - The final snapshot must expose only both replacements.
  - This verifies version-specific commit filtering rather than applying the final transaction union to every snapshot.
- Run `make format-check`, rebuild the relevant debug target, and execute focused fixture-backed tests serially.

## Assumptions

- Alter index is the invalidation version because it is stable across commit retries and independent of catalog sequence-number changes.
- Re-invalidating the same path keeps its earliest version; Iceberg file paths identify immutable files and are not expected to be reused.
- The ledger is transaction-internal and introduces no public API or metadata-format change.

paultmathew:
This makes sense — thanks for the reframe.

I'd framed it as "attribute the `IcebergManifestDeletes` to the statement that created
it," which pushed me toward making the statement the *key* and partitioning the deletes
per snapshot. That's what made the read-side check O(snapshots) and left #1301 as a pure
relocation with nothing to show for it.

I'll put it up, with the time-travel regression for two vectors replaced by separate
statements in one transaction. Closing #1301 in favour of it.

PR #1304, opened 2026-08-07
[Performance] Enable intra-file parallelism within Avro files

Tishj:
This PR sets duckdb-avro to a version that contains https://github.com/duckdb/duckdb-avro/pull/117
Serving as a final test for that PR before it gets merged.


Tishj:
The fact that this has **0** impact on the regression tests is a bit surprising.
If all avro files are written with only a single block, then maybe that explains it?

But regardless, that warrants investigation

PR #1310, opened 2026-08-10
Only retry deletes if isolation-level is `snapshot`

Tmonster:
Currently the delete retry functionality in DuckDB-Iceberg only provides snapshot level isolation semantics, but not serializable isolation level semantics. The default write.delete.isolation-level is serializable in the [iceberg spec](https://iceberg.apache.org/docs/latest/configuration/?h=write.delete.isolation+level#write-properties). [Serializable isolation](https://en.wikipedia.org/wiki/Isolation_(database_systems)#Serializable) for deletes effectively means a delete applies to all snapshots from before the delete 

This PR changes delete reapply logic so that deletes are only re-applied if the table has the table property`write.delete.isolation-level=snapshot`.

We could improve this logic by checking if snapshot information between the snapshot the delete applied to vs the snapshot the table is currently at. 

I.e if the delete only applied to an unaffected partition, we could theoretically still reapply the delete


Tmonster:
Interestingly enough there is a `write.update.isolation-level`, and `write.merge.isolation-level`, that I should probably address in this PR as well

Tishj:
Related issue: #1162

Tishj:
Thanks, I don't see us supporting serializable any time soon

PR #1313, opened 2026-08-11
Deprecate `ENDPOINT`, replaced with `URI` instead

Tishj:
This is both more in line with other Iceberg engines, and also more correct terminology.

The path is a base URI that Iceberg REST Catalog endpoints are built from.


Tmonster (approved):
thanks!

PR #1316, opened 2026-08-11
Extract delete-planning helpers (no behaviour change)

paultmathew:
Pure-refactor prep for #1287, split out at @Tishj's request so that PR's diff is more reviewable.

Three behaviour-preserving extractions, one per commit:

1. **`ResolveApplicableDeleteFiles`** — lift the delete-file resolution loop (the
   manifest-index bounds checks, `DeleteEntryMatchesFilters`,
   `DeleteEntryAppliesToDataFile`) out of `ProcessDeletes` into a method returning
   the applicable delete-file references. `ProcessDeletes` iterates that list to
   build its loads.
2. **`ApplyNullCounts`** — move the per-column value/null-count derivation in
   `FileMatchesFilter` into a file-local helper.
3. **`PlanDeleteInternal`** — split the delete-planning body out of `PlanDelete`
   (which now forwards to it), and route MERGE's delete action through it too, so
   the standalone DELETE and MERGE entry points share one implementation.

No behaviour change. These are the code moves #1287 currently carries; once this
merges, #1287 rebases on top and shrinks to just the new metadata-only-DELETE
logic.

Verified locally with a release build.

### DIFF
--- a/src/execution/operator/merge_into/iceberg_merge_into.cpp
+++ b/src/execution/operator/merge_into/iceberg_merge_into.cpp
@@ -213,7 +213,9 @@ static unique_ptr<MergeIntoOperator> IcebergPlanMergeIntoAction(IcebergCatalog &
 			delete_op.expressions.push_back(std::move(ref));
 		}
 		delete_op.bound_constraints = std::move(bound_constraints);
-		result->op = catalog.PlanDelete(context, planner, delete_op, child_plan);
+		//! MERGE shares this scan with its other actions, so it plans its delete action through the shared
+		//! internal path rather than the standalone PlanDelete entry point.
+		result->op = catalog.PlanDeleteInternal(context, planner, delete_op, child_plan);
 		break;
 	}
 	case MergeActionType::MERGE_INSERT: {
--- a/src/execution/operator/merge_into/iceberg_merge_into.cpp
+++ b/src/execution/operator/merge_into/iceberg_merge_into.cpp
@@ -214,8 +214,8 @@ static unique_ptr<MergeIntoOperator> IcebergPlanMergeIntoAction(IcebergCatalog &
 		}
 		delete_op.bound_constraints = std::move(bound_constraints);
 		//! MERGE shares this scan with its other actions, so it plans its delete action through the shared
-		//! internal path rather than the standalone PlanDelete entry point.
-		result->op = catalog.PlanDeleteInternal(context, planner, delete_op, child_plan);
+		//! PlanDeleteOperation path rather than the standalone PlanDelete entry point.
+		result->op = catalog.PlanDeleteOperation(context, planner, delete_op, child_plan);
 		break;
 	}
 	case MergeActionType::MERGE_INSERT: {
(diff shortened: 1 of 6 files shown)


src/planning/iceberg_multi_file_list.cpp
```diff
@@ -637,39 +638,57 @@ IcebergDeletePlan IcebergMultiFileList::ProcessDeletes(const BoundIcebergManifes
...
 	vector<IcebergDeleteScanEntry> scan_entries;
 	vector<shared_ptr<IcebergDeleteFileLoadState>> required_loads;
 	vector<shared_ptr<IcebergDeleteFileLoadState>> new_loads;
 	unique_ptr<IcebergDeletePlanningContext> delete_context;
 	{
 		annotated_lock_guard<annotated_mutex> guard(shared_state->lock);
 		annotated_lock_guard<annotated_mutex> delete_guard(shared_state->delete_lock);
-		auto delete_files = provider->GetDeleteFiles(manifest_indexes);
+		provider = scan_plan_provider.get();
+		if (!provider) {
```
Tishj, line 685: We call `InitializeView` in `ResolveApplicableDeleteFiles`, so `provider` should never realistically be nullptr. That means this shouldn't just return if that is the case, it's an error.

Hmm, I guess it's because of the atomic?
Then keep it `optional_ptr<IcebergScanPlanProvider> provider;` and populate it within the loop

`if (!provider) { ... grab provider / throw if it's not present }`

The loop's body won't get hit for the case where we returned nothing

src/execution/operator/merge_into/iceberg_merge_into.cpp
```diff
@@ -213,7 +213,9 @@ static unique_ptr<MergeIntoOperator> IcebergPlanMergeIntoAction(IcebergCatalog &
 			delete_op.expressions.push_back(std::move(ref));
 		}
 		delete_op.bound_constraints = std::move(bound_constraints);
-		result->op = catalog.PlanDelete(context, planner, delete_op, child_plan);
+		//! MERGE shares this scan with its other actions, so it plans its delete action through the shared
+		//! internal path rather than the standalone PlanDelete entry point.
+		result->op = catalog.PlanDeleteInternal(context, planner, delete_op, child_plan);
```
Tishj: I'm not a huge fan of `PlanDeleteInternal`, it's not really clear what portion of the delete logic is handled by this, and what isn't.

I don't have a great suggestion for improvement though, all I could think of was `PlanDeleteOperation`, to try to make it clearer that it includes all of the delete logic.

Ideally we rename `PlanDelete` to something more descriptive like `PlanDeleteStatement`, to make it clearer that we don't want to call that same path as we do for a MERGE operator, but we can't do that since it's a virtual method we need to override :/

src/planning/iceberg_multi_file_list.cpp
paultmathew: Agreed the null case should be an error. Updated it to throw.

I kept the fetch up front rather than populating lazily inside the loop, because `provider` is dereferenced again after the loop in the positional-delete block. If `ResolveApplicableDeleteFiles` returns empty `provider` would still be null at that positional-delete deref.

src/execution/operator/merge_into/iceberg_merge_into.cpp
paultmathew: Agreed it's not a great name and I'd also rather rename `PlanDelete` itself. Since that's not an option, I think `PlanDeleteOperation` works.

PR #1318, opened 2026-08-12
Support Iceberg REST catalog remote request signing

### LINKED ISSUE #670: Support remote signing instead of vended credentials for storage access
[Pyiceberg](https://github.com/apache/iceberg-python) supports remote signing instead of vended credentials for access to S3 storage. This is helpful when the storage solution does not support STS/vended credentials. [Lakekeeper](https://docs.lakekeeper.io/docs/0.5.x/storage/#s3) supports remote signing, so this will make querying iceberg with duckdb nice and straightforward.

Would this be possible to implement in duckdb and the iceberg extension?

N-Clerkx:
Adds support for the `remote-signing` storage access delegation mode, so Iceberg tables can be read from object stores that have no STS endpoint and therefore can't vend credentials. Closes #670.

## Why a separate filesystem

With remote signing the client never holds storage credentials: it POSTs the request details to the catalog's signer endpoint and gets back the `Authorization`, `x-amz-date` and `x-amz-content-sha256` headers to replay against the object store. `httpfs`'s `S3FileSystem` computes SigV4 locally and has no hook to substitute a signer, so this adds an iceberg-owned filesystem that claims only the storage locations the catalog told us to sign for. Everything else keeps going through `httpfs`.

I guess creating a parallel S3 filesystem long-term is undesireable. If you'd take a pluggable request-signer hook in `S3FileSystem` ( the way the AWS SDKs and PyIceberg implement this ) I'm happy to redo it that way, and it would give us connection pooling and multipart uploads for free. This PR is the version that needs no changes outside this repo.

## Honoring it only when asked

`ACCESS_DELEGATION_MODE 'remote_signing'` sends `X-Iceberg-Access-Delegation: remote-signing`, and signing is used only when the attach asked for it.

That gating matters. Catalogs report `s3.remote-signing-enabled` whenever their storage profile *supports* signing, not when the client chose it — Nessie advertises it unconditionally. Treating the flag alone as consent hijacks storage access from an attach that manages its own credentials, and the existing Nessie tests catch it.

## Details worth a look

- The signer URL comes from `signer.uri`/`signer.endpoint`, falling back to the deprecated `s3.signer.*` aliases and then to the catalog URI plus the spec default `v1/aws/s3/sign`. An absolute `signer.endpoint` replaces the URI rather than extending it, matching `RESTUtil.resolveEndpoint`.
- Region is required, since it's what the request is signed for.
- The sign request goes through the catalog's existing auth handler, so OAuth2 / SigV4 / none all work unchanged.
- Object keys are percent-encoded before signing and left untouched afterwards, because the signer computes the canonical request over the URL as it is sent.
- Signatures are cached per `(signer, method, url)` for 30s, matching `S3V4RestSignerClient`. `Cache-Control` is dropped from the applied headers.
- `X-Iceberg-Access-Delegation` was written inline at six call sites; those now go through `ICUtils::AddAccessDelegationHeader`.

## Reads only

Writes fail with an explicit error rather than a signature rejection. The cleanup entry points (`RemoveFile`, `CreateDirectory`) are no-ops instead of throwing, since they run while a write error is already in flight and nothing can be written through this filesystem anyway.

Reads use signed ranged GETs. `force_full_download`, which Iceberg sets on manifests and manifest lists, is honored with a single whole-object GET, and the `HEAD` for the file size is skipped when Iceberg already reported `file_size`. The external file cache and the parquet reader's prefetch coalescing both sit above the filesystem, so they keep working unchanged.

## Testing

`make lakekeeper` now also creates a `demo_remote_signing` warehouse with `sts-enabled: false` and `remote-signing-enabled: true` and seeds a table into it, via the Spark client already used for the other Lakekeeper fixtures. With STS off the catalog can only answer with signing information, so the test can't quietly pass on vended credentials instead.

`test_remote_signing.test` attaches that warehouse, reads the table, asserts that no S3 secret exists in the process, reads the table metadata, and asserts the write path errors. It also pins two behaviours that otherwise hold only implicitly, where nothing would fail if they stopped working: that data file reads participate in the external file cache, and that a signature is reused across all reads of an object rather than fetched per read.

The Nessie setup test was updated for the new access-mode error message.

## Known gaps

- Writes need PUT and multipart-upload support in the new filesystem.
- No HTTP connection reuse. The filesystem default-constructs a client per request and never returns it through `HTTPUtil::CloseClient`, so every request pays a fresh handshake.
- `FileSystem::TryGetNetworkThroughput` isn't implemented, so the parquet prefetch cost model stays unmeasured and the coalescing gap stays at its 16 KiB default. Largely masked by the external file cache's 2 MiB blocks, so it matters most when that cache is off.

## AI Disclaimer

I wrote this code with extensive help of AI next to my own effort. I'm happy to spend more time on this if you think the quality is not good enough. Thanks for taking the time to check it

### DIFF
--- a/make/catalogs/lakekeeper.mk
+++ b/make/catalogs/lakekeeper.mk
@@ -25,6 +25,17 @@ lakekeeper-configure-auth:
 	docker compose exec -T keycloak /opt/keycloak/bin/kcadm.sh update realms/iceberg \
 		-s sslRequired=NONE)
 
+lakekeeper-remote-signing:
+	@echo "Creating the Lakekeeper warehouse used by the remote signing tests..."
+	cp scripts/lakekeeper/remote_signing_setup.py .catalogs/lakekeeper/examples/access-control-simple/notebooks/
+	(cd .catalogs/lakekeeper/examples/access-control-simple && \
+	docker compose run --rm --no-deps \
+		-e SPARK_LOCAL_DIRS='$(LAKEKEEPER_SPARK_LOCAL_DIR)' \
+		jupyter bash -lc '\
+			rm -rf "$$SPARK_LOCAL_DIRS" /tmp/blockmgr-* /tmp/spark-* && \
+			mkdir -p "$$SPARK_LOCAL_DIRS" && \
+			python /home/jovyan/examples/remote_signing_setup.py')
+
 lakekeeper: lakekeeper-clone lakekeeper-stop
 	$(call stop_active_catalog)
 	@echo "Starting Lakekeeper catalog..."
@@ -41,6 +52,7 @@ lakekeeper: lakekeeper-clone lakekeeper-stop
 			jupyter nbconvert --to notebook --execute --output-dir=/tmp /home/jovyan/examples/01-Bootstrap.ipynb && \
 			jupyter nbconvert --to notebook --execute --output-dir=/tmp /home/jovyan/examples/02-Create-Warehouse.ipynb && \
 			jupyter nbconvert --to notebook --execute --output-dir=/tmp /home/jovyan/examples/03-01-Spark.ipynb'
+	$(MAKE) lakekeeper-remote-signing
 	$(call set_active_catalog,lakekeeper)
 
 lakekeeper-data: lakekeeper
--- a/scripts/lakekeeper/remote_signing_setup.py
+++ b/scripts/lakekeeper/remote_signing_setup.py
@@ -1,0 +1,110 @@
+"""Create the Lakekeeper warehouse the remote signing tests read from.
+
+Runs inside the Lakekeeper example 'jupyter' container, so all service names are the ones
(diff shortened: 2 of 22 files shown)


Tishj:
I skimmed through the PR, it looks interesting for sure, and less intrusive than I was expecting (apart from the filesystem).

On the topic of the filesystem, like you said I think it is ideally added onto `S3FileSystem`, but I wonder if there's precedence for one extension extending another extension.

It might be interesting to see a sketch of the functionality directly implemented in `S3FileSystem` (in `duckdb-httpfs`) to get a better understanding of what the scope of the change would be, and get a better idea of the needed extension-mechanism there.

N-Clerkx:
I can try to make a outline of how that would work in `duckdb-httpfs`, what would need to be implemented there and how that would change the implementation here!
I'll let you know once i have something

PR #1323, opened 2026-08-13
Fix parallel Avro manifest publication race

nicosuave:
After #1304, multiple Avro blocks from one manifest can be decoded concurrently. Decode chunks privately and serialize metadata and entry publication per manifest so workers cannot mutate the same shared vector at once.

This was causing a segmentation fault for me

### DIFF
--- a/src/planning/metadata_io/avro/iceberg_avro_multi_file_reader.cpp
+++ b/src/planning/metadata_io/avro/iceberg_avro_multi_file_reader.cpp
@@ -603,11 +603,13 @@ ReaderInitializeType IcebergAvroMultiFileReader::InitializeReader(
 			auto file_idx = manifest_scan_info.GetManifestIndex(reader_data.reader->file_list_idx.GetIndex());
 			lock_guard<mutex> manifest_guard(manifest_scan_info.GetManifestLock(file_idx));
 			auto &manifest_list_entry = manifest_scan_info.manifest_files[file_idx];
-			auto manifest_path = manifest_list_entry.file.manifest_path.empty()
-			                         ? reader_data.reader->GetFileName()
-			                         : manifest_list_entry.file.manifest_path;
-			manifest_list_entry.manifest_metadata.emplace(
-			    ParseManifestMetadata(reader_data.reader->GetMetadata(), manifest_path));
+			if (!manifest_list_entry.manifest_metadata) {
+				auto manifest_path = manifest_list_entry.file.manifest_path.empty()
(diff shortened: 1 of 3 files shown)


src/planning/metadata_io/avro/iceberg_avro_multi_file_reader.cpp
```diff
@@ -660,10 +668,26 @@ void IcebergAvroMultiFileReader::FinalizeChunk(ClientContext &context, const Mul
...
 		auto &manifest_entries = manifest_file.GetOrCreateManifestEntries();
 		idx_t start_index = manifest_entries.size();
-		manifest_file::ManifestReader::ReadChunk(output_chunk, manifest_scan_info.partition_field_id_to_type, input,
-		                                         manifest_entries);
+		if (manifest_scan_info.read_state) {
+			//! Streaming consumers retain references into this vector after publication. LoadManifestList
+			//! reserves the exact manifest-list file count before it enables this path, so exceeding the
+			//! reserved capacity means the advertised count was wrong; appending would invalidate those
+			//! references.
+			if (decoded_entries.size() > manifest_entries.capacity() - manifest_entries.size()) {
```
Tishj, line 681: Thanks for the run-time check, we should have already had this tbh 👍

src/planning/metadata_io/avro/iceberg_avro_multi_file_reader.cpp
```diff
@@ -648,7 +649,14 @@ void IcebergAvroMultiFileReader::FinalizeChunk(ClientContext &context, const Mul
 		auto manifest_file_idx = manifest_scan_info.GetManifestIndex(reader.file_list_idx.GetIndex());
 		auto &manifest_file = manifest_scan_info.manifest_files[manifest_file_idx];
 
-		auto &manifest_metadata = *manifest_file.manifest_metadata;
+		auto manifest_metadata = [&]() {
```
Tishj: This feels excessive, can we really not assume the metadata is created at this point?
I thought since we're scanning committed entries here that we would already have parses the file's metadata before we even start to scan?

This seems to imply that InitializeReader is called once per local state 🤔 If that's the case perhaps we should move this initialization into the file state somehow, because now it would be getting done multiple times which is wasteful

nicosuave: I changed it to initialize the shared metadata just once under the per manifest lock

Tishj:
Thanks!

Issue #626, opened 2025-12-11
Query performance slow when a large number of equality delete records exist


CodingJun:
### Describe

When the table contains a large number of equality delete records, query performance degrades significantly. Even simple queries that should complete quickly take an unexpectedly long time to finish.

Meanwhile, when reading the same table using Spark, the query completes much faster.

### Environment

**Duckdb version:** 1.4.1 (also occurs in 1.4.3)
**Iceberg version:** 1.5.2
**Spark version:** 3.5.4
**OS:** MacOS26
**CPU:** M1 pro (aarch64)

### Quick Reproduction Steps

1. Download the attachment **[equality_delete_performance.zip](https://github.com/user-attachments/files/24101169/equality_delete_performance.zip)**
2. Unzip and place files under: /tmp/data
3. Execute the SQL in DuckDB: `select count(1) from iceberg_scan('/tmp/data/equality_delete_performance/metadata/v9.metadata.json')`
4. The query took over 2 minutes with high cpu and memory usage.

### Query Explain

<img width="1700" height="1328" alt="Image" src="https://github.com/user-attachments/assets/df74a8c0-5f66-417d-8b59-f00094c19774" />

### Other Engine (Spark)

<img width="416" height="312" alt="Image" src="https://github.com/user-attachments/assets/a7515fb1-15f1-468f-b644-eff83d34bbac" />

Tmonster:
Thank you for the issue! We will try to get a fix for this in ASAP

james-aus:
There's not much to my original suggestion, just a "dumb" way to apply the existing filters more efficiently in a way that may have been low-effort and low-impact -- it would not have addressed #624 and would not have taken advantage of the infrastructure that currently supports joins in the event that the number of deletes grew large.

Spending a bit more time today, my dataset may be suffering from an unrelated issue with deletes (I don't *think* it's #624 as I'm not deleting a high enough percentage of my data or in consistent enough runs to expect whole batches to become empty, but I have no proof either way). A manual anti-join over the appropriate parquet files seems to be performant enough for my use-case and feels like it would be the most straight-forward in terms of validation if we moved forward with it as a solution.

Based on that, I'm more inclined towards your suggestion of an anti-join, as it may lead to a simpler solution for supporting avro-formatted equality delete files in future too.

I'll need to finish checking in with my employer about spending time in this area and getting the required internal approvals for IP release.

Tishj:
[...]

PR #1324, opened 2026-08-13
Add equality-delete hash filter

nicosuave:
Avoid constructing a bound-expression tree for every equality-delete row when primitive key layouts can be encoded and probed directly. Keep unsupported layouts and physically absent fields on the existing expression fallback.

For every equality delete file in `accelerated_files`, push every row into a hash table (per set of delete columns) in `InitializeReader`, check for every produced chunk in `FinalizeChunk` whether that row is already contained in the hash table, if it is then we know that row should be deleted.

This implements the hash membership approach proposed in #626 using a sharded flat hash filter instead of `std::unordered_map`.

On a ~15 billion row delete-heavy table, I could not get queries to complete at all (after 10+ minutes) without this fix.

After, peak RSS dropped by ~5x, network transfer remained the same, and the query completed in 28 seconds.


src/iceberg_extension.cpp
```diff
@@ -118,6 +118,11 @@ static void LoadInternal(ExtensionLoader &loader) {
 	    "Maximum number of characters of a REST catalog POST body to include in Iceberg log messages. "
 	    "Bodies longer than this are truncated with a trailing '... (truncated)' marker. Set to 0 to omit the body.",
 	    LogicalType::UBIGINT, Value::UBIGINT(10000));
+	config.AddExtensionOption(
+	    "iceberg_equality_delete_fast_filter",
+	    "Apply byte-comparable equality deletes through a shared columnar flat hash filter instead of a bound "
+	    "expression per delete row. Unsupported types retain the upstream expression path. Off by default for A/B.",
+	    LogicalType::BOOLEAN, Value::BOOLEAN(false), nullptr, SetScope::GLOBAL);
```
Tishj: On by default please. I imagine we're having much less coverage tested this way.
Optimizations shouldn't be feature flagged, but if they are they should be on, so we can be confident they work in all cases.

If not, the tests added by this PR would need to be much more extensive to ensure coverage

nicosuave: Thanks; updated

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
```diff
@@ -0,0 +1,341 @@
...
+namespace duckdb {
+
+void IcebergEqualityDeleteFastFilter::FlatShard::Initialize(idx_t expected_count) {
+	if (expected_count == 0) {
+		return;
+	}
+	if (expected_count > NumericLimits<idx_t>::Maximum() / 2) {
+		throw OutOfMemoryException("Too many Iceberg equality-delete keys");
+	}
+	auto capacity = NumericCast<idx_t>(NextPowerOfTwo(MaxValue<idx_t>(expected_count * 2, 16)));
```
NiclasHaderer: I think it would be possible to still overflow because of `NextPowerOfTwo`

NiclasHaderer (commented):
I think I am worried about all the untracked memory the PR is introducing. Maybe we can use the ddb Allocator for the more memory intense datastructures

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
nicosuave: Thanks @NiclasHaderer this is all good feedback

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
```diff
@@ -0,0 +1,387 @@
...
+		for (idx_t row = 0; row < count; row++) {
+			if (layout.types.size() == 1) {
+				auto index = formats[0].sel->get_index(row);
+				if (!formats[0].validity.RowIsValid(index)) {
+					layout.has_null_key = true;
+					continue;
+				}
+				uint32_t length;
+				auto bytes = RowBytes(formats[0], layout.types[0], index, length);
+				auto target = filter.arena.Allocate(MaxValue<idx_t>(length, 1));
```
Tishj: We can use StringHeap for this, this is kind of reinventing that

Tishj:
I have a sneaking suspicion this is implementing a hash table, which we already have in core, and can probably be used for this purpose.

The pattern reminds me of this recent PR: https://github.com/duckdb/duckdb/pull/24271

I might be way of base, as I only glanced at the code, if that's the case I apologize for jumping to conclusions

nicosuave:
Thank you for that context @Tishj I adjusted

Tishj:
I see 3 places where we construct DataChunks in the hot path, it's worth checking if we can cache some of those in some sort of local state, or at least create them out of the loop and use Reset() between iterations.

The Initialize call of a DataChunk does a decent bit of allocation, because it's built for reuse. So a VectorCache is created for each column of the chunk, with the capacity given (default STANDARD_VECTOR_SIZE (2048))

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
```diff
@@ -0,0 +1,198 @@
...
+		}
+		groups.data[column_idx].Reference(cast_columns.back());
+	}
+	groups.SetCardinalityUnsafe(count);
+
+	//! GroupedAggregateHashTable's probe state is vector-sized. Delete-file
+	//! state can contain many vectors, so add it in standard-sized slices.
+	for (idx_t offset = 0; offset < count; offset += STANDARD_VECTOR_SIZE) {
+		auto end = MinValue<idx_t>(offset + STANDARD_VECTOR_SIZE, count);
+		DataChunk slice;
```
Tishj: nit: maybe outside of the loop, using Reset() for each iteration in the loop
But since it's `InitializeEmpty`, it's probably fine

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
```diff
@@ -0,0 +1,198 @@
...
+		slice.InitializeEmpty(layout.types);
+		slice.Slice(groups, offset, end);
+		Vector addresses(LogicalType::POINTER);
```
Tishj: This could actually be moved outside of the loop, so we allocate it once and reuse it

Tishj:
Can you add logging so we can verify in tests that files are accelerated?

Tishj:
An optimization for SelectionVector creation, and also preventing doing a hash table lookup for already-deleted rows from checks with earlier layout(s):

(and also avoiding the use of https://en.cppreference.com/cpp/container/vector_bool which doesn't have great performance characteristics)

```patch
diff --git a/src/include/planning/deletes/iceberg_equality_delete_fast_filter.hpp b/src/include/planning/deletes/iceberg_equality_delete_fast_filter.hpp
index 925f013d6..5971a4f8f 100644
--- a/src/include/planning/deletes/iceberg_equality_delete_fast_filter.hpp
+++ b/src/include/planning/deletes/iceberg_equality_delete_fast_filter.hpp
@@ -35,8 +35,8 @@ public:
 	                         const unordered_map<int32_t, idx_t> &field_indexes,
 	                         const vector<LogicalType> &target_types, ClientContext &context, Allocator &allocator);
 
-	//! Marks matching rows in `deleted`; it never clears an existing mark.
-	void MarkDeleted(DataChunk &keys, vector<bool> &deleted) const;
+	//! Removes matching rows from `sel` and returns the number of remaining rows.
+	idx_t Filter(DataChunk &keys, SelectionVector &sel, idx_t count) const;
 
 private:
 	vector<Layout> layouts;
diff --git a/src/planning/deletes/iceberg_equality_delete_fast_filter.cpp b/src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
index 9705768f5..37063b3ff 100644
--- a/src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
+++ b/src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
@@ -173,10 +173,11 @@ IcebergEqualityDeleteFastFilter::Build(const vector<reference<const IcebergEqual
 	return result;
 }
 
-void IcebergEqualityDeleteFastFilter::MarkDeleted(DataChunk &keys, vector<bool> &deleted) const {
-	auto count = keys.size();
-	D_ASSERT(deleted.size() == count);
+idx_t IcebergEqualityDeleteFastFilter::Filter(DataChunk &keys, SelectionVector &sel, idx_t count) const {
 	for (auto &layout : layouts) {
+		if (count == 0) {
+			break;
+		}
 		if (layout.hash_table->Count() == 0) {
 			continue;
 		}
@@ -185,15 +186,24 @@ void IcebergEqualityDeleteFastFilter::MarkDeleted(DataChunk &keys, vector<bool>
 		for (idx_t column_idx = 0; column_idx < layout.column_indices.size(); column_idx++) {
 			groups.data[column_idx].Reference(keys.data[layout.column_indices[column_idx]]);
 		}
-		groups.SetCardinalityUnsafe(count);
+		groups.SetCardinalityUnsafe(keys.size());
+		groups.Slice(sel, count);
 
 		AggregateHTLookupState lookup_state;
 		SelectionVector found_groups(count);
 		auto found_count = layout.hash_table->LookupGroups(groups, lookup_state, found_groups);
-		for (idx_t found_idx = 0; found_idx < found_count; found_idx++) {
-			deleted[found_groups.get_index(found_idx)] = true;
+		if (found_count == 0) {
+			continue;
 		}
+
+		//! LookupGroups returns matching positions in sorted order. Invert those
+		//! positions, then compose the survivors with the existing selection.
+		SelectionVector survivors(count);
+		auto survivor_count = SelectionVector::Inverted(found_groups, survivors, found_count, count);
+		sel.SliceInPlace(survivors, survivor_count);
+		count = survivor_count;
 	}
+	return count;
 }
 
 } // namespace duckdb
diff --git a/src/planning/iceberg_multi_file_reader.cpp b/src/planning/iceberg_multi_file_reader.cpp
index 79144e857..6287ab313 100644
--- a/src/planning/iceberg_multi_file_reader.cpp
+++ b/src/planning/iceberg_multi_file_reader.cpp
@@ -24,7 +24,6 @@
 #include "core/metadata/iceberg_table_metadata.hpp"
 #include "duckdb/common/multi_file/multi_file_states.hpp"
 
-#include <algorithm>
 #include <cstdint>
 #include <cstring>
 
@@ -621,26 +620,14 @@ void IcebergMultiFileReader::FinalizeChunk(ClientContext &context, const MultiFi
 		equality_delete_executor.Execute(input_chunk, equality_delete_chunk);
 
 		auto input_count = equality_delete_chunk.size();
-		vector<bool> keep(input_count, true);
+		auto sel_vec = SelectionVector::Incremental(input_count);
+		idx_t result_count = input_count;
 		if (equality_delete_expression) {
 			ExpressionExecutor filter_executor(context, *equality_delete_expression);
-			SelectionVector expression_sel(input_count);
-			auto expression_count = filter_executor.SelectExpression(equality_delete_chunk, expression_sel);
-			std::fill(keep.begin(), keep.end(), false);
-			for (idx_t i = 0; i < expression_count; i++) {
-				keep[expression_sel.get_index(i)] = true;
-			}
+			result_count = filter_executor.SelectExpression(equality_delete_chunk, sel_vec);
 		}
-		vector<bool> deleted(input_count, false);
 		if (equality_delete_state.fast_filter) {
-			equality_delete_state.fast_filter->MarkDeleted(equality_delete_chunk, deleted);
-		}
-		SelectionVector sel_vec(input_count);
-		idx_t result_count = 0;
-		for (idx_t i = 0; i < input_count; i++) {
-			if (keep[i] && !deleted[i]) {
-				sel_vec.set_index(result_count++, i);
-			}
+			result_count = equality_delete_state.fast_filter->Filter(equality_delete_chunk, sel_vec, result_count);
 		}
 		output_chunk.Slice(sel_vec, result_count);
 	}
```

src/planning/iceberg_multi_file_reader.cpp
```diff
@@ -24,8 +24,62 @@
...
+				eligible = false;
+				break;
+			}
+		}
+		if (eligible) {
+			eligible_files.push_back(delete_file);
+			eligible_file_indexes.push_back(file_idx);
+		}
+	}
+	lock_guard<mutex> guard(equality_delete_fast_filter_lock);
```
Tishj: We are holding this lock while building the fast-filter, perhaps it makes sense to release it while building, so other builds can run concurrently

src/planning/iceberg_multi_file_reader.cpp
```diff
@@ -24,8 +24,62 @@
...
+#include <cstdint>
+#include <cstring>
+
 namespace duckdb {
 
+IcebergEqualityDeleteFastFilter::BuildResult IcebergMultiFileReaderGlobalState::GetOrCreateEqualityDeleteFastFilter(
+    const vector<reference<const IcebergEqualityDeleteFile>> &delete_files,
+    const IcebergEqualityDeleteReadState &read_state, const set<int32_t> &local_field_ids, ClientContext &context,
+    Allocator &allocator) {
+	string key;
```
Tishj: Can we replace the key creation+lookup with:
```c++
struct ProbeColumn {
    int32_t field_id;
    LogicalType type;
    bool locally_present;
};

struct EqualityDeleteFastFilterKey {
    vector<const IcebergEqualityDeleteFile *> delete_files;
    vector<ProbeColumn> columns;
};
```

With custom `std::hash` implementation + equality operator overload?
See for example `duckdb/src/include/duckdb/common/types/timestamp_base.hpp`


Also, I wonder if we should even cache at this level, shouldn't we be caching at the `Layout` level?

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
```diff
@@ -0,0 +1,258 @@
...
+		auto &file = delete_files[file_idx].get();
+		if (file.equality_values.size() == 0) {
+			result.accelerated_files[file_idx] = true;
+			continue;
+		}
+		if (file.equality_values.ColumnCount() != file.equality_ids.size()) {
+			throw InvalidConfigurationException("Equality delete file contains an unexpected number of columns");
+		}
+
+		struct Field {
```
Tmonster: Why declare this here?

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
```diff
@@ -0,0 +1,258 @@
...
+				for (idx_t file_idx = 0; file_idx < group.files.size(); file_idx++) {
+					auto &descriptor = group.files[file_idx];
+					auto &file = delete_files[descriptor.file_index].get();
+					built.accelerated_files[file_idx] =
+					    AddDeleteFile(*layout->hash_table, layout->types, file, descriptor.source_indices);
+				}
+			} catch (std::exception &ex) {
+				error = ErrorData(ex);
+			} catch (...) { // LCOV_EXCL_START
+				error = ErrorData("Unknown exception while building an Iceberg equality-delete hash filter");
```
Tmonster: Is it possible to get the filename here as well? This can help with debugging later on

src/planning/iceberg_multi_file_reader.cpp
```diff
@@ -417,8 +427,30 @@ ReaderInitializeType IcebergMultiFileReader::InitializeReader(MultiFileReaderDat
 	ApplyPartitionConstants(manifest_file, bound_manifest_entry, metadata, reader_data, scan_columns, scan_column_ids,
 	                        context);
 
-	equality_delete_state->expression =
-	    CreateEqualityDeleteExpression(delete_plan.equality_deletes, local_columns, *equality_delete_state);
+	vector<bool> accelerated_files;
+	Value fast_filter_setting;
+	if (context.TryGetCurrentSetting("iceberg_equality_delete_fast_filter", fast_filter_setting) &&
```
Tmonster: Is it worth it to have this on by default?

src/include/planning/deletes/iceberg_equality_delete_fast_filter.hpp
```diff
@@ -0,0 +1,125 @@
...
+#include "duckdb/execution/aggregate_hashtable.hpp"
+
+#include <condition_variable>
+
+namespace duckdb {
+
+struct ProbeColumn {
+	int32_t field_id;
+	LogicalType type;
+	bool locally_present;
```
Tmonster: Why is the significance of `locally_present` a quick search and it seems to only be read, but never written to.

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
```diff
@@ -0,0 +1,258 @@
...
+IcebergEqualityDeleteFastFilter::BuildResult IcebergEqualityDeleteFastFilterCache::GetOrCreate(
+    const vector<reference<const IcebergEqualityDeleteFile>> &delete_files,
+    const unordered_map<int32_t, idx_t> &field_indexes, const vector<LogicalType> &target_types,
+    const set<int32_t> &local_field_ids, ClientContext &context, Allocator &allocator) {
+	IcebergEqualityDeleteFastFilter::BuildResult result;
+	result.accelerated_files.resize(delete_files.size(), false);
+
+	map<vector<idx_t>, idx_t> layout_indexes;
+	vector<LayoutGroup> layout_groups;
+	for (idx_t file_idx = 0; file_idx < delete_files.size(); file_idx++) {
```
Tmonster, line 103: This is coming from `IntializeReader` for each data file, and delete files is a list of delete files to apply to the data file.
I feel like we might be doing a lot of extra work here? The only thing that changes per equality delete is the data file position.

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
Tmonster: Any reason `Field` needs to be declared here?

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
```diff
@@ -0,0 +1,258 @@
...
+		auto &file = delete_files[file_idx].get();
+		if (file.equality_values.size() == 0) {
```
Tmonster, line 105: If the delete file is a positional_delete does it still show up here? If yes, will it break anything downstream?

Tmonster (commented):
Thanks! A couple of notes from my side

Building the Hash table in `IcebergEqualityDeleteFastFilterCache::GetOrCreate` and understanding the need for`Layout`, `ProbeLayout`, `LayoutBuildResult` etc. can get pretty confusing. Could you add some comments in `iceberg_equality_delete_fast_filter.hpp`, just one liners is fine.

We also have a way to support writing equality deletes if you build with 
```
 BUILD_BENCHMARK=1 \
  GEN=ninja EXT_FLAGS=-DICEBERG_ENABLE_EQUALITY_DELETE_WRITES=1 \
  CATALOG_TEST_CONFIG_SETUP=1 FORCE_DEBUG=1 FORCE_ASSERT=1 make relassert
```

It would be nice to see a test where there is a lot of schema evolution with added/deleted columns + type evolution. 

It would also be nice to see a regression test for this, see `.github/workflows/Regression.yml`. You can put them in a general "performance" folder. The workflow will fail to start since the benchmark is not currently on main, just a heads up

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
nicosuave: No, the deletes are split between equality and positional earlier in planning

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
nicosuave: While true, the more expensive part is reading & casting delete values and building the hash table. My thinking was to avoid add more complexity for now and potentially re-evaluate in the future.

src/include/planning/deletes/iceberg_equality_delete_fast_filter.hpp
nicosuave: You're right, this is stale, thank you for noticing.

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
nicosuave: Good call tks

src/planning/deletes/iceberg_equality_delete_fast_filter.cpp
nicosuave: Moved it to the anonymous namespace

nicosuave:
> Building the Hash table in `IcebergEqualityDeleteFastFilterCache::GetOrCreate` and understanding the need for`Layout`, `ProbeLayout`, `LayoutBuildResult` etc. can get pretty confusing. Could you add some comments in `iceberg_equality_delete_fast_filter.hpp`, just one liners is fine.

Thanks, added this

> We also have a way to support writing equality deletes if you build with
> 
> ```
>  BUILD_BENCHMARK=1 \
>   GEN=ninja EXT_FLAGS=-DICEBERG_ENABLE_EQUALITY_DELETE_WRITES=1 \
>   CATALOG_TEST_CONFIG_SETUP=1 FORCE_DEBUG=1 FORCE_ASSERT=1 make relassert
> ```
> 
> It would be nice to see a test where there is a lot of schema evolution with added/deleted columns + type evolution.

Sure, added

> It would also be nice to see a regression test for this, see `.github/workflows/Regression.yml`. You can put them in a general "performance" folder. The workflow will fail to start since the benchmark is not currently on main, just a heads up

Appreciate the notes here, added

nicosuave:
Those failures look like some noisy machines during CI

nicosuave:
> The workflow will fail to start since the benchmark is not currently on main, just a heads up

@Tmonster any recommendations on how to proceed with the failing test?

Tmonster:
> @Tmonster any recommendations on how to proceed with the failing test?

Failing Regression test is fine if you added one. I hope to eventually find a way to fix the regression workflow so that it stays green in these scenarios. I plan to review again Monday or Tuesday 👍

nicosuave:
ping @Tmonster

Tishj:
Thanks!

nicosuave:
Thank _you_ for the test addition @Tishj

PR #1328, opened 2026-08-14
Forward extra_http_headers to OAuth2 token endpoint (fixes #978)

### LINKED ISSUE #978: Duckdb Polaris Integration does not support multi-tenant with Polaris-Realm header
**Is your feature request related to a problem? Please describe.**
I deployed duckdb and polaris to query iceberg table. The Polaris manage multi-tenant with Polaris Realms. But when using EXTRA_HTTP_HEADERS MAP {'Polaris-Realm': 'REALM1'}, it call Polaris instance with default realm rather than realm in the header.

**Describe the solution you'd like**
It need to handle Polaris-Realm to enable multi-tenant in Polaris integration.

**Additional context**
duckdb.sql('''
ATTACH 'my_catalog' AS my_catalog (
    TYPE iceberg,
    ACCESS_DELEGATION_MODE 'vended_credentials',
    ENDPOINT 'xxx',
    CLIENT_ID 'xxx',
    CLIENT_SECRET 'xxx',
    OAUTH2_SERVER_URI 'xxx/v1/oauth/tokens',
    EXTRA_HTTP_HEADERS MAP {'Polaris-Realm': 'REALM1'}
);
''')

But from polaris log, it handle with default polaris is POLARIS.

tehdima:
This PR fixes #978

extra_http_headers were only attached to catalog requests, not to the OAuth2
token request, so attaching a Polaris catalog with
EXTRA_HTTP_HEADERS MAP {'Polaris-Realm': '...'} still hit the default realm.

FetchOAuth2TokenResponse now also takes the extra headers and inserts them
before the service headers (Content-Type, Authorization) so those can't be
overridden. They're forwarded from the initial token fetch in
CreateCatalogSecretFunction and from both refresh paths in
RefreshAccessTokenUnlocked.

I'v verified this against a live multi-tenant Polaris on an earlier version of the
code - the OAuth2 path was refactored since, so I rebased onto current main (same
behaviour).

### DIFF
--- a/src/catalog/rest/storage/authorization/oauth2.cpp
+++ b/src/catalog/rest/storage/authorization/oauth2.cpp
@@ -130,10 +130,10 @@ static void ExtractOAuth2CredentialsFromOptions(const case_insensitive_map_t<Val
 
 //! Helper function to fetch OAuth2 token and parse full response (RFC 6749).
 //! Relies on DuckDB's built-in HTTP retry infrastructure (RunRequestWithRetry) for transient errors.
-static rest_api_objects::OAuthTokenResponse FetchOAuth2TokenResponse(ClientContext &context, const string &grant_type,
-                                                                     const string &uri, const string &client_id,
-                                                                     const string &client_secret, const string &scope,
-                                                                     const string &refresh_token_param = "") {
+static rest_api_objects::OAuthTokenResponse
+FetchOAuth2TokenResponse(ClientContext &context, const string &grant_type, const string &uri, const string &client_id,
+                         const string &client_secret, const string &scope, const string &refresh_token_param = "",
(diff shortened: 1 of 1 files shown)


Tmonster (commented):
Thanks!

We run tests against polaris in our CI, would you mind adding a test case for this?
Somewhere under `test/sql/local/catalog_test_config_setup/catalog_specifc/polaris/` is good. 

You can also take a look at `test/sql/local/catalog_custom_setup/fixture/enable_logging_http/test_table_information_requests.test` to see how headers are verified in tests

tehdima:
Hello!

I'm working on it, I think I'll fill in everything in the next couple of days, thanks for the tip!

Tishj:
Thanks!

PR #1331, opened 2026-08-18
Preserve endpoint suffix from vended Azure SAS config key

djouallah:
Everything after the account name in the `adls.sas-token.<host>` key is discarded, so the Azure SDK defaults to `core.windows.net` and reads fail with `AuthenticationFailed` for other endpoints — OneLake IRC vends `adls.sas-token.onelake.dfs.fabric.microsoft.com`. This preserves the suffix as `EndpointSuffix=` in the connection string.

Tested against a live OneLake Iceberg REST catalog on my fork: [run](https://github.com/djouallah/duckdb-iceberg/actions/runs/32104921695). Written with Claude Code.


Tmonster (approved):
Thanks!

Tmonster:
Can you also cherry-pick this to `main`?

PR #1336, opened 2026-08-19
Fix iceberg_bucket UUID hashing to use RFC 4122 big-endian bytes

### LINKED ISSUE #1325: iceberg_bucket on UUID hashes DuckDB's internal representation, not the RFC bytes — disagrees with Iceberg spec test vector
### What happens?

`iceberg_bucket(N, col)` with a `UUID` argument computes bucket values that violate the Iceberg spec and disagree with all other engines (Java, Spark, PyIceberg, iceberg-go). Per the spec ([Appendix B](https://iceberg.apache.org/spec/#appendix-b-32-bit-hash-requirements)), `bucket` on UUID must murmur3-hash the 16 big-endian RFC bytes. The UUID overload instead appears to hash DuckDB's internal UUID storage (the sign-flipped int128 used for ordering): hashing the same bytes with the most-significant bit flipped reproduces the overload's output exactly (see reproduction below).

Consequences: tables written by DuckDB-Iceberg with `PARTITIONED BY (bucket(N, uuid_col))` (supported since v1.5.3) carry non-conformant partition values, and bucket predicate pushdown applied to files written by conformant engines can prune files that contain matching rows, silently dropping results. No catalog is involved; the scalar function alone reproduces it.

### To Reproduce

```sql
LOAD iceberg;

-- Iceberg spec Appendix B test vector: this UUID hashes to 1488055340,
-- so bucket[16] = 1488055340 % 16 = 12.
SELECT
    iceberg_bucket(16, UUID 'f79c3e09-677c-4bbd-a479-3f349cb785e7') AS uuid_overload,
    iceberg_bucket(16, CAST(UUID 'f79c3e09-677c-4bbd-a479-3f349cb785e7' AS BLOB)) AS rfc_bytes;
```

```
uuid_overload = 2, rfc_bytes = 12   -- spec requires 12
```

Evidence for the root cause — hashing the sign-flipped bytes (MSB XOR 0x80, DuckDB's internal UUID form) reproduces the UUID overload's answers:

```sql
LOAD iceberg;
SELECT
    iceberg_bucket(16, from_hex('779c3e09677c4bbda4793f349cb785e7')) AS signflipped_vector, -- f7^80=77
    iceberg_bucket(16, UUID '00000000-0000-0000-0000-000000000000') AS uuid_zero,
    iceberg_bucket(16, from_hex('80000000000000000000000000000000')) AS signflipped_zero,
    iceberg_bucket(16, from_hex('00000000000000000000000000000000')) AS rfc_zero;
```

```
signflipped_vector = 2 (matches UUID overload), uuid_zero = 1, signflipped_zero = 1, rfc_zero = 8
```

The long and string overloads match the spec vectors on the same binary (`iceberg_bucket(16, 34::BIGINT)` = 3, `iceberg_bucket(16, 'iceberg')` = 9), so the mismatch is specific to UUID.

### OS:

macOS 14 (aarch64)

### DuckDB Version:

v1.5.5 (Variegata) d8cdaa33fd

### DuckDB Client:

CLI

### Hardware:

n/a

### Full Name:

Ryan Worl

### Affiliation:

Confluent

- [x] Yes, I have

- [x] Yes, I have

tom-kita:
Fixes #1325

## Problem

`iceberg_bucket` on UUID values hashes DuckDB's internal `hugeint_t` representation. DuckDB stores UUIDs with the MSB of the upper 64 bits flipped so they sort correctly as signed int128 (see `BaseUUID::FromString` / `ToString`). Hashing that representation disagrees with the Iceberg spec, which requires hashing the RFC 4122 big-endian bytes ([Appendix B, note 4](https://iceberg.apache.org/spec/#appendix-b-32-bit-hash-requirements)).

Spec test vector: `hash(uuid "f79c3e09-677c-4bbd-a479-3f349cb785e7") = 1488055340`. Before this change DuckDB produced `2132039410`. Integer and string hashing already match the spec vectors, so only the UUID byte serialization was wrong.

## Fix

Convert with `BaseUUID::ToBlob` before hashing — the same utility the Parquet writer (`ParquetUUIDOperator::Operation`) uses for the identical conversion.

This single change fixes all three paths that share `IcebergHash::HashUUID`: the `iceberg_bucket` scalar function, partition value computation on writes (`iceberg_insert.cpp` binds the same scalar function), and bucket predicate pushdown when pruning files on reads (`BucketTransform::ApplyTransform` → `IcebergHash::BucketValue`).

Note that this changes the bucket values produced for UUIDs (they were spec-incorrect); see #1325 for the impact on existing tables.

## Testing

Added the spec test vector for UUID to `test_iceberg_bucket.test` (NULL handling, return type, bucket values, and the raw 31-bit hash value).

### DIFF
--- a/scripts/data_generators/tests/default/bucket_partitioned_uuid/__init__.py
+++ b/scripts/data_generators/tests/default/bucket_partitioned_uuid/__init__.py
@@ -1,0 +1,36 @@
+from scripts.data_generators.tests.base import IcebergTest
+import pathlib
+
+
+@IcebergTest.register()
+class Test(IcebergTest):
+    """Bucket-partitioned UUID table.
+
+    Spark SQL has no UUID type, so the table cannot be created with `CREATE TABLE ... (val UUID)`.
+    Instead the table is created through the Iceberg Java API (reached through the Spark catalog),
+    after which Spark can insert into it using the string representation of the UUIDs.
+    """
+
+    def __init__(self):
+        path = pathlib.PurePath(__file__)
+        super().__init__(__file__)
+
+    def setup(self, con):
+        spark = con.con
+        jvm = spark.sparkContext._jvm
+        gateway = spark.sparkContext._gateway
+
+        types = jvm.org.apache.iceberg.types.Types
+        fields = gateway.new_array(types.NestedField, 2)
+        fields[0] = types.NestedField.required(1, "id", types.IntegerType.get())
+        fields[1] = types.NestedField.optional(2, "val", types.UUIDType.get())
+        schema = jvm.org.apache.iceberg.Schema(fields)
+        spec = jvm.org.apache.iceberg.PartitionSpec.builderFor(schema).bucket("val", 32).build()
+
+        spark_catalog = spark._jsparkSession.sessionState().catalogManager().catalog(con.catalog)
+        iceberg_catalog = spark_catalog.icebergCatalog()
+        identifier = jvm.org.apache.iceberg.catalog.TableIdentifier.parse(self.qualified_name)
+
+        properties = jvm.java.util.HashMap()
+        properties.put("format-version", "2")
+        iceberg_catalog.createTable(identifier, schema, spec, properties)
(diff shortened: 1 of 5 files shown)


Tishj:
Thanks!
I still need to verify and cross-reference with the spec, but it looks correct at first glance :+1:

Tmonster:
I have a test case [here](https://github.com/Tmonster/duckdb-iceberg/tree/add_test_uuid) if you want to add verification with spark.

Tmonster (approved):
Thanks!

PR #1338, opened 2026-08-19
iceberg_rewrite_data_files: add min_file_size_bytes and max_file_size_bytes

The-Alchemist:
Align config with Spark's `SizeBasedFileRewriter` defaults (75%/180% of target) and allow overrides via `min_file_size_bytes` and `max_file_size_bytes`.

## Source

Option | Descrition | Details
-- | -- | --
min-file-size-bytes | 75% of target file size | Files under this threshold will be considered for rewriting regardless of any other criteria
max-file-size-bytes | 180% of target file size | Files with sizes above this threshold will be considered for rewriting regardless of any other criteria

### DIFF
--- a/src/maintenance/rewrite_data_files_planner.cpp
+++ b/src/maintenance/rewrite_data_files_planner.cpp
@@ -19,7 +19,8 @@ namespace duckdb {
 
 namespace {
 
-constexpr int64_t DEFAULT_TARGET_FILE_SIZE_BYTES = 134217728;
+//! Iceberg spec default for write.target-file-size-bytes (same as IcebergCopyOptions::file_size_bytes).
+constexpr int64_t DEFAULT_TARGET_FILE_SIZE_BYTES = 512LL * 1024 * 1024;
 constexpr int64_t MIN_TARGET_FILE_SIZE_BYTES = 100;
 
 static int64_t ParseTargetFileSizeProperty(const string &value, const string &property) {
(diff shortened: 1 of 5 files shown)


test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rewrite_data_files_min_max_file_size.test
```diff
@@ -0,0 +1,110 @@
...
+call iceberg_rewrite_data_files(
+	'my_datalake.default.rewrite_min_max_size',
+	min_file_size_bytes => 1000000,
+	max_file_size_bytes => 1000000000,
+	min_input_files => 5
+);
+----
+5	1	<REGEX>:[1-9][0-9]*
+
+query I
```
Tmonster, line 111: can we also check the number of data files here again?

Tmonster (commented):
Thanks. Could you also update or remove the `DEFAULT_TARGET_FILE_SIZE_BYTES` setting in rewrite_data_files_planner.cpp? `IcebergCopyOptions` sets it by default to 512mb, which is in the iceberg spec.

test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rewrite_data_files_min_max_file_size.test
The-Alchemist: done

src/maintenance/rewrite_data_files_planner.cpp
Tmonster, line 215: Would it be better to filter out files based on file size here? Keeping around the full list of data files only to prune them later could cause performance degradation? 

Actually, getting the candidates could be made even better if we group them by partition here since we have the `partition` value in `entry.data_file.partition` I think.

Tmonster (approved):
Thanks! My comment would could be a good follow up PR I think

PR #1339, opened 2026-08-19
iceberg_rewrite_data_files: support Spark's min_input_files parameter

The-Alchemist:
Match Spark's `SizeBasedFileRewriter`: select a partition group when it has at least two files whose combined size meets the rewrite target.

## Source

min-input-files | 5 | Any file group with this number of files or more will be rewritten regardless of other criteria (the file group should have at least two files)
-- | -- | --

### DIFF
--- a/test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rewrite_data_files_enough_bytes.test
+++ b/test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rewrite_data_files_enough_bytes.test
@@ -43,12 +43,19 @@ where content = 'DATA' and status <> 'DELETED';
 ----
 3
 
-# Files are ~1.5KiB each (under target 2000), so they pass the undersized gate.
-# Three files is below min_input_files=5, but combined size (~4.4KiB) meets the target.
+# Spark's default min is 75% of target (1500 here). Catalogs write slightly
+# different parquet sizes for the same rows (Nessie ~1.7KiB, fixture ~1.5KiB),
+# so files can land in the "leave-alone" category. Pin min_file_size_bytes above
+# actual file size so this test hits enough_bytes, not catalog writer variation.
+# Also pin max_file_size_bytes: Spark's default max is 180% of target (3600),
+# which would reject min 10000.
+# Three files is below min_input_files=5, but combined size meets the 2KiB target.
 query III
 call iceberg_rewrite_data_files(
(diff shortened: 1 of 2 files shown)


test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rewrite_data_files_enough_bytes.test
```diff
@@ -0,0 +1,70 @@
...
+----
+3	<REGEX>:[1-9][0-9]*	<REGEX>:[1-9][0-9]*
+
+query I
+select count(*) from my_datalake.default.rewrite_enough_bytes;
+----
+60
+
+query I
+select count(*) > 0
```
Tmonster: Can we get an exact count? I think it would be 1 in this case right?
Can we also check the size of the re-written files? You can do that with something like 

This should be double checked, but the SQL to check would look like this I think.
```
set variable file_path = select file_path from iceberg_metadata(<table_name>) where sequence_number = (3?) limit 1;
select file_size from read_blob(getvariable('file_path'));
```

You could also try the new [variable syntax](https://github.com/duckdb/duckdb/pull/21194)? I haven't played around with that yet

Tmonster (commented):
Thanks! Looks good to me except for a testing case.

If you are up for it, would you also like to add a `max-file-group-size-bytes` parameter similar to the iceberg options? That way DuckDB-Iceberg won't always try to compact ALL candidate files in a table, which for peta byte size tables would not be feasible for either resource or time constraints. This could be a follow up PR, or added here since this PR is already pretty small

test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rewrite_data_files_enough_bytes.test
The-Alchemist: looks like the new variable syntax works!

the new tests might be too specific, as they depend on file size estimates.  let me know what you think 🙏

Tmonster (approved):
I think the tests should be fine. We will take a look if they start to fail.
I restarted the regression test, it seems to be flaky on CI for some reason

Tmonster:
I think I fixed the merge conflict correctly. We shall see if the tests pass

The-Alchemist:
@Tmonster 
> I think I fixed the merge conflict correctly. We shall see if the tests pass

I think I got the last one!

Tmonster (approved):
thanks!

PR #1340, opened 2026-08-19
iceberg_rewrite_data_files: Add `max_files_to_rewrite` parameter

The-Alchemist:
Adds the Spark-compatible `max_files_to_rewrite` parameter to `iceberg_rewrite_data_files`, an upper bound on how many eligible data files a single call will compact.

Without it, compaction is all-or-nothing: a table-wide rewrite touches every eligible file in one commit. The cap lets callers break that into smaller, incremental, resumable batches, which keeps each commit cheap and bounds the blast radius if something goes wrong.

```sql
call iceberg_rewrite_data_files('my_datalake.default.events', max_files_to_rewrite => 100);
```

## Source

max-files-to-rewrite | null | This option sets an upper limit on the number of eligible files that will be rewritten. If this option is not specified, all eligible files will be rewritten.
-- | -- | --

# LLM Notes

## Behavior

The cap bounds how many files are rewritten. It never changes *which* partitions are considered worth rewriting.

That distinction is the core of the change. A partition qualifies for rewrite based on the size and count of **all** its eligible files, and only after it qualifies does the cap limit how many of those files this call actually takes. A low cap can therefore leave a partition partly compacted, but it can never cause a partition to be skipped that would otherwise have been selected.

`max_files_to_rewrite` is rejected at bind time when it is `NULL` or below `1`. When it is omitted, planning and selection behave exactly as before.

## Implementation notes

**Planning no longer materializes every live data file.** `PlanRewrite` used to collect all live DATA files into `RewritePlan::candidates` and group them in a second pass. Candidates are now bucketed by partition key as they are read from the manifests, and `RewritePlan::candidates` is removed. For an unpartitioned table, planning can also stop scanning once the cap and the gating thresholds are both satisfied, rather than always walking every manifest entry. Manifests are still checked for partition spec evolution even after that early exit.

**Gating is evaluated against the full eligible set, not the capped prefix.** Each partition bucket tracks `eligible_count` and `eligible_bytes` across every eligible file, separately from the `retained` prefix that the cap truncates. This is what makes the guarantee above hold. The rule lives in a single `PartitionQualifies` helper shared by the selection loop and the unpartitioned early-exit check, so the two cannot drift apart.

This matters because of how it interacts with the `enough_bytes` rule from #1339, which landed on `main` while this branch was open. That rule selects a partition holding at least two files whose combined size already meets the target, even when the file count is below `min_input_files`. Had gating been evaluated against the cap-truncated prefix instead of the full eligible set, a low cap could have suppressed a partition that should have been rewritten.

## Limitations compared to Spark

Spark's `RewriteDataFiles` bin-packs eligible files into ordered *file groups* and truncates that list. This implementation has no equivalent grouping step: every selected file across every qualifying partition feeds a single `COPY`. Two consequences follow.

* **The cap is filled in partition-key order, not by benefit.** Partitions are visited in sorted key order and the budget is consumed greedily by the first qualifying ones; within a partition, files are taken in manifest-scan order. Spark orders its file groups before truncating, so it can spend a small cap where compaction pays off most. Here a low cap deterministically lands on the lexicographically-first partitions.
* **A partition is never split across calls in a size-aware way.** There is no way to bin-pack one very large partition into several groups and take the cap from more than one of them.

The ordering is at least deterministic, so repeated capped calls make steady forward progress rather than revisiting the same files.

### DIFF
--- a/src/function/metadata/iceberg_rewrite_data_files.cpp
+++ b/src/function/metadata/iceberg_rewrite_data_files.cpp
@@ -85,6 +85,16 @@ static RewriteDataFilesPlanInput ParseRewritePlanInput(TableFunctionBindInput &i
 				                            value);
 			}
 			result.min_input_files = value;
+		} else if (opt == "max_files_to_rewrite") {
+			if (val.IsNull()) {
+				throw InvalidInputException("iceberg_rewrite_data_files: 'max_files_to_rewrite' cannot be NULL");
+			}
+			auto value = val.GetValue<int64_t>();
+			if (value < 1) {
+				throw InvalidInputException("iceberg_rewrite_data_files: 'max_files_to_rewrite' must be >= 1, got %lld",
+				                            value);
+			}
+			result.max_files_to_rewrite = value;
 		} else if (opt == "rewrite_all") {
 			result.rewrite_all = BooleanValue::Get(val);
 		}
@@ -183,6 +193,7 @@ TableFunctionSet IcebergFunctions::GetIcebergRewriteDataFilesFunction() {
 	function.named_parameters["min_file_size_bytes"] = LogicalType::ANY;
 	function.named_parameters["max_file_size_bytes"] = LogicalType::ANY;
 	function.named_parameters["min_input_files"] = LogicalType::BIGINT;
+	function.named_parameters["max_files_to_rewrite"] = LogicalType::BIGINT;
 	function.named_parameters["rewrite_all"] = LogicalType::BOOLEAN;
 	function_set.AddFunction(function);
 	return function_set;
--- a/src/maintenance/rewrite_data_files_planner.cpp
+++ b/src/maintenance/rewrite_data_files_planner.cpp
@@ -74,7 +74,7 @@ static int64_t ResolveMaxFileSizeBytes(const RewriteDataFilesPlanInput &input, i
 }
 
 bool BucketQualifies(const RewriteBucket &bucket, const RewriteDataFilesPlanInput &input,
-                        int64_t target_file_size_bytes) {
+                     int64_t target_file_size_bytes) {
 	if (input.rewrite_all) {
 		return true;
 	}
--- a/src/maintenance/rewrite_data_files_planner.cpp
+++ b/src/maintenance/rewrite_data_files_planner.cpp
@@ -23,7 +23,7 @@ namespace {
 constexpr int64_t DEFAULT_TARGET_FILE_SIZE_BYTES = 512LL * 1024 * 1024;
 constexpr int64_t MIN_TARGET_FILE_SIZE_BYTES = 100;
 
-static int64_t ParseTargetFileSizeProperty(const string &value, const string &property) {
+int64_t ParseTargetFileSizeProperty(const string &value, const string &property) {
 	idx_t parsed_value;
 	if (!TryCast::Operation<string_t, idx_t>(string_t(value), parsed_value)) {
 		auto error = StringUtil::TryParseFormattedBytes(value, parsed_value);
@@ -41,8 +41,7 @@ static int64_t ParseTargetFileSizeProperty(const string &value, const string &pr
 	return static_cast<int64_t>(parsed_value);
 }
 
-static int64_t ResolveTargetFileSizeBytes(const RewriteDataFilesPlanInput &input,
-                                          const IcebergTableMetadata &metadata) {
+int64_t ResolveTargetFileSizeBytes(const RewriteDataFilesPlanInput &input, const IcebergTableMetadata &metadata) {
 	if (input.target_file_size_bytes) {
 		return input.target_file_size_bytes.value();
 	}
@@ -59,14 +58,14 @@ static int64_t ResolveTargetFileSizeBytes(const RewriteDataFilesPlanInput &input
 }
 
 //! Spark SizeBasedFileRewriter defaults: min = 75% of target, max = 180% of target.
-static int64_t ResolveMinFileSizeBytes(const RewriteDataFilesPlanInput &input, int64_t target_file_size_bytes) {
+int64_t ResolveMinFileSizeBytes(const RewriteDataFilesPlanInput &input, int64_t target_file_size_bytes) {
 	if (input.min_file_size_bytes) {
 		return input.min_file_size_bytes.value();
 	}
 	return (target_file_size_bytes * 3) / 4;
 }
 
-static int64_t ResolveMaxFileSizeBytes(const RewriteDataFilesPlanInput &input, int64_t target_file_size_bytes) {
+int64_t ResolveMaxFileSizeBytes(const RewriteDataFilesPlanInput &input, int64_t target_file_size_bytes) {
 	if (input.max_file_size_bytes) {
 		return input.max_file_size_bytes.value();
 	}
@@ -206,29 +205,7 @@ void CollectUnpartitionedCandidates(RewriteBucket &bucket, const vector<IcebergM
 	}
 }
 
-void CollectPartitionedCandidates(std::map<string, RewriteBucket> &per_partition,
-                                  const vector<IcebergManifestListEntry> &manifest_files,
-                                  const RewriteDataFilesPlanInput &input, int32_t default_spec_id) {
-	for (const auto &list_entry : manifest_files) {
-		if (list_entry.file.content != IcebergManifestContentType::DATA) {
-			continue;
-		}
-		AssertCurrentPartitionSpec(list_entry, default_spec_id);
-		for (const auto &entry : list_entry.GetManifestEntries()) {
-			auto cand = TryMakeRewriteCandidate(entry);
-			if (!cand) {
-				continue;
-			}
-			auto &bucket = per_partition[rewrite_planner_internal::PartitionBucketKey(cand->partition_info)];
-			ConsiderCandidate(bucket, std::move(*cand), input);
-		}
-	}
-}
-
-} // namespace
-
-namespace rewrite_planner_internal {
-
+//! Canonical partition key used by the bin-packer.
 string PartitionBucketKey(const vector<IcebergPartitionInfo> &partition_info) {
 	if (partition_info.empty()) {
 		return "";
@@ -254,7 +231,26 @@ string PartitionBucketKey(const vector<IcebergPartitionInfo> &partition_info) {
 	return out;
 }
 
-} // namespace rewrite_planner_internal
+void CollectPartitionedCandidates(std::map<string, RewriteBucket> &per_partition,
+                                  const vector<IcebergManifestListEntry> &manifest_files,
+                                  const RewriteDataFilesPlanInput &input, int32_t default_spec_id) {
+	for (const auto &list_entry : manifest_files) {
+		if (list_entry.file.content != IcebergManifestContentType::DATA) {
+			continue;
+		}
+		AssertCurrentPartitionSpec(list_entry, default_spec_id);
+		for (const auto &entry : list_entry.GetManifestEntries()) {
+			auto cand = TryMakeRewriteCandidate(entry);
+			if (!cand) {
+				continue;
+			}
+			auto &bucket = per_partition[PartitionBucketKey(cand->partition_info)];
+			ConsiderCandidate(bucket, std::move(*cand), input);
+		}
+	}
+}
+
+} // namespace
 
 RewritePlan PlanRewrite(ClientContext &context, RewriteDataFilesPlanInput input) {
 	RewritePlan plan;
--- a/src/maintenance/rewrite_data_files_planner.cpp
+++ b/src/maintenance/rewrite_data_files_planner.cpp
@@ -156,11 +156,7 @@ void SelectCandidates(RewritePlan &plan, const RewriteDataFilesPlanInput &input,
 }
 
 void AssertCurrentPartitionSpec(const IcebergManifestListEntry &list_entry, int32_t default_spec_id) {
-	//! Guard against partition spec evolution: if a manifest was written
-	//! under a different partition spec, its partition tuples may not match
-	//! the current default spec. Mixing specs in one rewrite group would
-	//! produce incorrect manifest metadata. Reject until multi-spec support
-	//! is implemented.
+	//! Guard against partition spec evolution: reject until multi-spec support is implemented.
 	if (list_entry.file.partition_spec_id != default_spec_id) {
 		throw NotImplementedException(
 		    "iceberg_rewrite_data_files: table has data files written under partition spec %d "
--- a/src/maintenance/rewrite_data_files_planner.cpp
+++ b/src/maintenance/rewrite_data_files_planner.cpp
@@ -73,45 +73,75 @@ static int64_t ResolveMaxFileSizeBytes(const RewriteDataFilesPlanInput &input, i
 	return (target_file_size_bytes * 9) / 5;
 }
 
-void GroupCandidates(RewritePlan &plan, const RewriteDataFilesPlanInput &input) {
-	if (plan.candidates.empty()) {
+bool PartitionQualifies(const PartitionRewriteBucket &bucket, const RewriteDataFilesPlanInput &input,
+                        int64_t target_file_size_bytes) {
+	if (input.rewrite_all) {
+		return true;
+	}
+	//! Spark SizeBasedFileRewriter: rewrite when the group has enough files,
+	//! or at least two files whose total size already meets the target.
+	if (bucket.eligible_count >= input.min_input_files) {
+		return true;
+	}
+	return bucket.eligible_count >= 2 && bucket.eligible_bytes >= target_file_size_bytes;
+}
+
+void ConsiderCandidate(PartitionRewriteBucket &bucket, RewriteCandidate cand, const RewriteDataFilesPlanInput &input,
+                       int64_t min_file_size_bytes, int64_t max_file_size_bytes) {
+	//! Match Spark: rewrite undersized or oversized files; leave the band alone.
+	if (!input.rewrite_all && cand.file_size_in_bytes >= min_file_size_bytes &&
+	    cand.file_size_in_bytes <= max_file_size_bytes) {
 		return;
 	}
-
-	const auto min_file_size_bytes = ResolveMinFileSizeBytes(input, plan.target_file_size_bytes);
-	const auto max_file_size_bytes = ResolveMaxFileSizeBytes(input, plan.target_file_size_bytes);
-	if (min_file_size_bytes > max_file_size_bytes) {
-		throw InvalidInputException("iceberg_rewrite_data_files: resolved 'min_file_size_bytes' (%lld) must be <= "
-		                            "'max_file_size_bytes' (%lld)",
-		                            min_file_size_bytes, max_file_size_bytes);
+	bucket.eligible_count++;
+	bucket.eligible_bytes += cand.file_size_in_bytes;
+	if (input.max_files_to_rewrite &&
+	    bucket.retained.size() >= static_cast<idx_t>(input.max_files_to_rewrite.value())) {
+		//! Partition already has a Spark-style subList prefix; keep counting so gating stays cap-independent.
+		return;
 	}
+	bucket.retained.push_back(std::move(cand));
+}
 
-	std::map<string, vector<RewriteCandidate>> per_partition;
-	for (auto &cand : plan.candidates) {
-		//! Match Spark: rewrite undersized or oversized files; leave the band alone.
-		if (!input.rewrite_all && cand.file_size_in_bytes >= min_file_size_bytes &&
-		    cand.file_size_in_bytes <= max_file_size_bytes) {
+bool UnpartitionedCollectionComplete(const std::map<string, PartitionRewriteBucket> &per_partition,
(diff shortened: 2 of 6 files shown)


src/function/metadata/iceberg_rewrite_data_files.cpp
```diff
@@ -83,6 +83,13 @@ static RewriteDataFilesPlanInput ParseRewritePlanInput(TableFunctionBindInput &i
 				                            value);
 			}
 			result.min_input_files = value;
+		} else if (opt == "max_files_to_rewrite") {
+			auto value = val.GetValue<int64_t>();
```
Tmonster, line 92: Can we get a null check on this as well?

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -80,6 +80,24 @@ void GroupCandidates(RewritePlan &plan, int64_t target_file_size_bytes, int64_t
 	}
 }
 
+void LimitSelectedCandidates(RewritePlan &plan, const RewriteDataFilesPlanInput &input) {
+	if (!input.max_files_to_rewrite ||
```
Tmonster: If the table is not partitioned, is there a reason to still sort the candidates?

`PlanRewrite` could also help with this? For unpartitioned tables, we just stop collecting at one point, for partitioned tables, while collecting candidates, if a partition has more than max_input_files, we could short circuit?

test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rewrite_data_files_max_files_to_rewrite.test
```diff
@@ -0,0 +1,70 @@
...
+from iceberg_metadata('my_datalake.default.rewrite_max_files')
+where content = 'DATA' and status <> 'DELETED';
+----
+6
+
+# Six eligible files, but only rewrite three of them.
+query III
+call iceberg_rewrite_data_files(
+	'my_datalake.default.rewrite_max_files',
+	min_input_files => 5,
```
Tmonster, line 50: Is this also sparks behavior? Seems confusing to require a group has 5 candidates, but then only rewrite 3 of them. 
If this is what spark does, ok, otherwise I think we can throw an error is `min_input_files > max_files_to_rewrite`

Tmonster (commented):
Thanks!
Just a few comments

src/function/metadata/iceberg_rewrite_data_files.cpp
The-Alchemist: done.

i'm assuming we'd want to skip on `null`, rather than throw an error.

added a test too; might be overkill

test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rewrite_data_files_max_files_to_rewrite.test
The-Alchemist: my reading of Spark’s pipeline in [SizeBasedFileRewritePlanner](https://raw.githubusercontent.com/apache/iceberg/main/core/src/main/java/org/apache/iceberg/actions/SizeBasedFileRewritePlanner.java) then [BinPackRewriteFilePlanner.plan()](https://raw.githubusercontent.com/apache/iceberg/main/core/src/main/java/org/apache/iceberg/actions/BinPackRewriteFilePlanner.java) is that :
* Spark does not compare `min_input_files` and `max_files_to_rewrite`
    * they're used in separate stages of the pipeline
    * `min-input-files` (default is 5): any file group with this many files (and at least 2) is rewritten
    * `max-files-to-rewrite` (default is null): upper bound on how many eligible files this run will rewrite
 * my LLM says:
     * [PlanRewrite](src/maintenance/rewrite_data_files_planner.cpp) matches Spark’s order:
         * GroupCandidates: drop undersized files (unless rewrite_all); keep a partition only if count >= min_input_files
        * LimitSelectedCandidates: sort deterministically, then resize to max_files_to_rewrite
     * That is why the test expects 3  1  <bytes> and 4 remaining data files (1 new + 3 untouched).

IMHO, it's confusing, but Spark's logic looks to be quite complex with so many compaction parameters.

https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/actions/SizeBasedFileRewritePlanner.java#L189-L195

So Spark will compact two files if together they are bigger than the target, even when `min-input-files` is 5.

Work for a future PR?

src/function/metadata/iceberg_rewrite_data_files.cpp
Tmonster: I think actually on NULL we should error. Just like in the regression tests, if someone wants to pass a variable, if the variable does not exist, we want to notify the user they passed NULL

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -159,6 +198,8 @@ RewritePlan PlanRewrite(ClientContext &context, const RewriteDataFilesPlanInput
 	}
 
 	auto default_spec_id = table_metadata.default_spec_id;
+	const bool unpartitioned = !table_metadata.HasPartitionSpec();
+	std::map<string, PartitionRewriteBucket> per_partition;
```
Tmonster: Looks like this is already implementing [this comment](https://github.com/duckdb/duckdb-iceberg/pull/1338#discussion_r3828836187) a bit.

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -57,25 +57,64 @@ static int64_t ResolveTargetFileSizeBytes(const RewriteDataFilesPlanInput &input
...
-void GroupCandidates(RewritePlan &plan, int64_t target_file_size_bytes, int64_t min_input_files, bool rewrite_all) {
-	if (plan.candidates.empty()) {
+struct PartitionRewriteBucket {
+	vector<RewriteCandidate> retained;
+	int64_t eligible_count = 0;
+};
+
+void ConsiderCandidate(PartitionRewriteBucket &bucket, RewriteCandidate cand, const RewriteDataFilesPlanInput &input,
+                       int64_t target_file_size_bytes) {
+	if (!input.rewrite_all && cand.file_size_in_bytes >= target_file_size_bytes) {
```
Tmonster: looks like here you have to add back in `min_file_size_bytes` `max_file_size_bytes` logic?

Tmonster (commented):
Seems like the merge conflicts here need some attention since https://github.com/duckdb/duckdb-iceberg/pull/1338 got merged. Thank you for being so active though, really appreciate it!

src/function/metadata/iceberg_rewrite_data_files.cpp
The-Alchemist: done!

src/maintenance/rewrite_data_files_planner.cpp
The-Alchemist: yeah, let's see how the merge/rebase looks

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -73,30 +73,58 @@ static int64_t ResolveMaxFileSizeBytes(const RewriteDataFilesPlanInput &input, i
 	return (target_file_size_bytes * 9) / 5;
 }
 
-void GroupCandidates(RewritePlan &plan, const RewriteDataFilesPlanInput &input) {
-	if (plan.candidates.empty()) {
+struct PartitionRewriteBucket {
```
Tmonster: nit: This could also go in the header file

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -73,30 +73,58 @@ static int64_t ResolveMaxFileSizeBytes(const RewriteDataFilesPlanInput &input, i
...
-		    cand.file_size_in_bytes <= max_file_size_bytes) {
-			continue;
-		}
-		per_partition[rewrite_planner_internal::PartitionBucketKey(cand.partition_info)].push_back(cand);
+bool UnpartitionedCollectionComplete(const std::map<string, PartitionRewriteBucket> &per_partition,
+                                     const RewriteDataFilesPlanInput &input) {
+	if (!input.max_files_to_rewrite) {
+		return false;
+	}
+	auto it = per_partition.find("");
```
Tmonster: This scares me a little bit. What happens a table is partitioned on a varchar column where one of the values is `""`?

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -73,30 +73,58 @@ static int64_t ResolveMaxFileSizeBytes(const RewriteDataFilesPlanInput &input, i
...
+		remaining = NumericCast<idx_t>(input.max_files_to_rewrite.value());
+	}
 	for (auto &kv : per_partition) {
+		if (capped && remaining == 0) {
+			return;
+		}
+		auto &bucket = kv.second;
+		if (!input.rewrite_all && bucket.eligible_count < input.min_input_files) {
+			continue;
+    }
```
Tmonster: nit: the indentation here looks a little funky? I think one is tabs, one is spaces

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -212,6 +256,12 @@ RewritePlan PlanRewrite(ClientContext &context, const RewriteDataFilesPlanInput
 			    list_entry.file.partition_spec_id, default_spec_id);
 		}
 
+		if (unpartitioned && UnpartitionedCollectionComplete(per_partition, input)) {
```
Tmonster: Could we rename `PartitionRewriteBucket` to `RewriteBucket`? 
And we can check for partitioned-ness outside of this loop, and pass the `RewriteBucket` to `UnpartitionedCollectionComplete`? Then `UnpartitionedCollectionComplete` will read a bit nicer in the function. 

Then there is more clear separation between how partitioned and non-partitioned tables are handled when selecting candidates. Then if we want to expand this to multiple partition specs (with potentially unpartitioned specs mixed in), we will have a clear distinction when collecting candidates from different `partitioned_ids` and  candidates that are from an unpartitioned spec

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -73,30 +73,58 @@ static int64_t ResolveMaxFileSizeBytes(const RewriteDataFilesPlanInput &input, i
...
+	}
 	for (auto &kv : per_partition) {
+		if (capped && remaining == 0) {
+			return;
+		}
+		auto &bucket = kv.second;
+		if (!input.rewrite_all && bucket.eligible_count < input.min_input_files) {
+			continue;
+    }
 		if (!input.rewrite_all) {
```
Tmonster: Could this now also go in `ConsiderCandidate`?

Tmonster (commented):
Thanks! almost ready, just need to have better handling for partitioned vs unpartitioned specs.

Do we have any tests on compacting temporal transforms for partitions? i.e `day()`, `month()` etc?

src/maintenance/rewrite_data_files_planner.cpp
The-Alchemist: done

src/maintenance/rewrite_data_files_planner.cpp
The-Alchemist: i think i fixed it with the rebase. needs to check

src/maintenance/rewrite_data_files_planner.cpp
The-Alchemist: done

src/maintenance/rewrite_data_files_planner.cpp
The-Alchemist: yeah, threw me off as well.

```

string PartitionBucketKey(const vector<IcebergPartitionInfo> &partition_info) {
	if (partition_info.empty()) {
		return "";
	}
```


LLM:
> So it only returns the empty string when the partition_info vector itself is empty, i.e. there are no partition fields at all. Any actual field contributes at least field_id, a colon, and a trailing semicolon.

> A table partitioned on a varchar where the value is "" yields something like 1000:0:; — field id, length 0, empty value, terminator. That is not "", so it lands in its own bucket exactly like any other value. The length prefix is also what keeps "" from colliding with a value containing ; or :.

> Worth noting the same encoding handles the other collision you'd worry about: a NULL is written as 1000:NULL; while the literal string "NULL" is written as 1000:4:NULL;. Since a non-null value always emits digits after the first colon and NULL starts with N, those can never collide either.

But I agree it's confusing.  I added a clearer early-exist for unpartitioned tables

src/maintenance/rewrite_data_files_planner.cpp
The-Alchemist: ok, done.  let me know what you think, it's a bigger change.  we can move it to a separate PR if that makes more sense

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -73,45 +73,160 @@ static int64_t ResolveMaxFileSizeBytes(const RewriteDataFilesPlanInput &input, i
 	return (target_file_size_bytes * 9) / 5;
 }
 
-void GroupCandidates(RewritePlan &plan, const RewriteDataFilesPlanInput &input) {
-	if (plan.candidates.empty()) {
+bool PartitionQualifies(const RewriteBucket &bucket, const RewriteDataFilesPlanInput &input,
```
Tmonster: nit: `BucketQualifies`

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -73,45 +73,160 @@ static int64_t ResolveMaxFileSizeBytes(const RewriteDataFilesPlanInput &input, i
...
+	//! is implemented.
+	if (list_entry.file.partition_spec_id != default_spec_id) {
+		throw NotImplementedException(
+		    "iceberg_rewrite_data_files: table has data files written under partition spec %d "
+		    "but current default spec is %d; partition spec evolution is not yet supported",
+		    list_entry.file.partition_spec_id, default_spec_id);
+	}
+}
+
+bool TryMakeRewriteCandidate(const IcebergManifestEntry &entry, RewriteCandidate &cand) {
```
Tmonster: Could we return `optional<RewriteCandidate>`?

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -73,45 +73,160 @@ static int64_t ResolveMaxFileSizeBytes(const RewriteDataFilesPlanInput &input, i
...
+	//! Spark SizeBasedFileRewriter: rewrite when the group has enough files,
+	//! or at least two files whose total size already meets the target.
+	if (bucket.eligible_count >= input.min_input_files) {
+		return true;
+	}
+	return bucket.eligible_count >= 2 && bucket.eligible_bytes >= target_file_size_bytes;
+}
+
+void ConsiderCandidate(RewriteBucket &bucket, RewriteCandidate cand, const RewriteDataFilesPlanInput &input,
+                       int64_t min_file_size_bytes, int64_t max_file_size_bytes) {
```
Tmonster: Can we modify `PlanRewrite` so that it ensure values for `min_files_size_bytes` and `max_file_size_bytes` exist? Then `RewriteDataFilesPlanInput` is guaranteed to already have parsed values of `min_file_size_bytes` and `max_file_size_bytes`, and we don't need to pass them around to all the functions

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -73,45 +73,160 @@ static int64_t ResolveMaxFileSizeBytes(const RewriteDataFilesPlanInput &input, i
...
-			if (!enough_files && !enough_bytes) {
+		if (capped && remaining == 0) {
+			return;
+		}
+		SelectFromBucket(plan, kv.second, input, remaining, capped);
+	}
+}
+
+void AssertCurrentPartitionSpec(const IcebergManifestListEntry &list_entry, int32_t default_spec_id) {
+	//! Guard against partition spec evolution: if a manifest was written
```
Tmonster: nit: comment is too verbose here

src/maintenance/rewrite_data_files_planner.cpp
```diff
@@ -73,45 +73,160 @@ static int64_t ResolveMaxFileSizeBytes(const RewriteDataFilesPlanInput &input, i
...
+			continue;
+		}
+		AssertCurrentPartitionSpec(list_entry, default_spec_id);
+		for (const auto &entry : list_entry.GetManifestEntries()) {
+			RewriteCandidate cand;
+			if (!TryMakeRewriteCandidate(entry, cand)) {
+				continue;
+			}
+			auto &bucket = per_partition[rewrite_planner_internal::PartitionBucketKey(cand.partition_info)];
+			ConsiderCandidate(bucket, std::move(cand), input, min_file_size_bytes, max_file_size_bytes);
```
Tmonster: Should we also have a check here if the bucket is complete? Similar to `CollectUnpartitionedCandidates`?

Tmonster (commented):
thanks! Almost there again, the more important thing is the comment about not passing all the `min_file_size_bytes` and `max_file_size_bytes` arguments everywhere

src/maintenance/rewrite_data_files_planner.cpp
The-Alchemist: it looks like that won't work.  i'm not very familiar with this part of the code, so here's the LLM analysis.  i'll leave you to interpret.

# LLM analysis

**No.** `CollectPartitionedCandidates` should not skip remaining manifests the way `CollectUnpartitionedCandidates` does.

## Why the unpartitioned check exists

[`UnpartitionedCollectionComplete`](src/maintenance/rewrite_data_files_planner.cpp) is an early-exit for the table's **one** partition (the unpartitioned table is collected into a single `RewriteBucket`):

- `max_files_to_rewrite` is set
- `retained` already has that many files
- `BucketQualifies` is already true (`min_input_files` / byte-target / `rewrite_all`)

After that, extra eligible files only inflate `eligible_count` / `eligible_bytes`, which no longer changes the decision. [`ConsiderCandidate`](src/maintenance/rewrite_data_files_planner.cpp) already refuses to retain more than the cap. So the unpartitioned collector can skip `GetManifestEntries()` on later DATA manifests (it still walks the list to run `AssertCurrentPartitionSpec`).

```195:208:src/maintenance/rewrite_data_files_planner.cpp
		if (UnpartitionedCollectionComplete(bucket, input, target_file_size_bytes)) {
			//! Cap and group gating are already satisfied; remaining DATA manifests
			//! are still checked above for partition spec evolution.
			continue;
		}
		// ...
			if (UnpartitionedCollectionComplete(bucket, input, target_file_size_bytes)) {
				break;
			}
```

## Why that does not transfer to partitioned collection

Partitioned planning has **many partitions** and a **global** cap applied later:

1. Files are grouped by partition key in `per_partition`.
2. Each partition retains at most `max_files_to_rewrite` files (scan-order prefix).
3. [`SelectCandidates`](src/maintenance/rewrite_data_files_planner.cpp) walks the `std::map` in **lexicographic partition-key order** and spends a **global** remaining cap.

Manifest order is not partition-key order. Tests such as [`rewrite_data_files_max_files_to_rewrite.test`](test/sql/local/catalog_test_config_setup/catalog_agnostic/maintenance/rewrite_data_files_max_files_to_rewrite.test) rely on filling the cap from the **first key** (`category=a`), not from whichever partition was scanned first.

If collection stopped after one partition already had a qualifying, cap-sized `retained` prefix:

- A later manifest could still introduce a lexicographically earlier partition that should be selected first.
- The same manifest can mix partitions, so you cannot skip the rest of a DATA manifest just because one partition is complete.

Skipping remaining manifests would also skip `AssertCurrentPartitionSpec` for those manifests, which the unpartitioned path deliberately still does.

```mermaid
flowchart TD
  manifests[DATA manifests in snapshot order]
  collect[Collect into per_partition map]
  select[SelectFromBucket in map key order]
  cap[Global remaining max_files_to_rewrite]
  manifests --> collect
  collect --> select
  select --> cap
```

## Per-partition skip (optional, not worth matching the unpartitioned shape)

A check of the form “this partition is already complete, do not call `ConsiderCandidate`” would be **correct**: that partition already qualifies and will not retain more files.

It would **not** save the expensive work. You still must:

- iterate every DATA manifest (spec-evolution guard)
- parse every entry (`GetManifestEntries`, `TryMakeRewriteCandidate`)
- compute `PartitionBucketKey` to know which partition the file belongs to

`ConsiderCandidate` is already a cheap no-op on retain once that partition's prefix is full. Copying `UnpartitionedCollectionComplete` into this loop would look symmetric and be misleading.

## Recommendation

Leave [`CollectPartitionedCandidates`](src/maintenance/rewrite_data_files_planner.cpp) as-is. The completeness helper is intentionally named and used only on the unpartitioned path.

No code or test changes unless you later want a clearly named per-partition micro-opt (skip `ConsiderCandidate` only), which should not skip manifests or break out of the entry loop.

src/maintenance/rewrite_data_files_planner.cpp
The-Alchemist: done

src/maintenance/rewrite_data_files_planner.cpp
The-Alchemist: done

Tmonster (approved):
Thanks!

PR #1342, opened 2026-08-20
Preserve endpoint suffix from vended Azure SAS config key

djouallah:
Cherry-pick of #1331 to main, as requested there. Same change; applies to
iceberg_table.cpp after the rename in b6115a9b.

Verified against a live OneLake Iceberg REST catalog: reads return data with
the fix and fail with AuthenticationFailed without it.


Tmonster:
thanks!

PR #1345, opened 2026-08-20
Enable file pruning on nested columns

jvansanten:
This PR solves the same problem as #820, but uses the current TableFilter API and test infrastructure, following [this advice](https://github.com/duckdb/duckdb-iceberg/pull/820#issuecomment-4726602987) from @Tishj 

To recap: currently, `IcebergFilePruner` assumes that all filters refer to top-level columns. Filters on fields of struct columns are ignored during scan planning, because Iceberg metadata only stores min/max statistics for leaf columns. This makes it impossible to prune files from a scan based on a nested column, which is a thing I need to do in order to handle an externally-specified schema.

This PR introduces the following changes:
- IcebergTableSchema::GetColumnFromIndex can look up columns based on a pushdown-extract ColumnIndex (i.e. a reference to a nested column)
- IcebergFilePruner uses IcebergTableSchema::GetColumnFromIndex to look up columns from the schema
- IcebergFilePruner::ManifestMatchesFilter walks filters, then looks up the corresponding column and spec, just like IcebergFilePruner::FileMatchesFilter
- IcebergMultiFileList converts struct_extract() expressions to bound column references, and adds the corresponding leaf columns to a copy of the column projection. This allows for file pruning even when the scan itself reads a parent of the filtered column.


jvansanten:
Tests now seem to be broken by minio disappearing (presumably fixed by https://github.com/duckdb/duckdb-iceberg/pull/1387).

Other than that, any thoughts on this approach, @Tishj?

PR #1346, opened 2026-08-20
Bump duckdb to 3be83b480f

Tmonster:
Settings/catalog lookups now take `Identifier`: config constants 
Value::DefaultTryCastAs` returns `optional<Value>` 
`Value::Reinterpret` has been renamed to `WithType`.
duckdb #24834 made `MultiFileBindInternal` call `IsEmpty()` on every bind, before `Bind()` applies options or resolves metadata expansion warm-up in `GetExpandResult()` only happens if `have_bound` is true


Tmonster:
gonna bump even further

PR #1347, opened 2026-08-20
Fixes around `DEFAULT_SCHEMA`

carlopi:
Fixes around `DEFAULT_SCHEMA`.

1. A bug around:
```sql
ATTACH '' AS db (TYPE ICEBERG, SECRET s, DEFAULT_SCHEMA 'bronze');
CREATE SCHEMA db.main;
CREATE TABLE db.main.T AS SELECT 42;
```
would currently land table T in schema `bronze` (like, in the actual catalog, not only in our views, making basically schema named `main` not at the same level of functionality.

2. Make:
```
ATTACH '' AS db (TYPE ICEBERG, SECRET s, DEFAULT_SCHEMA '');
```
viable, with `''` basically meaning "disable default schema behaviour, and don't validate it"

3. And improve error message around:
```sql
ATTACH '' AS db (TYPE ICEBERG, SECRET s);
CREATE TABLE db.T AS SELECT 42;
```
```
--- before
Catalog Error: Schema with name "" not found

--- after
Catalog Error: No default namespace for catalog "db" - fully qualify the name, or re-attach with DEFAULT_SCHEMA '<namespace>'
```

### DIFF
--- a/test/sql/local/catalog_custom_setup/fixture/main_namespace.test
+++ b/test/sql/local/catalog_custom_setup/fixture/main_namespace.test
@@ -1,6 +1,5 @@
 # name: test/sql/local/catalog_custom_setup/fixture/main_namespace.test
-# description: namespaces are resolved verbatim - 'main' is an ordinary namespace and is never
-#              an alias for the catalog's default namespace
+# description: namespaces are resolved verbatim - 'main' is an ordinary namespace and is never an alias for the catalog's default namespace
 # group: [fixture]
 
 require-env FIXTURE_SERVER_AVAILABLE
--- a/test/sql/local/catalog_custom_setup/fixture/main_namespace.test
+++ b/test/sql/local/catalog_custom_setup/fixture/main_namespace.test
@@ -1,0 +1,142 @@
+# name: test/sql/local/catalog_custom_setup/fixture/main_namespace.test
+# description: namespaces are resolved verbatim - 'main' is an ordinary namespace and is never
+#              an alias for the catalog's default namespace
+# group: [fixture]
+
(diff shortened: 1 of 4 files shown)


test/sql/local/catalog_custom_setup/fixture/main_namespace.test
```diff
@@ -0,0 +1,141 @@
+# name: test/sql/local/catalog_custom_setup/fixture/main_namespace.test
+# description: namespaces are resolved verbatim - 'main' is an ordinary namespace and is never an alias for the catalog's default namespace
+# group: [fixture]
+
+require-env FIXTURE_SERVER_AVAILABLE
```
Tmonster, line 5: This can be a `CATALOG_TEST_CONFIG_SETUP`, then it runs on all catalogs

Tmonster (commented):
This might be ok? I want to make sure there are not any adverse affects from core if default_schema = "".
In `Catalog::TryLookupEntry` in core has `IsInvalidSchema(schema))` which returns true when the default schema is empty. When that happens, all other known schemas are tried, so I worry we might be making a ton of extra requests that we should not be making? Also do you have any tests when you run 
```
ATTACH ... as my_datalake.
USE my_datalake;
---
USE my_datalake.main;
```
To see what the effects are there?

Also if core ends up throwing an error when `default_schema = ""` the error message could have bad UX.

Tmonster:
With https://github.com/duckdb/duckdb/pull/24979 and the duckdb submodule now merged, I think we can close this in favor of setting Default schema to an empty schema

carlopi:
Sure, makes sense, I would consider re-using the provided tests to check the current bug is fixed, cool this is moving forward.

PR #1350, opened 2026-08-21
iceberg_rewrite_data_files: size-filter and group candidates while scanning

The-Alchemist:
- Follow-up to the review on #1338 ([comment](https://github.com/duckdb/duckdb-iceberg/pull/1338#discussion_r3828836187)): size-filter rewrite candidates while scanning manifests instead of keeping every live data file and pruning later.
- Group candidates by partition during that same scan using `entry.data_file.partition_info`, then apply `min_input_files` on the per-partition groups.
- Drop the unused `RewritePlan::candidates` vector; selection semantics are unchanged.


The-Alchemist:
already done, basically

Issue #1128, opened 2026-07-02
CREATE TABLE against AWS Glue Iceberg REST catalog fails: location omitted from createTable request


malon64:
### What happens?

When the `iceberg` extension is attached to an **AWS Glue** Iceberg REST catalog
(`ENDPOINT_TYPE GLUE`), `CREATE TABLE` fails because DuckDB sends a `createTable` request with **no
table `location`**. AWS Glue requires a location and rejects the request:

```
Request to '.../namespaces/<ns>/tables' returned a non-200 status code (BadRequest_400) …
{"error":{"code":400,"message":"Location information cannot be null while creating an iceberg table","type":"InvalidInputException"}}
```

The same code path works against Apache Polaris because Polaris auto-assigns the table location from
the namespace/warehouse when the client omits it. Glue does not auto-assign — it requires the client
to send `location` — so DuckDB's omission surfaces as a hard failure. Reads (loadTable / scans) work
fine; only table creation is affected.

### To Reproduce

```sql
INSTALL aws; LOAD aws; INSTALL httpfs; LOAD httpfs; INSTALL iceberg; LOAD iceberg;
CREATE OR REPLACE SECRET s (TYPE S3, PROVIDER credential_chain, REGION 'eu-west-1');

ATTACH '<account_id>' AS lakehouse (
  TYPE ICEBERG,
  ENDPOINT 'https://glue.eu-west-1.amazonaws.com/iceberg',
  ENDPOINT_TYPE GLUE,
  ACCESS_DELEGATION_MODE none
);

-- namespace already exists with a location property (verified via loadNamespaceMetadata):
--   "location": "s3://my-bucket/my_namespace/"
CREATE TABLE lakehouse.my_namespace.t AS SELECT 1 AS a;   -- 400: Location information cannot be null
```

### OS:

aarch64

### DuckDB Version:

DuckDB 1.4.5

### DuckDB Client:

iceberg extension build `2f229463`

### Full Name:

Alexis Metwalli

### Affiliation:

Onepoint

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

malon64:
[...]

PR #1351, opened 2026-08-22
Add default_table_location_from_namespace attach option

### LINKED ISSUE #1299: Glue: derive a default table location on CREATE TABLE (or add a DEFAULT_TABLE_LOCATION attach option)
### Feature / gap

When attached to AWS Glue's Iceberg REST catalog, every `CREATE TABLE` must carry an explicit `WITH ('location' = 's3://…')`, or Glue rejects it:

```
POST …/iceberg/v1/catalogs/<id>/namespaces/<ns>/tables
400 InvalidInputException: "Location information cannot be null while creating an iceberg table"
```

This was discussed in #1128 (closed with the `WITH ('location'…)` workaround), where @Tmonster commented:

> Hmm, potentially glue catalogs return a default location that can be used when attaching, I'll look into this.

Requesting exactly that: **derive a default table location for Glue when the user doesn't pass one** — or, failing that, a `DEFAULT_TABLE_LOCATION` ATTACH option the client can set once.

### Why the workaround isn't always available

The `WITH ('location'…)` workaround assumes the tool emitting the SQL can add the clause. Higher-level tools often can't: **dbt-duckdb** generates `CREATE TABLE … AS` from model definitions and has no mechanism to pass per-table options through to an attached Iceberg catalog. Net effect today: a dbt project that runs fine against OneLake / S3 Tables / Polaris / Snowflake Horizon REST catalogs (where the server assigns locations, via stage-create or otherwise) cannot run against Glue at all — Glue both rejects `stage-create` (`400 "Stage create is not supported"`, so `STAGE_CREATE_TABLES false` is mandatory) and refuses to assign a location on direct creates. The two behaviors together mean the *client* is the only place a location can come from.

### Repro

duckdb 1.5.4, iceberg extension 75726455 (also current `main` by code inspection — `IcebergTableSet::CreateTable` only reads `location` from the SQL options):

```sql
CREATE SECRET glue_secret (TYPE s3, KEY_ID '…', SECRET '…', REGION 'xx-xxxx-1');
ATTACH '<account_id>' AS glue (TYPE iceberg, ENDPOINT_TYPE 'glue', SECRET glue_secret);
CREATE TABLE glue.demo.t (a INTEGER);
-- 400: Location information cannot be null while creating an iceberg table
CREATE TABLE glue.demo.t (a INTEGER) WITH ('location' = 's3://bucket/demo/t');
-- works
```

### Possible shapes

1. On create without a location, call `LoadNamespaceMetadata` (Glue returns the database `LocationUri` in the namespace properties when one is set) and default to `<namespace location>/<table_name>`. This mirrors what Spark's GlueCatalog does.
2. And/or an ATTACH option, e.g. `DEFAULT_TABLE_LOCATION 's3://bucket/warehouse'`, used as `<value>/<namespace>/<table_name>` whenever `location` is absent — works even for namespaces without a `LocationUri`, and is settable from tools that only expose ATTACH-level configuration (dbt-duckdb profiles, for instance).

Happy to test either against a real Glue catalog — public CI repro setup at https://github.com/djouallah/testing-iceberg-rest-catalog.

tom-kita:
Fixes #1299

### Problem

AWS Glue's Iceberg REST endpoint does not assign a table location server-side. `CREATE TABLE` without an explicit `WITH ('location' = ...)` fails:

```
400 InvalidInputException: "Location information cannot be null while creating an iceberg table"
```

The `WITH ('location')` workaround (from #1128) is per-table, which tools like dbt-duckdb cannot inject.

### Solution

A new boolean attach option, `default_table_location_from_namespace` (default `false`). When enabled and `CREATE TABLE` has no explicit location, the client derives `<namespace location>/<table name>` from the namespace's `location` property (fetched via the existing `IcebergSchemaEntry::LoadProperties`, which caches the `GET /namespaces/{ns}` response). Trailing slashes are trimmed. If the namespace has no `location` property, behavior is unchanged.

The derivation rule matches the default behavior of Java's [`GlueCatalog.defaultWarehouseLocation`](https://github.com/apache/iceberg/blob/7f879b11366e17a676a03f15247a821751415529/aws/src/main/java/org/apache/iceberg/aws/glue/GlueCatalog.java#L287) (`dbLocationUri + "/" + tableName`). Java additionally supports a `unique-table-location` catalog property that appends a UUID suffix to the table name; it is off by default and is not implemented here.

Because this is a per-connection attach option rather than a per-table clause, tools that cannot emit `WITH ('location')` should be able to set it once when configuring the attachment (dbt-duckdb, for example, passes arbitrary attach options through to `ATTACH`), which should resolve the dbt use case from the issue.

### Why opt-in (not defaulted on for `ENDPOINT_TYPE 'glue'`)

Verified against a real Glue account (ap-northeast-1):

- Native Glue namespace **with** `locationUri`: `GET /namespaces/{ns}` returns a `location` property. Before: 400. After (option enabled): create succeeds at `<locationUri>/<table>`, and the table is readable/writable.
- Native Glue namespace **without** `locationUri`: no `location` property, so derivation is skipped and behavior is unchanged.
- **Glue-federated S3 Tables** (`<account>:s3tablescatalog/<bucket>`, the setup used by this repo's Glue cloud tests): `GET /namespaces/{ns}` returns only `ownerAccountId`/`createdBy` and no `location` property, so derivation is a no-op there and server-side assignment keeps working as today.

Other Glue federated catalog types (e.g. Redshift/RMS) are unverified, so the option is opt-in everywhere for now. Defaulting it on for `ENDPOINT_TYPE 'glue'` could be a follow-up once those are verified.

### Testing

New fixture test (placed under `catalog_specifc/fixture/` following the existing pattern of tests that need their own `ATTACH` with non-default options, like `test_polaris.test`): covers derivation including trailing-slash trimming, reads/writes through the derived location, explicit `WITH ('location')` still taking precedence, and no behavior change without the option. The fixture is discriminative here because the server side (JdbcCatalog) assigns locations from the warehouse root and ignores the namespace's `location` property.

### DIFF
--- a/src/catalog/rest/iceberg_table_set.cpp
+++ b/src/catalog/rest/iceberg_table_set.cpp
@@ -262,6 +262,15 @@ IcebergTable &IcebergTableSet::CreateNewEntry(ClientContext &context, IcebergCat
 		location = ParseTableProperty(property_binder, context, *location_it->second, "location", LogicalType::VARCHAR)
 		               .GetValue<string>();
 	}
+	if (location.empty() && catalog.attach_options.default_table_location_from_namespace) {
+		schema.LoadProperties(context);
+		auto ns_location_it = schema.schema_info.properties.find("location");
+		if (ns_location_it != schema.schema_info.properties.end() && !ns_location_it->second.empty()) {
+			location = ns_location_it->second;
+			StringUtil::RTrim(location, "/");
+			location += "/" + info.GetTableName().GetIdentifierName();
+		}
+	}
 
 	IcebergTableMetadata bootstrap_metadata;
 	bootstrap_metadata.iceberg_version = iceberg_version.GetIndex();
--- a/src/iceberg_attach.cpp
+++ b/src/iceberg_attach.cpp
@@ -246,6 +246,10 @@ unique_ptr<Catalog> IcebergAttach::Attach(optional_ptr<StorageExtensionInfo> sto
 		} else if (lower_name == "purge_requested") {
 			attach_options.purge_requested = entry.second.DefaultCastAs(LogicalType::BOOLEAN).GetValue<bool>();
 			set_by_attach_options.insert("purge_requested");
+		} else if (lower_name == "default_table_location_from_namespace") {
+			attach_options.default_table_location_from_namespace =
+			    entry.second.DefaultCastAs(LogicalType::BOOLEAN).GetValue<bool>();
+			set_by_attach_options.insert("default_table_location_from_namespace");
 		} else if (lower_name == "default_schema") {
 			default_schema = Identifier(entry.second.ToString());
 		} else if (lower_name == "encode_entire_prefix") {
--- a/src/include/iceberg_attach.hpp
+++ b/src/include/iceberg_attach.hpp
@@ -32,6 +32,9 @@ struct IcebergAttachOptions {
 	bool encode_entire_prefix = false;
 	// in rest api spec, purge requested defaults to false.
 	bool purge_requested = false;
+	// some catalogs (e.g. AWS Glue) do not assign a table location server-side; derive one from the namespace's
+	// 'location' property
+	bool default_table_location_from_namespace = false;
 	IRCAccessDelegationMode access_mode = IRCAccessDelegationMode::VENDED_CREDENTIALS;
 	IcebergAuthorizationType authorization_type = IcebergAuthorizationType::INVALID;
 	unordered_map<string, Value> options;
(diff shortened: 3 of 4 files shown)


Tmonster (commented):
Thanks for the PR!

My understanding is the location is inferred from the schema properties only if the given location is false. 

Is there any reason to not have this be default functionality? If we remove the setting and fall back to `schema.schema_info.properties["location"]/table.name` whenever location is empty and `schema.schema_info.properties["location"]` exists, what use cases would have worse UX?

I'm just wondering because I think it could clean up the PR a bit if we remove the setting.

edit: 
to make removing the setting even better, we could update the error message when no location is provided to include a statement like 
"""
To infer a table location, set the schema 'location' property for the table with call iceberg_schema_set_property('my_datalake.default', {'location': 's3://path/to/schema'});
"""

Tmonster:
Nevermind, I got my order of operations mixed up, you have to pass the location on table create, so it can't default to on. If so, duckdb-iceberg will use locations provided by schema properties instead of locations that catalogs create and provide for users

Tmonster (approved):
Thanks!

tom-kita:
@Tmonster I'm sorry for my late response. Thanks for review!
I will work on document update.

PR #1353, opened 2026-08-23
Fix credential refresh for catalogs without the credentials endpoint

### LINKED ISSUE #1352: Expired vended storage credentials are reused on subsequent table scans
### What happens?

When an Iceberg REST catalog vends temporary storage credentials, the credentials work for the initial table scan.

After the credentials expire, a subsequent scan of the same table continues to use the previously cached credentials instead of obtaining a fresh set from the catalog. The object-storage request is therefore signed with the expired key and fails with an authentication error.

This can affect credentials containing expiration properties such as:

- `s3.session-token-expires-at-ms`
- `gcs.oauth2.token-expires-at`
- `adls.sas-token-expires-at-ms.<host>`

Expected behavior:

Before reusing vended credentials, DuckDB-Iceberg should check their advertised expiration time. If the credentials have expired or are close to expiry, the extension should obtain and persist replacement credentials before accessing object storage.

Concurrent scans of the same table should share the refreshed credential state and should not independently perform duplicate refreshes.

If the standard table credentials request fails or returns no credentials, the extension should fall back to reloading the table because some catalogs return replacement credentials only through `LoadTable`.

### To Reproduce

Use an Iceberg REST catalog that vends temporary S3 credentials.

1. Configure the catalog to return valid temporary credentials when the table is initially loaded:

   ```json
   {
     "storage-credentials": [
       {
         "prefix": "s3://warehouse/default/example",
         "config": {
           "s3.access-key-id": "TEMPORARY_KEY_1",
           "s3.secret-access-key": "TEMPORARY_SECRET_1",
           "s3.session-token": "TEMPORARY_TOKEN_1",
           "s3.session-token-expires-at-ms": "<a near-future epoch timestamp in milliseconds>",
           "s3.region": "us-east-1",
           "s3.endpoint": "http://127.0.0.1:9000",
           "s3.path-style-access": "true"
         }
       }
     ]
   }
   ```

2. Attach the catalog:

   ```sql
   ATTACH '' AS iceberg_catalog (
       TYPE ICEBERG,
       URI 'http://127.0.0.1:8181',
       CLIENT_ID 'admin',
       CLIENT_SECRET 'password'
   );
   ```

3. Scan the table while the credentials are still valid:

   ```sql
   SELECT * FROM iceberg_catalog.default.example;
   ```

   The query succeeds.

4. Wait until `TEMPORARY_KEY_1` expires.

5. Configure the catalog to return a new valid credential, such as `TEMPORARY_KEY_2`, when the table credentials are requested or the table is loaded again.

6. Scan the same table again:

   ```sql
   SELECT * FROM iceberg_catalog.default.example;
   ```

Actual result:

The second scan still signs the object-storage request using `TEMPORARY_KEY_1`. The request fails because the credential has expired.

Example failure:

```text
HTTP 403 Forbidden
InvalidAccessKeyId / ExpiredToken
```

The catalog is not asked for replacement credentials before the expired credential is used.

Expected result:

The extension detects that `TEMPORARY_KEY_1` has expired, obtains `TEMPORARY_KEY_2`, updates its cached credential state, and completes the second scan successfully.

### OS:

macOS, Apple Silicon (arm64)

### DuckDB Version:

1.5

### DuckDB Client:

cli

### Hardware:

_No response_

### Full Name:

yoogo chen

### Affiliation:

JDIX

- [ ] Yes, I have

- [ ] Yes, I have

yoogoc:
## Summary

Use LoadTable to refresh vended storage credentials when a REST catalog does not advertise support for the table credentials endpoint.

Related to #1352.

## Changes

Update `IRCAPI::GetTableCredentials` to select the request path using the catalog's supported endpoints:

- If `/credentials` is advertised, use the existing credentials endpoint.
- Otherwise, call LoadTable and construct a `LoadCredentialsResponse` from its `storage-credentials`, discarding the remaining table metadata.
- Preserve errors from an advertised credentials endpoint rather than silently falling back to LoadTable.

## Scope

This PR only addresses endpoint compatibility.

Persisting refreshed credentials and handling their advertised expiration will be addressed in a separate PR.

## Testing

Added fixture-backed coverage for:

- An explicit endpoint list without `/credentials`.
- An omitted endpoint list, using the legacy default endpoints.
- An advertised credentials endpoint returning an error without triggering a fallback.

Existing coverage exercises successful refresh through `/credentials`.

Validation completed:
- Formatting checks.
- C++ syntax checks against the expected DuckDB headers.
- Proxy behavior checks.

The full fixture-backed integration suite has not yet been run locally.

### DIFF
--- a/src/catalog/rest/catalog_entry/table/iceberg_table.cpp
+++ b/src/catalog/rest/catalog_entry/table/iceberg_table.cpp
@@ -179,11 +179,22 @@ static void ParseConfigOptions(const case_insensitive_map_t<string> &config, cas
 }
 
 IRCAPITableCredentials IcebergTable::GetVendedCredentials(ClientContext &context) const {
-	return GetVendedCredentials(context, storage_credentials);
+	return GetCurrentVendedCredentials(context);
+}
+
+IRCAPITableCredentials IcebergTable::GetCurrentVendedCredentials(ClientContext &context) const {
+	lock_guard<mutex> guard(credential_state->lock);
+	return GetVendedCredentials(context, credential_state->config, credential_state->storage_credentials);
 }
 
 IRCAPITableCredentials
 IcebergTable::GetVendedCredentials(ClientContext &context,
+                                   const vector<rest_api_objects::StorageCredential> &storage_credentials) const {
+	return GetVendedCredentials(context, config, storage_credentials);
+}
+
+IRCAPITableCredentials
+IcebergTable::GetVendedCredentials(ClientContext &context, const case_insensitive_map_t<string> &table_config,
                                    const vector<rest_api_objects::StorageCredential> &storage_credentials) const {
 	IRCAPITableCredentials result;
 	auto schema_component = IRCPathComponent::NamespaceComponent(schema.namespace_items, catalog.namespace_separator);
@@ -223,7 +234,7 @@ IcebergTable::GetVendedCredentials(ClientContext &context,
 	case_insensitive_map_t<Value> config_options;
 	//! TODO: apply the 'defaults' retrieved from the /v1/config endpoint
 	config_options.insert(user_defaults.begin(), user_defaults.end());
-	ParseConfigOptions(config, config_options, context, storage_type);
+	ParseConfigOptions(table_config, config_options, context, storage_type);
 
 	//! If there is only one credential listed, we don't really care about the prefix,
 	//! we can use the table_location instead.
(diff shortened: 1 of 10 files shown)


Tmonster (commented):
Can you provide a PR description please?

`IcebergTable` has a `credential_state` and IcebergCatalog has `credential_states`, could you provide more information as to why we need both? 

We already have functionality for refreshing vended credentials, it may only be for SigV4 catalogs now, but we should easily be able to extend it for the other authorization types. Did you look into that? If so, what was the blocker?

yoogoc:
> Can you provide a PR description please?
> 
> `IcebergTable` has a `credential_state` and IcebergCatalog has `credential_states`, could you provide more information as to why we need both?
> 
> We already have functionality for refreshing vended credentials, it may only be for SigV4 catalogs now, but we should easily be able to extend it for the other authorization types. Did you look into that? If so, what was the blocker?

Thanks. I have added a detailed PR description.

`IcebergTable::credential_state` is the table's reference to the mutable state, while `IcebergCatalog::credential_states` owns and deduplicates that state by table key. The catalog-level registry is needed because concurrent connections can construct independent `IcebergTable` instances without going through `Copy()`.

I also looked into #1051. It refreshes the Catalog AWS Secret used for SigV4 REST requests. This issue concerns table-scoped `storage-credentials` used by HTTPFS, including credentials vended by OAuth2 catalogs. The PR therefore extends the existing `IcebergTableSecretProvider` refresh path rather than the SigV4 authorization path.

src/catalog/rest/catalog_entry/table/iceberg_table.cpp
```diff
@@ -508,12 +519,94 @@ static void AddHTTPSecretsToOptions(SecretEntry &http_secret_entry, case_insensi
 	                            : http_kv_secret.TryGetValue("verify_ssl").DefaultCastAs(LogicalType::BOOLEAN);
 }
 
+bool IcebergTable::VendedCredentialsExpired() const {
```
Tmonster, line 522: This is never called it seems like?

src/catalog/rest/iceberg_catalog.cpp
```diff
@@ -45,6 +84,17 @@ IcebergCatalog::IcebergCatalog(AttachedDatabase &db_p, AccessMode access_mode,
...
 IcebergCatalog::~IcebergCatalog() = default;
 
+shared_ptr<IcebergVendedCredentialState> IcebergCatalog::GetVendedCredentialState(const string &table_key) {
+	lock_guard<mutex> guard(credential_states_lock);
+	auto entry = credential_states.find(table_key);
+	if (entry != credential_states.end()) {
+		return entry->second;
+	}
+	auto state = make_shared_ptr<IcebergVendedCredentialState>();
+	credential_states[table_key] = state;
```
Tmonster, line 94: I'm seeing the iceberg catalog having `credential_states` and each iceberg table having a `credential_state`. To me this smells bad

Tmonster (commented):
I feel like this PR is fixing 2 different issues, and the solution for both is not very clean.

What I've identified as one issue, is that we always request the `/credentials` url when attempting to Revend credentials when they've expired. If a catalog does not support this url, we should just hit the regular `GetTableInformation` endpoint. 

That can be 1 PR. 

To address it, I think we can make some changes in the function 
```
IRCAPI::GetTableCredentials(ClientContext &context, IcebergCatalog &catalog, const IcebergSchemaEntry &schema,
                            const string &table_name) {
``` 
in `catalog_api.cpp`

You can check if the catalog supports the `/credentials` url. If it doesn't, you can just hit the `LoadTableResult` url again and construct the `LoadCredentialsResponse` from the `LoadTableResult` and throw away the rest of the information. 

 
Another PR can be storing this new vended credential since we do not update the credential when we revend it. Along side that you can add readers for the 
```
s3.session-token-expires-at-ms
gcs.oauth2.token-expires-at
adls.sas-token-expires-at-ms.<host>
```
prefixes. 

I think this 2nd PR would be relatively small.

yoogoc:
> I feel like this PR is fixing 2 different issues, and the solution for both is not very clean.
> 
> What I've identified as one issue, is that we always request the `/credentials` url when attempting to Revend credentials when they've expired. If a catalog does not support this url, we should just hit the regular `GetTableInformation` endpoint.
> 
> That can be 1 PR.
> 
> To address it, I think we can make some changes in the function
> 
> ```
> IRCAPI::GetTableCredentials(ClientContext &context, IcebergCatalog &catalog, const IcebergSchemaEntry &schema,
>                             const string &table_name) {
> ```
> 
> in `catalog_api.cpp`
> 
> You can check if the catalog supports the `/credentials` url. If it doesn't, you can just hit the `LoadTableResult` url again and construct the `LoadCredentialsResponse` from the `LoadTableResult` and throw away the rest of the information.
> 
> Another PR can be storing this new vended credential since we do not update the credential when we revend it. Along side that you can add readers for the
> 
> ```
> s3.session-token-expires-at-ms
> gcs.oauth2.token-expires-at
> adls.sas-token-expires-at-ms.<host>
> ```
> 
> prefixes.
> 
> I think this 2nd PR would be relatively small.

Thanks, that makes sense. I'll narrow this PR to endpoint selection in IRCAPI::GetTableCredentials, using LoadTable when the catalog doesn't support /credentials. I'll handle persisting refreshed credentials and parsing expiration fields in a separate PR, and revisit the state management to keep it simpler. You're also right that VendedCredentialsExpired() is unused; I'll remove it.

PR #1357, opened 2026-08-27
Return a tri-state pushdown result from FileMatchesFilter

paultmathew:
Split out of #1287, implementing @Tmonster's review suggestion there:

> We have `FileMatchesFilter` functionality right now that returns `true/false`. Potentially `FileFullyCoveredByFilter` could merge with `FileMatchesFilter` and instead of returning a bool you can return an enum value

## What changed

`FileMatchesFilter` answered one question with a `bool`: could any row in this data file match the pushed-down filters. Deciding whether *every* row matches — what a metadata-only DELETE needs — meant a second function re-walking the same manifest entry and the same filter expression.

The two are now one `METADATA_STATS_PUSHDOWN` result. `IcebergPredicate::MatchBounds` walks the filter **once** and answers both halves per node, using Iceberg's own vocabulary: the *inclusive* evaluator asks whether any row can match, the *strict* one whether all do. `FileMatchesFilter` gathers both kinds of evidence a manifest entry carries — the file's partition values and its own column bounds — in a single filter-driven loop, replacing the separate spec-field pass that used to run before the bounds loop.

Exclusion short-circuits, since one filter ruling the file out settles it; a strict match needs every filter, and within a filter any one evidence source suffices (a column can source several partition fields, e.g. `bucket` and `truncate`).

## Reviewability

**Nothing consumes `ALL_ROWS_MATCH` yet.** Both call sites treat it as `SOME_ROWS_MATCH`, so file selection is unchanged apart from the deltas below. The consumer lands in #1287.

## Behavior deltas

Three widenings, all of which prune more rather than less:

- A non-NULL partition value now clears `has_null`. Every transform carries NULL through to NULL, so a file gathered under a non-NULL partition value holds no NULL rows for that column, and `IS NULL` can rule it out. This one is load-bearing: with `has_null` left at its default `true`, the strict evaluator rejects every partition-value proof.
- A spec field whose `source_id` is missing from the schema map is skipped rather than throwing out of `unordered_map::at`.
- A VARIANT column with an absent bound yields no evidence for that column instead of abandoning the walk over the whole file, so the remaining filters are still evaluated.

Proving a strict match no longer requires a partitioned spec, nor a partition field for every filtered column, because a file's own bounds can establish it alone. That removes the need for a separate spec-covers-every-filter-column guard and the per-spec memo cache it required — which is what @Tmonster's comment on `iceberg_multi_file_list.cpp:193` in #1287 was about.

## Tests

Two new sqllogictests, both covering gaps that existed before this change:

- `test_is_null_partition_pruning.test` — `IS NULL` against a partitioned table. A single insert spanning three partition values puts three data files in one manifest, so the manifest summary's `contains_null` doesn't prune first and evaluation reaches the data-file level. Also pins the asymmetry: `IS NOT NULL` rules out the NULL-partition file on its null counts, not its partition value.
- `test_multi_field_partition_pruning.test` — `PARTITIONED BY (bucket(4, id), truncate(100, id))`, two fields sharing one source column. The suite had no coverage for this layout, and it is exactly what the per-column vector of partition fields exists for.


src/include/planning/pruning/iceberg_predicate.hpp
```diff
@@ -7,13 +7,25 @@
 
 namespace duckdb {
 
+//! What a set of bounds proves about a filter.
+enum class METADATA_STATS_PUSHDOWN : uint8_t {
```
Tishj, line 11: Please use PascalCase for the enum class name, not fully capitalized, we reserve that style for defines and constants. Note that the members of the enum themselves are fine

src/planning/pruning/iceberg_file_pruner.cpp
```diff
@@ -34,12 +34,145 @@ void ApplyNullCounts(const IcebergDataFile &data_file, int32_t column_id, Iceber
...
+	           stats.upper_bound ? stats.upper_bound->ToString() : "N/A", filter.ToString(column.name));
+}
+
 } // namespace
 
-bool IcebergFilePruner::FilePartitionMatchesFilter(const IcebergDataFile &data_file,
-                                                   const IcebergManifestFile &manifest_file) const {
+//! Partition values a data file carries are few (rarely more than a handful of fields), so a linear scan
+//! to find one by field id is cheaper than hashing it.
+const Value *TryGetPartitionValue(const IcebergDataFile &data_file, uint64_t partition_field_id) {
```
Tishj, line 153: `optional_ptr<Value>`

src/planning/pruning/iceberg_predicate.cpp
```diff
@@ -87,6 +88,112 @@ static bool MatchBoundsConstantTemplated(const Value &constant, ExpressionType c
 	}
 }
 
+//! Identity keeps the source value, so the predicate applies to the bounds unchanged - no projection.
+static bool AllRowsMatchIdentityBounds(const Value &constant, ExpressionType comparison_type,
+                                       const IcebergPredicateStats &stats) {
+	switch (comparison_type) {
+	case ExpressionType::COMPARE_EQUAL:
+		return *stats.lower_bound == constant && *stats.upper_bound == constant;
```
Tishj, line 96: This isn't checking that `lower_bound` and `upper_bound` are actually set.
It looks like that is checked by the calling code, this needs assertions to indicate that

src/include/planning/pruning/iceberg_predicate.hpp
Tishj, line 11: On a different note, I wonder if using flags isn't more appropriate here, since surely `all_rows_match` is a superset of `some_rows_match` right?
But we could make that change in a separate PR if that makes sense, just wanted to raise it here

paultmathew: Conceptually this makes sense to me and I think the current `SOME_ROWS_MATCH` enum isn't quite accurate because we need to scan the file to determine this. I think it makes sense to reframe this to model the output as 2 booleans:
| Relation between predicate and file | `match_possible` | `all_rows_match` |
| --- | --- | --- |
| The predicate definitely matches no rows | `false` | `false` |
| The predicate may match rows, but metadata cannot confirm it | `true` | `false` |
| The predicate definitely matches every row | `true` | `true` |

The caller then maps those boolean outputs to a scan decision based on whether it needs row values:
| `match_possible` | `all_rows_match` | Predicate relation | Caller consumes rows (`SELECT`, `UPDATE`, etc.) | Caller does not consume rows (metadata-only `DELETE`) |
| --- | --- | --- | --- | --- |
| `false` | `false` | The predicate definitely matches no rows | Do not scan | Do not scan: there is nothing to delete |
| `true` | `false` | The predicate may match rows | Scan and evaluate the predicate | Scan and perform a row-level delete |
| `true` | `true` | The predicate definitely matches every row | Scan because the caller needs rows | Do not scan; drop the data file through metadata |

paultmathew: Lmk if this makes sense to you or if you had something different in mind. Either way happy to implement it in this PR or a separate one.

test/configs/fixture-latest.json
```diff
@@ -22,6 +22,9 @@
 			"paths": [
 				"test/sql/local/catalog_test_config_setup/catalog_agnostic/delete/test_no_head_request_for_deletes.test",
 				"test/sql/local/catalog_test_config_setup/catalog_agnostic/insert/partitions/combined/test_identity_and_year.test",
+				"test/sql/local/catalog_test_config_setup/catalog_agnostic/reads/filtering/test_is_null_partition_pruning.test",
```
Tmonster: any reason these are skipped?

PR #1358, opened 2026-08-27
Accept bare scheme names as vended credential prefixes

plusplusjiajia:
### Problem

  A storage credential's `prefix` is a plain string matched against file paths — the Java `S3FileIO` filters with `c.prefix().startsWith(ROOT_PREFIX)` — so catalogs write it either as a url (`s3://bucket/path`) or as a bare scheme name (`s3`). Both are legal, but the bare form breaks twice:

  1. `CredentialMatchesStorageType` only accepts the url form, so every credential from such a catalog is filtered out and reads fall back to long-term credentials. Against buckets that only accept the vended session credentials that fails with `403 AccessDenied`, after a successful `ATTACH` and table listing.
  2. Even if it survives the filter, the bare name is used as the secret scope, and `BaseSecret::MatchScore` scores by matched prefix length — scope `s3` (2) loses to any secret scoped `s3://` (5). The fallback that appends the table location does not fire, because `s3://bucket/path` does start with `s3`.

  This only bites when a catalog vends more than one credential; with exactly one, `ignore_credential_prefix` skips the filter.

  ### Fix

  Normalize a prefix to its scheme before comparing (url-form behaviour unchanged), and scope such credentials to the table location, as already happens for a single vended credential — a bare scheme says which storage the credential is for, not which paths it covers.

### DIFF
--- a/src/catalog/rest/catalog_entry/table/iceberg_table.cpp
+++ b/src/catalog/rest/catalog_entry/table/iceberg_table.cpp
@@ -241,14 +241,18 @@ IcebergTable::GetVendedCredentials(ClientContext &context,
 		create_secret_input.on_conflict = OnCreateConflict::REPLACE_ON_CONFLICT;
 		create_secret_input.persist_type = SecretPersistType::TRANSACTION;
 
+		//! A bare scheme name would score below any secret scoped to the full "scheme://" prefix, so
+		//! widen it to that prefix: it stays the least specific credential, as in Java's S3FileIO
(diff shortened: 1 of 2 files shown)


Tmonster (approved):
Thanks!
@plusplusjiajia What catalog are you testing against? Is it something that can be added to the Iceberg Test suite? Or would we need to set up an online account somewhere? Seems like it would be beneficial to run all duckdb tests suite against this catalog to see if there are any other missing features

PR #1360, opened 2026-08-27
Fix UTC semantics for Iceberg TIMESTAMPTZ temporal transforms

p1p1bear:
Fix Iceberg temporal transforms for `TIMESTAMPTZ` and `TIMESTAMPTZ_NS` to use the encoded UTC instant during writes.

Previously, `date_diff` used ICU and the session time zone, while partition pruning used UTC epoch semantics. Reinterpreting zoned timestamps as their timezone-free physical equivalents keeps write and pruning values consistent.


Tmonster (approved):
Thanks!

PR #1361, opened 2026-08-27
fix: handle pre-epoch Iceberg month and day transforms

p1p1bear:
Fix Iceberg month and day transforms for timestamps before the Unix epoch.

Month now follows the existing year transform by using calendar components, while day follows the existing hour transform by using floor division. This ensures correct partition pruning for negative timestamps.

### DIFF
--- a/src/include/core/expression/iceberg_transform.hpp
+++ b/src/include/core/expression/iceberg_transform.hpp
@@ -2,6 +2,7 @@
 
 #include "common/iceberg_math.hpp"
 #include "core/expression/iceberg_predicate_stats.hpp"
+#include "duckdb/common/types/date.hpp"
 #include "duckdb/common/types/interval.hpp"
 
 namespace duckdb {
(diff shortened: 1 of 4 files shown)


src/include/core/expression/iceberg_transform.hpp
```diff
@@ -160,25 +160,25 @@ struct MonthTransform {
...
+			return Value::INTEGER((components.year - 1970) * 12 + (components.month - 1));
 		}
 		case LogicalTypeId::TIMESTAMP_TZ_NS: {
 			auto val = constant.GetValue<timestamp_tz_ns_t>();
-			auto ts = timestamp_t(IcebergNanosToMicrosFloor(val.value));
-			auto diff = Interval::GetAge(ts, timestamp_t::epoch());
-			return Value::INTEGER(diff.months);
+			auto micros = IcebergNanosToMicrosFloor(val.value);
+			auto components = Timestamp::GetComponents(timestamp_t(micros));
+			return Value::INTEGER((components.year - 1970) * 12 + (components.month - 1));
```
Tmonster: Can we get 1970 from `Date::EPOCH_YEAR` instead?  Available in `duckdb/common/types/date.hpp` ?

Tmonster (commented):
Thanks!
Looks good for me except the one nit

p1p1bear:
@Tmonster  Thanks!
I've updated the related places to consistently use `Date::EPOCH_YEAR` instead of 1970.

PR #1362, opened 2026-08-27
fix(credentials): Load vended credentials manually when LoadTable omits them

### LINKED ISSUE #1349: iceberg_rewrite_data_files does not load catalog credentials
`iceberg_rewrite_data_files` (#1035) doesn't fetch the catalog's vended storage credentials. It only works if something else has already read the table in the same session.

duckdb `1.6.0.dev365`, iceberg from core, Cloudflare R2 Data Catalog.

```sql
CREATE SECRET (TYPE ICEBERG, TOKEN '...');
ATTACH 'warehouse' AS catalog (TYPE ICEBERG, ENDPOINT '...');

-- fails: HTTP 403 AccessDenied, "Authentication Failure ... No credentials are provided"
SELECT * FROM iceberg_rewrite_data_files('catalog.mart.dim_duid');

-- works: same session, one metadata read first
SELECT count(*) FROM iceberg_metadata('catalog.mart.dim_duid');
SELECT * FROM iceberg_rewrite_data_files('catalog.mart.dim_duid');  -- 61 -> 1 files
```

`information_schema.tables` does not prime it; only a per-table read does.

This is very confusing for a job that does nothing but compaction — there's no reason for it to scan the table first, so it fails every time with an auth error that looks like a bad token.

btw, it is an awesome work

====================================================================================================
Iceberg compaction (duckdb 1.6.0.dev365)
----------------------------------------------------------------------------------------------------
landing.fct_price_today         OK (10 -> 1 files, 14.9 MB rewritten)
landing.fct_scada_today         OK (4707 -> 1 files, 83.4 MB rewritten)
mart.dim_calendar               skipped (<5 eligible files)
mart.dim_duid                   skipped (<5 eligible files)
landing.stg_csv_archive_log     OK (10 -> 1 files, 3.8 MB rewritten)
landing.fct_price               OK (239 -> 8 files, 684.0 MB rewritten)
landing.fct_scada               OK (640 -> 67 files, 4626.1 MB rewritten)
====================================================================================================

The-Alchemist:
## Summary
- Fixes `iceberg_rewrite_data_files` failing in a catalog-only session when LoadTable does not include usable storage credentials. This matches the [Cloudflare R2 report](https://github.com/duckdb/duckdb-iceberg/issues/1349#issuecomment-5423023604): compaction returned HTTP 403 / "No credentials are provided" until a prior `iceberg_metadata` read primed the session.
- `IcebergTable::LoadCredentials` now falls back to the advertised `GET /v1/{prefix}/namespaces/{namespace}/tables/{table}/credentials` endpoint when LoadTable `storage-credentials` / `config` do not contain usable storage authentication. LoadTable credentials still take precedence.
- Not sure if there is a cleaner solution
- Feedback requested

Fixes #1349

### DIFF
--- a/scripts/data_generators/tests/default/vended_credentials_refresh_tables/__init__.py
+++ b/scripts/data_generators/tests/default/vended_credentials_refresh_tables/__init__.py
@@ -54,6 +54,23 @@ class Test(IcebergTest):
             """
         )
 
+        con.con.sql(
+            """
+            CREATE OR REPLACE TABLE default.vended_rewrite_table (
+                id integer,
+                payload string
+            )
+            USING iceberg
+            """
+        )
+        for i in range(1, 7):
+            con.con.sql(
+                f"""
+                INSERT INTO default.vended_rewrite_table
+                VALUES ({i}, 'p{i}')
+                """
+            )
+
         for table in REFRESH_METADATA_TABLES:
             con.con.sql(
                 f"""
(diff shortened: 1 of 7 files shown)


Tmonster:
Seems like the original issue is no longer reproducible, so closing this

PR #1363, opened 2026-09-01
iceberg_rewrite_data_files: Add  physical sort-order test

The-Alchemist:
## Summary
- We were lacking a general test that verifies sort-order after compaction, aka, `iceberg_rewrite_data_files` 
- Add catalog-agnostic coverage that `iceberg_rewrite_data_files` writes physical row order from the table's current default sort order.
- Inserts six unordered files first, then sets `a DESC NULLS FIRST, b ASC NULLS LAST` so rewrite (not INSERT) must establish ordering.
- Asserts 6→1 compaction, exact `file_row_number` sequence, and rewritten manifest `sort_order_id`.


Tishj:
Thanks!

PR #1364, opened 2026-09-02
No longer check if schema is Default Schema.

Tmonster:
This is a little bit of cleanup now that https://github.com/duckdb/duckdb/pull/24979 has merged. 
Errors are a bit cleaner around Default schema behavior, and users no longer need to create one or know that one exists in their catalog

CC @carlopi

### DIFF
--- a/src/catalog/rest/iceberg_catalog.cpp
+++ b/src/catalog/rest/iceberg_catalog.cpp
@@ -8,7 +8,6 @@
 #include "duckdb/main/attached_database.hpp"
 #include "duckdb/planner/operator/logical_create_table.hpp"
 #include "duckdb/common/exception/conversion_exception.hpp"
-#include "duckdb/common/string_util.hpp"
 
 #include "catalog/rest/catalog_entry/schema/iceberg_schema_entry.hpp"
 #include "catalog/rest/catalog_entry/table/iceberg_table_schema_version.hpp"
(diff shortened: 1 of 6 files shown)


carlopi:
What happens if `DEFAULT_SCHEMA` is not present in advance?
Like, failure is at `ATTACH` or at query time?

```sql
ATTACH '' AS db (......, DEFAULT_SCHEMA 'not_existing');
FROM db.T;
```
When does it fail? First statement or second?

And unsure, but that would be for a separate PR, if there would not be value in a `(DEFAULT_SCHEMA 'some_schema_name', CREATE_DEFAULT_SCHEMA_IF_NOT_EXISTING)` to guarantee a default schema is there.

Tmonster:
First statement will fail, most recent commit has added more context to the error message

carlopi:
Great, I would consider, also as follow up, a test for that behaviour. And given the workaround is a bit ugly (attach anon + create + detach + attach), I would consider a `CREATE_DEFAULT_SCHEMA_IF_NOT_EXISTING` option that can do the creation directly inline.

Tmonster:
We can do that, I think we can actually downgrade the error to just a warning, and make the default schema a nullopt. Then a user can retroactively set the default schema via a setting or a table function.

carlopi (approved):

PR #1365, opened 2026-09-03
Map write.parquet.page-size-bytes to the parquet writer's data_page_size_limit

djouallah:
One-line addition to `ICEBERG_TABLE_PROPERTY_MAPPING`: the standard Iceberg write property `write.parquet.page-size-bytes` now reaches the parquet writer's `data_page_size_limit` option (added in duckdb/duckdb#24645). Follows the same passthrough as `write.parquet.dict-size-bytes`.

Why it matters: page/layout details of the written parquet have a large measured effect on downstream readers. On a 144M-row table read by Power BI Direct Lake, fixing the writer's footer/page layout halved query cost — 7,923 → 3,132 capacity units (~2.5×) on byte-identical data. Without this mapping, `data_page_size_limit` is reachable from plain `COPY TO` but not from an Iceberg write.

### DIFF
--- a/src/execution/operator/iceberg_insert.cpp
+++ b/src/execution/operator/iceberg_insert.cpp
@@ -608,6 +608,7 @@ static const IcebergParquetOptionMapping ICEBERG_TABLE_PROPERTY_MAPPING[] = {
     {"write.parquet.compression-codec", "codec"},
     {"write.parquet.compression-level", "compression_level"},
     {"write.parquet.dict-size-bytes", "string_dictionary_page_size_limit"},
+    {"write.parquet.page-size-bytes", "data_page_size_limit"},
     {"write.parquet.row-group-size-bytes", "row_group_size_bytes"},
     {"write.parquet.row-group-size", "row_group_size"},
     {"write.parquet.row-groups-per-file", "row_groups_per_file"}};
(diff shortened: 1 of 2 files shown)


Tmonster:
thanks! Have you also thought about a way to verify this? We have tests in `test/sql/local/catalog_test_config_setup/catalog_agnostic/file_properties/` that check file properties of iceberg tables

djouallah:
Added a test in `test/sql/local/catalog_test_config_setup/catalog_agnostic/file_properties/test_parquet_page_size_bytes.test`.

It writes two Iceberg tables that differ only in `write.parquet.page-size-bytes`, both with `'write.parquet.compression-codec' = 'uncompressed'` and the same 100k rows, then asserts that `sum(total_compressed_size)` from `parquet_metadata()` is strictly larger for the 4096-byte-page file — more data pages means more page headers. That is the assertion that fails if the mapping is dropped. It also checks the property round-trips through `iceberg_table_properties` and that the data itself is unaffected.

Same trick duckdb uses for the option itself in `test/sql/copy/parquet/writer/parquet_write_data_page_size.test`.

The Polaris job failed for infrastructure reasons, not the test — it died in `make polaris-data` before anything ran:

```
Starting Polaris catalog...
(cd .catalogs/polaris/site/content/guides/minio && docker compose up -d)
 minio  Interrupted
 setup_bucket  Interrupted
received unexpected HTTP status: 502 Bad Gateway
make: *** [make/catalogs/polaris.mk:23: polaris] Error 1
```

I have pushed an empty commit to retrigger it. Every catalog that actually started passed: fixture, latest fixture, fixture + equality deletes, Nessie, Lakekeeper, Gravitino, local, SigV4 mock, and the duckdb/duckdb suite against fixture.

Tmonster:
Thanks!

djouallah:
no, thank you @Tmonster

PR #1366, opened 2026-09-03
Enable TopN late materialization for Iceberg scans

Liuzw1998:
## Summary

Enable DuckDB's TopN late materialization for eligible Iceberg V2
`ORDER BY ... LIMIT` queries. A narrow candidate scan selects rows, and a
payload scan joins them using `(filename, file_row_number)`.

- Preserve the bound snapshot, schema, options, and transaction view, with
  independent scan state.
- Support equality/position deletes and transaction-local changes.
- Keep `IcebergOptimizerRoutine` unchanged and detect LM candidates after it.
- Fall back for scan predicates referencing virtual columns. Virtual-column
  projections and ordering remain eligible.

No DuckDB core changes are required.

## Performance

Measured on a 500,000-row Iceberg V2 table stored in 10 Parquet files (about 145 MB). The table has two numeric columns and eight 64-character payload columns. Each configuration used four threads, one warmup, and three interleaved measured runs.

```sql
SELECT *
FROM iceberg_scan('topn_wide')
ORDER BY abs(sort_key) DESC, id DESC
LIMIT 100;
```

| | Time (median of 3 runs) | Bytes read |
|---|---:|---:|
| Before (disabled) | 147.27 ms | 144.67 MB |
| After (enabled) | 80.09 ms | 17.52 MB |

This reduces median latency by 45.6% (1.84x speedup) and bytes read by 87.9%.

### DIFF
--- a/src/function/scan/iceberg_scan.cpp
+++ b/src/function/scan/iceberg_scan.cpp
@@ -78,6 +78,13 @@ static void IcebergSetScanOrder(unique_ptr<RowGroupOrderOptions> order_options, 
 	file_list.SetScanOrder(std::move(order_options));
 }
 
+static vector<column_t> IcebergGetRowIdColumns(ClientContext &, optional_ptr<FunctionData>) {
+	vector<column_t> result;
+	result.push_back(MultiFileReader::COLUMN_IDENTIFIER_FILENAME);
+	result.push_back(MultiFileReader::COLUMN_IDENTIFIER_FILE_ROW_NUMBER);
+	return result;
+}
+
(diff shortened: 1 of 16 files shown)


Tishj:
Thanks for the PR, looks good from my side, could you just look into the failing CI? I imagine these tests just need to get skipped on Nessie because it can't facilitate the time travel required.

Liuzw1998:
@Tishj Done! Could you rerun the CI?

Liuzw1998:
@Tishj I skipped certain tests for Nessie and Fixture catalogs. I realized that server-side planning conficts with my implementation of late materialization, which request different columns. Could you please have a look again?

PR #1368, opened 2026-09-03
Enable eager table resolution in attach option

Tmonster:
This PR enables eager resolution for table metadata as an attach option. 

Currently the attach option is `TABLE_RESOLUTION 'eager|lazy'` with the default being lazy so `SHOW ALL TABLES` and `select * from information_schema.columns` are still fast and snappy. Looking for feedback on if the naming here is correct, I'm not in love with it, but other options seems a bit lengthy 
- `COLUMN_RESOLUTION` (unclear)
- `TABLE_COLUMN_METADATA_RESOLUTION`, 
- `METADATA_RESOLUTION` (also unclear, since it's really only for columns)
- `COLUMN_LOADING` (clunky)
- etc.

A regression test has also been added. It currently fails since it is not available in the base branch but it will exist to track progress on how fast table metadata can be resolved.

I know tests pass for fixture, lets see if they also pass for nessie and lakekeeper.

### DIFF
--- a/.github/workflows/Regression.yml
+++ b/.github/workflows/Regression.yml
@@ -99,17 +99,6 @@ jobs:
           sudo apt update -y -qq
           sudo apt install -y -qq python3-venv tar pkg-config
 
-      # Required by the table_resolution benchmarks: the fixture catalog runs on localhost, where a
-      # round trip is sub-millisecond scheduling noise, so toxiproxy fronts it with a fixed,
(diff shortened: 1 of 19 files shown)


Tmonster:
@Tishj do you have any preferences on what the attach option name should be? I'm not in love with `TABLE_RESOLUTION`, but can't think of anything easier

PR #1375, opened 2026-09-08
[ServerSideScanPlanning] Send IN lists and OR-of-equalities to the plan endpoint

JohnEarle:
`IN` lists never reach the plan request. `WHERE id IN (5, 3500)` plans the scan with **no filter
at all**, so the server returns every data file. The rows are right — DuckDB re-applies the
filter — but the pruning is lost.

DuckDB wraps `IN` in an optional filter, and `TryConvertFilter` does not look inside the wrapper:

```
│ Filters: optional: (id IN (5, 3500)) │   -- dropped
│ Filters: id = 3500                   │   -- sent
```

The extension already unwraps that wrapper in `ExtractFilterExpressionForPath`
(`src/planning/pruning/iceberg_table_filter.cpp:90-101`) for client-side pruning, so client
planning uses these filters while server planning discards them. This does the same before
converting, and maps `COMPARE_IN` to the spec's `in` `SetExpression`. DuckDB wraps
`col = a OR col = b` identically, so that shape starts pruning too.

Files returned on a 4-file table, from the catalog's own accounting:

| query | before | after |
|---|---|---|
| `WHERE id IN (5, 3500)` | 4 | **2** |
| `WHERE msg IN ('…', '…')` | 4 | **2** |
| `WHERE id = 5 OR id = 3500` | 4 | **2** |

Row counts unchanged — they were already correct.

Conversion bails out on a non-constant or NULL member of the set. `NOT IN` is not covered:
DuckDB canonicalises it to `NOT(IN)`, and negating a converted child is only sound when the
child converted *exactly*, which `TryConvertFilter` cannot signal today. `starts-with` is not
added either — `FilterCombiner::TryPushdownPrefixFilter` rewrites prefixes into range
comparisons before the extension sees them, so the code would be unreachable.

The first commit logs the scan-planning POST body. It was the only REST POST not doing so, which
is why the filter a scan sends was not observable from `duckdb_logs()` and this could not be
tested in-repo.

One skip added to `test/configs/fixture-latest.json`:
`catalog_agnostic/delete/partitions/test_delete_manifest_partitions.test` now sends `in` on a
string partition column against a table with a NULL partition, and the pinned fixture answers
HTTP 500 — it is built from apache/iceberg `b0df3ca` (2026-04-29) and predates
[apache/iceberg#17014](https://github.com/apache/iceberg/pull/17014), which added the null guard
to `Evaluator`'s IN/NOT_IN on 2026-07-12. Docker Hub has published no newer fixture tag. Happy to
handle that differently if you would rather.

### DIFF
--- a/src/catalog/rest/api/catalog_api.cpp
+++ b/src/catalog/rest/api/catalog_api.cpp
@@ -74,26 +74,6 @@ vector<string> IRCAPI::ParseSchemaName(const string &namespace_name) {
 	                    int(response.status));
 }
 
-static void LogPostBody(ClientContext &context, const IRCEndpointBuilder &url_builder, const string &body) {
-	if (!Logger::Get(context).ShouldLog(IcebergLogType::NAME, IcebergLogType::LEVEL)) {
-		return;
-	}
-	idx_t truncate_limit = 10000;
-	Value limit_value;
-	if (context.TryGetCurrentSetting("iceberg_logging_post_body_truncate_limit", limit_value)) {
-		truncate_limit = limit_value.GetValue<idx_t>();
-	}
-	string body_to_log;
-	if (truncate_limit == 0) {
-		body_to_log = "<body omitted>";
-	} else if (body.size() > truncate_limit) {
-		body_to_log = body.substr(0, truncate_limit) + "... (truncated)";
-	} else {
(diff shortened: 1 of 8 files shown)


Tishj (approved):
Thanks!

PR #1376, opened 2026-09-09
Improve iceberg_bucket test coverage

tom-kita:
## What

Raises test coverage in `test_iceberg_bucket.test`. Test-only change.

**NULL and return-type checks.** The four NULL cases were the identical query `iceberg_bucket(16, NULL)`, so only one overload was reached. Both sections now cover every supported type, including all four DECIMAL physical widths, which dispatch to separate kernels.

**Spec vectors.** All 12 bucketable types are now covered, up from 5. `num_buckets` = INT32 max exposes the sign-discarded hash itself, since the masked hash cannot exceed the modulus, so each vector asserts the hash from the spec table rather than a reduced bucket. This extends the technique already used for the UUID vector to the rest of the types, and removes the need for a second modulus: a wrong hash cannot reach the expected value by collision. Reduction to smaller bucket counts stays covered by the range section. Also drops a `long 0` vector, whose value comes from the float/double row of the non-bucketable table rather than the long row.

**Nanosecond check.** The previous values (1ns and 1001ns past a microsecond) produce the same microsecond under floor, truncation toward zero and rounding alike, so the section verified nothing about the rounding direction despite its name. It now uses a pre-epoch value where only floor matches:

| | µs | hash |
| --- | --- | --- |
| floor | -2 | 709407621 |
| truncate toward zero | -1 | 1651860712 |
| round to nearest | -1 | 1651860712 |

**Negative decimals.** The minimum big-endian form strips leading `0xFF` for negatives and `0x00` for positives, so the positive `14.20` vector leaves half of that logic unexercised. `-14.20` is added across all four widths.

**Removed.** A determinism check of the form `count(DISTINCT f(v)) = count(DISTINCT f(v))`, which compares one aggregate to itself and therefore held for any return value.

## Verification

Expected values were cross-checked against pyiceberg 0.11.1 rather than against this implementation's output. The file needs no REST catalog:

```
./build/release/test/unittest test/sql/local/catalog_test_config_setup/catalog_agnostic/functions/test_iceberg_bucket.test
```

59 assertions pass, up from 32.

### DIFF
--- a/test/sql/local/catalog_test_config_setup/catalog_agnostic/functions/test_iceberg_bucket.test
+++ b/test/sql/local/catalog_test_config_setup/catalog_agnostic/functions/test_iceberg_bucket.test
@@ -17,27 +17,80 @@ require httpfs
 # Iceberg spec: all transforms return null for null input
 
 query I
-SELECT iceberg_bucket(16, NULL);
+SELECT iceberg_bucket(16, NULL::INTEGER);
 ----
 NULL
 
 query I
-SELECT iceberg_bucket(16, NULL);
+SELECT iceberg_bucket(16, NULL::BIGINT);
 ----
 NULL
 
 query I
-SELECT iceberg_bucket(16, NULL);
+SELECT iceberg_bucket(16, NULL::VARCHAR);
 ----
 NULL
 
 query I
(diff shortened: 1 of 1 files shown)


Tishj:
Thanks!

PR #1383, opened 2026-09-11
Remove iceberg_via_aws_sdk_for_catalog_interactions option, and remove aws-sdk-cpp dependency

carlopi:
Proper responsibility split: credentials to be managed by duckdb-aws extension, catalog interactions via iceberg

`iceberg_via_aws_sdk_for_catalog_interactions` has been a legacy codepath for a while now, now properly removed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,21 +23,10 @@ endif()
 # Resolve dependencies before descending into src/ so that every per-directory OBJECT library
 # inherits the include directories it needs (these commands set directory-scoped properties that
 # are inherited by subdirectories added afterwards).
-if(NOT CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
-    find_package(CURL REQUIRED)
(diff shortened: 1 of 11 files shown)


carlopi:
@Tmonster and @Tishj: I opened this as experiment, let's discuss on Monday (live?).

Tmonster:
@Tishj I think we can just merge even though Lakekeeper is failing? I don't think this will have an effect. Lakekeeper also seems to be down due to minio not being on Docker anymore

Tishj:
Somehow this fails the regression test, already reran it so I don't think this is a fluke.
I imagine this refactor has broken some caching strategy?

Tmonster:
I'm re-running the Regression tests again. I've also tested locally and found no regressions. I think the metadata_late_materialization benchmarks should also introduce some delay similar to the tpch queries. Something like `SET debug_fs_delay_mean_ms=3;`.

Also re-ran regression testing again and it passed, so I think this is g2g

Tishj:
Interesting, I still see the regression failure on the PR

Issue #1382, opened 2026-09-10
allow_moved_paths doesn't rewrite puffin (position delete) file paths in V3 Iceberg tables


eduard-daily:
### What happens?

1. On any catalog
2. When running an `iceberg_scan()` with `allow_moved_paths=true`, the delete vectors paths are not being rebased. Only the absolute original path is being used for delete puffin vectors.

I believe, this is similar issue to the #685 issue, which was fixed in #693 PR. 

### To Reproduce

To reproduce, we would need a V3 Iceberg table with at least 1 data file and at least 1 delete(puffin) file. I generated a table with a Python script (noted below).

I'm putting the table into:
```
/tmp/old_location
```

After that, open DuckDB connection to verify the table: 
```sql
memory D SELECT COUNT(*) FROM iceberg_scan('/tmp/old_location/default/deletion_vectors/metadata/v2.metadata.json');
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│           99 │
└──────────────┘

memory D SELECT * FROM iceberg_metadata('/tmp/old_location/default/deletion_vectors/metadata/v2.metadata.json');
┌────────────────────────────────────┬──────────────────────────┬──────────────────┬─────────┬──────────────────┬───────────────────────────────────┬─────────────┬──────────────┐
│           manifest_path            │ manifest_sequence_number │ manifest_content │ status  │     content      │             file_path             │ file_format │ record_count │
│              varchar               │          int64           │     varchar      │ varchar │     varchar      │              varchar              │   varchar   │    int64     │
├────────────────────────────────────┼──────────────────────────┼──────────────────┼─────────┼──────────────────┼───────────────────────────────────┼─────────────┼──────────────┤
│ /tmp/old_location/default/deletion │                        1 │ DATA             │ ADDED   │ EXISTING         │ /tmp/old_location/default/deletio │ PARQUET     │          100 │
│ _vectors/metadata/63d69540-eb5e-49 │                          │                  │         │                  │ n_vectors/data/00000-0-35a34621-5 │             │              │
│ 7c-a06d-745c0219dd1e-m0.avro       │                          │                  │         │                  │ c63-4498-8dd2-b61cb03f0740-0-0000 │             │              │
│                                    │                          │                  │         │                  │ 1.parquet                         │             │              │
├────────────────────────────────────┼──────────────────────────┼──────────────────┼─────────┼──────────────────┼───────────────────────────────────┼─────────────┼──────────────┤
[...]

Issue #685, opened 2026-02-05
allow_moved_paths doesn't rewrite position delete file paths


troyfendall:
## Summary

When using `iceberg_scan()` with `allow_moved_paths=true` on a moved/copied Iceberg table, data file paths are correctly rewritten to the new location, but **position delete file paths (content=1) are not rewritten** and still reference the original location.

## Environment

- DuckDB: v1.4.4
- Iceberg extension: latest (force installed)
- Iceberg table format: v2 with position deletes

## Steps to Reproduce

1. Download and extract the attached reproduction zip
2. Run:

```bash
cd duckdb-iceberg-issue

duckdb -c "
INSTALL iceberg; LOAD iceberg;
SELECT count(*) FROM iceberg_scan('./table', allow_moved_paths=true);
"
```

## Expected Behavior

The query should succeed, with `allow_moved_paths=true` rewriting all file paths (both data files and delete files) to be relative to `./table`.

## Actual Behavior

```
HTTP Error: HTTP GET error on 'https://original-bucket.s3.amazonaws.com/warehouse/products/data/1DPzzQ/...' (HTTP 403)
```

DuckDB attempts to fetch the position delete file from the original S3 path stored in the metadata, instead of rewriting it to the local path.

## Analysis

- **Data files (content=0)**: Paths are correctly rewritten by `allow_moved_paths`
- **Position delete files (content=1)**: Paths are NOT rewritten, causing 403 errors

## Proof that it works with correct paths

The reproduction includes a `table-fixed/` directory with manually corrected paths:

```sql
SELECT count(*) FROM iceberg_scan('./table-fixed');
-- Returns: 13136 (with deletes correctly applied)
```

## Reproduction

See attached zip file containing:
- `table/` - Broken table with S3 paths in metadata
- `table-fixed/` - Working table with local paths manually corrected
- `README.md` - Detailed documentation

troyfendall:
[duckdb-iceberg-allow-moved-paths-bug.zip](https://github.com/user-attachments/files/25083991/duckdb-iceberg-allow-moved-paths-bug.zip)

Here's a folder that reproduces the issue

PR #1384, opened 2026-09-13
Apply allow_moved_paths to puffin delete vectors

eduard-daily:
Fix for issue #1382 (similar to issue https://github.com/duckdb/duckdb-iceberg/issues/685, which was fixed in https://github.com/duckdb/duckdb-iceberg/pull/693 PR).

Problem: reading a relocated Iceberg v3 table with `allow_moved_paths = true` fails when the table's snapshot has puffin deletion vector.

Fix: rebase the puffin path before opening it, in the same way as the parquet delete branch.

Added a test case `moved_deletion_vector_path.test` with a v3 table, that has 1 data file and 1 deletion file. The test scans the table with `allow_moved_paths = true` and asserts both the row count and that the deleted row is gone.


Tishj (approved):
Thanks!

Tishj:
Thanks!

PR #1387, opened 2026-09-14
Replace MinIO with SeaweedFS 

NiclasHaderer:
MinIO's community edition is archived and `minio/minio` is gone from Docker Hub. This swaps the fixture's object store for SeaweedFS and replaces the mc bootstrap container with aws-cli, and moves the Lakekeeper catalog off MinIO the same way by bumping its pinned upstream commit to one where they already use SeaweedFS.


Tishj:
Interesting, is this not a long term viable solution then? https://github.com/duckdb/duckdb-iceberg/pull/1386

NiclasHaderer:
I dont think so. https://github.com/minio/mc has been archived and not been updated in 10 months

NiclasHaderer:
Same goes for the main https://github.com/minio/minio

Tishj:
Thanks!
We'd like to postpone this fix until after v1.5.6, using the quay.io workaround for now, just because it's less likely to have unforeseen issues

NiclasHaderer:
I personally feel like this is not a high risk change, since it only touches the testing infra, but happy to retarget to main, or should I just wait until 1.5.6 was released?

Tishj:
A random CI failure might still impact release candidate building afaik, but yea, if you could retarget this to `main` that would be great 👍

NiclasHaderer:
I regtargetted to main. I also bumped the Polaris catalog to their respective replacement which is RustFS.  
Nessie I had to patch, but I think it is worth it. quay.io has been pretty unstable for us in the last 24h since everyone moved to it after the minio image was pulled. We've been seeing 502 Bad Gateway multiple times.

Tishj:
I can't really place the Polaris issues, I assume they're related to this being a newer Polaris version.
`test/sql/local/catalog_test_config_setup/catalog_agnostic/partitioning/hour/hour_nested_struct_timestamp.test` for example looks like it's a credential-scoping problem, where somehow it doesn't have access to the bucket.

We've fixed something maybe similar for #1194 where we were making an additional request for the metadata.json file which was outside of the scope of the credentials given. Though this looks like the created credentials don't get scoped correctly making even requests to `s3://bucket123/default/partition_hour_timestamp_nested_struct_field/data/` fail.

Tishj:
I'll have a look, we might want to separate that fix into its own PR

EDIT:
The test is wrong, we shouldn't be listing files on S3 with those credentials, Polaris hasn't given us authorization to do that listing with the credentials we were given.

Still looking into the other failure, but I suspect that Polaris doesn't give us permissions to remove files, so we just have to skip the test on Polaris now.

Tishj:
I suspect that should fix all 4 of the test failures, hopefully no others pop up

NiclasHaderer:
ill take a look at the next failing tests, except if you want to

Tishj:
Thanks!

PR #1393, opened 2026-09-16
Sign and send X-Iceberg-Access-Delegation on the SigV4 path

fyang623:
## Problem

`AWSInput::ExecuteRequest` builds its outbound header set from scratch — `host`,
`x-amz-date`, `x-amz-content-sha256`, `x-amz-security-token` and `Content-Type` —
so every other header in the `HTTPHeaders` argument is dropped before the request
is signed and sent.

`X-Iceberg-Access-Delegation` is one of those headers. It is dropped whether it
arrives from the `extra_http_headers` ATTACH option (parsed and merged by
`SIGV4Authorization::Request`) or from the `IRCAPI` call sites that insert
`vended-credentials`. So on a SigV4 catalog, `extra_http_headers` parses,
validates, and has no effect on the wire.

## Why it matters

Against an AWS Glue REST catalog whose tables are registered with Lake Formation,
an **absent** delegation header is read as a request to vend scoped credentials, which
Lake Formation refuses for a table carrying column or row restrictions:

```
403  Insufficient Lake Formation permission(s). Unable to fetch credentials for table: <table>
```

so Glue `LoadTable` fails before any data is read. Sending the header with value
`none` asks for metadata only and succeeds. Measured against a Glue catalog with
a column-restricted grant (3-column table, 2 columns granted):

| `X-Iceberg-Access-Delegation` | result |
|---|---|
| `none` | 200, schema pruned to the granted columns |
| `vended-credentials` | 403, cannot fetch credentials |
| absent | 403, cannot fetch credentials |

## Change

Add the header to `signed_headers`, to the canonical request, and to the outbound
headers together — all three must agree or the signature will not match.

The change is deliberately narrow and purely additive. It cannot change the request
for any catalog that does not send this header.

## Testing

The signing algorithm is confirmed against live Glue with an equivalent
hand-rolled SigV4 signer (same sorted `SignedHeaders` + canonical block): with the
header, `LoadTable` returns 200 and a column-pruned schema; without it, 403.


Tishj:
Thanks!

PR #1399, opened 2026-09-17
Disable external file cache and add delay to metadata regression tests

Tmonster:
Add a delay to some of the metadata regression tests. 

These regression tests were flaky and I think it was because the external file cache was enabled so tests would run under <0.01s (at least locally) and any kind of regression measuring at that significance is really hard. 

This disabled external file cache and adds a httfs delay. Should stabilize the benchmark


Tishj:
I'll let you decide whether to merge this
Since we're adding a delay, I would expect a regression failure? The fact that it's not failing is suspicious

Issue #1359, opened 2026-08-27
Iceberg query results show NULL for nested structs when using attached catalog but works fine with iceberg_scan()


xin-luo-dragon:
### What happens?

On an S3 / Glue Catalog, when querying a table with a struct field, the values show up as NULL.
If querying the same struct field with iceberg_scan('s3://bucket/path/to/table/metadata/abc.metadata.json'), the column returns correctly.
Additionally, when using PyIceberg, querying from the S3/Glue catalog returns the column correctly as well.

### To Reproduce

1. Create an Iceberg table on S3/Glue with a regular string column and a struct column.
2. Prepopulate the table with valid values.
3. Attach the Iceberg table on duckdb: ```
    CREATE OR REPLACE SECRET (
         TYPE s3,
         PROVIDER credential_chain,
         CHAIN 'config',
         REGION '{region_name}',
         USE_SSL true
    );
    ATTACH '{account_id}' AS glue (
        TYPE iceberg,
        ENDPOINT 'glue.ap-east-1.amazonaws.com/iceberg',
        AUTHORIZATION_Type 'SIGV4'
    );
    USE glue.iceberg_db;
```
4. Query the table ```
SELECT str_column, struct_column FROM glue.iceberg_db.sample_table
```
5. Observe that columns inside struct_columns are all NULL.
6. Compare against the same results with iceberg_scan() which has said columns populated normally
```
SELECT str_column, struct_column FROM iceberg_scan('s3://bucket/path/to/table/metadata/current-table-id.metadata.json');
```

### OS:

mac_os, linux x86_64

### DuckDB Version:

1.5.5

### DuckDB Client:

CLI, Python

### Full Name:

Xin Luo

### Affiliation:

Mox Bank

### Did you include all relevant data sets for reproducing the issue?

No - I cannot share the data sets because they are confidential

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

xin-luo-dragon:
[...]

PR #1403, opened 2026-09-18
REST catalog: add 'use_metadata_location' attach option to load authoritative table metadata

u70b3:
## Problem

When querying a table through an ATTACH'ed catalog, DuckDB initializes the table from the `metadata` embedded in the REST `LoadTable` response. Some catalogs do not return the contents of the table's metadata file there — AWS Glue's Iceberg REST endpoint reconstructs the table metadata from its own catalog state (independently documented with SigV4 captures in [malon64/floe#433](https://github.com/malon64/floe/issues/433), where the reconstructed response even lacked `schemas` entirely).

For tables registered through Hive-style Glue APIs (e.g. Spark's `GlueCatalog`), Glue's `StorageDescriptor` columns carry no Iceberg field ids for nested fields, so the reconstructed schema's nested struct field ids don't match the field ids written in the parquet files. DuckDB maps file columns strictly by field id (`BY_FIELD_ID`), so every struct child misses and falls back to its (NULL) default — nested structs silently read as NULL, while top-level columns keep working. `iceberg_scan()` and PyIceberg read the actual metadata file, which is why they are unaffected (#1359, Spark 3.5.2 vs. Athena CTAS split included).

## Fix

New ATTACH option `USE_METADATA_LOCATION` (default **false**). When enabled and the `LoadTable` response carries `metadata-location`, the catalog loads the authoritative metadata file from there instead of trusting the embedded metadata. If the file cannot be read (permissions, GC'd file), we log and fall back to the embedded metadata — i.e. previous behavior.

The option defaults to false because the IRC spec requires the embedded metadata to match the metadata file: deviating is only justified for catalogs known to reconstruct the response, and default-on would add a file read to table resolution on spec-compliant catalogs for no benefit.

When the option is on, loaded metadata files are cached per catalog, keyed by `metadata-location`. Metadata files are immutable (a commit writes a new file), so an unchanged table costs no extra file requests on refresh; the only added traffic is one metadata file read when a table's metadata location actually changes.

## Testing

- `scripts/reconstructed_metadata_mock_server.py` mimics Glue's reconstruction (reassigned nested field ids, `schema.name-mapping.default` dropped) over a checked-in pyiceberg-generated table. The test asserts both behaviors: default ATTACH reads struct children as NULL (spec-compliant trust of the embedded metadata), ATTACH with `USE_METADATA_LOCATION true` matches `iceberg_scan()`. Wired into CI via a reusable workflow modeled on the SigV4 mock one.
- Request-counting tests (`test_minimal_head_requests`, `test_max_table_staleness`, `test_table_information_requests`) are unchanged: with the option off by default, no extra requests are issued on any catalog.

## Notes for reviewers

- I don't have Glue access; the repro simulates Glue's reconstruction behavior based on the evidence linked above. Verification steps against a real Glue catalog are in the review thread; happy to add a `test/sql/cloud/glue/` test if a Hive-registered table is made available in the test account.
- `iceberg_load_table_response()` intentionally still returns the raw catalog response.
- The metadata file cache (option on) has no eviction bound beyond the catalog lifetime; happy to add one if preferred.


Tmonster:
@u70b3 do you have an example of how to reproduce this error in glue/AWS? 

I have a hard time wrapping my head around the fact that the Glue IRC endpoints don't serve valid metadata.json. I can verify that the Glue::GetTable API doesn't return the correct schema information, but that API is not hit via this extension, and I would expect many more failures if that was how this extension was used. 

The IRC spec also says that the `LoadTableInformation` should [match the contents of the](https://github.com/apache/iceberg/blob/778d103c0fa802be5e85e4cb2e4e4b4bcd8ac891/open-api/rest-catalog-open-api.yaml#L1047) `metadata.json` file, so I don't love the fact that we now default to reading the `metadata.json` file for **all** catalogs, even the ones that do follow the spec

I would like to see this behind an attach option because it deviates from the IRC spec, and also steps for me to verify this using my own aws account. Potentially this is a case that can be included in our cloud tests as well, since I don't love the idea of adding an extra mock just for these tests.

u70b3:
The two fixture failures were caused by this change itself, not pre-existing: resolving a table now fetches the authoritative `.metadata.json` from `metadata-location` (one HEAD + one GET), and both `test_max_table_staleness.test` and `test_table_information_requests.test` assert the exact set of HTTP requests. The extra rows in the actual output were precisely those metadata file requests, and main's fixture job is green — my earlier claim that these reproduce on main was wrong (my local fixture stack had a broken S3 credential, so the comparison run never got past the first scan).

Fixed in 11b6a1f by excluding `%.metadata.json` requests from those assertions, mirroring what `test_minimal_head_requests.test` already does. The catalog-side assertions (LoadTable calls, staleness/refresh behavior) are unchanged.

Tmonster:
> The two fixture failures were caused by this change itself, not pre-existing: resolving a table now fetches the authoritative `.metadata.json` from `metadata-location` (one HEAD + one GET), and both `test_max_table_staleness.test` and `test_table_information_requests.test` assert the exact set of HTTP requests. The extra rows in the actual output were precisely those metadata file requests, and main's fixture job is green — my earlier claim that these reproduce on main was wrong (my local fixture stack had a broken S3 credential, so the comparison run never got past the first scan).
> 
> Fixed in [11b6a1f](https://github.com/duckdb/duckdb-iceberg/commit/11b6a1f7f5caf0aa7fbd7377a30752d72440a7de) by excluding `%.metadata.json` requests from those assertions, mirroring what `test_minimal_head_requests.test` already does. The catalog-side assertions (LoadTable calls, staleness/refresh behavior) are unchanged.

@u70b3 this is not a response to my question. I would appreciate some thoughts by you here. The response feels LLM generated, and that is not a productive way to collaborate and solve an actual issue

u70b3:
my previous comment was only documenting the CI fix; I never actually responded to your questions. sorry for delay reply

u70b3:
On the main question — whether Glue's IRC endpoint can return metadata that doesn't match the metadata file. The sharpest evidence I have is the SigV4 capture in malon64/floe#433: the LoadTable response there contains no `schemas` field at all. That cannot be the contents of a metadata file (a metadata file always has schemas), so the endpoint must have synthesized the response from something else — plausibly the same Glue catalog state you inspected via Glue::GetTable, which, as you verified, does not carry Iceberg field ids for nested fields. I can't confirm this against a live Glue catalog myself (no AWS access), so this inference is exactly what I'd like your help verifying.

On "I would expect many more failures": I think this failure is narrow and silent by construction. It only bites clients that trust the embedded response metadata — Spark, PyIceberg, Athena and iceberg_scan all read the metadata file directly and are unaffected, which matches #1359, where only the ATTACH path failed. It only bites tables registered through Hive-style Glue APIs; Athena-created tables are fine. It only bites nested columns — top-level columns keep working. And it fails as silent NULLs rather than an error, so affected users have no signal that points back to the catalog response. The absence of widespread reports doesn't seem like evidence against it to me.

The attach option you asked for is implemented and pushed (d551c6be): USE_METADATA_LOCATION, default false. The default path is exactly the spec behavior — embedded metadata trusted, no extra file read on any catalog. You were right that always-on was the wrong default; it was also what broke the Gravitino/Lakekeeper request-count tests, and those test changes are reverted now.

To see the response-vs-file difference on your own account, two queries are enough for any table registered through Hive-style Glue APIs (e.g. Spark's GlueCatalog) that has a nested struct column:

```sql
SELECT metadata_location FROM iceberg_load_table_response('my_datalake.db.t');
SELECT metadata."schemas" FROM iceberg_load_table_response('my_datalake.db.t');
```

Compare the embedded schema's nested field ids against the file at metadata-location (iceberg_scan or pyiceberg reads it). Reading the table via ATTACH with and without USE_METADATA_LOCATION then shows the behavioral difference directly.

Happy to add a cloud test under test/sql/cloud/glue/ if such a Hive-registered table can be made available in the test account — I can't create one myself. The mock server exists only because this behavior is otherwise untestable without AWS; I'd gladly drop it once a cloud test covers the scenario.

src/include/iceberg_attach.hpp
```diff
@@ -41,6 +41,11 @@ struct IcebergAttachOptions {
 	// some catalogs (e.g. AWS Glue) do not assign a table location server-side; derive one from the namespace's
 	// 'location' property
 	bool default_table_location_from_namespace = false;
+	// load the authoritative table metadata from the LoadTableResult's 'metadata-location' instead of trusting the
+	// metadata embedded in the response. Needed for catalogs (e.g. AWS Glue's Iceberg REST endpoint) that
+	// reconstruct the response metadata from their own catalog state, which can lose nested field ids. The IRC
```
Tmonster: comment is too verbose. Potentially just the first line is needed and a link to this PR number or an issue number

test/sql/local/catalog_custom_setup/reconstructed_metadata_mock/attach_uses_metadata_location.test
```diff
@@ -0,0 +1,63 @@
+# name: test/sql/local/catalog_custom_setup/reconstructed_metadata_mock/attach_uses_metadata_location.test
+# description: ATTACH with 'use_metadata_location' reads the authoritative metadata file at 'metadata-location' when the catalog response contains reconstructed metadata (duckdb-iceberg#1359)
+# group: [reconstructed_metadata_mock]
+
+# Requires: python3 scripts/reconstructed_metadata_mock_server.py 19131 (run from the repository root)
+require-env RECONSTRUCTED_METADATA_MOCK_SERVER_AVAILABLE
```
Tmonster: I'm pretty sure the rest-fixture will also return the metadata_location,so you don't need to run a reconstructed metadata mock server for this

test/sql/local/catalog_custom_setup/reconstructed_metadata_mock/attach_uses_metadata_location.test
```diff
@@ -0,0 +1,63 @@
...
+DETACH reconstructed_embedded;
+
+# With 'use_metadata_location', the authoritative metadata file at
+# 'metadata-location' is loaded instead
+statement ok
+ATTACH '' AS reconstructed_catalog (TYPE ICEBERG, URI 'http://127.0.0.1:19131', AUTHORIZATION_TYPE 'none', USE_METADATA_LOCATION true);
+
+# Struct children must read back their values, not NULL
+query III
+SELECT id, s.a, s.b FROM reconstructed_catalog.db.t ORDER BY id;
```
Tmonster: The test here just needs to read a table and then verify that the table metadata.json is read, which can be done with `from duckdb_logs_parsed('HTTP') where request.url like '%metadata.json'`

Tmonster (commented):
Thanks, I also think you can remove the `data/persistent/reconstructed` table as well now

src/include/iceberg_attach.hpp
u70b3: Shortened to one line plus the PR link.

test/sql/local/catalog_custom_setup/reconstructed_metadata_mock/attach_uses_metadata_location.test
u70b3: Confirmed, the rest-fixture does return metadata-location. Mock server, workflow, mock test and the checked-in table are all removed; this coverage now lives in test/sql/local/catalog_custom_setup/fixture/test_use_metadata_location.test.

test/sql/local/catalog_custom_setup/reconstructed_metadata_mock/attach_uses_metadata_location.test
u70b3: Done in the fixture test with exactly this assertion (GET count on '%.metadata.json%' after a scan). It also covers the cache (re-scan issues no new request) and the default attach (no metadata file request at all).

u70b3:
Removed the checked-in reconstructed table as well — the PR is now just the option, the gated load, and one fixture test.

u70b3:
https://github.com/duckdb/duckdb-iceberg/issues/1359#issuecomment-5769742109 — someone reproduced this on a real Glue catalog and posted the diff. Their case is the other variant: field ids all survive, but Glue strips `schema.name-mapping.default` down to the top-level entries (856 vs 3385 chars), so id-less parquet files read back NULL the same way. Reading the metadata file covers both.

PR #1404, opened 2026-09-18
Iceberg view support

nicosuave:
Based on https://github.com/duckdb/duckdb-iceberg/pull/887 by @LucaSoato

I adapted the code to current main, fixed transaction isolation & foreign view solution. Also added test coverage via existing catalog & Spark integration suites.


Tmonster:
Thanks! 
Can you take a look at the failing fixture tests? Seems like spark is having a problem creating views? Or potentially something is wrong with how spark gets initialized

Tishj:
Thanks, here's some feedback:

1. We should not be supporting non-duckdb dialects without an opt-in setting, and I'd add that in a follow up. For now only accept `duckdb` dialect.

2. The binding/qualifying logic seems extremely brittle and a constant source of headaches, can we please find a different way to achieve this? Perhaps we can make use of replacement scans because that's essentially what this is.

2.1 It seems that this qualifying is only necessary to support views that specify a different default catalog and/or namespace. Let's not support those initially to cut out all of this qualifying logic in this initial PR.

nicosuave:
This should be in a good spot again for review @Tishj 

Was passing CI, then I rebased to resolve a conflict

src/catalog/rest/api/catalog_api.cpp
```diff
@@ -313,9 +313,13 @@ IcebergListTablesResult IcebergListTablesRequest::Execute(ClientContext &context
...
 		auto response = catalog.auth_handler->Request(RequestType::GET_REQUEST, context, url_builder, headers);
 		if (!response->Success()) {
+			if (response->status == HTTPStatusCode::NotFound_404) {
+				// A missing namespace has no entries. Native view binding may probe
+				// the catalog's default namespace before trying replacement scans.
+				return vector<rest_api_objects::TableIdentifier>();
+			}
 			if (response->status == HTTPStatusCode::Forbidden_403 ||
-			    response->status == HTTPStatusCode::Unauthorized_401 ||
-			    response->status == HTTPStatusCode::NotFound_404) {
```
Tishj, line 318: I don't think this was placed here without good reason, I'm not sure about removing it @Tmonster

Tmonster: Different catalogs respond with different error codes when attempting to list a namespace that doesn't exist / you don't have access too. Throwing will interrupt the listing process which we don't want. 
Seems like the catch for a 404 was moved, but I'm not sure returning an empty vector of TableIdentifers is correct instead of a nullopt?

PR #1405, opened 2026-09-18
bump duckdb and apply patches

Tishj:


Tishj:
Made https://github.com/duckdb/duckdb/pull/25908 in core to fix the test failures

PR #1407, opened 2026-09-18
Omit unknown columns and struct fields from data files

### LINKED ISSUE #1406: Columns and struct fields of type `unknown` are written to data files
## What the spec says

`unknown` has no physical Parquet representation. Writers leave it out of the data file, and readers produce nulls for it.

- [Appendix A, Parquet](https://iceberg.apache.org/spec/#parquet): physical type "None", "Omit from data files", and "When reading an `unknown` column, any corresponding column must be ignored and replaced with `null` values"
- [Primitive Types](https://iceberg.apache.org/spec/#primitive-types): "Must be optional with `null` defaults; not stored in data files"

## Problem

DuckDB writes such a column as an all-null `INT32` column carrying the column's field id, both at the top level and for a field nested inside a struct. Data files written by DuckDB therefore cannot be read by engines that follow the spec.

### Environment

- DuckDB v2.0.0-alpha42069 with the released `iceberg` extension, Glue catalog (`ENDPOINT_TYPE 'glue'`)
- EMR with Spark 4.1.1-amzn-0 and Iceberg 1.11.0-amzn-0, same Glue catalog and warehouse

### Case 1: top-level column

Spark creates the table and writes the first row:

```python
spark.sql("""
CREATE TABLE my_catalog.duckdb_unknown_test.test (
    id INT,
    data VOID
) USING iceberg
TBLPROPERTIES ('format-version'='3');
""")
spark.sql("INSERT INTO my_catalog.duckdb_unknown_test.test VALUES (1, null)")
```

The table metadata records the column as `unknown`:

```json
"schemas":[{"type":"struct","schema-id":0,"fields":[{"id":1,"name":"id","required":false,"type":"int"},{"id":2,"name":"data","required":false,"type":"unknown"}]}]
```

The file Spark wrote has no column for it:

```python
>>> import pyarrow.parquet as pq
>>> metadata = pq.read_metadata("00000-0-0e95c5e7-7894-42f9-9438-9d1660047cb0-0-00001.parquet")
>>> parquet_schema = metadata.schema
>>> print(parquet_schema)
<pyarrow._parquet.ParquetSchema object at 0xffff8b6628c0>
required group field_id=-1 table {
  required int32 field_id=1 id;
}
```

DuckDB reads that table correctly:

```
D SELECT * FROM glue_catalog.duckdb_unknown_test.test;
┌───────┬────────┐
│  id   │  data  │
│ int32 │ "null" │
├───────┼────────┤
│     1 │ NULL   │
└───────┴────────┘
```

But the file DuckDB writes contains the column:

```
D INSERT INTO glue_catalog.duckdb_unknown_test.test VALUES (2, null);
```

```python
>>> metadata = pq.read_metadata("01a0b28e-5600-7ef6-97f5-8efecd98122a.parquet")
>>> parquet_schema = metadata.schema
>>> print(parquet_schema)
<pyarrow._parquet.ParquetSchema object at 0xffff9810d880>
required group field_id=-1 duckdb_schema {
  optional int32 field_id=1 id (Int(bitWidth=32, isSigned=true));
  optional int32 field_id=2 data (Null);
}
```

Spark can no longer read the table:

```
>>> spark.sql("SELECT * FROM my_catalog.duckdb_unknown_test.test").show()
java.lang.UnsupportedOperationException: Unsupported primitive type: unknown
        at org.apache.iceberg.arrow.ArrowSchemaUtil$IcebergToArrowTypeConverter.primitive(ArrowSchemaUtil.java:184)
        at org.apache.iceberg.arrow.ArrowSchemaUtil$IcebergToArrowTypeConverter.primitive(ArrowSchemaUtil.java:68)
        at org.apache.iceberg.types.TypeUtil.visit(TypeUtil.java:795)
        at org.apache.iceberg.arrow.ArrowSchemaUtil.convert(ArrowSchemaUtil.java:65)
        at org.apache.iceberg.arrow.vectorized.VectorizedArrowReader.allocateFieldVector(VectorizedArrowReader.java:287)
        at org.apache.iceberg.arrow.vectorized.VectorizedArrowReader.read(VectorizedArrowReader.java:203)
        at org.apache.iceberg.spark.data.vectorized.ColumnarBatchReader$ColumnBatchLoader.readDataToColumnVectors(ColumnarBatchReader.java:150)
      	...
```

### Case 2: field of a struct

```python
spark.sql("""
CREATE TABLE my_catalog.duckdb_unknown_test.test2 (
    id INT,
    data struct<a: VOID, b: INT>
) USING iceberg
TBLPROPERTIES ('format-version'='3');
""")
spark.sql("""INSERT INTO my_catalog.duckdb_unknown_test.test2 VALUES (1, named_struct("a", null, "b", 1))""")
```

Spark omits the field and keeps the rest of the struct (note the absent field id 3):

```python
>>> import pyarrow.parquet as pq
>>> metadata = pq.read_metadata("01a0b2a0-02b6-722c-801f-c5403dd38585.parquet")
>>> parquet_schema = metadata.schema
>>> print(parquet_schema)
<pyarrow._parquet.ParquetSchema object at 0xffff92c1f3c0>
required group field_id=-1 table {
  required int32 field_id=1 id;
  required group field_id=2 data {
    required int32 field_id=4 b;
  }
}
```

DuckDB writes the field:

```
D INSERT INTO glue_catalog.duckdb_unknown_test.test2 VALUES (2, {"a":null, "b":1});
```

```python
>>> metadata = pq.read_metadata("01a0b2a0-02b6-722c-801f-c5403dd38585.parquet")
>>> parquet_schema = metadata.schema
>>> print(parquet_schema)
<pyarrow._parquet.ParquetSchema object at 0xffff9cfefe80>
required group field_id=-1 duckdb_schema {
  optional int32 field_id=1 id (Int(bitWidth=32, isSigned=true));
  optional group field_id=2 data {
    optional int32 field_id=3 a (Null);
    optional int32 field_id=4 b (Int(bitWidth=32, isSigned=true));
  }
}
```

Spark fails the same way, now through the struct reader:

```
>>> spark.sql("SELECT * FROM my_catalog.duckdb_unknown_test.test2").show()
java.lang.UnsupportedOperationException: Unsupported primitive type: unknown
        at org.apache.iceberg.arrow.ArrowSchemaUtil$IcebergToArrowTypeConverter.primitive(ArrowSchemaUtil.java:184)
        at org.apache.iceberg.arrow.ArrowSchemaUtil$IcebergToArrowTypeConverter.primitive(ArrowSchemaUtil.java:68)
        at org.apache.iceberg.types.TypeUtil.visit(TypeUtil.java:795)
        at org.apache.iceberg.arrow.ArrowSchemaUtil.convert(ArrowSchemaUtil.java:65)
        at org.apache.iceberg.arrow.vectorized.VectorizedArrowReader.allocateFieldVector(VectorizedArrowReader.java:287)
        at org.apache.iceberg.arrow.vectorized.VectorizedArrowReader.read(VectorizedArrowReader.java:203)
        at org.apache.iceberg.arrow.vectorized.VectorizedArrowReader.read(VectorizedArrowReader.java:64)
        at org.apache.iceberg.arrow.vectorized.VectorizedStructReader.read(VectorizedStructReader.java:131)
        at org.apache.iceberg.spark.data.vectorized.ColumnarBatchReader$ColumnBatchLoader.readDataToColumnVectors(ColumnarBatchReader.java:150)
	      ...
```

### OS:

macOS Tahoe 26.6.2, aarch

### DuckDB Version:

v2.0.0-alpha42069

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Tomoya Kitazato

### Affiliation:

Amazon Web Services Japan

- [x] Yes, I have

- [x] Yes, I have

tom-kita:
Fixes #1406.

## What the spec requires

`unknown` has no physical Parquet representation. Appendix A lists its physical type as "None" with "Omit from data files", and requires a reader to ignore any corresponding column and replace it with `null`. The primitive type table records that an unknown column is "not stored in data files".

We wrote such a column as an all-null `INT32` column carrying its field id, both at the top level and for a field nested in a struct. A reader that resolves columns by field id then builds a reader for a type that has no physical representation, so data files we write cannot be read by a spec-compliant engine (see #1406 for the failure against Spark 4.1 / Iceberg 1.11 on a Glue catalog).

## What this PR does

Two commits, the first of which does not change behaviour:

1. **`refactor: Build the iceberg write layout once`** — `GetCopyOptions` described the chunk that reaches the copy operator in four places (the parquet bind names and types, the `FIELD_IDS` option, the partition column positions, and the partition transform projection), each deriving its positions independently. They are now derived from a single `IcebergWriteLayout`, which is what makes omitting a column possible without the four disagreeing. The sort is planned before the projection, since sort expressions reference the child plan output. `IcebergCopyInput::get_table_index`, which nothing has ever set, is dropped along with the branch it guarded.

2. **`fix: Omit unknown columns and struct fields from data files`** — the written type of a column is derived from its logical type with the unknown fields of nested structs dropped, and the value is cast onto that type, which drops the fields the target type does not name. `FIELD_IDS` follows the same rule, so it never names a field the written type lacks.

Nothing in DuckDB's parquet writer changes; the columns and field ids we hand it do.

## Behaviour changes

- Data files no longer contain a column for an `unknown` column, or a field for an `unknown` struct field. Reads are unaffected: the column or field comes back as `null`, which is what the spec asks a reader to do. A table can hold files written before and after this change; both read identically.
- A struct whose fields are all `unknown` is omitted entirely, because a Parquet group needs at least one field. Such a struct reads back as `null` rather than a struct of nulls.
- `PARTITIONED BY` an `unknown` column keeps working, now through a computed partition value, since the column it would route by is no longer in the file. The partition value is always `null`, as the spec prescribes for a transform of a `null` input.
- A write that would leave no columns at all is rejected with a clear error: a table whose columns are all `unknown`, and a list element or map key/value of type `unknown`, which Parquet cannot express once the element type is omitted.

## Tests

`insert/test_unknown_type_omitted_from_data_files.test` covers the written file contents and the read-back values for: an unpartitioned table, identity partitioning on a later column, a transformed partition, identity partitioning on the unknown column itself, `UPDATE` (which adds `_row_id`), a sort order, an unknown first column, `CREATE TABLE AS`, a table of only unknown columns, nested structs (including one nested two levels deep, one inside a list, one inside a map, and one whose only field is unknown), and `CREATE TABLE AS` with a nested unknown field.

The test is skipped for Lakekeeper and Nessie for the same reasons as the existing `create/test_create_unknown_type.test`, and for S3 Tables, which rejects a table containing an `unknown` column.

## Verification

Built and run against the local `fixture` catalog in both release and debug configurations; `make format-check` passes. `create`, `insert`, `sorting`, `merge`, `update`, `delete`, `maintenance`, `alter` and `other_engines` show the same failures before and after the change, all of them from test data this environment does not generate.

Also confirmed against a Glue catalog with Spark 4.1 / Iceberg 1.11 that a table written by this branch reads back correctly from Spark, both for a top-level `unknown` column and for an `unknown` field inside a struct: the failure reported in #1406 no longer occurs.

## Not covered

- An `unknown` field nested in a list element or map key/value is rejected rather than supported, matching what the spec can express in Parquet.
- `iceberg_rewrite_data_files` does not support V3 tables, so compaction cannot be exercised for `unknown` yet; it resolves its copy options through the same path and needs no separate change.


src/execution/operator/iceberg_insert.cpp
```diff
@@ -130,15 +130,96 @@ static string GetColumnNameBySourceId(const IcebergTableSchema &schema, idx_t so
 	return schema.GetColumnByFieldId(source_id).name;
 }
 
-//! Check if all partition fields use identity transforms
+//! SPEC (Appendix A): `unknown` has no physical Parquet type and must be omitted from data files;
+//! readers replace the column with nulls. Iceberg maps `unknown` to SQLNULL, so the type says which fields
+//! are omitted. A struct whose fields are all omitted has no representation either, because a Parquet group
+//! needs at least one field.
+static bool IsWrittenToDataFile(const LogicalType &type) {
```
Tishj: Use `TypeVisitor::Contains` instead please 👍

src/execution/operator/iceberg_insert.cpp
```diff
@@ -517,71 +574,158 @@ static void GenerateSortOrderExpressions(ClientContext &context, const IcebergCo
...
-		// Just set up the correct references to the partition columns
-		for (auto &field : spec.fields) {
-			if (field.transform.Type() == IcebergTransformType::VOID) {
-				continue;
-			}
-			auto col_idx = GetColumnIndexBySourceId(copy_input.schema, field.source_id).GetPrimaryIndex();
-			partition_columns.push_back(col_idx);
+//! The layout of the chunk that reaches the copy operator. Built once, so that the copy bind
+//! (names/types/FIELD_IDS), the projection and the partition column positions cannot disagree.
+struct IcebergWriteLayout {
```
Tishj, line 596: Let's wrap these structs (both `IcebergWriteLayout` and `IcebergWriteColumn`) in an anonymous namespace just to be safe, i.e:

```c++
namespace {
<body>
} // namespace
```

src/execution/operator/iceberg_insert.cpp
```diff
@@ -517,71 +574,158 @@ static void GenerateSortOrderExpressions(ClientContext &context, const IcebergCo
...
+		write_column.source = make_uniq<BoundReferenceExpression>(LogicalType::BIGINT, child_idx++);
+		write_column.field_id = Value::BIGINT(MultiFileReader::ROW_ID_FIELD_ID);
+		layout.columns.push_back(std::move(write_column));
 	}
+	if (WriteSequenceNumber(copy_input.virtual_columns)) {
+		IcebergWriteColumn write_column;
+		write_column.name = "_last_updated_sequence_number";
+		write_column.type = LogicalType::BIGINT;
+		write_column.source = make_uniq<BoundReferenceExpression>(LogicalType::BIGINT, child_idx++);
+		layout.columns.push_back(std::move(write_column));
```
Tishj, line 692: Shouldn't this also set the `field_id`?

src/execution/operator/iceberg_insert.cpp
tom-kita: Done in [b0d7dcf](https://github.com/duckdb/duckdb-iceberg/pull/1407/commits/b0d7dcf57335a3f24e89dcdf5a1eedc803cf7967).

src/execution/operator/iceberg_insert.cpp
tom-kita: It should.
Fixed in [a291f3a](https://github.com/duckdb/duckdb-iceberg/pull/1407/commits/a291f3a4b332f133fd789cef3607fb6bc3a9d88d) using MultiFileReader::LAST_UPDATED_SEQUENCE_NUMBER_ID.
Note that the previous implementation only emits field-id for `_row_id` : https://github.com/duckdb/duckdb-iceberg/blob/887a560a8b70cae07ecdcd6a69ff3cedb6d011bb/src/execution/operator/iceberg_insert.cpp#L355-L357

Tishj: Thanks for fixing that!

tom-kita:
Confirmed the two failures: an `unknown` list element or map key/value (this PR rejects the write) and a struct whose fields are all `unknown` (this PR omits the column).
  
For list and map I followed Iceberg Java, which rejects them:
https://github.com/apache/iceberg/blob/apache-iceberg-1.11.0/parquet/src/main/java/org/apache/iceberg/parquet/TypeToMessageType.java#L138-L164
  
For the struct case I omitted the column because writing the `unknown` field keeps a field id in the data file, which is what makes Spark fail in #1406.
  
I would like to check the Iceberg table spec and what Iceberg Java writes for these types, and whether Spark can read it, before deciding which behavior we want.

tom-kita:
The spec does not cover these cases. Following the Java implementation looks like the reasonable choice to me: it keeps the data files spec-compliant, and it avoids the cases where Parquet cannot tell `NULL` apart from `{'v': NULL}`. I verified what it actually does before proposing a rule.

## What Iceberg Java does

Spark 4.1.1-amzn-0 with Iceberg 1.11.0-amzn-0. `CREATE TABLE` succeeds in all three cases; every write fails:

| column type | write fails with | raised in |
| --- | --- | --- |
| `STRUCT<n: void>` | `InvalidSchemaException: Cannot write a schema with an empty group: optional group v = 2 { }` | parquet-java `TypeUtil.checkValidWriteSchema`, from `ParquetFileWriter.<init>` |
| `ARRAY<void>` | `IllegalArgumentException: Cannot convert element Parquet: unknown` | `TypeToMessageType.list` |
| `MAP<STRING, void>` | `IllegalArgumentException: Cannot convert value Parquet: unknown` | `TypeToMessageType.map` |

So Iceberg Java cannot write a container that holds only `unknown`. `TypeToMessageType.struct()` skips every `unknown` field and hands parquet-java an empty group, which parquet-java refuses to write. Note that empty groups can be *read* (`MessageColumnIO` falls back to `EmptyRecordReader`), but never written.

## What I propose

Omit `unknown` from data files as the spec requires, and reject the write only where omitting it would leave
something Parquet cannot represent:

| | behavior | Iceberg Java |
| --- | --- | --- |
| top-level `unknown` column | omitted | same (`convert()` skips it) |
| `unknown` struct field with a written sibling | omitted | same (`struct()` skips it) |
| struct whose fields are all `unknown` | rejected | same (empty group is refused) |
| `unknown` list element | rejected | same |
| `unknown` map key or value | rejected | same |
| table whose columns are all `unknown` | rejected | same (empty message is refused) |

The alternative is to keep writing the `unknown` field, which is what main does today. That is exactly the shape #1406 is about: the file carries a field id for an `unknown` field, and Spark's vectorized reader fails on it. It also cannot be made to round-trip: a struct's nullness lives in the definition levels of its leaves, so once the only leaf is gone there is nowhere left to record whether the struct was `NULL` or `{'v': NULL}`.

`alter_unknown_type.test` and `alter_unknown_type_variant.test` write rows into the containers while the fields are still `unknown`, so they expect writes that this rule rejects. Before touching them I would like to know whether you agree with the rule itself: is rejecting these writes the behavior we want?

## Appendix: how I verified this

EMR, Spark 4.1.1-amzn-0, Iceberg 1.11.0-amzn-0, Glue catalog. Spark task frames are elided at the end of each
stack trace.

```python
CATALOG, NS = "my_catalog", "duckdb_unknown_test"
T = lambda name: f"{CATALOG}.{NS}.{name}"

print("spark:", spark.version)
print("iceberg:", spark._jvm.org.apache.iceberg.IcebergBuild.fullVersion())
# spark: 4.1.1-amzn-0
# iceberg: Apache Iceberg 1.11.0-amzn-0 (commit amzn)
```

### A struct whose only field is `unknown`

```python
spark.sql(f"""CREATE TABLE {T('u_struct')} (id INT, v STRUCT<n: void>)
              USING iceberg TBLPROPERTIES ('format-version' = '3')""")
spark.sql(f"INSERT INTO {T('u_struct')} VALUES (1, named_struct('n', CAST(NULL AS void))), (2, NULL)")
```

```
org.apache.iceberg.shaded.org.apache.parquet.schema.InvalidSchemaException: Cannot write a schema with an empty group: optional group v = 2 {
}
        at org.apache.iceberg.shaded.org.apache.parquet.schema.TypeUtil$1.visit(TypeUtil.java:27)
        at org.apache.iceberg.shaded.org.apache.parquet.schema.GroupType.accept(GroupType.java:259)
        at org.apache.iceberg.shaded.org.apache.parquet.schema.TypeUtil$1.visit(TypeUtil.java:30)
        at org.apache.iceberg.shaded.org.apache.parquet.schema.TypeUtil$1.visit(TypeUtil.java:36)
        at org.apache.iceberg.shaded.org.apache.parquet.schema.MessageType.accept(MessageType.java:52)
        at org.apache.iceberg.shaded.org.apache.parquet.schema.TypeUtil.checkValidWriteSchema(TypeUtil.java:23)
        at org.apache.iceberg.shaded.org.apache.parquet.hadoop.ParquetFileWriter.<init>(ParquetFileWriter.java:460)
        at org.apache.iceberg.shaded.org.apache.parquet.hadoop.ParquetFileWriter.<init>(ParquetFileWriter.java:433)
        at org.apache.iceberg.parquet.ParquetWriter.ensureWriterInitialized(ParquetWriter.java:114)
        at org.apache.iceberg.parquet.ParquetWriter.flushRowGroup(ParquetWriter.java:240)
        at org.apache.iceberg.parquet.ParquetWriter.close(ParquetWriter.java:301)
        at org.apache.iceberg.io.DataWriter.close(DataWriter.java:83)
	...
```

### An `unknown` list element

```python
spark.sql(f"""CREATE TABLE {T('u_list')} (id INT, l ARRAY<void>)
              USING iceberg TBLPROPERTIES ('format-version' = '3')""")
spark.sql(f"INSERT INTO {T('u_list')} VALUES (1, array(CAST(NULL AS void), CAST(NULL AS void)))")
```

```
java.lang.IllegalArgumentException: Cannot convert element Parquet: unknown
        at org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkArgument(Preconditions.java:220)
        at org.apache.iceberg.parquet.TypeToMessageType.list(TypeToMessageType.java:144)
        at org.apache.iceberg.parquet.TypeToMessageType.field(TypeToMessageType.java:135)
        at org.apache.iceberg.parquet.TypeToMessageType.convert(TypeToMessageType.java:90)
        at org.apache.iceberg.parquet.ParquetSchemaUtil.convert(ParquetSchemaUtil.java:79)
        at org.apache.iceberg.parquet.Parquet$WriteBuilder.build(Parquet.java:397)
        at org.apache.iceberg.parquet.ParquetFormatModel$WriteBuilderWrapper.build(ParquetFormatModel.java:306)
        at org.apache.iceberg.formats.FileWriterBuilderImpl$DataFileWriterBuilder.build(FileWriterBuilderImpl.java:231)
        at org.apache.iceberg.formats.FileWriterBuilderImpl$DataFileWriterBuilder.build(FileWriterBuilderImpl.java:219)
        at org.apache.iceberg.data.RegistryBasedFileWriterFactory.newDataWriter(RegistryBasedFileWriterFactory.java:121)
	...
```

### An `unknown` map value

```python
spark.sql(f"""CREATE TABLE {T('u_map')} (id INT, m MAP<STRING, void>)
              USING iceberg TBLPROPERTIES ('format-version' = '3')""")
spark.sql(f"INSERT INTO {T('u_map')} VALUES (1, map('old', CAST(NULL AS void)))")
```

```
java.lang.IllegalArgumentException: Cannot convert value Parquet: unknown
        at org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkArgument(Preconditions.java:220)
        at org.apache.iceberg.parquet.TypeToMessageType.map(TypeToMessageType.java:159)
        at org.apache.iceberg.parquet.TypeToMessageType.field(TypeToMessageType.java:133)
        at org.apache.iceberg.parquet.TypeToMessageType.convert(TypeToMessageType.java:90)
        at org.apache.iceberg.parquet.ParquetSchemaUtil.convert(ParquetSchemaUtil.java:79)
        at org.apache.iceberg.parquet.Parquet$WriteBuilder.build(Parquet.java:397)
        at org.apache.iceberg.parquet.ParquetFormatModel$WriteBuilderWrapper.build(ParquetFormatModel.java:306)
        at org.apache.iceberg.formats.FileWriterBuilderImpl$DataFileWriterBuilder.build(FileWriterBuilderImpl.java:231)
        at org.apache.iceberg.formats.FileWriterBuilderImpl$DataFileWriterBuilder.build(FileWriterBuilderImpl.java:219)
        at org.apache.iceberg.data.RegistryBasedFileWriterFactory.newDataWriter(RegistryBasedFileWriterFactory.java:121)
	...
```

Line numbers in these traces are from the `amzn` build and are shifted by a few lines from the `apache-iceberg-1.11.0` tag.

Tishj:
Personally I think that the *requirement* for unknown fields column data to be omitted from parquet files is too strong, I interpreted this as merely an optimization when I implemented the `UNKNOWN` write support.

But we want to be compatible and I think it's wishful thinking to assume that Spark will fix their end, so that kind of forces our hand.

TL;DR
Your proposal is great, we should reject shapes that can't be represented correctly by parquet, while omitting UNKNOWN from written files to be compatible with Spark.

Small note, I didn't quite understand this part, could you elaborate on this? Or at least, I do understand that that's how parquet's struct representation works but I'm having trouble visualizing how this problem applies here.
> It also cannot be made to round-trip: a struct's nullness lives in the definition levels of its leaves, so once the only leaf is gone there is nowhere left to record whether the struct was NULL or {'v': NULL}

I take it to mean that a struct where all fields are omitted can't be represented, because NULL-ness is carried by the children, so an empty struct wouldn't be able to represent it. Which in my mind would require this to be replaced by some sentinel leaf, just to carry this information. Which comes with its own can of worms.

Tishj:
Not that it changes the decision, but I wanted to bring it up regardless;

By throwing on this, Spark is actually not spec-compliant, because the spec has this in its requirements for the Parquet format in Appendix A:
> When reading an `unknown` column, any corresponding column must be ignored and replaced with null values.

Something to note:
There exists a `LogicalType.NULL` in Parquet *for this exact reason*. It has 0 of the issues that omitting the field has, and comes up with the same space-savings, as no data has to be represented.

This cuts both ways though, I knew of its existence as I implemented the reader for it, but I mistakenly thought that we had write support for it as well.
Instead we write `INT32`, which makes Spark's refusal more justifiable.

On this topic, I've made https://github.com/apache/iceberg/issues/18190 to hopefully fix this upstream in the spec, removing the limitations that come from this note.

tom-kita:
You read it right. Here is the case I had in mind:

```sql
CREATE TABLE t (id INTEGER, v STRUCT(n "NULL")) WITH ('format-version' = '3');
INSERT INTO t VALUES (1, NULL), (2, {'n': NULL});
```

Row 1 writes `v` as null, row 2 writes `v` as present with a null inside. `n` is the only field, so once it is omitted the file says nothing about `v`, and both rows read back as `NULL`.

I think your proposal (https://github.com/apache/iceberg/issues/18190) is reasonable, and it solves exactly this problem.

About `INT32`:
That impression came from how I worded #1406: the Problem section describes the column as an all-null `INT32` column and never mentions the annotation. Sorry about that. `INT32` is only the physical type; the logical type is already `Null`, as the schema later in the same issue shows:

```
optional int32 field_id=2 data (Null);
```

Tishj:
Thanks!

PR #1410, opened 2026-09-19
Make `DEFAULT_SCHEMA` non-optional, defaults to `main`, add `CREATE_DEFAULT_SCHEMA`

Tishj:
DuckDB only functions to a lesser degree for Catalog's that don't have any schema at all.
And to do anything meaningful with the extension, you need a schema anyways.
To avoid this limbo altogether, this PR makes the `DEFAULT_SCHEMA` set by default (previously it was empty) to DuckDB's default schema (`main`).

If the schema referenced by `DEFAULT_SCHEMA` doesn't exist, we'll throw on `ATTACH`, alluding to both the current `DEFAULT_SCHEMA` setting, and to the new `CREATE_DEFAULT_SCHEMA` option.

That way users can choose whether to change the `DEFAULT_SCHEMA` to whatever schema they know exists, or to the schema they want to create, paired with the `CREATE_DEFAULT_SCHEMA` option to create it on `ATTACH`.

This default of `main` could in the future still be changed to be influenced by the REST Catalog we're attaching to, in case a standardized way emerges in which a Catalog can communicate this, probably through the `v1/config` endpoint in the shape of `defaults` and/or `overrides`.

cc @benfleis 

### Future Work

Support specifying nested namespaces as `DEFAULT_SCHEMA` through a `VARCHAR[]` (to avoid ambiguity in interpretation of the string)


Tishj:
Needs https://github.com/duckdb/duckdb/pull/25934 to fix one of the CI failures

Tishj:
Actually, thanks to @Tmonster making (<https://github.com/duckdb/duckdb/pull/24979>) in core, the `DEFAULT_SCHEMA` being optional is now properly supported by core.

One small concern is that it can't be tested by core, because the default-schema of core isn't configurable.
So this PR (I believe) relies entirely on duckdb-iceberg to test this.

I'm in favor of closing this PR, but maybe we should make `DEFAULT_SCHEMA` a core ATTACH option, and also make duckdb/duckdb respect it, to standardize it, rather than have every Catalog implementation have to re-implement this behavior.

PR #1412, opened 2026-09-20
Send X-Iceberg-Access-Delegation on commit requests when vending credentials

zh3w4ng:
## Summary
- `IRCAPI::CommitTableUpdate` and `IRCAPI::CommitMultiTableUpdate` never set the `X-Iceberg-Access-Delegation` header, even though `GetTable`, `GetTableCredentials`, `GetNamespace`, `GetTables`, and `CommitNewTable` all set it when `ACCESS_DELEGATION_MODE` is `vended_credentials`.
- BigLake's REST catalog strictly enforces the delegation header on every request while in vended-credentials mode, including the commit itself, so committing a transaction failed with:
  ```
  X-Iceberg-Access-Delegation header must be present and contain vended-credentials when credential mode is CREDENTIAL_MODE_VENDED_CREDENTIALS.
  ```
- This mirrors the symptom fixed for AWS in #1393, but has a different root cause: on AWS the header was being silently dropped during SigV4 signing (`ExecuteRequest` rebuilds headers from scratch); here the header was simply never added to the two commit code paths.
- Fix: add the same `access_mode == VENDED_CREDENTIALS` header insertion to both commit functions, matching the existing pattern used elsewhere in `catalog_api.cpp`.

## Test plan
- [x] Extended `test/sql/local/catalog_custom_setup/fixture/vended_credentials_refresh.test` to insert a row (exercising `CommitTableUpdate`) and assert the commit `POST` request carries `X-Iceberg-Access-Delegation: vended-credentials`.
- [ ] Full build + fixture-catalog test run (`FIXTURE_SERVER_AVAILABLE`) not performed in this environment — please run CI / verify against a real BigLake catalog before merging.

🤖 Generated with Claude Code


Tishj:
Thanks for the PR, could you fix the failing CI?

zhe-serve:
@Tishj Thank you for looking into this PR. I haven't got a chance to test this end to end for my BigLake catalog via with credential vending. At the moment I find a workaround to query (both read and write) with end user Mode.

Once I move forward with credential vending in the near future, I'll let you know.

PR #1414, opened 2026-09-20
[Testing] Improved `fixture_duckdb_tests.json` config

Tishj:
Added cleanup to the catalog that runs on the start of every test, to remove the false-negative skips on colliding explicit schema/table creation.


Tishj:
The extra cost to this cleanup is enormous (CI took 2.5x as long)

The only teachings we gained from this run is:
```
1. test/sql/attach/attach_hidden.test:35
Query failed, but error message did not match expected error message: schema "hidden_db.main" does not exist (test/sql/attach/attach_hidden.test:35)!
SELECT * FROM hidden_db.main.tbl;
Actual result:
Catalog Error: Table with name tbl does not exist!
Did you mean "information_schema.tables"?

LINE 1: SELECT * FROM hidden_db.main.tbl
                      ^^^^^^^^^^^^^^^^^^

2. test/sql/settings/test_disabled_file_systems.test:50
Query failed, but error message did not match expected error message: schema "abc.main" does not exist (test/sql/settings/test_disabled_file_systems.test:50)!
SELECT * FROM abc.main.t1;
Actual result:
Catalog Error: Table with name t1 does not exist!
Did you mean "information_schema.tables"?

LINE 1: SELECT * FROM abc.main.t1;
```

None of those are very interesting if we dig into their root cause.

PR #1421, opened 2026-09-22
[Experiment][Testing] Use `require-env[-not]` to check for capability presence/absence instead of explicit skips

Tishj:
Use environment variables defined in the test config to drive enabling/disabling of tests.

One noticeable impact of this:
Our tests used to be opt-out from the Catalog's perspective, opt-out went through the test-config's skipped-tests list.
Now that tests are explicitly enabled through their capability requirement, tests are opt-in.

To make sure coverage doesn't fall through the cracks for a Catalog if it forgets to declare support for functionality that it has, we need to add explicit failure tests using `require-env-not <FEATURE>_SUPPORT`.
That way we're aware of a Catalog's undefined support through the failure of that/those tests.


Tishj:
I think my locally installed/running version of Polaris is out of sync, because I can't reproduce this failure. Nuked everything and recloning/rebuilding now

Tmonster (approved):
Thanks!

PR #1423, opened 2026-09-22
Apply patches from duckdb/duckdb

maiadegraaf:
This PR applies patches from duckdb/duckdb:
- `0001-update-aws-c-io.patch`
- `0003-literal-constant-expression.patch`
- `0004-manifest-read-task-execute-step.patch`

It also bumps the duckdb submodules and `.github/workflows/MainDistributionPipeline.yml` to v2.0.0.


Tishj:
All of these are already applied

